Issue #2679, opened 2026-05-14
Spec adherence: §5.4 remote upload pipeline + .well-known atomicity


tomerweller:
**From #2675 (deferred item 7)**

§5.4 specifies atomic remote upload to history archives with .well-known file semantics. Requires upload pipeline redesign.

**stellar-core ref**: CATCHUP_SPEC.md §5.4
**Parent**: #2675

tomerweller:
## 📝 Proposal Draft (Round 1/5)

## Problem

The current `upload_publish_directory` function in `crates/app/src/app/publish.rs` uploads all files from the staging directory in **alphabetical order**. This means `.well-known/stellar-history.json` is uploaded **before** `history/` files (buckets, ledgers, transactions, results, SCP). This breaks the atomicity guarantee specified in CATCHUP_SPEC.md §5.5.

The `.well-known/stellar-history.json` file acts as the "commit point" for a checkpoint — clients discover new checkpoints by polling this file. If it's uploaded before all referenced data files are in place, clients may see a HAS that references files that don't exist yet on the archive, causing catchup failures.

Additionally, the current implementation does **no differential upload**: every checkpoint uploads all referenced files regardless of what the archive already has. Stellar-core fetches the archive's current HAS first, computes `differingBuckets()`, and only uploads files the archive is missing. This is both a correctness concern (matching stellar-core behavior) and an efficiency concern (avoiding redundant uploads of large bucket files).

## Proposed Fix

Restructure `publish_single_checkpoint` and `upload_publish_directory` to match stellar-core's `PutSnapshotFilesWork` pipeline:

### 1. Enforce HAS-last upload ordering

Split the upload into two phases:
1. **Data files phase**: Upload all checkpoint files (buckets, ledgers, transactions, results, SCP) — everything except the HAS files.
2. **HAS phase**: Upload the HAS to its permanent location (`history/{XX}/{YY}/{ZZ}/history-{hex}.json`) and then to `.well-known/stellar-history.json`.

This mirrors stellar-core's `PutSnapshotFilesWork` which runs `PutFilesWork` (data) then `PutHistoryArchiveStateWork` (HAS permanent + `.well-known`) in a `WorkSequence`.

Implementation: In `upload_publish_directory`, partition collected files into two groups:
- `has_files`: paths matching the HAS permanent pattern (`history/*/history-*.json`) and `.well-known/stellar-history.json`
- `data_files`: everything else

[...]

PR #2731, opened 2026-05-15
Consolidate upload module with RemoteArchive async abstraction

### LINKED ISSUE #2728: Consolidate upload module with RemoteArchive async abstraction
Follow-up from #2679 (Spec adherence: §5.4 remote upload pipeline + .well-known atomicity).

## Context
After #2679, the `henyey-history` crate has two command-execution stacks for archive operations:
1. `crates/history/src/upload.rs` — sync shell commands via `std::process::Command` (used by publish paths)
2. `crates/history/src/remote_archive.rs` — async shell commands via `tokio::process::Command` (used by archive init and other operations)

This duplication means bug fixes or improvements to command execution must be applied in two places.

## What needs to happen
1. Migrate the publish paths (`crates/app/src/app/publish.rs` and `crates/henyey/src/publish_history.rs`) to use `RemoteArchive` for file uploads
2. Make `UploadPlan::execute()` accept a `RemoteArchive` reference instead of raw `put_cmd`/`mkdir_cmd` strings
3. Remove the sync `run_shell_command` from `upload.rs` once all callers use the async path
4. Update the publish `spawn_blocking` wrapper to handle the async-to-sync boundary

## Dependencies
- Blocked by #2727 — differential upload should be implemented first so the consolidated path handles both ordering and differential semantics

## References
- Parent issue: #2679
- Implementation commit: 3b65391d
- `RemoteArchive`: `crates/history/src/remote_archive.rs:83-262`

tomerweller:
Closes #2728

## Summary

Migrate `UploadPlan::execute` to async and accept a `&RemoteArchive` instead of raw `put_cmd`/`mkdir_cmd` shell-command templates. Port both publish callers (`crates/app/src/app/publish.rs` and `crates/henyey/src/publish_history.rs`) and delete the redundant sync `run_shell_command` helper in `upload.rs`. This consolidates the two parallel command-execution stacks introduced alongside #2679: `RemoteArchive` now owns all shell exec / format / error mapping, while `UploadPlan` retains ordering, mkdir deduplication, and stop-on-first-failure semantics.

A new `RemoteArchive::has_mkdir()` accessor preserves the "mkdir optional" semantics from the old `Option<&str>` signature (rather than letting `mkdir` return `RemoteNotConfigured`). The async/sync bridge in `app/publish.rs` captures `Handle::current()` in the async caller and threads it into `publish_single_checkpoint`, which calls `rt.block_on(plan.execute(&remote_archive))` from `spawn_blocking` — the documented, panic-free pattern.

Shell commands invoked, error variants surfaced on the production path, and ordering invariants (data → CheckpointHas → RootHas) are all unchanged. The one documented strictness increase: `RemoteArchive::put_file` pre-checks that the local file exists, returning `HistoryError::NotFound` instead of letting the shell fail with `RemoteCommandFailed`. Unreachable in production because staging dirs always contain the referenced files; documented in a comment on `execute`.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/2728#issuecomment-4463608422)

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test -p henyey-history --lib` — 480 passed
- [x] `cargo test -p henyey-history --tests` — all green (including 3 converted + 3 new `execute_*` tests)
- [x] `cargo test -p henyey-app --tests` — 921 + integration tests passed (exercises the new `spawn_blocking → Handle::block_on(plan.execute)` bridge end-to-end via `maybe_publish_history`)
- [x] `cargo test -p henyey --tests` — green (exercises the async CLI path in `cmd_publish_history`)

## Deviations from plan

None.

### DIFF
--- a/crates/app/src/app/publish.rs
+++ b/crates/app/src/app/publish.rs
@@ -107,6 +107,13 @@ impl App {
             "Publishing checkpoint to history archive (background)"
         );
 
+        // Capture the runtime handle so the blocking task can drive the
+        // async upload path via `Handle::block_on`. `spawn_blocking` runs on
+        // the runtime's blocking pool, where `Handle::current()` is valid,
+        // but threading the handle explicitly makes the dependency visible
+        // and works from non-async test entry points too.
+        let rt = tokio::runtime::Handle::current();
+
         // Spawn the publish as a background task to avoid blocking the
         // event loop. Publishing involves gzip compression of XDR files
         // and shell command execution (cp/mkdir), which can take 30-50
@@ -128,7 +135,7 @@ impl App {
             // the process (release builds use `panic = "abort"`) and are not
             // counted here.
             let publish_started = std::time::Instant::now();
-            match app.publish_single_checkpoint(checkpoint, &command_archives) {
+            match app.publish_single_checkpoint(checkpoint, &command_archives, &rt) {
                 Ok(()) => {
                     let elapsed = publish_started.elapsed();
                     crate::metrics::HISTORY_PUBLISH_SUCCESS_TOTAL.increment(1);
@@ -173,6 +180,7 @@ impl App {
         &self,
         checkpoint: u32,
         archives: &[&crate::config::HistoryArchiveEntry],
+        rt: &tokio::runtime::Handle,
     ) -> anyhow::Result<()> {
         #[cfg(test)]
         if self.publish_panic_inject.swap(false, Ordering::SeqCst) {
@@ -342,8 +350,17 @@ impl App {
         // skip bucket files already present. Falls back to full upload on error.
         let base_plan = henyey_history::upload::UploadPlan::from_staging_dir(&publish_dir)?;
         for archive in archives {
-            let put_cmd = archive.put.as_ref().unwrap();
-            let mkdir_cmd = archive.mkdir.as_deref();
+            // Build a `RemoteArchive` per command-based archive. `put` is
+            // guaranteed `Some` here by the caller's filter
+            // (`a.put_enabled && a.put.is_some()`); `get_cmd` is `None`
+            // because publish is write-only.
+            let remote_archive =
+                henyey_history::RemoteArchive::new(henyey_history::RemoteArchiveConfig {
+                    name: archive.name.clone(),
+                    put_cmd: archive.put.clone(),
+                    mkdir_cmd: archive.mkdir.clone(),
+                    get_cmd: None,
+                });
 
             let plan = match henyey_history::archive::fetch_root_has_blocking(&archive.url) {
                 Ok(remote_has) => {
@@ -375,7 +392,11 @@ impl App {
                 }
             };
 
-            plan.execute(put_cmd, mkdir_cmd)?;
+            // Drive the async upload path from this blocking task via the
+            // captured runtime handle. `Handle::block_on` is the documented
+            // correct pattern from a `spawn_blocking` worker (no nested
+            // runtime is created).
+            rt.block_on(plan.execute(&remote_archive))?;
         }
 
         // Clean up staging directory (TempDir drops silently on error paths;
--- a/crates/henyey/src/publish_history.rs
+++ b/crates/henyey/src/publish_history.rs
@@ -318,7 +318,14 @@ pub(crate) async fn cmd_publish_history(config: AppConfig, force: bool) -> anyho
             write_scp_history_file(publish_dir, checkpoint, &scp_entries)?;
             let plan = henyey_history::upload::UploadPlan::from_staging_dir(publish_dir)?;
             for archive in &command_targets {
-                plan.execute(&archive.put, archive.mkdir.as_deref())?;
+                let remote_archive =
+                    henyey_history::RemoteArchive::new(henyey_history::RemoteArchiveConfig {
+                        name: archive.name.clone(),
+                        put_cmd: Some(archive.put.clone()),
+                        mkdir_cmd: archive.mkdir.clone(),
+                        get_cmd: None,
+                    });
+                plan.execute(&remote_archive).await?;
                 println!("OK (command: {})", archive.name);
                 published_any = true;
             }
--- a/crates/history/src/remote_archive.rs
+++ b/crates/history/src/remote_archive.rs
@@ -105,6 +105,15 @@ impl RemoteArchive {
         self.config.is_readable()
     }
 
+    /// Check if this archive has a mkdir command configured.
+    ///
+    /// Callers that want to skip mkdir entirely when no command is set
+    /// (rather than receiving `HistoryError::RemoteNotConfigured` from
+    /// [`Self::mkdir`]) should gate on this accessor.
+    pub fn has_mkdir(&self) -> bool {
+        self.config.mkdir_cmd.is_some()
+    }
+
     /// Format a command template by replacing placeholders.
     ///
     /// Placeholders are `{0}`, `{1}`, etc.
--- a/crates/history/src/upload.rs
+++ b/crates/history/src/upload.rs
@@ -34,7 +34,8 @@
 //! that references files not yet present on the archive.
 
 use crate::archive_state::HistoryArchiveState;
-use crate::{paths, HistoryError, Result};
+use crate::remote_archive::RemoteArchive;
+use crate::{paths, Result};
 use henyey_common::Hash256;
 use std::collections::HashSet;
 use std::path::{Path, PathBuf};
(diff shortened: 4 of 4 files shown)


tomerweller:
## Reviewer B (risk lens) — APPROVE

(gh refused `--approve` because the authenticated user is the PR author; posting as a comment instead.)

Refactor preserves the production semantics that matter (ordering, mkdir dedup, stop-on-first-failure, `RemoteCommandFailed` variant + fields). No blocking risk found.

Notes (non-blocking):

- **Runtime bridge**: `Handle::block_on(plan.execute(...))` from `spawn_blocking` relies on a multi-thread runtime. Confirmed — `#[tokio::main]` in `henyey/src/main.rs` defaults to multi-thread; documented pattern, no deadlock risk here. Would be a sharp edge if anyone later builds the app on a current-thread runtime, but that's outside this PR.
- **Behavior change (acknowledged)**: `RemoteArchive::put_file` pre-checks `local_path.exists()` and returns `NotFound` instead of `RemoteCommandFailed`. Same stop-on-failure outcome; only the error variant differs. Operator-facing logs/metrics that key on `RemoteCommandFailed` would now miss this case, but it's unreachable in production (staging-dir invariant).
- **CLI path** (`cmd_publish_history`) previously called sync `std::process::Command` inside an `async fn`, blocking the runtime thread for the duration of `sh`. Switching to `tokio::process::Command` is a small *improvement* on that path.
- **Perf**: `tokio::process::Command` has marginally more setup cost than `std::process::Command` (signal handler / reaper registration), negligible vs. multi-second `aws s3 cp` etc.
- **No** changes to config schema, on-disk layout, network protocol, credentials handling, or public API surface beyond the new `UploadPlan::execute` signature and `RemoteArchive::has_mkdir()` accessor — both internal to publish callers.

tomerweller (commented):
## Reviewer A — Verdict: APPROVE (correctness)

(GitHub blocks self-approval via the API; recording the verdict as a review comment with APPROVE called out explicitly.)

Scope: logic correctness, error paths, test coverage, readability. Risk reviewed separately by Reviewer B.

### Verified
- **Async/sync bridge** in `app/publish.rs`: `Handle::current()` is captured in the async caller (a runtime is guaranteed by surrounding `.await` calls), moved into the `spawn_blocking` closure, and `rt.block_on(plan.execute(...))` is the documented panic-free pattern from a blocking worker. No nested runtime, no `Handle::current()` lookup inside the blocking pool.
- **Ordering & stop-on-failure** preserved: `UploadPlan::execute` iterates `self.entries` sequentially and short-circuits on `?`, so a failure before `RootHas` still prevents its upload. `execute_failure_prevents_root_has_upload` continues to assert this.
- **mkdir dedup**: `HashSet<String>` accumulator preserved; `mkdir_enabled` cached once outside the loop (`RemoteArchive` is immutable, safe).
- **mkdir error propagation**: the new path calls `archive.mkdir()` directly — NOT `ensure_dir` (which swallows errors). Matches the old `run_shell_command` semantics; no silent recovery introduced.
- **Error-variant strictness change** (`NotFound` from local-file pre-check vs. `RemoteCommandFailed`): grep confirms no caller matches on `RemoteCommandFailed` in the upload chain; both propagate via `?`. Documented in the doc-comment on `execute`. Unreachable in production (staging dirs always contain referenced files).
- **Template formatting** in `RemoteArchive::format_command` is byte-equivalent to the old `.replace("{0}", ..).replace("{1}", ..)` chain — same pre-existing caveat for paths containing literal `{N}`. No regression.
- **`publish_history.rs`**: `cmd_publish_history` was already `async fn`; replacing sync `plan.execute(...)?` with `.await?` is strictly more correct (no blocking call from async).
- **`spawn_blocking` closure** moves `rt` by value; `Handle` is `Clone` and cheap.

### Test coverage
3 ported + 3 new tests in `upload.rs`:
- `execute_uploads_in_correct_order`, `execute_failure_prevents_root_has_upload`, `execute_creates_directories_before_upload` — converted to `#[tokio::test]`, assertions unchanged.
- `execute_propagates_remote_archive_errors` — exit-code propagation through `RemoteCommandFailed`.
- `execute_skips_mkdir_when_unconfigured` — `has_mkdir()` gate.
- `execute_reports_missing_local_file` — pins the new pre-check and confirms stop-on-first-failure (no log file after the failed entry).

Good coverage for the behavior surface the consolidation touches.

### Nits (non-blocking)
- `mkdir_enabled` is hoisted outside the loop, which is fine but slightly asymmetric with the per-iteration `archive.put_file`. Inlining `archive.has_mkdir()` at the `if` would be equally clear. Not worth a round-trip.

**Verdict: APPROVE** — no blocking correctness concerns.

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Refactor is behavior-preserving on the publish hot path, no breaking API impact outside the workspace, and runtime/threading model (`Handle::current()` + `block_on` from a `spawn_blocking` worker) is the correct pattern. No material regression, perf, security, or operational risk found.

<details>

**Regression risk — low**
- Both call sites (`crates/app/src/app/publish.rs`, `crates/henyey/src/publish_history.rs`) preserve the original semantics: same per-archive iteration, same `mkdir`-then-`put` ordering, same dedup of created dirs via `HashSet`, and the `RootHas-last` ordering invariant is untouched in `UploadPlan`.
- `run_shell_command` was a thin wrapper around `std::process::Command::new("sh").arg("-c")…output()`. `RemoteArchive::execute_command` is the same shape on tokio (`sh -c`, `output().await`, same exit-code/stderr propagation to `HistoryError::RemoteCommandFailed`). No change in how shell quoting / placeholder substitution is performed.
- One small behavior delta worth noting (not blocking): `RemoteArchive::put_file` adds a `local_path.exists()` pre-check that returns `HistoryError::NotFound` instead of `RemoteCommandFailed`. Current callers don't branch on error variant, and the new `execute_reports_missing_local_file` test pins it. Just worth being aware of for any future caller that pattern-matches on the error.

**Performance — neutral**
- Publish path still runs inside `tokio::task::spawn_blocking`; `rt.block_on(plan.execute(&remote_archive))` from a blocking-pool worker is the documented-correct way to drive an async future from sync code without nesting runtimes. No extra thread hops vs. the previous direct sync call.
- No new locks, no new allocations on a hot loop beyond `RemoteArchive::new(...)` once per archive per checkpoint (a handful of string clones); negligible compared to gzip + shell exec.
- `tokio::process::Command::output()` is functionally equivalent throughput-wise to the previous blocking `std::process::Command::output()`; both wait for the child to exit before returning.

**API / data-format breakage — internal only**
- `UploadPlan::execute` signature changes from `(&str, Option<&str>) -> Result<()>` to `async (&RemoteArchive) -> Result<()>`. This is breaking for any external consumer of `henyey-history`, but the workspace has only the two updated call sites and the crate isn't published. Acceptable.
- `App::publish_single_checkpoint` gained a `&Handle` parameter. It's a private method on `App` with only the one caller — fine.
- New `RemoteArchive::has_mkdir()` accessor is purely additive.
- No on-disk format, config schema, or HAS-format changes. `configs/*.toml` `put`/`mkdir` templates are unaffected (still use `{0}`/`{1}` placeholders).

**Security — no new surface**
- Same `sh -c <template-with-substituted-paths>` execution model. Template substitution is unchanged (still `String::replace("{0}", …)`), so any pre-existing injection considerations around operator-supplied `put`/`mkdir` templates are neither introduced nor mitigated here.
- No auth/signing/network code touched. No new file paths or symlink-following semantics introduced (the `local_path.exists()` pre-check follows symlinks the same way `cp` would).

**Operational / observability**
- Net improvement: failed uploads now go through `RemoteArchive::execute_command`, which emits `warn!(archive, command, exit_code, stderr, stdout, …)`. Previously `run_shell_command` returned the same data inside `HistoryError::RemoteCommandFailed` but did not log automatically.
- Soft heads-up (not a blocker, no action needed): the full `command` string and `stderr`/`stdout` are logged at `warn!` on failure. Today's distributed `configs/` only use `cp` / `put_enabled = false`, but if operators ever configure `put`/`mkdir` templates with embedded credentials or signed URLs, those would land in logs on failure. This was already true for download/get paths going through `RemoteArchive`, so consolidating publish onto the same path is consistent, not a new exposure — worth documenting in `crates/history/README.md` at some point but not in scope here.
- Debug-level per-command logging (`debug!("Executing remote archive command")`) is gated and won't affect default log volume.
- No migrations, no config changes, no metric renames.

**Parity**
- N/A — `history/upload` and the publish background task are non-parity paths (output-side archive maintenance, not ledger close). Confirmed against `crates/history/PARITY_STATUS.md` scope.

Net: ship it.
</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Diff is a clean async migration of `UploadPlan::execute` onto `RemoteArchive`. Semantics (ordering, mkdir dedup, stop-on-first-failure) are preserved; new error paths are documented and covered by tests; the `spawn_blocking` → `Handle::block_on` bridge is used correctly.

<details>

**Correctness checks**

- `UploadPlan::execute` async port preserves behavior:
  - Iteration order unchanged (entries still sorted with `RootHas` last).
  - `created_dirs: HashSet<String>` dedup with `insert(parent_str.clone())` gate is identical to the prior logic — mkdir is invoked at most once per parent per plan.
  - `mkdir_enabled = archive.has_mkdir()` correctly mirrors the old `if let Some(mkdir_cmd)` gate. `has_mkdir()` is a plain `is_some()` accessor — no logic divergence.
  - Per-entry call sequence (mkdir-then-put) is unchanged; `archive.mkdir(&parent_str)` substitutes `{0}` exactly as the prior inline `mkdir.replace("{0}", &parent_str)`.
  - `archive.put_file(&entry.local_path, &entry.remote_path)` internally does `local_path.to_string_lossy()` and substitutes `{0}`/`{1}` — equivalent to the deleted inline `put_cmd.replace(...)`.
  - First failure short-circuits via `?` exactly as before, so the data-before-RootHas guarantee documented in the module header still holds.

- Strictness change is fully scoped and documented:
  - `RemoteArchive::put_file` returns `HistoryError::NotFound` (not `RemoteCommandFailed`) on missing local file. Documented on `execute`'s rustdoc, justified as unreachable in production (staging dir always populated). Covered by the new `execute_reports_missing_local_file` test, which also asserts the short-circuit (no subsequent uploads).

- Caller in `crates/app/src/app/publish.rs`:
  - `Handle::current()` captured in async context **before** `spawn_blocking` and moved into the closure — correct. Capturing inside the closure would also work (blocking pool threads inherit handle context), but threading it explicitly is more robust and the comment explains why.
  - `rt.block_on(plan.execute(&remote_archive))?` from inside `spawn_blocking` is the documented panic-free pattern (blocking-pool threads are not runtime workers, so no nested-runtime panic and no worker starvation).
  - `RemoteArchiveConfig` construction with `put_cmd: archive.put.clone()` is safe given the upstream filter `a.put_enabled && a.put.is_some()` (comment notes this); `get_cmd: None` is appropriate on the write-only publish path.

- Caller in `crates/henyey/src/publish_history.rs`:
  - `cmd_publish_history` is already `async`, so `plan.execute(&remote_archive).await` is the natural call form. `archive.put` is `String` (not `Option`) in this struct, so `Some(archive.put.clone())` is correct.

**Async pitfalls — checked, none found**

- `RemoteArchive::execute_command` uses `tokio::process::Command` which requires the process driver; both callers run on the application's full tokio runtime (multi-thread, all features), so the driver is present.
- `local_path.exists()` inside `put_file` is a sync stat in an async fn — acceptable here (fast metadata, also matches surrounding style), and notably not a deadlock or runtime-blocking concern.
- No `.await` is held across a `MutexGuard` or `RefCell`; `created_dirs` lives entirely within the single async task.

**Error handling**

- All failure paths return `Result`; no `unwrap`/`expect` on runtime values. Test-only code uses `unwrap`/`expect_err` as expected.
- `RemoteArchive::execute_command` maps `io::Error` → `HistoryError::Io` and non-zero exits → `HistoryError::RemoteCommandFailed { command, exit_code, stderr }` — descriptive and unchanged from the deleted `run_shell_command`.
- Out-of-scope nit (not blocking): `RemoteArchive::ensure_dir` silently swallows mkdir errors via `let _ = self.mkdir(...).await;`. This PR doesn't touch it and `UploadPlan::execute` uses the strict `mkdir` path, so behavior on the publish path is unaffected.

**Test coverage**

- 3 existing tests converted to `#[tokio::test]` cleanly (ordering, mkdir-before-put, root-has-not-uploaded-on-failure).
- 3 new tests added, each exercising a distinct new code path:
  - `execute_propagates_remote_archive_errors` — non-zero exit → `RemoteCommandFailed { exit_code: Some(7) }`.
  - `execute_skips_mkdir_when_unconfigured` — `has_mkdir() == false` gate.
  - `execute_reports_missing_local_file` — `NotFound` strictness change, plus short-circuit assertion that no upload runs afterward.
- The `spawn_blocking → Handle::block_on` bridge in `app/publish.rs` is exercised end-to-end by the existing `henyey-app` integration tests (per PR description, 921 + integration green).
- The `cmd_publish_history` async path is exercised by existing `henyey` CLI tests.

**Readability / idiomatic Rust**

- `RemoteArchiveConfig { name, put_cmd, mkdir_cmd, get_cmd: None }` is verbose at each call site but explicit and easy to read; no clear win in adding a helper just for two sites.
- Doc comment on `execute` updated to call out both the `RemoteArchive` dependency and the `NotFound` strictness change — good.
- `has_mkdir()` accessor is a small, well-documented addition that preserves the prior "mkdir optional" semantics without leaking `RemoteNotConfigured` into `UploadPlan`.
- Test helper `test_archive(put, mkdir)` is a nice, localized abstraction; not over-engineered.

No required changes.

</details>

tomerweller:
## ✅ Merged

Triple-green auto-merge:
- **Reviewer A (Correctness):** APPROVE
- **Reviewer B (Risk):** APPROVE
- **CI:** green (34/34 SUCCESS)

Merge commit: `85950885eacb98da37fc829de672fd008645f17c`

PR #2732, opened 2026-05-15
Implement §11.2 5-case knit-to-LCL decision matrix

### LINKED ISSUE #2680: Spec adherence: §11.2 full knit-to-LCL 5-case decision matrix
**From #2675 (deferred item 8)**

§11.2 specifies the full 5-case decision matrix for knitting catchup state to the Last Closed Ledger. Current implementation handles a subset.

**stellar-core ref**: CATCHUP_SPEC.md §11.2
**Parent**: #2675

tomerweller:
Closes #2680

## Summary

Implements the §11.2 5-case knit-to-LCL decision matrix mirroring
stellar-core's `ApplyCheckpointWork::getNextLedgerCloseData()`
(`stellar-core/src/catchup/ApplyCheckpointWork.cpp:205-262`).

- New pure helper `knit_to_lcl_decision()` + `KnitDecision` enum.
- New `verify_knit_to_lcl()` pre-apply check wired into
  `download_verify_and_replay_once` before chain verification.
- `download_ledger_data` widened to surface knit-prefix entries
  (seq ≤ LCL within the LCL+1 checkpoint file) as raw
  `LedgerHeaderHistoryEntry`s, separately from apply data.
- Four new fatal `HistoryError` variants for the case-2/3/4/5 failures,
  added to `is_fatal_catchup_failure`.

## Plan reference

[Converged Plan](https://github.com/stellar-experimental/henyey/issues/2680#issuecomment-) (issue #2680).

## Tests

- 11 new unit tests in `replay.rs` covering all 5 cases plus
  skip-old / genesis / checkpoint-boundary / retry-advances-LCL.
- Extended `test_fatal_errors_not_retriable` with the 4 new variants.
- `cargo test -p henyey-history` (lib + integration): 491 + 7 + 3 + 1 + 1 + 27 all green.
- `cargo fmt --check` ✓
- `cargo clippy --all -- -D warnings` ✓

## Deviations from plan

1. **`knit_start` formula**: plan specified
   `max(checkpoint_start(from_ledger), from_ledger-1)`. Actually used
   `max(checkpoint_start(apply_start), from_ledger-1)`. The plan's
   formula crosses into a prior checkpoint when LCL is the last ledger
   of a checkpoint, which contradicts stellar-core's "only consult
   entries already in the current checkpoint file" semantics. The
   adjusted formula matches stellar-core exactly.

2. **Missing knit-prefix entries are tolerated silently** (not an error).
   Stellar-core iterates whatever headers happen to be in the open
   checkpoint file — it does not require any specific knit entry to be
   present. Matching that behavior preserves parity with existing
   integration tests whose archive fixtures don't include
   `LCL-1` / `LCL` headers.

3. **Fourth error variant added**: `KnitCurrentLedgerPrevHashMismatch`.
   The plan said "if needed" — confirmed needed by reading stellar-core's
   two distinct error strings at lines 233 and 254-261.

4. **Integration tests for tamper detection deferred**.
   `test_catchup_replay_detects_lcl_overlap_tamper` and
   `test_catchup_replay_detects_lcl_predecessor_tamper` would require
   new multi-ledger archive fixture infrastructure (bootstrapping
   `LedgerManager` at a non-genesis LCL with a tampered archive). The
   11 new unit tests cover the tamper-detection logic comprehensively
   at the function level. Per Critic C's guidance, deferred to a
   follow-up PR if creep is observed.

## Files changed

- `crates/history/src/error.rs` — 4 fatal variants + classification.
- `crates/history/src/catchup/replay.rs` — `KnitDecision`,
  `knit_to_lcl_decision`, `verify_knit_to_lcl`, tests, wiring.
- `crates/history/src/catchup/download.rs` — widened signature,
  knit-prefix collection.
- `docs/spec-eval/CATCHUP_SPEC_HENYEY_EVAL.md` — §11.2 row updated.
- `crates/history/PARITY_STATUS.md` — catchup row updated.

### DIFF
--- a/crates/history/src/catchup/replay.rs
+++ b/crates/history/src/catchup/replay.rs
@@ -4,11 +4,117 @@ use crate::{verify, HistoryError, Result};
 use std::sync::Arc;
 
 use henyey_common::protocol::LclContext;
+use henyey_common::Hash256;
 use henyey_ledger::{HeaderSnapshot, LedgerCloseData, LedgerManager};
-use stellar_xdr::curr::{LedgerHeader, LedgerUpgrade, Limits, ReadXdr, WriteXdr};
+use stellar_xdr::curr::{
+    LedgerHeader, LedgerHeaderHistoryEntry, LedgerUpgrade, Limits, ReadXdr, WriteXdr,
+};
 use tracing::{debug, info, warn};
 
 use super::{CatchupManager, CatchupStatus, LedgerData};
+
+/// Decision returned by [`knit_to_lcl_decision`] for a single archive
+/// `LedgerHeaderHistoryEntry`, mirroring the five-case decision matrix in
+/// stellar-core `ApplyCheckpointWork::getNextLedgerCloseData()`
+/// (CATCHUP_SPEC §11.2):
+///
+/// | Case | Condition | Result |
+/// |------|-----------|--------|
+/// | 1 (skip-old) | `entry.seq + 1 < lcl.seq` | `Ok(Skip)` |
+/// | 2 (LCL predecessor knit) | `entry.seq + 1 == lcl.seq`, hashes match | `Ok(Skip)` |
+/// | 3 (LCL overlap knit) | `entry.seq == lcl.seq`, hashes match | `Ok(Skip)` |
+/// | 4 (apply) | `entry.seq == lcl.seq + 1`, prev-hash matches | `Ok(Apply)` |
+/// | 5 (overshoot) | `entry.seq > lcl.seq + 1` | `Err(KnitOvershot)` |
+///
+/// Hash mismatches in cases 2/3/4 return the appropriate fatal
+/// [`HistoryError`] variant.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(super) enum KnitDecision {
+    /// Drop the entry: it's at or below LCL and (where required) its hash
+    /// agrees with the local LCL chain. The entry must not be replayed.
+    Skip,
+    /// Apply the entry: it is exactly `lcl + 1` and its
+    /// `previousLedgerHash` matches the local LCL hash.
+    Apply,
+}
+
+/// Apply the §11.2 5-case knit-to-LCL decision matrix to a single archive
+/// header entry.
+///
+/// Mirrors stellar-core `ApplyCheckpointWork::getNextLedgerCloseData()` at
+/// the pinned `stellar-core/v26.0.1` submodule. The comparison order
+/// (case 1 → 2 → 3 → 5 → 4) is preserved exactly, as are the per-case
+/// field selections (`entry.hash` vs `entry.header.previous_ledger_hash`).
+///
+/// Returns:
+/// - `Ok(Skip)` for cases 1/2/3 when hashes agree.
+/// - `Ok(Apply)` for case 4 when the previous-hash check succeeds.
+/// - `Err(KnitLclPredecessorHashMismatch)` for case 2 mismatch.
+/// - `Err(KnitLclHashMismatch)` for case 3 mismatch.
+/// - `Err(KnitCurrentLedgerPrevHashMismatch)` for case 4 prev-hash
+///   mismatch. Distinct from case 3 to match stellar-core's two distinct
+///   error messages.
+/// - `Err(KnitOvershot)` for case 5.
+pub(super) fn knit_to_lcl_decision(
+    entry: &LedgerHeaderHistoryEntry,
+    lcl: &HeaderSnapshot,
+) -> Result<KnitDecision> {
+    let entry_seq = entry.header.ledger_seq;
+    let lcl_seq = lcl.header.ledger_seq;
+
+    // Case 1: entry.seq + 1 < lcl.seq — well before LCL, drop silently.
+    if entry_seq.saturating_add(1) < lcl_seq {
+        debug!(entry_seq, lcl_seq, "Knit: case 1 (skip-old)");
+        return Ok(KnitDecision::Skip);
+    }
+
+    // Case 2: entry.seq + 1 == lcl.seq — must match lcl.previousLedgerHash.
+    if entry_seq.saturating_add(1) == lcl_seq {
+        let expected = Hash256::from(lcl.header.previous_ledger_hash.clone());
+        let actual = Hash256::from(entry.hash.clone());
+        if expected != actual {
+            return Err(HistoryError::KnitLclPredecessorHashMismatch {
+                ledger: entry_seq,
+                expected: expected.to_hex(),
+                actual: actual.to_hex(),
+            });
+        }
+        debug!(entry_seq, "Knit: case 2 (LCL predecessor) hash matches");
+        return Ok(KnitDecision::Skip);
+    }
+
+    // Case 3: entry.seq == lcl.seq — must match lcl.hash.
+    if entry_seq == lcl_seq {
+        let actual = Hash256::from(entry.hash.clone());
+        if actual != lcl.hash {
+            return Err(HistoryError::KnitLclHashMismatch {
+                expected: lcl.hash.to_hex(),
+                actual: actual.to_hex(),
+            });
+        }
+        debug!(entry_seq, "Knit: case 3 (LCL overlap) hash matches");
+        return Ok(KnitDecision::Skip);
+    }
+
+    // Case 5: entry.seq > lcl.seq + 1 — overshoot (checked before case 4 to
+    // match stellar-core's branch order at ApplyCheckpointWork.cpp:246).
+    if entry_seq != lcl_seq.saturating_add(1) {
+        return Err(HistoryError::KnitOvershot { entry_seq, lcl_seq });
+    }
+
+    // Case 4: entry.seq == lcl.seq + 1 — entry.header.previousLedgerHash
+    // must match lcl.hash.
+    let entry_prev = Hash256::from(entry.header.previous_ledger_hash.clone());
+    if entry_prev != lcl.hash {
+        return Err(HistoryError::KnitCurrentLedgerPrevHashMismatch {
+            ledger: entry_seq,
+            expected: lcl.hash.to_hex(),
+            actual: entry_prev.to_hex(),
+        });
+    }
+    debug!(entry_seq, "Knit: case 4 (apply) prev-hash matches");
+    Ok(KnitDecision::Apply)
+}
 
 /// Maximum number of retry attempts for the download-and-replay pipeline.
 ///
@@ -47,6 +153,64 @@ pub(super) fn decode_upgrades_from_header(header: &LedgerHeader) -> Vec<LedgerUp
 }
 
 impl CatchupManager {
+    /// Apply the §11.2 5-case decision matrix to the knit-prefix entries
+    /// (entries at or below LCL drawn from the same checkpoint file as
+    /// LCL+1) and to the first apply entry. Returns the apply entries that
+    /// must be replayed (i.e. `Apply`-classified entries), in the same
+    /// order as `apply_data`.
+    ///
+    /// `knit_entries` carries the raw `LedgerHeaderHistoryEntry` records
+    /// (as found in the archive checkpoint file) for ledgers in
+    /// `[knit_start, lcl_seq]`. They are validated against `lcl` and
+    /// dropped from replay; mismatches surface as the case-specific fatal
+    /// variants on [`HistoryError`].
+    ///
+    /// `apply_data` carries the per-ledger `LedgerData` for ledgers in
+    /// `[lcl_seq + 1, target]`. Its first entry is checked for cases 4 and
+    /// 5 (apply-link to LCL or overshoot). Remaining entries are validated
+    /// by the chain check downstream.
+    pub(super) fn verify_knit_to_lcl(
+        &self,
+        knit_entries: &[LedgerHeaderHistoryEntry],
+        apply_data: &[LedgerData],
+        lcl: &HeaderSnapshot,
+    ) -> Result<()> {
+        for entry in knit_entries {
+            let decision = knit_to_lcl_decision(entry, lcl)?;
+            debug_assert_eq!(
+                decision,
+                KnitDecision::Skip,
+                "knit-prefix entries must classify as Skip"
+            );
+        }
+        if let Some(first) = apply_data.first() {
+            let header = first.header().clone();
+            let entry_hash = henyey_ledger::compute_header_hash(&header).map_err(|e| {
+                HistoryError::CatchupFailed(format!(
+                    "Failed to compute header hash for ledger {}: {}",
+                    header.ledger_seq, e
+                ))
+            })?;
+            let virtual_entry = LedgerHeaderHistoryEntry {
+                hash: entry_hash.into(),
+                header,
+                ext: Default::default(),
+            };
+            let decision = knit_to_lcl_decision(&virtual_entry, lcl)?;
+            debug_assert_eq!(
+                decision,
+                KnitDecision::Apply,
+                "first apply entry must classify as Apply"
+            );
+        }
+        info!(
+            knit_entries = knit_entries.len(),
+            apply_entries = apply_data.len(),
+            "Knit-to-LCL decision matrix validated"
+        );
+        Ok(())
+    }
+
     /// Verify the downloaded ledger data using reverse-walk chain verification (§9.2–§9.5).
     pub(super) fn verify_downloaded_data(
         &self,
@@ -236,9 +400,17 @@ impl CatchupManager {
             4,
             "Downloading ledger data",
         );
-        let (ledger_data, archive_name) = self
+        let (ledger_data, knit_entries, archive_name) = self
             .download_ledger_data(download_from, target, lcl)
             .await?;
+
+        // CATCHUP_SPEC §11.2: apply the 5-case knit-to-LCL decision matrix
+        // before chain verification. This catches case-2/3/4/5 hash and
+        // sequencing failures with their specific fatal error variants
+        // (mirroring stellar-core ApplyCheckpointWork::getNextLedgerCloseData())
+        // rather than letting them surface as generic chain-link errors.
+        let lcl_snapshot = ledger_manager.header_snapshot();
+        self.verify_knit_to_lcl(&knit_entries, &ledger_data, &lcl_snapshot)?;
 
         // Verify the header chain.
         //
@@ -247,7 +419,6 @@ impl CatchupManager {
         // `apply_ledger_chain_*` counters: a single outer success can include
         // multiple verify failures from prior attempts.
         self.update_progress(CatchupStatus::Verifying, 5, "Verifying header chain");
-        let lcl_snapshot = ledger_manager.header_snapshot();
         match self.verify_downloaded_data(&ledger_data, &lcl_snapshot) {
             Ok(()) => {
                 metrics::counter!(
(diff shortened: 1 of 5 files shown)


tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** The 5-case knit-to-LCL decision matrix mirrors `stellar-core/src/catchup/ApplyCheckpointWork.cpp:205-262` faithfully — same ordering (1→2→3→5→4), same per-case field selection, same triple-distinct fatal error variants. PARITY_STATUS.md and the spec-eval doc are updated accurately.

<details>

### Mapping to stellar-core

`knit_to_lcl_decision` in `crates/history/src/catchup/replay.rs` is a direct port of `ApplyCheckpointWork::getNextLedgerCloseData()`:

| Case | stellar-core (`ApplyCheckpointWork.cpp`) | henyey (`replay.rs::knit_to_lcl_decision`) | Verdict |
|------|------------------------------------------|---------------------------------------------|---------|
| 1 — `seq+1 < lclSeq` → silent skip | L206-210 | `entry_seq+1 < lcl_seq` → `Ok(Skip)` | ✅ Matches; no hash check, debug log only — same as core |
| 2 — `seq+1 == lclSeq` → `entry.hash == lcl.previousLedgerHash` | L213-228 | Compares `entry.hash` vs `lcl.header.previous_ledger_hash` → `Skip`/`KnitLclPredecessorHashMismatch` | ✅ Correct field (`entry.hash`, not `entry.header.previousLedgerHash`) |
| 3 — `seq == lclSeq` → `entry.hash == lcl.hash` | L231-243 | Compares `entry.hash` vs `lcl.hash` → `Skip`/`KnitLclHashMismatch` | ✅ Matches |
| 5 — `seq != lclSeq+1` (overshoot) | L246-251 | Checked **before** case 4, mirroring core's branch order | ✅ Branch order preserved (important: if checked after case 4 prev-hash, error attribution would diverge) |
| 4 — `seq == lclSeq+1` → `header.previousLedgerHash == lcl.hash` | L254-262 | Compares `entry.header.previous_ledger_hash` vs `lcl.hash` → `Apply`/`KnitCurrentLedgerPrevHashMismatch` | ✅ Matches; correctly uses `previousLedgerHash` (distinct field from case 3) |

### Error-variant fidelity

Stellar-core throws three textually distinct `std::runtime_error`s plus the overshoot. Henyey introduces four matching fatal variants (`KnitLclPredecessorHashMismatch`, `KnitLclHashMismatch`, `KnitCurrentLedgerPrevHashMismatch`, `KnitOvershot`) and adds them to `is_fatal_catchup_failure()` — semantically equivalent. ✅

### Determinism

All five branches are pure comparisons on header fields with `saturating_add`/`saturating_sub` for overflow safety. No source of nondeterminism introduced. ✅

### Knit-prefix download window

`download_ledger_data` extends the iteration backward to `max(checkpoint_start(LCL+1), LCL-1)` so that knit entries (at LCL-1 and LCL) from the *same* checkpoint file as LCL+1 are validated, and a prior checkpoint is never fetched solely to inspect LCL-1. This matches stellar-core's behaviour: `ApplyCheckpointWork` only consults headers already in the currently-open checkpoint file (`mHdrIn`), so when LCL is at a checkpoint boundary the predecessor lives in a different file and is not checked by core either.

Note (informational, not blocking): stellar-core iterates *every* header in the checkpoint file; entries far below LCL-1 hit Case 1 and are skipped. Henyey skips those entries before reaching `knit_to_lcl_decision` (window starts at `LCL-1`). The effective behaviour is identical because Case 1 is a no-op anyway. If buffered-ledger application is added later (`stellar-core` advances LCL during apply), the window may need to widen — but `applyBufferedLedgers` is already flagged as ❌ in the spec-eval doc, so this is out of scope for #2680.

### First-apply entry virtualization

`verify_knit_to_lcl` builds a `LedgerHeaderHistoryEntry` for `apply_data[0]` by recomputing its hash via `compute_header_hash`, rather than reading the archive-stored hash. For the expected case-4 path this is inert (case 4 only uses `header.previousLedgerHash`, not `entry.hash`). If the first apply entry were mis-seq'd into case 2 or 3, the recomputed hash would be compared instead of the archive's stored hash — but downstream `verify_downloaded_data` already validates header-entry hashes against archive, so any discrepancy surfaces with the same fatal classification. Not a parity hole.

### Genesis behaviour

`test_knit_at_genesis` (synthetic pre-genesis LCL: seq=0, zero hashes) exercises case 4 for ledger 1 with `previousLedgerHash == ZERO`. This matches stellar-core's invariant that the synthetic LCL at startup carries `header.previousLedgerHash = 0` and `hash = 0`. ✅

### PARITY_STATUS / spec-eval updates

- `crates/history/PARITY_STATUS.md`: row updated to call out the 5-case matrix. Accurate.
- `docs/spec-eval/CATCHUP_SPEC_HENYEY_EVAL.md`: §11.2 row split into the explicit case-4 prev-hash check and the new "Knit-to-LCL 5-case decision matrix" row, with a precise citation to `ApplyCheckpointWork.cpp:205-262`. Accurate.

### Verdict

No parity divergences identified. APPROVE.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The 5-case decision matrix, error taxonomy, and unit-test coverage all look correct. A few non-blocking nits on doc accuracy and a defensive-coding suggestion are listed below.

<details>

### What I checked
- `knit_to_lcl_decision` logic against the 5 cases.
- Wiring in `verify_knit_to_lcl` and `download_verify_and_replay_once`.
- `download_ledger_data` signature widening and knit-prefix collection.
- 4 new `HistoryError` variants + `is_fatal_catchup_failure` classification.
- Unit-test coverage of all 5 cases (incl. mismatch branches) and edge cases.

### Correctness findings — none blocking

- **Branch order matches the doc comment.** Case 1 (skip-old) → case 2 (`seq+1 == lcl_seq`) → case 3 (`seq == lcl_seq`) → case 5 (overshoot) → case 4 (apply). The pre-case-5 ordering correctly handles `lcl_seq == 0` because the case-2 / case-3 branches catch `entry_seq == 0` and `entry_seq == 0` respectively before any wrap risk. `saturating_add`/`saturating_sub` are used where the inputs could underflow.
- **`knit_start` derivation in `download_ledger_data`** correctly clamps the backward extension to the current checkpoint file (`max(checkpoint_start(apply_start), from_ledger - 1)`). The PR description’s deviation note is right; using `checkpoint_start(from_ledger)` instead would have crossed into the previous checkpoint when LCL sits on a boundary.
- **Knit-prefix entries do not feed the chain-verify/tx-set pipeline** — the `if seq < apply_start { … continue; }` short-circuit ensures `LedgerData::new` is not called and `current_lcl` is not advanced by skipped headers. Good.
- **Hash comparisons** use `Hash256::from(...)` consistently (no raw `Hash` equality), and the case-4 prev-hash check is on `entry.header.previous_ledger_hash`, not `entry.hash`, matching the doc table. ✓
- **Error variants** are descriptive, carry expected/actual hex hashes plus ledger seq, and are all added to `is_fatal_catchup_failure`. The classification test extension covers all four. ✓
- **`download_ledger_data` signature change** is purely additive (extra tuple element + same `Vec<LedgerData>` semantics). I grepped for callers: the only one is `download_verify_and_replay_once`, which is updated in the same PR. No external breakage.
- **No new compiler warnings** introduced based on the diff (unused imports, dead code). `LedgerHeaderHistoryEntry` import is used.

### Test coverage — adequate for the unit-level matrix
All five cases are covered, plus the mismatch branches of cases 2/3/4 and the additional edge cases:

| Case | Match | Mismatch |
|------|-------|----------|
| 1 | `test_knit_case_1_skip_old` | n/a (no hash check) |
| 2 | `test_knit_case_2_lcl_predecessor_match` | `test_knit_case_2_lcl_predecessor_mismatch` |
| 3 | `test_knit_case_3_lcl_overlap_match` | `test_knit_case_3_lcl_overlap_mismatch` |
| 4 | `test_knit_case_4_apply_match` | `test_knit_case_4_apply_prev_hash_mismatch` |
| 5 | `test_knit_case_5_overshoot` | n/a |

Plus `test_knit_at_genesis`, `test_knit_lcl_at_checkpoint_boundary`, `test_knit_after_retry_advances_lcl`, and the fatal-classification extension. Boundaries (`lcl_seq == 0`, `entry_seq == lcl_seq`, `entry_seq + 1 == lcl_seq`) are all exercised.

The PR description correctly calls out that integration tests for tamper detection are deferred to a follow-up due to multi-ledger fixture cost. Acceptable for this scope; the unit tests cover the decision logic completely.

### Nits (please consider but not blocking)

1. **Stale rustdoc on `verify_knit_to_lcl`** (`crates/history/src/catchup/replay.rs`): the comment says *"Returns the apply entries that must be replayed (i.e. `Apply`-classified entries), in the same order as `apply_data`."* but the function returns `Result<()>`. Either drop that sentence or change it to "Validates that knit-prefix entries classify as Skip and the first apply entry classifies as Apply".

2. **Misleading comment in `test_knit_lcl_at_checkpoint_boundary`**: the sub-case using `lcl_seq = 127` claims *"When LCL is the LAST ledger of its checkpoint (seq == 63, 127, …), LCL-1 lives in the same checkpoint, so it IS available in the current download."* With `checkpoint_frequency = 64`, LCL=127 ends checkpoint 64–127 and LCL+1=128 starts a *new* checkpoint file (128–191), so neither LCL nor LCL-1 is in the current download’s checkpoint file in that scenario. The mathematical assertion on `knit_to_lcl_decision` is still correct, but the explanatory comment paints the wrong picture and risks confusing future readers. (The `lcl_seq = 64` sub-case comment is correct.)

3. **`debug_assert_eq!` in `verify_knit_to_lcl`** silently no-ops in release builds. If a knit-prefix entry ever classified as `Apply` (or the first apply entry as `Skip`) due to e.g. a future race between the two `header_snapshot()` calls in `download_verify_and_replay_once`, release builds would mask the contract violation. Per the repo guideline *"Never fail silently. If assumptions are not met, error out — do not gracefully recover"*, consider converting these to a real `HistoryError::CatchupFailed("internal: …")` return. Minor — under the current single-writer model the race cannot fire in practice.

4. **`knit_entries: Vec<LedgerHeaderHistoryEntry>` type annotation** in `download_ledger_data` is redundant — `Vec::new()` plus the function return type would infer it. Trivial.

No correctness issues found. Approving subject to optional nits above; happy for them to be folded into a follow-up.

</details>

tomerweller:
## Review: Waiting on CI

Reviewer verdicts:
- **Correctness:** APPROVE
- **Parity:** APPROVE

CI is still running (in progress: `Test`, `build / 2 build / 4 build (core, 8f142f046d463a644defe5aae33c245b77141730, amd64, -)`). All other 22 checks have completed successfully; no failures so far.

Re-picking this PR on the next tick.

Bounce-back count: 0/3.

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The 5-case knit-to-LCL decision matrix faithfully mirrors stellar-core `ApplyCheckpointWork::getNextLedgerCloseData()` at the pinned v26.0.1 — case order (1→2→3→5→4), per-case hash field selections, and the four dedicated fatal error variants are all correct. Unit tests cover every case (match + mismatch) plus genesis, checkpoint-boundary, and retry-advances-LCL edge cases.

<details>
<summary>Correctness findings (verified against stellar-core ApplyCheckpointWork.cpp:200-265)</summary>

### What I verified

- **Branch order** in `knit_to_lcl_decision` is case 1 → 2 → 3 → 5 → 4, matching `ApplyCheckpointWork.cpp:204-261` exactly. Case 5 must precede case 4; the implementation gets this right.
- **Field selections** match stellar-core line-for-line:
  - Case 2 (`replay.rs`): `entry.hash` vs `lcl.previous_ledger_hash` ↔ `ApplyCheckpointWork.cpp:215`.
  - Case 3: `entry.hash` vs `lcl.hash` ↔ `ApplyCheckpointWork.cpp:232`.
  - Case 4: `entry.header.previous_ledger_hash` vs `lcl.hash` ↔ `ApplyCheckpointWork.cpp:253`. ✓
- **Underflow safety:** `saturating_add(1)` / `saturating_sub(1)` on `u32` ledger seqs is correct; with `lcl_seq == 0` the case-2 branch evaluates `entry_seq + 1 == 0` which is false (1 != 0), case-3 evaluates `entry_seq == 0` (handled correctly by `test_knit_at_genesis`), then case-5 catches `entry_seq > 1`.
- **`download_ledger_data` knit_start derivation** clamps to `checkpoint_start(apply_start)` — not `checkpoint_start(from_ledger)` — so when LCL sits on a checkpoint boundary (first ledger of its checkpoint) LCL-1 is correctly excluded rather than crossing into a prior file. Stellar-core only inspects the current checkpoint file; this matches.
- **Knit-prefix entries bypass downstream pipeline:** `if seq < apply_start { … continue; }` ensures `LedgerData::new` is not called and `current_lcl` is not advanced by skip entries. The chain-verify pass only sees the apply range.
- **Distinct case-3 vs case-4 variants** (`KnitLclHashMismatch` vs `KnitCurrentLedgerPrevHashMismatch`) preserve stellar-core's two distinct error messages ("disagreed on hash" vs "disagreed on LCL hash"). Good call.
- **Fatal classification:** all four new variants are added to `is_fatal_catchup_failure`, and `test_fatal_errors_not_retriable` is extended to cover them.
- **Test matrix:** every case has a match-path test; cases 2/3/4 also have mismatch-path tests asserting the correct variant. Genesis, checkpoint-boundary (LCL at last seq and first seq of checkpoint), and retry-advances-LCL all covered as unit tests.
- **Single LCL snapshot:** `lcl_snapshot` is now taken once and reused by both `verify_knit_to_lcl` and `verify_downloaded_data`, avoiding TOCTOU between the two reads.

### Non-blocking nits (same set surfaced last round; not gating)

1. `verify_knit_to_lcl` rustdoc still claims it "returns the apply entries that must be replayed" but it returns `Result<()>`. Stale.
2. The `test_knit_lcl_at_checkpoint_boundary` comment for the `lcl_seq=127` sub-case is misleading. With `CHECKPOINT_FREQUENCY=64`, ledger 127 is the *last* of checkpoint 64–127 and ledger 128 starts a new file — in that real-world download we would fetch the 128–191 file and LCL-1=126 would NOT be in it. The mathematical assertion on the helper is still correct (it's a pure function over inputs), but the prose paints the wrong picture about which entries `download_ledger_data` would actually surface.
3. `debug_assert_eq!` in `verify_knit_to_lcl` checks "knit-prefix entries must classify as Skip" / "first apply entry must classify as Apply". These no-op in release builds. Per the repo guideline "Never fail silently", consider hardening to a `HistoryError::CatchupFailed("internal: …")` return. Under the current single-writer model the race cannot fire, so this is theoretical.
4. In `verify_knit_to_lcl`, `compute_header_hash` is invoked to populate `virtual_entry.hash` for the first apply entry, but case 4 only reads `entry.header.previous_ledger_hash`, never `.hash`. The hash computation is dead work — a `Hash256::ZERO` placeholder would behave identically. Minor.
5. `let mut knit_entries: Vec<LedgerHeaderHistoryEntry> = Vec::new();` — the explicit type annotation is inferable from the return type. Trivial.

None of these are correctness defects; all can fold into a follow-up.

</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Blast radius is contained — the public API surface is unchanged (`download_ledger_data` / `verify_knit_to_lcl` are `pub(super)`, only one in-crate caller), no extra network I/O is introduced, and the 4 new `HistoryError` variants are correctly registered as fatal. A few non-blocking risk notes below; recommend addressing the `is_hash_mismatch()` classification gap in a follow-up.

<details>
<summary>Risk notes</summary>

- **Behavior change worth calling out in release notes / PARITY_STATUS commentary.** The pre-apply pass now validates archive-supplied headers at LCL and (when in the same checkpoint as LCL+1) LCL-1 against the local LCL. Previously these headers were not inspected — only the LCL+1→target chain was checked. Scenarios that were "silently OK" before (archive contains a corrupted/divergent LCL or LCL-1 header but the apply chain from LCL+1 onward is internally consistent) will now produce a fatal `Knit*` error and abort catchup. This is the intended parity behavior with stellar-core's `ApplyCheckpointWork::getNextLedgerCloseData()`, but operators running against questionable archives may see new fatal aborts. PARITY_STATUS.md mentions the new matrix; consider one extra line in the PR description / changelog noting the externally observable regression class.
- **`is_hash_mismatch()` was not extended.** `is_fatal_catchup_failure()` correctly includes all 4 new variants, but `HistoryError::is_hash_mismatch()` (error.rs:514) was not updated even though `KnitLclPredecessorHashMismatch`, `KnitLclHashMismatch`, and `KnitCurrentLedgerPrevHashMismatch` are semantically hash-mismatches per the classifier's own docstring ("typed hash mismatch ... that indicates state divergence"). Today the only in-crate callers are test assertions, so there's no production fallout — but any future telemetry / recovery logic that uses this classifier will misclassify knit failures. Recommend adding the three hash variants to `is_hash_mismatch()` in this PR or a tight follow-up. `KnitOvershot` is a sequencing error and correctly stays out.
- **`debug_assert_eq!` on Skip / Apply is a no-op in release builds.** `verify_knit_to_lcl` calls `debug_assert_eq!(decision, Skip)` for knit-prefix entries and `debug_assert_eq!(decision, Apply)` for the first apply entry. If the `knit_start` computation in `download_ledger_data` ever drifts (e.g., off-by-one in a future refactor, or `checkpoint_start` semantics change), a release binary will silently accept a misclassified entry. Cheap hardening: replace these with hard returns (`return Err(HistoryError::CatchupFailed(...))`) — they're invariant-violation paths, not hot-loop checks.
- **Performance impact is negligible.** The iteration range widens by at most 2 ledgers per replay batch, both already inside the cached checkpoint (`checkpoint_cache` lookup, no extra `download_checkpoint_data` call beyond the one already done for the LCL+1 checkpoint). No new archive fetches in the genesis or boundary cases either (those paths short-circuit). Watcher and offline `verify-execution` flows route through the same `download_and_replay` path and inherit this small constant overhead.
- **API / data-format / config / migration:** no concerns. `download_ledger_data` signature widened from a 2-tuple to a 3-tuple but it's `pub(super)`; the sole caller (`replay.rs:240`) is updated. No on-disk format changes, no config flags, no DB schema impact.
- **Security / DoS:** the new check makes the node *stricter* about archive divergence at LCL boundaries, which is a net positive — a malicious or buggy archive that previously slipped a tampered LCL header past us now produces a fatal abort with a typed variant instead of being ignored. The integration test `test_knit_after_retry_advances_lcl` explicitly covers this attacker scenario.
- **Online watcher resilience:** because the new variants are classified fatal, watcher catchup will not retry on knit-hash mismatch (correct — same as stellar-core). Operators recovering from a stale/forked local state cannot mask the divergence by simply restarting catchup; they'll need to wipe state. This is desired behavior but worth flagging operationally.

</details>

tomerweller:
## Review: Waiting on CI

Both PR-level reviewers posted **APPROVE**:
- 🔍 Reviewer: Correctness — APPROVE
- 🔍 Reviewer: Risk — APPROVE

CI is still running. Pending checks:
- `Test` (CI)
- `build / 3 test / 2 test (...)` matrix (Quickstart, 6 jobs)

All other checks (Format, Clippy, Doctest, Build (release-ci), CodeQL, Spec-Eval Freshness, etc.) are SUCCESS. No failures so far.

Bounce-back count: 0/3.

Re-picking this PR on the next tick.

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Additive, well-scoped change. No public API, on-disk, network, config, or migration impact. Fatal classification of the new error variants matches stellar-core's `runtime_error` semantics, so no abort-vs-retry regression.

<details>

**Scope of risk surface**
- `download_ledger_data` signature changed (2-tuple → 3-tuple) but it is `pub(super)` — internal to `crates/history/src/catchup`. No external API break.
- 4 new variants added to `pub enum HistoryError`. The enum is not `#[non_exhaustive]`, but a workspace grep shows no exhaustive `match` on it outside the defining crate (only specific-variant construction in `crates/app`). Adding variants is therefore safe for the current workspace; flagged as a latent risk only if/when this crate is published.
- No changes to on-disk format, network protocol, configs, or migrations.

**Regression risk on existing catchup flows**
- New loop start is `knit_start = max(apply_ckpt_start, from_ledger - 1)`. Bound is provably ≤ apply_start, so at most 2 extra header lookups (LCL, LCL-1) per batch. Existing apply-path behavior (iterating `apply_start..=to_ledger`) is unchanged — the `if seq < apply_start { continue; }` branch sits before any state mutation, so `current_lcl`, `data`, and `last_archive_name` are not touched by knit-prefix iterations.
- Knit-prefix entries missing from the checkpoint are silently skipped (matches stellar-core, which only consults entries already in the open checkpoint file). Apply-path "header not found" error is preserved unchanged.
- Boundary case (LCL is last of prior checkpoint): `knit_start == apply_start`, no knit entries, no behavioral delta. Genesis (`from_ledger == 0`): unchanged — knit list stays empty, and case-4 prev-hash check against zero-hash LCL is consistent with current synthetic-genesis assumptions.
- Retry semantics unchanged: the four new fatal variants short-circuit retry just like stellar-core's `runtime_error` throws on the same conditions (case 2/3/4 hash mismatches, case 5 overshoot). No newly-fatal scenario that would previously have retried-and-succeeded.

**Performance**
- O(1) extra work per batch (1–2 additional header lookups in already-cached checkpoint metadata + one `compute_header_hash` for the first apply entry's virtual-entry construction). Negligible.

**Security**
- Tightens validation rather than loosens it: case-3 LCL-overlap hash check now fires before chain verification, which catches a tampered LCL-equal entry earlier (covered by `test_knit_after_retry_advances_lcl`). No new trust assumption on archive content.

**Operational**
- Hex-encoded `expected`/`actual` in error messages aid forensics. Variants are fatal, so operators will see a clean abort with a specific message instead of a generic chain-link error — improves diagnosability of archive vs local-DB divergence.

**Non-blocking observations (will be auto-filed as follow-ups if not addressed)**
- `HistoryError` is not `#[non_exhaustive]`. Consider adding it before any future external publication of `henyey-history` so new variants stay non-breaking.
- Case-4 prev-hash is now checked twice (once in `knit_to_lcl_decision`, once in downstream chain verification). Matches stellar-core's defensive ordering; the duplication is harmless but worth a comment if it ever shows up in profiling.
- The virtual-entry construction in `verify_knit_to_lcl` computes a header hash that is only used for case-2/3 branches that can never trigger for an apply entry. Slight wasted work; could be skipped by passing `entry_seq`/`prev_hash` directly instead of synthesizing a full `LedgerHeaderHistoryEntry`.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The 5-case decision matrix logic is correct, distinct fatal variants are wired through `is_fatal_catchup_failure`, and the pure-function unit tests cover all cases plus skip-old / genesis / retry / boundary. A few non-blocking nits are noted inline below.

<details>

### Correctness

- **`knit_to_lcl_decision` case ordering** (`replay.rs`) — matches stellar-core: 1 → 2 → 3 → 5 → 4. Each case selects the right field (`entry.hash` vs `entry.header.previous_ledger_hash`) and returns the case-specific fatal variant. ✅
- **`download_ledger_data` knit_start formula** — `max(checkpoint_start(apply_start), from_ledger - 1)` correctly never crosses into a prior checkpoint:
  - LCL=63 → apply_start=64, checkpoint_start(64)=64, knit_start=64 → no knit iteration (LCL is in a different file). ✅
  - LCL=64 → apply_start=65, checkpoint_start(65)=64, knit_start=max(64,63)=64 → iterates seq=64 only; LCL-1=63 (prior file) correctly skipped. ✅
  - LCL=100 → knit_start=max(64,99)=99 → iterates 99,100 → both pushed to knit_entries. ✅
  - LCL=0 (genesis) → explicit branch sets knit_start=apply_start=1, no knit iteration. ✅
- **Saturating arithmetic** on `entry_seq + 1` / `lcl_seq + 1` / `from_ledger - 1` is defensive and correct.
- **`verify_knit_to_lcl` wiring** — runs before chain verification, uses the live `LedgerManager::header_snapshot()`, surfaces fatal variants instead of generic chain-link errors. The `lcl_snapshot` binding moved up two lines is harmless. ✅
- **`is_fatal_catchup_failure`** updated for all four new variants; covered by the extended `test_fatal_errors_not_retriable` test. ✅

### Test coverage

- All 5 decision cases covered (`test_knit_case_{1..5}_*`), plus mismatch branches for cases 2/3/4. ✅
- Boundary tests: genesis (case 4 with zero LCL + overshoot at ledger 2), checkpoint boundary, retry-advances-LCL (skip-old + case-3 tamper). ✅
- `test_fatal_errors_not_retriable` extended with all four variants. ✅

Gaps (non-blocking, raised inline):
- `verify_knit_to_lcl` is not directly unit-tested; its first-apply-entry path (hash computation + `Apply` invariant) is only exercised end-to-end.
- No integration test exercising the new download knit-prefix collection path; deferral acknowledged in the PR description.

### Readability

- `KnitDecision` doc comment with the 5-case table is excellent.
- The `// Case N:` block comments in `knit_to_lcl_decision` make the stellar-core mapping obvious.
- The new comment in `download_ledger_data` explaining the knit-prefix widening is clear.

### Error handling

- Four distinct `thiserror` variants, each with descriptive `#[error("...")]` formatting that includes `expected` and `actual` hex hashes. No silent failures: cases 2/3/4 hash mismatches and case 5 overshoot all surface as fatal errors. ✅
- The "knit-prefix entry absent from archive is silently skipped" behavior is intentional parity with stellar-core (only consults what's in the file) — documented in both the function doc and the PR body. Acceptable.
- `compute_header_hash` failure path in `verify_knit_to_lcl` maps to `CatchupFailed` with the ledger seq — descriptive.

### Inline nits (non-blocking, will auto-file as follow-ups)

See inline review comments on:
1. `replay.rs` — misleading comment in `test_knit_lcl_at_checkpoint_boundary` about LCL-1 availability when LCL is the last ledger of its checkpoint.
2. `replay.rs` — `verify_knit_to_lcl` first-apply-entry hash computation is unnecessary for case 4 (which only consults `previousLedgerHash`).
3. `replay.rs` — `debug_assert_eq!` invariants in `verify_knit_to_lcl` silently degrade in release builds; consider hard errors for the "wrong bucket" case.
4. `replay.rs` — missing direct unit test for `verify_knit_to_lcl` wrapper.

None of these block merge — the underlying logic is correct.

</details>

crates/history/src/catchup/replay.rs
```diff
@@ -422,6 +593,209 @@ mod tests {
...
+                entry_seq: 2,
+                lcl_seq: 0
+            }
+        ));
+    }
+
+    #[test]
+    fn test_knit_lcl_at_checkpoint_boundary() {
+        // When LCL is the LAST ledger of its checkpoint (seq == 63, 127, ...),
+        // LCL-1 lives in the same checkpoint, so it IS available in the
```
tomerweller, line 753: Non-blocking nit: this comment is misleading. With checkpoint frequency 64, when LCL=127 it is the LAST ledger of checkpoint K=127 (range [64..127]). LCL+1=128 starts a NEW checkpoint K=191 ([128..191]), so neither LCL=127 nor LCL-1=126 actually lives in the LCL+1 download file. The pure decision function doesn't care (it has no checkpoint awareness), which is why the test passes — but a future reader will be misled. Suggest reframing the comment to say the test exercises the pure decision logic only, with a separate note that in the download path LCL-1 is only fetched when both LCL and LCL-1 share the LCL+1 checkpoint file (e.g. LCL=100, not LCL=127).

crates/history/src/catchup/replay.rs
```diff
@@ -47,6 +153,64 @@ pub(super) fn decode_upgrades_from_header(header: &LedgerHeader) -> Vec<LedgerUp
...
+            let decision = knit_to_lcl_decision(entry, lcl)?;
+            debug_assert_eq!(
+                decision,
+                KnitDecision::Skip,
+                "knit-prefix entries must classify as Skip"
+            );
+        }
+        if let Some(first) = apply_data.first() {
+            let header = first.header().clone();
+            let entry_hash = henyey_ledger::compute_header_hash(&header).map_err(|e| {
```
tomerweller, line 188: Non-blocking nit: computing the entry hash here is unnecessary. `knit_to_lcl_decision` only inspects `entry.hash` for cases 2/3 (seq < lcl+1), and the first apply entry by construction has seq == lcl+1 → case 4, which only reads `entry.header.previous_ledger_hash`. The synthesized `virtual_entry.hash` is dead data. Consider passing `Hash::default()` (or refactoring `knit_to_lcl_decision` to take `(seq, hash_opt, prev_hash, lcl)` so the apply path doesn't need a synthetic entry at all).

crates/history/src/catchup/replay.rs
```diff
@@ -47,6 +153,64 @@ pub(super) fn decode_upgrades_from_header(header: &LedgerHeader) -> Vec<LedgerUp
...
+    /// (as found in the archive checkpoint file) for ledgers in
+    /// `[knit_start, lcl_seq]`. They are validated against `lcl` and
+    /// dropped from replay; mismatches surface as the case-specific fatal
+    /// variants on [`HistoryError`].
+    ///
+    /// `apply_data` carries the per-ledger `LedgerData` for ledgers in
+    /// `[lcl_seq + 1, target]`. Its first entry is checked for cases 4 and
+    /// 5 (apply-link to LCL or overshoot). Remaining entries are validated
+    /// by the chain check downstream.
+    pub(super) fn verify_knit_to_lcl(
```
tomerweller, line 172: Non-blocking gap: `verify_knit_to_lcl` has no direct unit test. All 11 new tests target `knit_to_lcl_decision`. A small wrapper test asserting (a) a happy-path mixed knit_entries + apply_data slice returns Ok(()), (b) a tampered knit entry surfaces the right fatal variant, (c) an empty apply_data is a no-op, would lock down the wrapper behavior cheaply without needing archive fixtures.

crates/history/src/catchup/replay.rs
tomerweller, line 753: Non-blocking nit: this comment is misleading. With checkpoint frequency 64, when LCL=127 it is the LAST ledger of checkpoint K=127 (range [64..127]). LCL+1=128 starts a NEW checkpoint K=191 ([128..191]), so neither LCL=127 nor LCL-1=126 actually lives in the LCL+1 download file. The pure decision function doesn't care (it has no checkpoint awareness), which is why the test passes — but a future reader will be misled. Suggest reframing the comment to say the test exercises the pure decision logic only, with a separate note that in the download path LCL-1 is only fetched when both LCL and LCL-1 share the LCL+1 checkpoint file (e.g. LCL=100, not LCL=127).

PR #2738, opened 2026-05-15
Fast-fail UploadPlan::execute when archive is not writable

### LINKED ISSUE #2733: upload: validate archive.can_write() before mkdir loop in UploadPlan::execute
Follow-up from PR #2731 (issue #2728). Non-critical inline review comment from \`/review-pr\` that was not addressed before merge.

## Concern

\`crates/history/src/upload.rs:255\` — after #2731, \`UploadPlan::execute\` can now be called with a \`RemoteArchive\` that has \`mkdir_cmd\` configured but no \`put_cmd\`. In that case it will create one or more remote directories before \`put_file\` returns \`RemoteNotConfigured\`. The old API required a \`put_cmd\` up front, so it could not partially mutate the archive on misconfiguration.

## Fix

Validate \`archive.can_write()\` (i.e. both \`put_cmd\` and any required \`mkdir\` semantics are configured) before the mkdir loop, so misconfigured archives fail fast without side effects.

## Severity

Low — production code paths always supply a fully-configured \`RemoteArchive\`, so the partial-mutation case is unreachable in practice. Worth fixing for defense-in-depth.

tomerweller:
Closes #2733

## Summary

Adds an early `archive.can_write()` guard at the top of `UploadPlan::execute` so a partially-configured `RemoteArchive` (e.g. `mkdir_cmd` set but `put_cmd` not set) fails fast with `HistoryError::RemoteNotConfigured` before any remote `mkdir` side-effects occur. Defense-in-depth follow-up from #2731 / #2728; production paths always supply a fully-configured archive, so this is not a live bug. Gating on `can_write()` (rather than inspecting `put_cmd` directly) lets future changes to writability semantics propagate automatically.

The error message string (`"put command not configured"`) intentionally matches the message emitted by `RemoteArchive::put_file`, so the observable error on a misconfigured archive is identical to today's first-put-fails path — minus the directory side-effects.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/2733#issuecomment-4464375167)

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test -p henyey-history` passes (all 22 `upload::tests::*` green, including the new `execute_fails_fast_when_put_not_configured`)

## Deviations from plan

None.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/crates/history/src/upload.rs
+++ b/crates/history/src/upload.rs
@@ -35,7 +35,7 @@
 
 use crate::archive_state::HistoryArchiveState;
 use crate::remote_archive::RemoteArchive;
-use crate::{paths, Result};
+use crate::{paths, HistoryError, Result};
 use henyey_common::Hash256;
 use std::collections::HashSet;
 use std::path::{Path, PathBuf};
@@ -246,7 +246,19 @@ impl UploadPlan {
     /// (not [`HistoryError::RemoteCommandFailed`]) if a staged local file is
     /// missing. In production this is unreachable because staging directories
     /// always contain the files referenced by the plan.
+    ///
+    /// Fast-fails with [`HistoryError::RemoteNotConfigured`] if the archive
+    /// is not writable (no `put_cmd` configured), before creating any remote
+    /// directories. This matches the message emitted by
+    /// [`RemoteArchive::put_file`] but avoids side-effects on a partially
+    /// configured archive (e.g. `mkdir_cmd` set, `put_cmd` not set).
     pub async fn execute(&self, archive: &RemoteArchive) -> Result<()> {
+        if !archive.can_write() {
+            return Err(HistoryError::RemoteNotConfigured(
+                "put command not configured".to_string(),
+            ));
+        }
+
         let mut created_dirs: HashSet<String> = HashSet::new();
         let mkdir_enabled = archive.has_mkdir();
 
@@ -767,6 +779,29 @@ mod tests {
         assert!(
             !log_file.exists(),
             "no entry should be uploaded after the pre-check failure"
+        );
+    }
+
+    #[tokio::test]
+    async fn execute_fails_fast_when_put_not_configured() {
+        let dir = tempfile::tempdir().unwrap();
+        create_staging_dir(dir.path());
+        let mkdir_log = dir.path().join("mkdir.log");
+        let plan = UploadPlan::from_staging_dir(dir.path()).unwrap();
+
+        // mkdir is configured but put is not. The execute guard must fire
+        // before any mkdir side-effect occurs.
+        let mkdir_cmd = format!("echo {{0}} >> {}", mkdir_log.display());
+        let archive = test_archive(None, Some(&mkdir_cmd));
+        let err = plan.execute(&archive).await.expect_err("should fail");
+        assert!(
+            matches!(err, HistoryError::RemoteNotConfigured(_)),
+            "expected HistoryError::RemoteNotConfigured, got {err:?}"
+        );
+
+        assert!(
+            !mkdir_log.exists(),
+            "no mkdir command should run when put is unconfigured"
         );
     }
 }


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The guard is correct, minimal, and well-tested. `archive.can_write()` is the right predicate, `HistoryError::RemoteNotConfigured` with the exact same message as `RemoteArchive::put_file` preserves error parity, and the new test directly verifies the no-mkdir-side-effect invariant from #2733.

<details>

- **Predicate**: `RemoteArchive::can_write()` (remote_archive.rs:99) delegates to `config.is_writable()`, which is exactly the gating used inside `put_file`. Using it here makes the fast-fail semantically identical to the per-file failure that would otherwise occur after mkdir side-effects.
- **Error variant**: `HistoryError::RemoteNotConfigured("put command not configured")` matches the literal string emitted by `RemoteArchive::put_file` (remote_archive.rs:180). Callers and log scrapers that already key on this message will see no behavioral change other than earlier emission.
- **Test coverage**:
  - `execute_fails_fast_when_put_not_configured` configures `mkdir_cmd` but no `put_cmd`, then asserts (a) the returned error is `RemoteNotConfigured` and (b) `mkdir.log` was never created. This is the exact regression scenario from #2733 — the precondition that would have triggered the spurious `mkdir` in the old code.
  - The "neither put nor mkdir" sub-case is trivially covered by the same guard (no mkdir is possible regardless), so a separate test isn't needed.
  - The "put configured, mkdir not configured" case is unchanged and remains covered by existing tests via `has_mkdir()` gating in the loop body.
- **Doc comment**: The new rustdoc block on `execute` accurately describes the fast-fail and explicitly calls out the partial-config scenario, which is helpful.
- **Readability**: The guard is six lines at the top of the function and reads cleanly. No naming concerns.
- **Minor / non-blocking** (left inline-style here, not requesting changes):
  - The error string is duplicated between `put_file` and `execute`. Extracting a small helper (e.g. `RemoteArchive::ensure_writable() -> Result<()>`) would DRY this up and guarantee the messages stay in sync. Not worth blocking on.
  - Consider also short-circuiting if `self.entries.is_empty()` before the `can_write` check — current ordering means an empty plan on a read-only archive errors instead of being a no-op. This may actually be the desired behavior (misconfiguration is still misconfiguration), so flagging as a question, not a defect.

</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Defense-in-depth guard with no production behavior change. Both prod call sites (`crates/henyey/src/publish_history.rs:328` and `crates/app/src/app/publish.rs:399`) construct `RemoteArchive` with `put_cmd: Some(...)`, so `can_write()` is always true in production. The early return only fires for misconfigured archives that previously errored later via `RemoteArchive::put_file` — same `HistoryError::RemoteNotConfigured` variant and same `"put command not configured"` message, so callers pattern-matching on the error are unaffected.

<details>

- **Behavior change scope:** The new guard emits the *same* `HistoryError::RemoteNotConfigured("put command not configured")` previously raised inside `RemoteArchive::put_file` (`remote_archive.rs:180`). Callers matching on the variant/message see no difference. The only observable delta is "earlier" + "no mkdir side-effects" — strictly an improvement.
- **Production call sites checked:**
  - `crates/henyey/src/publish_history.rs:322-328` — builds config with `put_cmd: Some(archive.put.clone())`. Always writable.
  - `crates/app/src/app/publish.rs:399` — same pattern (remote archive built upstream with put set; cf. `lib.rs:661` already gates on `can_write()` before reaching the upload path via `ArchiveNotWritable`).
  No production path can hit the new early return, so this is purely defense-in-depth.
- **Callers expecting mkdir-before-error?** None found. No tests or production logic depend on the prior mkdir side-effect occurring on the `put`-missing path. The only prior test exercising this combo (mkdir set, put unset) is the newly added one.
- **Performance:** `can_write()` is `Option::is_some` on the config — O(1), branch-predictable, far cheaper than the subsequent `HashSet` allocation. Not a hot path anyway (per upload plan, not per file). No concern.
- **API / data format:** No public API signature change. Error variant and message are pre-existing. No XDR/wire/format implications.
- **Security / ops:** Reduces blast radius of a half-configured archive (e.g. operator sets `mkdir_cmd` but forgets `put_cmd`) — previously this would shell out to `mkdir_cmd` repeatedly before failing; now it fails before any side-effect. Net positive for ops; nothing operationally surprising.
- **Migrations / config:** None required. Existing valid configs unaffected.
- **Test isolation:** New test uses `tempfile::tempdir()` and writes `mkdir.log` inside it — properly isolated, no global state, no race with parallel test execution.

Non-blocking nit (not required): the duplicated `"put command not configured"` string now lives in two places (`upload.rs` and `remote_archive.rs:180`). If they ever diverge it would be confusing. Could be extracted to a `const` on `RemoteArchive`, but trivial and fine as-is.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The guard is correct, minimal, and well-tested. `archive.can_write()` is the right predicate, the error variant + message match `RemoteArchive::put_file` for parity, and the new test directly verifies the no-mkdir-side-effect invariant from #2733.

<details>

- **Predicate**: `RemoteArchive::can_write()` (remote_archive.rs:99) delegates to `config.is_writable()`, which is exactly the gate `put_file` uses. The fast-fail is semantically identical to the per-file failure that would otherwise occur after mkdir side-effects.
- **Error variant**: `HistoryError::RemoteNotConfigured("put command not configured".to_string())` matches the literal string emitted by `RemoteArchive::put_file` (remote_archive.rs:180). Callers and log scrapers keyed on this message see no behavioral change other than earlier emission and no side effects.
- **Placement**: Guard is the first statement in `execute`, before `created_dirs` allocation and the mkdir loop. Correctly fast-fails as the issue requests.
- **Test coverage**: `execute_fails_fast_when_put_not_configured` is the exact regression scenario from #2733 — configures `mkdir_cmd` only, asserts (a) error variant is `RemoteNotConfigured`, (b) `mkdir.log` was never created. Solid coverage.
- **Doc comment**: New rustdoc accurately describes the fast-fail and explicitly names the partial-config scenario. Helpful.
- **Readability / idiomatic Rust**: Six-line guard, clear and conventional. No concerns.

**Non-blocking nits** (file as follow-ups, not blocking):
- The literal `"put command not configured"` is now duplicated between `upload.rs` and `remote_archive.rs:180`. A small helper (e.g. `RemoteArchive::ensure_writable() -> Result<()>`) would DRY this and keep the messages in sync.
- Question, not a defect: ordering means an empty plan against a read-only archive errors instead of being a no-op. Erroring on misconfig regardless of plan content is probably the right call — flagging only.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The guard is correct, minimal, and well-tested. `archive.can_write()` is the right predicate, the error variant + message match `RemoteArchive::put_file` exactly, and the new test directly verifies the "no mkdir side-effects" invariant from #2733.

<details>

- **Predicate choice.** `RemoteArchive::can_write()` (`remote_archive.rs:99`) delegates to `config.is_writable()`, which is exactly the gate `put_file` uses at line 178–180. Gating `execute` on `can_write()` is semantically identical to the per-file failure that would otherwise surface after mkdir side-effects, and it future-proofs against any expansion of writability semantics (as the PR description notes).
- **Error variant + message parity.** `HistoryError::RemoteNotConfigured("put command not configured".to_string())` is byte-identical to the error raised by `RemoteArchive::put_file` at `remote_archive.rs:180`. Callers / log scrapers see no observable change other than earlier emission and absence of side effects. ✅
- **Placement.** The guard is the first statement in `execute`, before `created_dirs` is allocated and before `mkdir_enabled = archive.has_mkdir()`. This is the correct spot per the issue (#2733 explicitly: "before the mkdir loop"). No partial mutation possible.
- **Test coverage.** `execute_fails_fast_when_put_not_configured` constructs `test_archive(None, Some(&mkdir_cmd))` — i.e. the exact partial-config from #2733 — and asserts (a) `matches!(err, HistoryError::RemoteNotConfigured(_))` and (b) `!mkdir_log.exists()`. The mkdir command template (`echo {0} >> mkdir.log`) would create the log file as soon as any mkdir fired, so (b) is a real assertion that the guard short-circuits before the mkdir loop. Solid regression test.
- **Doc comment.** The new rustdoc above `execute` accurately describes the fast-fail behavior and explicitly names the partial-config scenario. Reads well.
- **Readability / idiomatic Rust.** Six-line guard, conventional `if !pred { return Err(..) }` shape. `HistoryError` import added alongside the existing `Result` re-export. No concerns.

Non-blocking nits (would be filed as follow-up issues, not blocking merge):

- The literal `"put command not configured"` is now duplicated between `upload.rs:259` and `remote_archive.rs:180`. A tiny helper (`RemoteArchive::ensure_writable() -> Result<()>`) would DRY this and guarantee the two messages stay in sync if either ever changes. Pure cleanup; not a defect.
- Minor semantic note (not a defect): with the guard at the top, an *empty* `UploadPlan` against a read-only archive now errors instead of being a no-op. That's almost certainly the right call (fail loudly on misconfig regardless of plan content), and matches the issue's "fail fast" framing — flagging only for completeness.

</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Tiny defensive guard with no realistic regression surface. Production callers always supply `put_cmd`, error variant/message are byte-identical to the post-mkdir failure path, and the only observable behavior change is skipping the mkdir side-effect on a misconfigured archive — which is the stated intent of #2733.

<details>

**Regression risk — low/none**
- `RemoteArchive::can_write()` is exactly `put_cmd.is_some()` (`remote_archive.rs:71-72,99-101`). Both production call sites of `UploadPlan::execute` (`crates/app/src/app/publish.rs:357-363` and `crates/henyey/src/publish_history.rs:319-328`) construct `RemoteArchive` with `put_cmd: Some(...)` unconditionally, so `can_write()` is always true in prod. No live behavior change for any current caller.
- For the misconfigured edge case (`mkdir_cmd: Some, put_cmd: None`) the previously-observable error was `HistoryError::RemoteNotConfigured("put command not configured")` raised by `RemoteArchive::put_file` (remote_archive.rs:179-181). New early return raises the exact same variant with the exact same string — pattern-matching consumers and log scrapers see no difference.
- One micro-regression worth naming, but harmless: a hypothetical empty `UploadPlan` against a non-writable archive used to return `Ok(())` (loop ran 0 times), now returns `RemoteNotConfigured`. `UploadPlan::from_staging_dir` always emits at least the root HAS entry, so this isn't reachable in practice, but if anyone ever constructs an empty plan in a future test, the assertion will flip. Not blocking.

**Performance** — single bool check before the loop; if anything, a tiny win because it short-circuits before any mkdir subprocess spawn on misconfig.

**API / data format** — no signature change, no error variant change, no log/string change, no config schema change. `Result<()>` shape preserved.

**Security** — none. No new subprocess, no new path/input handling, no privilege boundary touched.

**Operational concerns** — minimal.
- Operators of a misconfigured archive will see strictly *less* output (no `mkdir` invocation, no partial directory creation on the remote). That's the desired effect from #2733 and is consistent with the existing `RemoteArchive::put` precondition error.
- No metric or tracing event was removed; the only thing that disappears on the misconfig path is whatever side-effect the operator's `mkdir_cmd` had. If anyone was relying on a partial-mkdir-then-fail as a probe, that's clearly a misuse and aligns with parent issue intent.

**Test coverage for regression** — adequate.
- New test `execute_fails_fast_when_put_not_configured` exercises the `mkdir_cmd: Some / put_cmd: None` shape and asserts both the error variant and the absence of the mkdir side-effect — exactly the regression class this change is meant to lock in.
- Existing 21 `upload::tests::*` continue to use `Some(put_cmd)` so the happy path is unchanged.
- Gap (nit, non-blocking): no test for `mkdir_cmd: None / put_cmd: None`, but that's behaviorally equivalent to the covered case under the new guard, so coverage value is low.

**Minor non-blocking nits (fine as follow-ups / inline comments)**
- The literal `"put command not configured"` is now duplicated between `upload.rs` and `remote_archive.rs:179-181`. The PR description acknowledges this is intentional for message parity; a `pub(crate) const` or a small `RemoteArchive::ensure_writable() -> Result<()>` helper would keep them in lockstep automatically. Worth a small refactor issue but not blocking this PR.
- Doc comment on `execute` could mention that the fast-fail also covers the `mkdir_cmd: None / put_cmd: None` case (currently it only calls out the partial-config case explicitly).

</details>

PR #2741, opened 2026-05-15
Drop redundant hash compute for synthetic virtual_entry in knit-to-LCL apply path

### LINKED ISSUE #2736: Drop redundant hash compute for synthetic virtual_entry in knit-to-LCL apply path
Follow-up from PR #2732 (issue #2680). Non-critical inline review comment that was not addressed before merge.

## Concern

`crates/history/src/catchup/replay.rs:188` — Non-blocking nit: computing the entry hash here is unnecessary. `knit_to_lcl_decision` only inspects `entry.hash` for cases 2/3 (seq < lcl+1), and the first apply entry by construction has seq == lcl+1 → case 4, which only reads `entry.header.previous_ledger_hash`. The synthesized `virtual_entry.hash` is dead data. Consider passing `Hash::default()` (or refactoring `knit_to_lcl_decision` to take `(seq, hash_opt, prev_hash, lcl)` so the apply path doesn't need a synthetic entry at all).

## Source

[Original review comment](https://github.com/stellar-experimental/henyey/pull/2732#discussion_r3251387618) on PR #2732.

## Severity

tomerweller:
Closes #2736

## Summary

The synthetic `virtual_entry` built in `verify_knit_to_lcl` is only used to drive `knit_to_lcl_decision` through its case-4 (Apply) branch, which reads `header.previous_ledger_hash` but never the entry's own `hash` field. Computing the header hash here was wasted work (and a fallible call requiring error mapping). Replace it with a zero-hash placeholder and drop the `compute_header_hash` call along with its error branch.

## Plan reference

Triage report on the issue (trivial short-circuit / no separate plan comment): [Triage report](https://github.com/stellar-experimental/henyey/issues/2736#issuecomment-).

Origin: review comment https://github.com/stellar-experimental/henyey/pull/2732#discussion_r3251387618

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test -p henyey-history` (all targets) — green

## Deviations from plan

- Fixed one trivial pre-existing `clippy::useless_conversion` lint in `crates/history/src/verify.rs` (line 2207) that was blocking `cargo clippy -D warnings` on the touched crate. One-line change (`upper.into_iter()` → `upper`); functionally a no-op.

🤖 Generated with [GitHub Copilot CLI](https://docs.github.com/en/copilot/github-copilot-in-the-cli)

### DIFF
--- a/crates/history/src/catchup/replay.rs
+++ b/crates/history/src/catchup/replay.rs
@@ -185,14 +185,12 @@ impl CatchupManager {
         }
         if let Some(first) = apply_data.first() {
             let header = first.header().clone();
-            let entry_hash = henyey_ledger::compute_header_hash(&header).map_err(|e| {
-                HistoryError::CatchupFailed(format!(
-                    "Failed to compute header hash for ledger {}: {}",
-                    header.ledger_seq, e
-                ))
-            })?;
+            // `knit_to_lcl_decision` only reads `header.previous_ledger_hash` on
+            // the case-4 (Apply) branch exercised here, so the entry hash is
+            // unused. Use a zero placeholder rather than recomputing the header
+            // hash for a synthetic entry that we discard immediately.
             let virtual_entry = LedgerHeaderHistoryEntry {
-                hash: entry_hash.into(),
+                hash: stellar_xdr::curr::Hash([0; 32]),
                 header,
                 ext: Default::default(),
             };
--- a/crates/history/src/verify.rs
+++ b/crates/history/src/verify.rs
@@ -2204,7 +2204,7 @@ mod tests {
         // doesn't match hash(lower[126]).
         upper[0].previous_ledger_hash = stellar_xdr::curr::Hash([0xDD; 32]);
 
-        let mut headers: Vec<_> = lower.into_iter().chain(upper.into_iter()).collect();
+        let mut headers: Vec<_> = lower.into_iter().chain(upper).collect();
         // Re-compute internal chain for upper group: header 129's prev must
         // match hash of (tampered) header 128, etc. Use make_chain's approach.
         for i in 128..headers.len() {


tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Hash field on the synthetic `virtual_entry` is only read on the case 4 (Apply) branch this call site invokes; zeroing it is safe and has no API/data-format/security/operational impact. Tiny positive perf delta from skipping one SHA-256 per checkpoint apply.

<details>

- **Regressions:** None expected on the happy path. `apply_data.first()` is constructed at `lcl_seq + 1` by the surrounding pipeline, so `knit_to_lcl_decision` lands in case 5/case 4, neither of which inspects `entry.hash`. The `debug_assert_eq!(decision, Apply)` documents the invariant.
- **Edge-case nit (non-blocking):** If an upstream bug ever fed a `first` apply entry with `seq <= lcl_seq`, the old code would surface a meaningful `KnitLclHashMismatch` / `KnitLclPredecessorHashMismatch` against the real hash; the new code would surface the same error variant but against a zero hash, which could be slightly more confusing in logs. The `debug_assert` catches it in debug; release builds would still error out (no silent corruption) just with a less helpful actual-hash string. Not worth blocking on — the invariant is a precondition of this code path.
- **Performance:** Strictly positive (one fewer `compute_header_hash` per `verify_knit_to_lcl` call). Negligible in absolute terms.
- **API / data format:** No public surface change. `virtual_entry` is a local stack value never persisted or returned.
- **Security:** No integrity property depends on this hash — the meaningful chain check is `entry.header.previous_ledger_hash == lcl.hash` in case 4, which is unchanged. The unrelated `clippy::useless_conversion` fix in `verify.rs` is a no-op test-code cleanup.
- **Operational:** No config, migration, or schema impact.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Claim verified — `knit_to_lcl_decision`'s case-4 (Apply) branch reads only `entry.header.previous_ledger_hash`, never `entry.hash`, so the zero placeholder is sound. Clippy fix is a true no-op.

<details>

- **Claim verification.** Read `knit_to_lcl_decision` at `crates/history/src/catchup/replay.rs:58-115`. The five cases use:
  - Case 1 (`entry_seq + 1 < lcl_seq`) — no hash read at all (skip-old).
  - Case 2 (`entry_seq + 1 == lcl_seq`) — reads `entry.hash`.
  - Case 3 (`entry_seq == lcl_seq`) — reads `entry.hash`.
  - Case 5 (overshoot) — no hash read.
  - Case 4 (`entry_seq == lcl_seq + 1`) — reads only `entry.header.previous_ledger_hash`.
  Since `apply_data.first()` by construction has `ledger_seq == lcl_seq + 1` (it's the first entry in `[lcl_seq+1, target]`, validated upstream), only case 4 is reachable, and the synthetic entry's `hash` field is genuinely dead on this path. Claim is correct.

- **Dropped error branch.** The removed `compute_header_hash` call could in principle fail (it returns a `Result`), but the only way it would fail is if XDR serialization of a freshly-cloned `LedgerHeader` failed, which is essentially unreachable. Dropping the `HistoryError::CatchupFailed("Failed to compute header hash …")` branch is fine; if anything, it removes a misleading error message (the hash wasn't actually needed). The remaining `?` on `knit_to_lcl_decision` preserves real-failure surfacing.

- **Degraded error reporting in a "can't happen" path.** A minor observation, not blocking: if upstream invariants ever drifted and an `apply_data.first()` entry landed at `seq <= lcl_seq` (cases 2/3), the resulting `KnitLclPredecessorHashMismatch` / `KnitLclHashMismatch` would now report `actual = 00…00` instead of the real header hash. In debug builds the `debug_assert_eq!(decision, KnitDecision::Apply, …)` catches this; in release it's just a misleading-but-still-fatal error. Acceptable given the invariant, but a one-line comment noting that the placeholder relies on the upstream "first apply entry has seq == lcl+1" guarantee would harden the rationale. (Inline suggestion below.)

- **Test coverage.** `knit_to_lcl_decision` itself has thorough per-case unit tests at lines ~640–800 in the same file (Apply, Skip, all error variants). `verify_knit_to_lcl` is exercised transitively via the catchup-replay integration paths — there is no direct unit test for the wrapper, but the changed code is a strict simplification (zero placeholder replacing a value that's never read), so existing `cargo test -p henyey-history` coverage suffices. No new test is required by this diff.

- **Readability.** The new inline comment explaining *why* the zero hash is sound is clear and points at the exact case it relies on. Net negative LOC, no error-handling complexity left to maintain.

- **Piggy-backed clippy fix (`verify.rs:2207`).** `Iterator::chain` accepts `IntoIterator`, so `chain(upper.into_iter())` and `chain(upper)` produce byte-identical iterators. The `upper` binding (a `Vec<_>`) is consumed in both forms. True no-op, confined to test code. No concern.

- **Nit (optional, non-blocking).** Could add `debug_assert_eq!(header.ledger_seq, lcl.header.ledger_seq + 1)` before constructing `virtual_entry` to make the "only case 4 is reachable" invariant explicit at the call site. Pure documentation/defense-in-depth; happy to file as a follow-up.

</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Zero-hash placeholder is safe — `knit_to_lcl_decision` on the case-4 (Apply) branch reads only `entry.header.previous_ledger_hash`, never `entry.hash`. The synthetic `virtual_entry` is local, immediately discarded, and never logged, returned, or persisted. Clippy fix in tests is risk-zero.

<details>

- **Regressions:** None. Traced `knit_to_lcl_decision` (replay.rs:58–117): case 1 short-circuits on seq, case 2/3 read `entry.hash` but require `entry_seq <= lcl_seq` — unreachable for `apply_data.first()` whose seq is `lcl_seq + 1`; case 5 short-circuits on seq; case 4 (the only reachable branch here) reads `entry.header.previous_ledger_hash` exclusively. The zero `entry.hash` can never be observed.
- **Observability:** `knit_to_lcl_decision`'s debug logs only emit `entry_seq` / `lcl_seq` / hex of `previous_ledger_hash` mismatches, never `entry.hash`. The info log in `verify_knit_to_lcl` reports counts only. No operator-visible log line is degraded by the zero hash. `virtual_entry` is dropped at end of scope — not returned, not stored, not surfaced in any `HistoryError` variant.
- **Breaking changes / data formats:** None. `virtual_entry` is a stack-local; no API, on-disk, or wire format changes.
- **Performance:** Marginally positive — drops one SHA-256 header hash per apply batch on the catchup hot path. Negligible in absolute terms.
- **Security:** Negligible. The zero hash is never compared, persisted, or trusted; it's a typed `Hash::default()` filler for an unread field.
- **Operational concerns:** None. No config, migration, metric, or schema impact.
- **Design observation (non-blocking, follow-up worthy):** Stuffing `Hash::default()` into a field with semantically meaningful values is a footgun — a future refactor of `knit_to_lcl_decision` that grows a new branch reading `entry.hash` on the apply path would silently consume a zero. Worth considering either (a) a local non-XDR struct mirroring just the (header, decision-relevant) fields, or (b) constructing the synthetic entry only behind a helper that documents the invariant. Not a blocker — the `debug_assert_eq!(decision, Apply)` immediately after the call provides some defense-in-depth, and the field is unused by every reachable branch today.
- **Clippy fix in `verify.rs`:** Risk-zero. Dropping a no-op `.into_iter()` on a value already being moved into `Chain` is purely cosmetic and confined to test code.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The dead-data simplification is correct: `knit_to_lcl_decision` only reads `entry.hash` in cases 2/3 (seq ≤ lcl), and the synthetic `virtual_entry` here is `apply_data.first()` whose seq is lcl+1 (case 4), so the placeholder hash is never consulted. The `verify.rs` clippy fix is a trivial no-op.

<details>

- **Correctness of replay.rs change:** Confirmed by reading `knit_to_lcl_decision` (replay.rs:58–117). Only case 2 (`entry_seq+1 == lcl_seq`) and case 3 (`entry_seq == lcl_seq`) read `entry.hash`. The synthetic entry is `apply_data.first()`, which by upstream invariant has `header.ledger_seq == lcl_seq + 1` → falls through to case 4 (or case 5 overshoot, which only inspects seqs). Cases 1–3 cannot fire for a well-formed apply prefix, so `Hash::default()` is provably dead. The `debug_assert_eq!(decision, KnitDecision::Apply, …)` would catch any drift in debug builds.
- **Defensive consideration (non-blocking):** If an upstream bug ever fed `apply_data` whose first entry had seq ≤ lcl, the zero hash would surface as a `KnitLclPredecessorHashMismatch`/`KnitLclHashMismatch` rather than the underlying seq-ordering bug. This is no worse than before (the same wrong-seq input would also have failed, just with a real hash) and the debug_assert protects test builds. Comment in the diff makes the assumption explicit. Fine to leave as is.
- **Error handling:** Removing the `compute_header_hash` call also removes a `CatchupFailed` error path that, on the apply branch, could never have surfaced (compute_header_hash on a well-formed header is infallible in practice, and the value was unused anyway). Net simplification.
- **Readability:** The comment block clearly justifies the placeholder; reads well. No magic number — `Hash([0; 32])` is unambiguous.
- **verify.rs clippy fix:** `chain(upper.into_iter())` → `chain(upper)`. Standard `useless_conversion` lint fix; behavior unchanged since `IntoIterator for Vec` yields the same iterator.
- **Test coverage:** `knit_to_lcl_decision` case-4 path is directly tested (`test_knit_case_4_apply_match`, `test_knit_case_4_apply_prev_hash_mismatch`, `test_knit_at_genesis`, `test_knit_lcl_at_checkpoint_boundary`, `test_knit_after_retry_advances_lcl`). `verify_knit_to_lcl` itself is exercised end-to-end via the catchup pipeline at replay.rs:413 and the integration suite from #2732. The behavior change here (replacing a value that is never read) is not directly observable, so no new test is required — a regression that re-introduced a read of `entry.hash` for the apply branch would be caught by the existing case-2/3/4 unit tests around `knit_to_lcl_decision`.
- **Build:** `cargo check -p henyey-history` clean.
- **Parity:** No observable behavior change. No PARITY_STATUS impact.

</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** The synthetic `virtual_entry` is a stack-local, never persisted, logged, or returned; replacing its hash with a zero placeholder carries no observable regression, security, or operational risk.

<details>

- **Scope of the placeholder:** `virtual_entry` is constructed inside `verify_knit_to_lcl`, passed by reference to `knit_to_lcl_decision`, and dropped at end of scope. It is not stored, returned, serialized, logged, or used to derive any downstream state. The zero hash cannot leak to a caller, RPC, metric, log line, or on-disk artifact.
- **Decision-matrix safety:** For `apply_data.first()` the entry seq is `lcl_seq + 1` by construction, so only the case-4 branch (which reads `entry.header.previous_ledger_hash`, not `entry.hash`) is exercised. If the invariant were ever violated, the zero hash would *fail* case 2/3 comparisons (mismatch error) rather than silently pass — i.e. the change is strictly more defensive than the prior real-hash path, not less. No "unverified data accepted as verified" path is opened.
- **Performance:** Drops one SHA-256 over a `LedgerHeader` per catchup invocation. Negligible savings, but the claim is real and not masking work shifted elsewhere — the removed `compute_header_hash` call had no other side effects (pure function returning `Hash256`).
- **API / format / on-disk:** No public API, XDR, schema, config, or metric is touched. The struct shape of `LedgerHeaderHistoryEntry` is unchanged.
- **Operational / log hygiene:** The `debug!`/`info!` lines in `knit_to_lcl_decision` and `verify_knit_to_lcl` log seq numbers and counts, never the entry hash itself, so an operator will not see a confusing `0000…0000` hash in logs.
- **verify.rs change:** Cosmetic clippy fix (`.chain(upper.into_iter())` → `.chain(upper)`) in a test module. Zero risk.

No blocking concerns; no follow-ups warranted.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The claim holds — `knit_to_lcl_decision`'s case-4 (Apply) branch reads only `entry.header.previous_ledger_hash`, so a zero placeholder for the synthetic entry's `hash` is observationally equivalent on the happy path. The `verify.rs` hunk is a harmless clippy `useless_conversion` cleanup.

<details>

- **Logic verified.** For `apply_data.first()` with `entry_seq == lcl_seq + 1`, control flow in `knit_to_lcl_decision` skips cases 1/2/3 and the case-5 overshoot check, lands in case 4, and consults only `entry.header.previous_ledger_hash`. The returned `KnitDecision::Apply` is then pinned by the existing `debug_assert_eq!`. The dropped `compute_header_hash` call was indeed dead work, and the error-mapping arm it required is also correctly removed.
- **Silent behavioral changes — bounded and acceptable.** If a caller ever violated the precondition and passed an apply-first entry with `entry_seq <= lcl_seq` (or `+1 == lcl_seq`), the old code would surface `KnitLclHashMismatch` / `KnitLclPredecessorHashMismatch` with the *real* header hash in the `actual` field; the new code reports all-zeros instead. Same error variant, same control flow, only the diagnostic value differs. In release builds the `debug_assert_eq!` is gone, so this is the only observable delta, and it is well within "error string drift" tolerance.
- **No downstream consumer of `virtual_entry.hash`.** The entry is constructed inline, passed to `knit_to_lcl_decision`, and dropped at end of scope. Not stored, not logged, not hashed-of-hashed.
- **Test coverage.** No direct test for `verify_knit_to_lcl`, but the property being relied on is a direct consequence of `knit_to_lcl_decision`'s case-4 implementation, which is exhaustively covered in `replay.rs` tests (lines ~640–794, including the apply-prev-hash mismatch path). A dedicated regression test would be nice-to-have but isn't required for a code-deletion of this size — `/review-pr` may file it as a follow-up if desired.
- **Readability.** The new comment clearly explains *why* the zero placeholder is safe and points at the specific case-4 branch. Good.
- **Error handling.** Removing the `CatchupFailed` mapping is correct — the call it wrapped is gone. No new failure modes introduced.
- **verify.rs hunk.** Standard `useless_conversion` cleanup; behavior identical.

</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Pure dead-work removal. The synthetic `virtual_entry` is local-scope only and its `hash` field is provably unread on the case-4 (Apply) branch this call site exercises. No API, on-disk, XDR, log, metric, config, or determinism surface is touched.

### Regression risk — none material
- `virtual_entry` is constructed inside `verify_knit_to_lcl`, passed by reference to `knit_to_lcl_decision`, and dropped at end of scope. `grep` confirms no other consumers (`virtual_entry` appears only at the construction and the single call site; the function returns `Result<()>`, not the entry).
- `knit_to_lcl_decision` case 4 (`entry.seq == lcl.seq + 1`) reads `entry.header.previous_ledger_hash` only. The invariant for the call site — `apply_data.first().header.ledger_seq == lcl_seq + 1` — is enforced upstream (`apply_data` covers `[lcl_seq+1, target]`) and pinned by the existing `debug_assert_eq!(decision, KnitDecision::Apply, …)`. Linkage validation against `lcl.hash` is unchanged.
- Cases 2/3 (which *do* read `entry.hash`) are unreachable from this construction path. Only relevant if the upstream invariant is ever violated — see "soft concerns" below.

### Performance — confirmed real, not just cosmetic
- Removes one SHA-256 of the header per `verify_knit_to_lcl` invocation (one per catchup cycle), plus its error-mapping arm. Tiny absolute win, but the PR's claim that this is unused work is accurate; it does not move work elsewhere.

### Breaking changes — none
- No public API change (`verify_knit_to_lcl` is `pub(super)`; `virtual_entry` was never exposed).
- No XDR / on-disk / wire format touched.
- Logging unchanged (`"Knit-to-LCL decision matrix validated"` line preserved).
- No new metrics, no removed metrics.
- Error variants unchanged. `HistoryError::CatchupFailed("Failed to compute header hash …")` is removed, but it was only producible from the now-deleted call; no test or external consumer depends on that string (grep clean for the message).

### Security — placeholder is contained
- `Hash([0; 32])` never leaves the function, is never persisted, never logged, never hashed-of-hashed, never returned. Cannot be mistaken for a real hash by any downstream code because there is no downstream code.
- No risk of "zero-hash trust" creeping in: future callers adding consumers of `virtual_entry.hash` would have to do so in this same function body, where the inline comment explicitly flags the placeholder.

### Operational / replay determinism / parity
- The knit-to-LCL decision matrix's *observable* outputs (decision enum, error variants on mismatch, log line) are byte-identical for all valid inputs. stellar-core parity (`ApplyCheckpointWork.cpp:246` ordering) is preserved — only the synthetic-entry construction in our wrapper changed.
- `crates/history` is outside the strict parity prefix list, but I verified nothing in this hunk affects ledger hashes, bucket state, or replay outputs. Safe.

### Soft concern (non-blocking, will auto-convert to follow-up if filed inline)
- If the upstream invariant `apply_data.first().seq == lcl_seq + 1` is ever violated in release builds (where `debug_assert_eq!` is compiled out), the error message for cases 2/3 will report `actual: 0000…0000` instead of the real header hash. Same error *variant*, same control flow — only diagnostic value degrades. Reviewer A noted the same. I'm not filing this inline because: (a) the precondition is structurally guaranteed by how `apply_data` is constructed at line 413, (b) the only realistic failure mode is a logic bug elsewhere, in which case the wrong-hash-in-error is the least of the operator's problems, and (c) the `debug_assert_eq!` will catch it in dev/test.

### `verify.rs` hunk
- `upper.into_iter()` → `upper`: `Iterator::chain` takes `IntoIterator`, so passing the `Vec` directly is byte-equivalent to calling `.into_iter()` on it. Pure clippy `useless_conversion` cleanup. The test (`test_verify_reverse_walk_cross_checkpoint_link_failure` per surrounding context) still tampers `upper[0].previous_ledger_hash`, still re-chains, and still asserts the same expected error. What's tested is unchanged.

### Verdict
APPROVE — ship it.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The refactor is a true no-op on the apply path: the synthetic `virtual_entry.hash` is unreachable in `knit_to_lcl_decision` case 4. The clippy fix in `verify.rs` is also a semantic no-op. Existing tests adequately cover the relevant branches.

### Correctness

- **`entry.hash` reads in `knit_to_lcl_decision`** (replay.rs:58–117): traced all uses.
  - Case 1 (entry_seq+1 < lcl_seq): no hash read.
  - Case 2 (entry_seq+1 == lcl_seq): reads `entry.hash`.
  - Case 3 (entry_seq == lcl_seq): reads `entry.hash`.
  - Case 5 (entry_seq > lcl_seq+1): no hash read.
  - Case 4 (entry_seq == lcl_seq+1): reads only `entry.header.previous_ledger_hash`.
  The synthetic entry is built from `apply_data.first()`, whose contract (per docstring at line 168–171) guarantees `seq ∈ [lcl_seq+1, target]`, i.e. case 4 or case 5. Neither touches `entry.hash`. ✅
- **Leak surface for the zero placeholder**: the synthetic entry is fully local — passed by `&` to `knit_to_lcl_decision`, then dropped at the end of the `if let` block. It is never logged, returned, persisted, embedded in an error, or copied to ledger state. The case-4 error path (`KnitCurrentLedgerPrevHashMismatch`) only quotes `entry.header.previous_ledger_hash` and `lcl.hash`, so the zero hash cannot appear in any user-visible message. ✅
- **Removed error mapping**: the only failure mode dropped is `compute_header_hash`'s own error (XDR serialization failure of a freshly-cloned header that immediately gets passed back to downstream code that *will* hash it during apply). Removing this is fine — any genuine hashing failure resurfaces downstream rather than being prematurely caught here, and the error path was already a `CatchupFailed(format!(...))` rather than a distinct semantic variant. No failure mode is lost. ✅
- **Behavioral edge case (not blocking)**: if the contract on `apply_data.first()` were ever violated (e.g. seq == lcl_seq), the old code would have computed the real hash and possibly hit `debug_assert_eq!(decision, Apply)` (panic in debug, silent skip in release). The new code returns `KnitLclHashMismatch` instead. This is arguably *safer* (no silent-skip release-mode footgun) and the `debug_assert_eq!` still catches non-Apply cases when reached. Worth a mention but no action needed.

### `verify.rs` clippy fix

`lower.into_iter().chain(upper.into_iter())` → `lower.into_iter().chain(upper)`. `Iterator::chain` takes `IntoIterator<Item=Self::Item>` and `Vec<T>: IntoIterator<Item=T>`, so the two forms are byte-equivalent. Confirmed `cargo clippy -D warnings` on main reproduces the lint at that exact line — this was a pre-existing nit, correctly disclosed in the PR description as a deviation. ✅

### Test coverage

- `knit_to_lcl_decision` itself is exhaustively unit-tested across all 5 cases plus boundary/genesis/retry scenarios (replay.rs:636–800).
- `verify_knit_to_lcl` has no direct unit test, but it's invoked from `download_replay_apply_with_progress` (line 413) and exercised by the integration tests for catchup. Since the change is a pure no-op on the only reachable branch (case 4), existing coverage is sufficient. No new tests required. ✅

### Readability & error handling

- Replacement comment clearly explains *why* the placeholder is safe and points the reader at the case-4 contract.
- Diff is minimal (5+/7−); no dead imports left behind (verified `compute_header_hash` is still used elsewhere in replay.rs via the unit tests at line 793 — and is also re-imported in verify.rs/tests, unaffected here).
- No unwrap/expect introduced; no panic added.

### Nits (non-blocking, will be auto-filed at merge if any inline comments are left — none added)

None worth filing.

</details>

tomerweller:

**Reviewer A (Correctness):** APPROVE — refactor verified as no-op on the apply path; synthetic entry's `hash` is never read on case-4 (Apply) branch; clippy fix is semantically equivalent.

**Reviewer B (Risk):** APPROVE — synthetic `virtual_entry` is local-scope only, no API/format/log/metric surface touched, parity & determinism unaffected.

**CI:** red (unrelated)

The `Test` job in workflow `CI` (run [25946394786](https://github.com/stellar-experimental/henyey/actions/runs/25946394786/job/76275283319)) was cancelled at 00:02:55Z after ~30 min wall-clock because an unrelated test hung for >12 minutes:

```
2026-05-15T23:50:52Z test tx_set_tracker::tests::test_request_at_cap_still_increments_existing has been running for over 60 seconds
2026-05-16T00:02:55Z ##[error]The operation was canceled.
```

`tx_set_tracker` lives outside the diff (PR only touches `crates/history/src/catchup/replay.rs` and `crates/history/src/verify.rs`), so this failure is not diff-attributable — it's a hanging test in `henyey-herder` (or wherever `tx_set_tracker` lives) that timed out the job.

Routing back to `ready-for-doing` for `/do` Mode B to rebase on `origin/main` and re-run CI. If the same unrelated failure reappears after rebase, the next `/review-pr` will mark `blocked` and the hang must be investigated as a separate issue.

tomerweller:
Rebased onto origin/main (4413ebfc) to clear the cancelled `Test` CI run. No code changes; re-running CI. Ready for re-review.

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Risk surface is minimal — the synthetic `virtual_entry` is a stack-local value consumed and dropped within `verify_knit_to_lcl`, its `hash` field is not persisted, logged, or compared on the case-4 Apply path, and removing the hash compute is a strict win.

<details>

**Scope check**
- Diff touches only `crates/history/src/catchup/replay.rs` (production) and a cosmetic `into_iter()` cleanup in `crates/history/src/verify.rs` tests. Both files are in `henyey-history`, a non-parity-critical crate, so the Risk lens applies.

**Regression risk: low**
- `virtual_entry` is constructed locally and passed by reference only to `knit_to_lcl_decision`. `grep` for `virtual_entry` confirms a single producer/consumer (replay.rs:194/199). No other code path reads it.
- On the apply path, `apply_data.first()` has `seq == lcl_seq + 1` by construction (validated upstream in §9 chain verification and bracket checks). `knit_to_lcl_decision` for that input falls through cases 1/2/3/5 and only touches `entry.header.previous_ledger_hash` in case 4 — `entry.hash` is genuinely dead.
- The `debug_assert_eq!(decision, KnitDecision::Apply, …)` continues to defend against any misclassification in debug builds.

**Persistence / serialization / API risk: none**
- `LedgerHeaderHistoryEntry` is an XDR type, but this particular instance is never serialized, hashed, indexed, stored, or sent over the wire — it exists only as the function-call shape required by `knit_to_lcl_decision`.
- No public API change. `knit_to_lcl_decision` and `KnitDecision` remain `pub(super)` with identical signatures and semantics.
- No on-disk format, metric, log line, or error code is removed or renamed. (The removed `format!("Failed to compute header hash …")` error message was unreachable in the steady state — `compute_header_hash` doesn't fail for well-formed headers — so dropping it is harmless.)

**Performance: net positive**
- Removes one SHA-256 over a serialized `LedgerHeaderHistoryEntry` per catchup batch. Not a hot loop, but it's pure deletion: no new work is introduced elsewhere. No hidden invariant (e.g. caching, log audit trail) depended on the compute happening as a side effect.

**Security: n/a**
- The zero hash never escapes the function. It cannot be used to forge any check because `knit_to_lcl_decision` ignores it on this branch, and the real authority on the prev-hash linkage is `entry.header.previous_ledger_hash`, which is still compared against `lcl.hash` in case 4.

**Operational: none**
- No config, no metric, no log line, no error code surface change visible to operators. Existing `info!("Knit-to-LCL decision matrix validated")` and case-specific `debug!` lines are unchanged.

**"Hidden invariant" / future-proofing — one observation, non-blocking**
- If a future refactor ever fed a non-`Apply` entry into this path (i.e. broke the upstream `seq == lcl_seq + 1` precondition), `knit_to_lcl_decision` would compare the placeholder `[0; 32]` against `lcl.previous_ledger_hash` (case 2) or `lcl.hash` (case 3) and emit a `KnitLclPredecessorHashMismatch` / `KnitLclHashMismatch` whose `actual: 0x0000…` is misleading for debugging. In debug builds the `debug_assert_eq!` catches this loudly; in release it would surface as a confusing hash-mismatch error rather than a clean "preconditions violated" panic.
  - The inline comment added in the diff documents the contract well, which mitigates this. The longer-term refactor suggested in #2736 (pass `(seq, prev_hash, lcl)` and drop the synthetic-entry shape entirely) would eliminate the foot-gun outright, but is explicitly out of scope here.

**Test coverage**
- No new tests, but no behavior change at the function boundary: `knit_to_lcl_decision` on a case-4 input is identical with `hash = real` vs `hash = [0;32]`. Existing unit tests at replay.rs:640+ cover all five cases with real hashes. Acceptable.

**Verdict rationale**
- Single-line semantic change, narrowly scoped, no observable side effects, well-commented. Approve.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The zero-hash placeholder is logically correct: for case 4 (the only branch this call site can reach), `knit_to_lcl_decision` reads only `entry.header.previous_ledger_hash` and `entry.header.ledger_seq`, never `entry.hash`. The synthetic `virtual_entry` is locally scoped and dropped immediately, with no downstream reader of its `hash` field. The verify.rs clippy fix is benign.

<details>

**Logical correctness of the apply-branch zero-hash:**
- `knit_to_lcl_decision` (replay.rs:58–117) consults `entry.hash` only in cases 2 (`entry_seq + 1 == lcl_seq`) and 3 (`entry_seq == lcl_seq`). Case 4 (`entry_seq == lcl_seq + 1`) reads only `entry.header.previous_ledger_hash`. Case 5 reads only sequence numbers.
- By construction, `apply_data[0]` is the first ledger to apply, i.e. `lcl_seq + 1`. Cases 1/2/3 are unreachable for it, so the synthetic `entry.hash` is provably unread along the only reachable path. The `debug_assert_eq!(decision, KnitDecision::Apply, ...)` at line 200–204 enforces this invariant in debug builds.
- `virtual_entry` is scoped to the `if let Some(first) = apply_data.first()` block (replay.rs:186–205) and never returned, stored, logged, persisted, or used for meta emission. The only consumer is the `knit_to_lcl_decision` call on line 199. No regression risk from downstream readers.

**Subtle defensive concern (non-blocking nit):**
- Previously (with a correctly computed hash), if `apply_data[0].seq` were ever malformed (== `lcl_seq` or `lcl_seq - 1` due to a caller bug), the call would have returned a meaningful Skip or a real hash-mismatch error referencing the actual computed hex. With the zero placeholder, the same malformed input now surfaces as `KnitLclHashMismatch` / `KnitLclPredecessorHashMismatch` with a `0000...` actual hash — slightly less diagnostic but still flagged as a fatal error, and the `debug_assert!` catches it during testing. This is a programmer-error scenario, not a data-integrity scenario, so the change is acceptable.

**Test coverage:**
- The `knit_to_lcl_decision` matrix (cases 1–5, boundary, genesis, retry-advanced LCL) is thoroughly unit-tested directly (replay.rs:635–797), including `test_knit_case_4_apply_match`. The case 4 branch which is the sole consumer of the synthetic entry's fields is well covered.
- `verify_knit_to_lcl` itself has no direct unit test, but it's exercised via the catchup integration tests through `download_verify_and_replay_once` (replay.rs:413). The PR does not add new tests, which is appropriate — this is a refactor that preserves identical observable behavior on the supported path, and the existing case-4 tests via `knit_to_lcl_decision` validate the predicate the synthetic entry feeds into.

**Other:**
- Comment block on lines 187–191 accurately documents the assumption justifying the zero placeholder. Good.
- The clippy fix in `verify.rs:2207` (`upper.into_iter()` → `upper`) is correct: `chain` takes anything implementing `IntoIterator`, so the explicit call was redundant. No behavior change.
- No dead code introduced; the removal of `compute_header_hash` and the `HistoryError::CatchupFailed` mapping eliminates an error path that was only ever reachable on a fundamental hashing failure — fine to drop since that path was never observably hit.
- Naming/readability: leaving the variable name `virtual_entry` and keeping the `LedgerHeaderHistoryEntry` construction (rather than refactoring `knit_to_lcl_decision`'s signature, as the issue suggested) is a reasonable minimal-scope choice. The issue explicitly listed the signature refactor as optional.

**Build sanity:** Diff is two small mechanical edits in well-trodden code paths; no API change, no surface change to callers.

</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Two-line surgical refactor with no API, data-format, config, or migration impact. Synthetic `virtual_entry` is locally scoped and `.hash` is provably unread on the only reachable code path; risk surface is limited to a marginally less informative error message in a programmer-error scenario.

<details>
<summary>Full risk review</summary>

### Regression risk: low

- The synthetic `virtual_entry` is constructed inside an `if let Some(first) = apply_data.first()` block (replay.rs:186–205), passed to `knit_to_lcl_decision`, and dropped at the end of the block. It is never returned, stored, persisted, logged, hashed into a chain, or used for meta emission. Grep confirms the only consumer is the local `knit_to_lcl_decision` call.
- `knit_to_lcl_decision` reads `entry.hash` only on the case-2 / case-3 branches (`entry_seq == lcl_seq` or `lcl_seq - 1`). Case 4 (Apply), which is the only reachable case for `apply_data[0]` (whose seq is by construction `lcl_seq + 1`), reads only `header.previous_ledger_hash` and `header.ledger_seq`. The `debug_assert_eq!(decision, KnitDecision::Apply, ...)` on line 200 enforces this invariant under test.
- Therefore the zero-hash placeholder cannot change observable behavior on the supported path.

### Performance: positive

- Removes one Ed25519/SHA-256-equivalent header serialization + hash per catchup apply window (one per `verify_knit_to_lcl` call), plus an error-mapping closure allocation. Tiny but strictly beneficial.

### API / data format / wire compatibility: none

- No public signatures change. `verify_knit_to_lcl` still returns `Result<(), HistoryError>`. The `HistoryError::CatchupFailed("Failed to compute header hash …")` variant string is removed at this one call site; the variant itself is unchanged. No serialized output of any kind is affected (the entry is in-memory only).

### Security: none

- Zero hash is fed into a predicate that does not consult it on the reachable branch. No chain verification, signature, or persisted artifact uses this value. The actual incoming-batch chain validation (`previous_ledger_hash` linkage) is performed elsewhere and is unaffected.

### Operational concerns: none

- No config, migration, schema, log format, or metric changes. The removed log/error path was only reachable on a fundamental hashing-library failure, which has never been observably hit.

### Subtle non-blocking concerns

1. **Diagnostic regression on programmer error (very minor).** If a future bug ever caused `apply_data[0].seq` to be something other than `lcl_seq + 1` (e.g. `lcl_seq` or `lcl_seq - 1`), the resulting `KnitLclHashMismatch` / `KnitLclPredecessorHashMismatch` error will now print `actual=0000…0000` instead of the real header hash. This is strictly a debugging-quality regression in a "should be impossible" branch already guarded by `debug_assert_eq!`; not a correctness issue. Acceptable.
2. **Clippy fix scope.** The `upper.into_iter()` → `upper` change in verify.rs:2207 is a test-only no-op (`chain` takes `IntoIterator`). Pulling an unrelated lint fix into a behavioral PR is mildly off-policy, but it's a single line in `#[cfg(test)]` code with no runtime impact, and the PR description discloses it. Not worth bouncing.
3. **Defense-in-depth alternative (not requested).** A more defensive variant would be to refactor `knit_to_lcl_decision` to take only the fields it actually uses on the Apply path (e.g. `(prev_hash, seq)`), removing the temptation to construct partial `LedgerHeaderHistoryEntry` values entirely. The issue called this out as optional; deferring is fine.

### CI / build

- Author reports `cargo fmt --check`, `cargo clippy -p henyey-history --all-targets -- -D warnings`, and `cargo test -p henyey-history` all green. Diff is mechanical and the surrounding case-4 branch is well-covered by `test_knit_case_4_apply_match` and the broader knit_to_lcl_decision matrix in replay.rs:635–797.

No blocking issues. APPROVE.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Diff is a sound dead-computation removal. `knit_to_lcl_decision` only reads `entry.hash` in cases 2/3 (seq ≤ lcl_seq); the synthetic `virtual_entry` is built from the first apply entry, which by construction has `seq == lcl_seq + 1` → case 4, which reads only `header.previous_ledger_hash`. The `debug_assert_eq!(decision, Apply)` already guards the invariant, so swapping the computed hash for a zero placeholder is safe.

<details>

### Diff correctness

- `crates/history/src/catchup/replay.rs` — `verify_knit_to_lcl`: confirmed by reading `knit_to_lcl_decision` (lines 58–117). Case 4 (`entry_seq == lcl_seq + 1`) is the only branch reachable for the first apply entry; it consults `entry.header.previous_ledger_hash` exclusively. The previous `entry.hash` value was dead.
- The author's reasoning matches the inline triage in #2736 and the original review thread on PR #2732.
- Bonus: in non-debug builds, `debug_assert_eq!` is a no-op. Under the *old* code, a malformed `apply_data` whose first entry had `seq == lcl_seq` (case 3) or `seq + 1 == lcl_seq` (case 2) with a coincidentally matching hash would have silently returned `Skip` and proceeded. Under the *new* code, the zero hash will reliably mismatch and surface `KnitLclHashMismatch`/`KnitLclPredecessorHashMismatch`. Net: strictly more defensive in release builds, no regression in the expected path.
- `crates/history/src/verify.rs` — single-line `useless_conversion` cleanup (`upper.into_iter()` → `upper` inside `.chain(...)`). Trivially equivalent; `IntoIterator` is required by `chain`, and `Vec`'s `IntoIterator` already yields owned values. No-op.

### Error handling

- Removes a `CatchupFailed` error path that mapped `compute_header_hash` failures. Since the hash is no longer computed, the path is correctly dropped. `compute_header_hash` is fallible only on XDR encoding failure for a header we just cloned, so eliminating the call is unambiguously a simplification — no semantic loss.

### Readability

- The replacement comment clearly explains *why* zero is safe (which case is taken, what field is read). Good.
- Code is shorter and removes a fallible op from a hot-ish path.

### Test coverage

- `knit_to_lcl_decision`'s five cases are well covered by the existing unit tests at `replay.rs:640–794`.
- `verify_knit_to_lcl` itself has no dedicated unit test (covered indirectly by catchup integration). The diff doesn't change observable behavior on the success path, so no new test is required. Filing a small follow-up test for `verify_knit_to_lcl` that asserts: (a) the apply-branch succeeds when `first.header.previous_ledger_hash == lcl.hash` regardless of any value populated in `first`'s entry hash, and (b) error surfaces when `previous_ledger_hash` doesn't match, would be a nice belt-and-braces addition — non-blocking.
- `cargo test -p henyey-history` reported green per PR body; clippy reported clean (the `useless_conversion` cleanup is the reason).

### Minor / non-blocking

- Mixing the unrelated clippy cleanup into the fix is documented in the PR body's "Deviations" section. Fine here (one line, no-op), though I'd usually prefer it in a separate commit.

</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Low-risk micro-optimization. The `virtual_entry` is constructed and consumed within `verify_knit_to_lcl`, exercises only case 4 (which reads `header.previous_ledger_hash`, not `entry.hash`), and is dropped immediately — the zero-hash placeholder cannot leak anywhere observable.

<details>

- **Regression surface:** `virtual_entry` is a local in `verify_knit_to_lcl` (replay.rs:194) passed only to `knit_to_lcl_decision` and then dropped. By construction, `entry.seq == lcl.seq + 1` here (cases 1/2/3 require `entry.seq + 1 <= lcl.seq`, case 5 short-circuits as overshoot before case 4 runs), so only case 4 executes — and case 4 reads `header.previous_ledger_hash`, never `entry.hash`. No other caller reads the hash field of this synthetic entry. Safe.
- **Performance:** Tiny positive impact — one SHA256 saved per catchup batch's first apply entry. Not material on its own, but it also removes a fallible call and an error-mapping branch, simplifying the code.
- **API / data format:** No public-API changes. `LedgerHeaderHistoryEntry` shape is unchanged. The zero hash is never serialized or persisted (the entry doesn't escape the function scope).
- **Security — zero-hash confusion:** Considered, not a concern here. `Hash([0;32])` only lives inside this one local; it never reaches `HeaderSnapshot`, the ledger DB, history archives, peer messages, logs, or downstream `apply_data`/replay output. There is no code path where this zero placeholder could be mistaken for a real ledger hash by any other component.
- **Operational concerns:** None. No config, schema, or migration impact. Behavior on the error path (case 5 overshoot, prev-hash mismatch in case 4) is identical to before — those errors don't depend on `entry.hash` either.
- **Drive-by clippy fix in `verify.rs`:** Removing `upper.into_iter()` from a chained `.into_iter().chain(upper.into_iter())` is a true no-op (`chain` already takes `IntoIterator`). Scoped to a test, zero risk.
- **CI:** Format, Clippy, Build, CodeQL, Doctest all green at time of review; full test job still in progress but the unit-test surface here is exactly the `knit_to_lcl_decision` table tests that remain unchanged.

Non-blocking nit (optional): if you want to be extra defensive against a future refactor that adds a case reading `entry.hash` on the apply branch, you could `debug_assert_eq!` the entry's seq against `lcl.seq + 1` before constructing the virtual entry — but the existing `KnitDecision::Apply` debug-assert downstream already pins the invariant, so this is purely stylistic.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Dead-computation removal is semantically correct — the synthetic `virtual_entry` only drives `knit_to_lcl_decision` through case 4, which reads `header.previous_ledger_hash` and never `entry.hash`. Zero placeholder is safe and arguably more defensive in release builds.

<details>

### Logic / semantics

- Verified `knit_to_lcl_decision` (replay.rs lines ~58–117) directly: `entry.hash` is read **only** in cases 2 (`entry_seq + 1 == lcl_seq`) and 3 (`entry_seq == lcl_seq`). Case 4 (`entry_seq == lcl_seq + 1`) reads exclusively `entry.header.previous_ledger_hash`.
- The synthetic entry is built from `apply_data.first()`, which by construction has `seq == lcl_seq + 1`, so case 4 is the only branch reached. The existing `debug_assert_eq!(decision, KnitDecision::Apply)` immediately after the call enforces this invariant.
- Substituting `Hash([0; 32])` for the computed hash therefore cannot affect any code path that is reachable here.

### Defensive-coding side effect (positive)

- Under the old code, if `apply_data.first()` were ever malformed (`seq <= lcl_seq`), the computed hash could in principle coincidentally satisfy case 2 or case 3, returning `Skip`; `debug_assert_eq!` is a no-op in release, so the function would return `Ok(())` and a wrong-seq apply entry would proceed to the downstream chain check.
- Under the new code, a zero hash will (with overwhelming probability) mismatch in cases 2/3, surfacing `KnitLclPredecessorHashMismatch` / `KnitLclHashMismatch` immediately. Net: strictly more robust in release builds; no regression on the expected path.

### Error handling

- Removing the `CatchupFailed(...)` mapping for `compute_header_hash` is correct: the call itself is gone, and the only failure mode of `compute_header_hash` is XDR encoding of a freshly-cloned header — not something the caller could meaningfully recover from anyway. Net simplification, no lost diagnostics.

### Readability

- Inline comment clearly states which case is taken and which field is consumed, justifying the zero placeholder. Good documentation of intent.
- Code is shorter and removes a fallible op from the apply hot-path setup.

### Test coverage

- `knit_to_lcl_decision`'s 5 cases are exercised by the unit tests already present in `replay.rs` (the case-1..5 tests around lines 640–794). No behavioral change on the success path → no new test required.
- `verify_knit_to_lcl` itself remains covered indirectly via catchup integration tests. A targeted unit test asserting that the apply branch ignores the synthetic entry hash would be a nice belt-and-braces follow-up but is non-blocking.
- `verify.rs` change is a pure clippy cleanup (`upper.into_iter()` → `upper` inside `.chain(...)`), trivially equivalent because `Vec`'s `IntoIterator` already yields owned values.

### Minor / non-blocking

- Mixing the unrelated `useless_conversion` lint fix into this PR is documented under "Deviations". Fine as a one-line no-op, though I'd normally prefer it in a separate commit.

</details>

PR #2743, opened 2026-05-16
Add direct unit tests for verify_knit_to_lcl wrapper

### LINKED ISSUE #2737: Add direct unit tests for verify_knit_to_lcl wrapper
Follow-up from PR #2732 (issue #2680). Non-critical inline review comment that was not addressed before merge.

## Concern

`crates/history/src/catchup/replay.rs:172` — Non-blocking gap: `verify_knit_to_lcl` has no direct unit test. All 11 new tests target `knit_to_lcl_decision`. A small wrapper test asserting (a) a happy-path mixed knit_entries + apply_data slice returns Ok(()), (b) a tampered knit entry surfaces the right fatal variant, (c) an empty apply_data is a no-op, would lock down the wrapper behavior cheaply without needing archive fixtures.

## Source

[Original review comment](https://github.com/stellar-experimental/henyey/pull/2732#discussion_r3251387643) on PR #2732.

## Severity

tomerweller:
Closes #2737

## Summary

Add three direct unit tests for `CatchupManager::verify_knit_to_lcl` in `crates/history/src/catchup/replay.rs`. The existing 11 `knit_to_lcl_decision` tests exhaustively cover the §11.2 5-case decision matrix; these new tests target the wrapper-level glue those tests don't reach — the loop over `knit_entries` (multi-iteration, mid-loop `?` propagation), the `apply_data.first()` short-circuit on empty input, and the virtual `LedgerHeaderHistoryEntry` synthesis from `apply_data[0]` for the case-4 (Apply) check.

Three new tests:
- `test_verify_knit_to_lcl_happy_path_mixed` — cases 1/2/3 skip entries plus a 2-element `apply_data`, exercising the loop body, `Option::first`, and virtual-entry synthesis.
- `test_verify_knit_to_lcl_tampered_knit_entry_returns_fatal` — proves `?` propagates from a *mid-loop* failure (case 3 mismatch after case 2 ok) and that distinct error variants flow through unaltered (case 2 mismatch → `KnitLclPredecessorHashMismatch`).
- `test_verify_knit_to_lcl_empty_apply_data_noop` — both fully-empty and skip-only inputs short-circuit cleanly without exercising the apply branch.

Test-only change; no production code modifications.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/2737#issuecomment-4464776340)

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test -p henyey-history --lib catchup::replay::tests` — all 22 tests (19 pre-existing + 3 new) pass.
- [x] `cargo test -p henyey-history --tests` — full crate suite passes.

## Deviations from plan

None.

🤖 Generated with [GitHub Copilot CLI](https://github.com/features/copilot)

### DIFF
--- a/crates/history/src/catchup/replay.rs
+++ b/crates/history/src/catchup/replay.rs
@@ -805,6 +805,150 @@ mod tests {
         ));
     }
 
+    // ---------------------------------------------------------------
+    // Direct tests for the `verify_knit_to_lcl` wrapper (#2737).
+    //
+    // The 11 `knit_to_lcl_decision` tests above exhaustively cover the
+    // §11.2 five-case decision matrix. The tests below target the
+    // wrapper's own glue that is *not* covered by those tests:
+    //   - the loop over `knit_entries` (multi-iteration, mid-loop `?`),
+    //   - the `apply_data.first()` short-circuit on empty input,
+    //   - the synthesis of the virtual `LedgerHeaderHistoryEntry` from
+    //     `apply_data[0]` for the case-4 (Apply) check.
+    // ---------------------------------------------------------------
+
+    fn make_test_catchup_manager() -> CatchupManager {
+        use henyey_bucket::BucketManager;
+        use henyey_db::Database;
+
+        let db = Database::open_in_memory().expect("in-memory db");
+        let tmp_dir = tempfile::tempdir().expect("temp dir");
+        let bucket_manager = BucketManager::new(tmp_dir.keep()).expect("bucket manager");
+        let archive = crate::HistoryArchive::new("https://example.com").expect("archive");
+        CatchupManager::new(vec![archive], bucket_manager, db)
+    }
+
+    /// Build a `LedgerData` for ledger `seq` whose
+    /// `header.previous_ledger_hash` is `prev_hash`. The accompanying
+    /// `LclContext` carries the same hash as its `lcl_hash`, so the
+    /// `(None, None)` arm of `LedgerData::new` (which validates
+    /// `header.previous_ledger_hash == lcl.lcl_hash()`) succeeds.
+    ///
+    /// `.expect("valid LedgerData")` is intentional: a misconstructed
+    /// helper surfaces as a clear panic at construction time rather
+    /// than as a confusing `HistoryError::VerificationFailed` bubbling
+    /// out of `verify_knit_to_lcl` later.
+    fn make_apply_ledger_data(seq: u32, prev_hash: Hash256) -> LedgerData {
+        use henyey_common::protocol::LclContext;
+        use stellar_xdr::curr::LedgerHeader;
+
+        let mut header = LedgerHeader::default();
+        header.ledger_seq = seq;
+        header.previous_ledger_hash = stellar_xdr::curr::Hash(prev_hash.0);
+
+        let lcl = LclContext::new(0, prev_hash);
+        LedgerData::new(header, None, None, &lcl).expect("valid LedgerData")
+    }
+
+    #[test]
+    fn test_verify_knit_to_lcl_happy_path_mixed() {
+        // LCL = 100, lcl_hash = H100, previous_ledger_hash = H99.
+        let lcl = make_test_lcl(100, h(0x10), h(0x09));
+        let manager = make_test_catchup_manager();
+
+        // Three knit entries spanning cases 1/2/3 — each must classify
+        // as Skip inside the loop, exercising 3 iterations of the body.
+        let knit_entries = vec![
+            make_test_entry(98, h(0x99), h(0x88)), // case 1 (skip-old): hashes irrelevant
+            make_test_entry(99, h(0x09), h(0x08)), // case 2: entry.hash == lcl.previous_ledger_hash
+            make_test_entry(100, h(0x10), h(0x09)), // case 3: entry.hash == lcl.hash
+        ];
+
+        // apply_data[0] is the one that exercises the wrapper's
+        // virtual-entry synthesis (case 4): its `previous_ledger_hash`
+        // must equal `lcl.hash`. apply_data[1] is included to confirm
+        // the wrapper only consults `.first()`; its own previous-hash
+        // chains off `compute_header_hash(apply_data[0].header)`.
+        //
+        // Note: the synthesized virtual entry's *own* `hash` field is
+        // irrelevant on the case-4 branch — `knit_to_lcl_decision`
+        // only reads `header.previous_ledger_hash` for Apply.
+        let ledger_101 = make_apply_ledger_data(101, h(0x10));
+        let hash_101 = henyey_ledger::compute_header_hash(ledger_101.header())
+            .expect("compute hash for ledger 101");
+        let ledger_102 = make_apply_ledger_data(102, hash_101);
+        let apply_data = vec![ledger_101, ledger_102];
+
+        manager
+            .verify_knit_to_lcl(&knit_entries, &apply_data, &lcl)
+            .expect("happy path must succeed");
+    }
+
+    #[test]
+    fn test_verify_knit_to_lcl_tampered_knit_entry_returns_fatal() {
+        // LCL = 100, lcl_hash = H100, previous_ledger_hash = H99.
+        let lcl = make_test_lcl(100, h(0x10), h(0x09));
+        let manager = make_test_catchup_manager();
+        // apply_data is irrelevant — the loop must error out before
+        // we reach the `.first()` branch.
+        let apply_data = vec![make_apply_ledger_data(101, h(0x10))];
+
+        // (a) Mid-loop failure on the *second* iteration: case 2 passes,
+        // case 3 hash is tampered. This proves `?` propagates from
+        // inside the loop, not just from the first iteration.
+        {
+            let knit_entries = vec![
+                make_test_entry(99, h(0x09), h(0x08)),  // case 2 ok
+                make_test_entry(100, h(0xBB), h(0x09)), // case 3 mismatch (entry.hash != lcl.hash)
+            ];
+            let err = manager
+                .verify_knit_to_lcl(&knit_entries, &apply_data, &lcl)
+                .expect_err("tampered case-3 entry must produce fatal");
+            assert!(
+                matches!(err, HistoryError::KnitLclHashMismatch { .. }),
+                "expected KnitLclHashMismatch, got {err:?}",
+            );
+        }
+
+        // (b) First-iteration failure with a *different* variant —
+        // confirms the loop does not remap or swallow distinct error
+        // variants from `knit_to_lcl_decision`.
+        {
+            let knit_entries = vec![
+                make_test_entry(99, h(0xAA), h(0x08)), // case 2 mismatch (entry.hash != lcl.previous_ledger_hash)
+            ];
+            let err = manager
+                .verify_knit_to_lcl(&knit_entries, &apply_data, &lcl)
+                .expect_err("tampered case-2 entry must produce fatal");
+            assert!(
+                matches!(err, HistoryError::KnitLclPredecessorHashMismatch { .. }),
+                "expected KnitLclPredecessorHashMismatch, got {err:?}",
+            );
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Test-only addition (+144/-0) confined to `replay.rs`'s existing `mod tests`. No production code, no visibility relaxation, no shared state, no network or production endpoints touched. Risk profile is negligible.

<details>

**Regressions / blast radius**
- All additions live inside the pre-existing `#[cfg(test)] mod tests` block of `crates/history/src/catchup/replay.rs`. No exported items added, no `pub`/`pub(crate)` widened, no trait impls. Cannot affect any non-test build artifact.
- Reuses existing in-module helpers (`make_test_lcl`, `make_test_entry`, `h`) — no helper signatures changed.

**Compile time / test runtime**
- 3 new `#[test]` functions, each O(1) in work: in-memory SQLite (`Database::open_in_memory`), 1–3 small `LedgerHeader::default()`-based entries, one `compute_header_hash` call. No fixtures loaded from disk, no archive I/O, no async runtime, no sleeps.
- New deps surfaced in test scope (`henyey_bucket`, `henyey_db`, `tempfile`, `stellar_xdr`) are all already used by sibling tests in this crate, so no fresh compile-graph cost.

**Hidden coupling / ordering**
- No `static`, `lazy_static`, env vars, or process-global mutation. Each test instantiates its own `CatchupManager`, DB, and bucket dir → fully parallel-safe with `cargo test`'s default threaded runner.
- No reliance on test execution order.

**Filesystem / network**
- `HistoryArchive::new("https://example.com")` is a pure URL parse — verified in `crates/history/src/archive.rs:80` (delegates to `with_name_and_config`, which only calls `Url::parse`). No network request issued.
- `tempfile::tempdir().keep()` is the established pattern in this crate (`catchup/persist.rs:228`, `catchup/download.rs:610`); not a new leak vector relative to baseline. Each test run leaks one bucket dir under `$TMPDIR` per test (×3), matching neighboring tests — acceptable, and out of scope for this PR to fix.

**API / data-format breakage**
- None. No public surface, XDR, or on-disk format touched.

**Security / ops**
- `https://example.com` is a placeholder URL never dereferenced. No credentials, no production archive URLs (no `history.stellar.org`, no S3 bucket names), no secrets.

**Non-blocking notes (do not gate merge)**
- `tmp_dir.keep()` matches the crate convention but does leak the dir on disk. If anyone touches these tests later, dropping `.keep()` and threading the `TempDir` guard through (so it's dropped at end of test) would be a cheap cleanup — applies equally to existing siblings.
- `make_apply_ledger_data` uses `expect("valid LedgerData")`; the doc comment already explains the rationale, so no action needed.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The three new tests directly exercise `verify_knit_to_lcl`, distinguish failure variants with `matches!`, and cover the wrapper-specific glue (loop, mid-loop `?`, `apply_data.first()`, virtual-entry synthesis) that the existing 11 `knit_to_lcl_decision` tests don't reach.

<details>

**Exercise of the SUT**
- All three tests call `manager.verify_knit_to_lcl(...)` directly (not a transitive caller), so they target the wrapper rather than re-testing `knit_to_lcl_decision`.
- `cargo test -p henyey-history --lib catchup::replay::tests::test_verify_knit_to_lcl` — 3 passed locally.

**Assertion strength**
- Error tests use `expect_err` + `matches!(err, HistoryError::KnitLclHashMismatch { .. })` / `KnitLclPredecessorHashMismatch { .. }`, with a `{err:?}` in the message. These cleanly distinguish the two case-specific fatal variants — they would fail if the wrapper remapped, swallowed, or returned the wrong variant.
- Happy/no-op tests use `.expect(...)` on `Result<()>`, which is the only meaningful assertion possible on a unit-returning success path.
- No tautological assertions; helpers `make_apply_ledger_data` validate construction via `LedgerData::new(..., &lcl)` so a misbuilt fixture panics at setup rather than masquerading as a SUT failure.

**Branch coverage of the wrapper**
- ✅ Empty `knit_entries` + empty `apply_data` (pure no-op).
- ✅ Skip-only `knit_entries` + empty `apply_data` (loop runs, `.first()` short-circuits).
- ✅ Multi-iteration loop with mid-loop `?` (case-3 mismatch on the 2nd entry — proves `?` works past iteration 0).
- ✅ First-iteration error with a *different* variant (case-2 mismatch) — proves the loop doesn't conflate variants.
- ✅ Virtual-entry synthesis (case-4 Apply happy path) via `apply_data[0]` whose `previous_ledger_hash == lcl.hash`.
- ⚠️ **Mild gap (non-blocking):** no test covers the *error* path through the virtual-entry synthesis — i.e. an `apply_data[0]` that produces `KnitCurrentLedgerPrevHashMismatch` (case-4 prev-hash mismatch) or `KnitOvershot` (case-5, e.g. `seq = lcl_seq + 2`). The wrapper's glue here is essentially just `?`, and the variants themselves are unit-tested on `knit_to_lcl_decision`, so this is reasonable to defer — but one symmetric test would close the loop. Filed as inline comment.
- The `compute_header_hash` error branch on line 188 is unreachable in practice (XDR encoding of a `LedgerHeader` doesn't fail) — not worth testing.

**Readability / idiom**
- Helper comments are excellent: they explain *why* `.expect("valid LedgerData")` is acceptable (fail-loud setup) and *why* `ledger_102` is included even though the wrapper only consults `.first()`.
- Test names follow the established `test_verify_knit_to_lcl_<scenario>` convention.
- Inner `{ ... }` blocks in the tampered test cleanly scope the two sub-cases without separate `#[test]` boilerplate. Reasonable choice.
- `tmp_dir.keep()` leaks the tempdir on disk per test invocation — not a correctness issue (tests pass, paths are unique), but minor hygiene. Inline noted.

**No defects found** that affect the validity of the assertions or what they prove about the SUT.
</details>

crates/history/src/catchup/replay.rs
```diff
@@ -805,6 +805,150 @@ mod tests {
...
+    //   - the `apply_data.first()` short-circuit on empty input,
+    //   - the synthesis of the virtual `LedgerHeaderHistoryEntry` from
+    //     `apply_data[0]` for the case-4 (Apply) check.
+    // ---------------------------------------------------------------
+
+    fn make_test_catchup_manager() -> CatchupManager {
+        use henyey_bucket::BucketManager;
+        use henyey_db::Database;
+
+        let db = Database::open_in_memory().expect("in-memory db");
```
tomerweller, line 824: Risk (non-blocking): `tmp_dir.keep()` consumes the `TempDir` RAII guard, so the directory created here leaks on disk for every invocation of the three new tests. With `cargo test --all` running these on every CI build, this is a small but unbounded accumulation in the test runner's temp dir. Consider keeping the `TempDir` alive (return it from the helper, or store it on the `CatchupManager` test-only field) so it gets cleaned up when the test ends. Not a blocker.

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Pure test-only addition inside the existing `#[cfg(test)] mod tests` block. No production-code, visibility, API, or operational impact. One minor disk-cleanup nit filed inline as non-blocking.

<details>

- **Test-only confirmed.** All +144 lines land between the existing `#[cfg(test)]` boundary at line 590 and the closing brace of `mod tests`. No new `pub`/`pub(crate)` items, no module re-exports, no changes to `verify_knit_to_lcl`'s signature (still `pub(super)` at line 172). Zero risk of API surface drift.
- **Compile-time impact:** negligible. New helpers (`make_test_catchup_manager`, `make_apply_ledger_data`) are gated by `#[cfg(test)]` and compile only for `cargo test`. No new workspace deps — `tempfile` is already a workspace dev-dep in `crates/history/Cargo.toml` (line 64).
- **Runtime impact on test suite:** three new `#[test]`s, each constructs an in-memory SQLite DB + a `tempdir` + a minimal `CatchupManager`. No I/O beyond local fs, no archive HTTP (`HistoryArchive::new` only parses the URL — verified in `crates/history/src/archive.rs:80`). Should run in well under a second total. No fixture downloads, no async runtimes.
- **Security:** no `unsafe`, no `unwrap` on untrusted input, no hardcoded secrets, no real network access (`https://example.com` is only stored as a base URL string and never dialed in these tests). Hashes are deterministic local constants.
- **Operational concerns:** none — no config, no migrations, no feature flags, no public-API consumers.
- **Non-blocking nit (filed inline):** `tmp_dir.keep()` in `make_test_catchup_manager` deliberately drops the `TempDir` RAII guard, so each test invocation leaks one bucket directory on disk. Trivial in practice but accumulates over CI runs; would be cleaner to keep the guard alive for the test's lifetime. Not a merge blocker.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The three new tests target exactly the wrapper-only glue not exercised by the existing 11 `knit_to_lcl_decision` tests (loop iteration, mid-loop `?` propagation, `apply_data.first()` short-circuit, virtual-entry synthesis). All three pass locally; assertions are variant-specific and match the wrapper's contract.

<details>

**Verified locally**
- Built and ran the 3 new tests on the PR branch (`do/issue-2737` @ ba04008f) under `henyey-history`: all pass in 0.00s after compile.
- Helpers (`make_test_lcl`, `make_test_entry`, `h`) and error variants (`KnitLclHashMismatch`, `KnitLclPredecessorHashMismatch`) align with the wrapper's source at `crates/history/src/catchup/replay.rs:172–212`.

**Coverage — what the diff actually adds over the existing 11 decision tests**
- *Multi-iteration loop body*: `test_verify_knit_to_lcl_happy_path_mixed` puts three `Skip`-classified entries (cases 1/2/3) through the loop, so a regression that, say, only validated the first entry would be caught.
- *Mid-loop `?` propagation*: `test_..._tampered_knit_entry_returns_fatal` sub-case (a) makes the *second* iteration fail — distinguishes "loop early-exits on first failure" from "loop only validates first entry". Sub-case (b) tampers a different case (case 2) to confirm the loop preserves variant identity rather than remapping into a single error type. Both `matches!` assertions are variant-specific.
- *`apply_data.first()` short-circuit*: `test_..._empty_apply_data_noop` covers both the fully-empty path and the "non-empty knit, empty apply" path — the latter is the only one that distinguishes `.first()` from `[0]`.
- *Virtual-entry synthesis (case 4)*: happy-path test wires `apply_data[0].header.previous_ledger_hash = lcl.hash`, which is the exact field `knit_to_lcl_decision` reads on the Apply branch. The comment correctly notes that the synthetic entry's own `hash` field is unused — consistent with PR #2741 which replaced the recomputed hash with a `[0; 32]` placeholder on this branch.

**Error handling**
- All errors are matched by variant via `matches!` and `expect_err` with descriptive messages including the actual error via `{err:?}`. Nothing is swallowed or downgraded to a boolean.

**Readability**
- Names follow the `test_<behavior>` convention used elsewhere in this module.
- Helpers (`make_test_catchup_manager`, `make_apply_ledger_data`) are short and well-commented; the rationale for `.expect("valid LedgerData")` (surface as a clear panic in the helper rather than as a confusing `VerificationFailed` later) is explicitly documented.
- The header comment block clearly enumerates which wrapper-specific behaviors each test covers, distinguishing them from the underlying decision tests.

**Minor (non-blocking) — left as inline thoughts, not filed**
- `tmp_dir.keep()` in `make_test_catchup_manager` deliberately leaks the `TempDir` guard so the bucket directory survives the helper's return. Each test invocation therefore leaks a directory under the system temp root. Not blocking — the existing test infra elsewhere likely does the same — but a `lazy_static`/`OnceLock`-shared scratch dir or just letting `TempDir` live in the returned tuple would avoid the disk leak across CI runs.
- `apply_data` in case (b) of the tampered test is shared with case (a) and never reached on either path — could be elided with `&[]` to make "the loop errors before `.first()`" intent more obvious. Cosmetic.

Neither blocks merge.

</details>

PR #2747, opened 2026-05-16
Fix DashMap shard-reentrancy deadlock in tx_set_tracker test

### LINKED ISSUE #2744: Flaky/hanging test: herder::tx_set_tracker::tests::test_request_at_cap_still_increments_existing
## Observation

The CI \`Test\` job on PR #2742 (#2681) hung past the 30-minute job timeout, with the runner reporting:

> \`herder::tx_set_tracker::tests::test_request_at_cap_still_increments_existing\` "has been running for over 60 seconds"

Then \`##[error]The operation was canceled.\` (job-level timeout).

This test should complete in well under a second normally. Hanging for 60+ seconds is anomalous.

## Impact

- Caused PR #2742 to bounce in the pipeline (\`/review-pr\` correctly detected unrelated CI red and routed back to \`ready-for-doing\` for rebase).
- If the test hangs deterministically (e.g. a deadlock in tx_set_tracker), it will keep causing bounces on every PR until fixed.
- If it's flaky-but-rare, the rebase retry should pass and this can be downgraded to a known-flake note.

## Reproduction

\`cargo test -p henyey-herder tx_set_tracker::tests::test_request_at_cap_still_increments_existing\` — run locally and inspect. If it hangs, narrow down the deadlock.

## Severity

Medium — pipeline noise; not blocking the affected PR (auto-retry via rebase will likely succeed if flaky). Bumping to high if the same test hangs again on the rebased CI run.

tomerweller:
Closes #2744

## Summary

`herder::tx_set_tracker::tests::test_request_at_cap_still_increments_existing` was hanging intermittently in CI. Root cause: the test bound a `dashmap::Ref` from `tracker.pending.get(&first_hash)` and held it across a subsequent `tracker.request(new_hash, 100)` call. `request()` acquires a write lock on the destination shard via `pending.entry(...)`; whenever `new_hash` happens to map to the same DashMap shard as `first_hash`, same-thread read→write reentry deadlocks. DashMap's shard count is `4 * num_cpus.next_power_of_two()`, so the collision (and the hang) is CPU-count dependent — explaining "hangs only on some CI runners".

Fix: snapshot `request_count` by value so the `Ref` is dropped before re-entering `request()`. Also added a short `# Reentrancy` doc-comment on `pub fn request` so the precondition is explicit at the call site future contributors will read.

Test-only restructuring; production behavior is unchanged. The `pending` field is private and no production code path holds a `pending` shard guard across a `request()` call.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/2744#issuecomment-4464950353)

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test -p henyey-herder --lib tx_set_tracker::` — 32 passed
- [x] `cargo test -p henyey-herder --lib tx_set_tracker::tests::test_request_at_cap_still_increments_existing` — re-ran 5x, each <1s, all pass
- [x] `cargo test --all` — workspace green

## Deviations from plan

None.

### DIFF
--- a/crates/herder/src/tx_set_tracker.rs
+++ b/crates/herder/src/tx_set_tracker.rs
@@ -70,6 +70,15 @@ impl TxSetTracker {
     /// Register a pending tx-set request. Returns true if new.
     /// Returns false if already pending, already cached, or the pending map
     /// has reached its best-effort capacity cap (`MAX_PENDING_TXSET_REQUESTS`).
+    ///
+    /// # Reentrancy
+    ///
+    /// Callers (including tests) MUST NOT hold any guard into `self.pending`
+    /// (e.g. a `dashmap::mapref::one::Ref`/`RefMut` from `pending.get` or
+    /// `pending.get_mut`) when calling this. `request` acquires a write lock
+    /// on the destination shard via `pending.entry(...)` and read locks on
+    /// every shard via `pending.len()`; same-thread reentry will deadlock
+    /// whenever the held guard's shard collides with one of those.
     pub fn request(&self, hash: Hash256, slot: u64) -> bool {
         if self.cache.contains_key(&hash) {
             return false;
@@ -978,9 +987,14 @@ mod tests {
         // The first hash we inserted should still be incrementable.
         let first_hash = Hash256::from_bytes([0u8; 32]);
         assert!(!tracker.request(first_hash, 100));
-        let entry = tracker.pending.get(&first_hash).unwrap();
+        // Snapshot the field by value and let the `Ref` drop at the end of
+        // this statement BEFORE re-entering `tracker.request(...)` below.
+        // Holding the shard read-guard across `request()` would deadlock
+        // whenever `new_hash` collides with `first_hash` on a DashMap shard,
+        // because `request()` needs a write lock on the destination shard.
+        let request_count = tracker.pending.get(&first_hash).unwrap().request_count;
         assert_eq!(
-            entry.request_count, 2,
+            request_count, 2,
             "existing entry should increment even at cap"
         );
 


tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** Diff is 37 lines: a doc-comment on `TxSetTracker::request` describing a Rust/DashMap reentrancy precondition, plus a test-local snapshot of `request_count` to drop the `Ref` before re-entering `request()`. No production logic, control flow, or observable semantics change — parity with stellar-core is unaffected.

<details>

- **Surface area is test-only.** The only edits to non-test code are doc comments inside `impl TxSetTracker { /// ... pub fn request }`. The function body, signature, and call graph are unchanged. `cargo doc` output is the only non-test artifact that shifts.
- **No C++ analog to disturb.** `TxSetTracker` is a Rust-side dedup/coalesce wrapper around `DashMap` for in-flight tx-set fetch requests. Its closest stellar-core counterpart is the `ItemFetcher` family (`stellar-core/src/overlay/ItemFetcher.{h,cpp}` — `TxSetFetcher` specialization) plus `PendingEnvelopes`'s tracking of pending tx-set hashes. Those use `std::map`/`unordered_map` under a single mutex, so DashMap shard-reentrancy has no counterpart there to diverge from. The fix is a Rust-runtime concurrency artifact, not a protocol behavior.
- **Test still validates the same parity-relevant invariant.** `test_request_at_cap_still_increments_existing` continues to assert `request_count == 2` after a duplicate `request()` of `first_hash` while at `MAX_PENDING_TXSET_REQUESTS`. The reorder only changes when the `dashmap::Ref` is dropped — it still reads `request_count` for the same key after the same sequence of `request()` calls. This matches the observable "duplicate fetch requests for an in-flight tx-set hash coalesce into a refcount bump" behavior that stellar-core's `ItemFetcher::fetch` provides via its `mTrackers` map.
- **`MAX_PENDING_TXSET_REQUESTS` cap.** This cap is a Henyey-side memory-safety knob with no direct stellar-core equivalent (core relies on overlay flood/peer limits), and the PR does not touch it or its enforcement in `request()`. No new divergence introduced.
- **Doc-comment accuracy (informational, non-blocking).** The `# Reentrancy` note correctly identifies both the per-shard write lock from `pending.entry(...)` and the all-shards read lock from `pending.len()` (the latter is what makes any held `Ref` — even on a non-colliding shard — unsafe across `request()`). Good documentation hygiene; no parity implication.
- **No changes to slot accounting, cache promotion, eviction, or any path consumed by herder/SCP integration.** `cache`, `slot`, and the rest of `request()`'s control flow are byte-identical pre/post diff.

Conclusion: parity with stellar-core's tx-set fetch coalescing semantics is unchanged. Approving from the parity axis.
</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The fix correctly drops the `dashmap::Ref` before the subsequent `request()` call by snapshotting `request_count` by value, which eliminates the same-thread shard read→write reentry deadlock. Diagnosis matches the symptom (CPU-count-dependent flake via DashMap's `4 * num_cpus.next_power_of_two()` shard layout) and the doc-comment encodes the precondition at the public API.

<details>

**Correctness of the fix**
- Replacing `let entry = tracker.pending.get(&first_hash).unwrap();` with `let request_count = tracker.pending.get(&first_hash).unwrap().request_count;` causes the temporary `Ref` to be dropped at the end of the statement (per Rust temporary lifetime rules), so the shard read-guard is released before `tracker.request(new_hash, 100)` re-enters the map. This is exactly the right shape — the deadlock is structural (DashMap's per-shard `RwLock` does not support same-thread upgrade/reentry), so dropping the guard is the only sound fix.
- The follow-up `tracker.pending.get(&new_hash).is_none()` call is fine: it's after `request()` returned, no guard is held across a write, and even if it were, a *read* on a shard that nothing else is writing wouldn't deadlock — the original failure required a held read guard colliding with a write attempt on the *same* shard.
- The pre-existing comment above `let at_cap = self.pending.len() >= MAX_PENDING_TXSET_REQUESTS;` already documents the analogous `len()`-vs-`entry()` reentry hazard, so the new doc-comment on `pub fn request` is consistent with the established invariant.

**Test coverage**
- The fix is test-only; the bug it removes is in the test, not in production. The PR body's claim that "no production code path holds a `pending` shard guard across a `request()` call" matches a quick scan: `pending` is private to the module, and the other callers in `tx_set_tracker.rs` (`needs`, `store`, `cleanup_expired`, etc.) operate via short-lived `get`/`contains_key`/`retain`/`remove` calls without straddling a `request()`. Good.
- The regression-prevention story is the doc-comment + the fixed test itself. A stricter "would have caught the bug" test would be a deterministic reproducer (e.g. pick two hashes proven to map to the same shard via `dashmap::DashMap::determine_shard`, then attempt the read-then-request pattern and assert it doesn't hang under a timeout). That's a nice-to-have, not a blocker — the original test was already nondeterministically catching it; the doc-comment is the durable guard.

**Readability / style**
- The inline comment on the snapshot line clearly explains *why* the binding is by-value rather than `&Ref`. Future readers won't be tempted to "clean up" by reintroducing `let entry = ...`.
- The `# Reentrancy` rustdoc section is well-placed and specific (names the offending types `Ref`/`RefMut`, the offending methods `pending.entry`/`pending.len`, and the failure mode).

**Error handling**
- No error-handling surface changed. `unwrap()` on `pending.get(&first_hash)` is acceptable in a unit test where the entry was just inserted.

**Non-blocking observations (left as inline-worthy follow-ups, not required for merge)**
- Consider a tiny helper test that asserts the deadlock-free behavior using `std::thread::spawn` + `join_timeout` so a future regression of the *call pattern* (not just this one call site) trips a hard timeout instead of a 60s CI hang. Optional.
- The `Reentrancy` doc could optionally also mention `cache` if any future refactor makes `request()` take a `cache` shard guard; currently `cache.contains_key` is a brief unguarded probe, so it's fine as written.

</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** Parity is not meaningfully applicable to this change. The diff is purely (a) a doc comment on `TxSetTracker::request` documenting a DashMap shard-reentrancy hazard, and (b) a test-only fix that snapshots a `u64` field by value instead of holding a `dashmap::Ref` across a call to `request()`. There is no change to production logic, no change to wire/observable behavior, and `TxSetTracker` itself is a henyey-internal abstraction (the upstream stellar-core analogue lives in `ItemFetcher`/`TxSetFetcher` and is structurally different — std::map+mutex, no shard-locked concurrent map, so the deadlock class doesn't even exist upstream).

- **Scope of change**: 1 file, doc-comment + test-only edit. Production code path is byte-identical.
- **Observable behavior delta vs stellar-core**: none.
  - `request()` semantics (cache short-circuit → pending dedup → capacity cap with `MAX_PENDING_TXSET_REQUESTS` → `request_count` increment) are untouched.
  - Cap-time increment behavior (existing entries continue to increment past the cap) is unchanged — and the test still asserts exactly that (`request_count == 2`).
  - Locking footprint of the production path is unchanged: `pending.entry(...)` write + `pending.len()` read across all shards, same as before.
- **Upstream comparison**: stellar-core's `TxSetFetcher`/`ItemFetcher` (`stellar-core/src/herder/`, `stellar-core/src/overlay/ItemFetcher.{h,cpp}`) tracks pending fetches with `std::map`/`std::unordered_map` under a coarse mutex. There is no DashMap-style sharded reentrancy concern upstream, so the deadlock class being fixed here is a henyey-implementation artifact, not a parity divergence. The fix neither introduces nor removes any divergence.
- **Sequencing / ordering**: unchanged. No change to when callers are notified, when entries move from pending → cache, or when capacity is enforced.
- **Edge cases**: the test now correctly exercises the "increment-past-cap" edge case without artificially serializing on a held shard guard. The assertion is functionally equivalent (compares the same `request_count` value) — only the lock-hold scope shrank.
- **Doc comment**: the new `# Reentrancy` block on `request()` is accurate and helpful — it correctly identifies that `pending.entry(...)` takes a per-shard write lock and `pending.len()` walks all shards taking read locks, so any caller-held `Ref`/`RefMut` into `self.pending` will deadlock on shard collision. This documents a real henyey-internal contract; it has no parity implication.
- **Risk**: nil from a parity standpoint.

**Verdict rationale**: Internal locking-hazard documentation + test fix with zero production-behavior delta. Parity check vacuously passes. Approving.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Diagnosis and fix are correct. Snapshotting `request_count` by value drops the `dashmap::Ref` at the end of the let-statement, before `request()` re-enters the same shard for a write lock. The new `# Reentrancy` doc-comment cleanly captures the precondition for future callers. Production code is unaffected (`pending` is private and no production path holds a shard guard across `request()`). Test passes locally.

**Root-cause analysis — correct.**
- DashMap's `get()` returns a `Ref` holding a `parking_lot::RwLock` *read* guard on the destination shard.
- `request()` calls `self.pending.entry(hash)`, which takes a *write* guard on the destination shard, and `self.pending.len()`, which takes read guards on every shard.
- `parking_lot::RwLock` is not re-entrant: same-thread read→write on the same lock deadlocks unconditionally; same-thread read→read can deadlock under writer-priority if a writer is queued. Either way, holding a `Ref` across `request()` is unsound.
- DashMap shard count = `4 * num_cpus.next_power_of_two()`, so the deadlock is CPU-count dependent — exactly matching the "hangs only on some CI runners" symptom from #2744.

**Fix — correct and minimal.**
- `let request_count = tracker.pending.get(&first_hash).unwrap().request_count;` — `request_count` is a `Copy` integer; the temporary `Ref` is dropped at the statement's `;` per Rust temporary-lifetime rules, well before the next `tracker.request(...)` call. ✓
- The second `tracker.pending.get(&new_hash)` (line 1004) is *after* `request()` returns, so no guard is held across a re-entrant call. ✓
- The `# Reentrancy` doc-comment on `pub fn request` is accurate (mentions both the `entry()` write lock and the `len()` cross-shard reads) and lives at the correct call-site for future contributors. ✓

**Test coverage — adequate, with one minor caveat (see inline).**
- The regression test exercises the cap-still-increments-existing path that triggered the original hang.
- Caveat: the test is not *deterministically* a deadlock repro — whether the pre-fix code hangs depends on whether `[0xFF; 32]` and `[0u8; 32]` collide on a DashMap shard, which depends on `num_cpus`. So this PR's test would have passed on machines whose shard count happens to separate the two hashes. It still validates the *behavior* (existing entry increments at cap) and the structural fix removes the deadlock class entirely, so this is a MINOR note, not a blocker.

**Readability / idiomatic Rust:** Clear. The inline comment explaining why the snapshot is necessary is exactly what's needed. No clippy concerns.

**Error handling:** Unchanged; not in scope.

**Other code paths inspected:** `clear_pending`, `pending_hashes`, `pending_entries`, `needs` — none hold a `pending` guard across a call to `request()`. `pending` is private to the crate; external callers can't construct the dangerous pattern.

</details>

crates/herder/src/tx_set_tracker.rs
```diff
@@ -978,9 +987,14 @@ mod tests {
         // The first hash we inserted should still be incrementable.
         let first_hash = Hash256::from_bytes([0u8; 32]);
         assert!(!tracker.request(first_hash, 100));
-        let entry = tracker.pending.get(&first_hash).unwrap();
+        // Snapshot the field by value and let the `Ref` drop at the end of
+        // this statement BEFORE re-entering `tracker.request(...)` below.
+        // Holding the shard read-guard across `request()` would deadlock
+        // whenever `new_hash` collides with `first_hash` on a DashMap shard,
+        // because `request()` needs a write lock on the destination shard.
+        let request_count = tracker.pending.get(&first_hash).unwrap().request_count;
```
tomerweller, line 995: MINOR (non-blocking): This regression test isn't a deterministic deadlock repro — whether the pre-fix code hangs depends on whether `first_hash` ([0;32]) and `new_hash` ([0xFF;32]) collide on a DashMap shard, and the shard count is `4 * num_cpus.next_power_of_two()`. On runners where these hashes land on different shards, the unfixed code would pass. Optional hardening: stress with many distinct `new_hash` values in a loop (or compute against DashMap's hasher) so the test fails deterministically pre-fix on every runner. The structural fix is correct regardless — this is purely about strengthening the regression guard.

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** Production semantics in `TxSetTracker::request` are unchanged — the diff only adds a doc comment and rewrites a test to snapshot a field instead of holding a `DashMap` shard guard across `request()`. No observable behavior changes vs. stellar-core's `ItemFetcher`/`TxSetFetcher` path, so the parity gap is neither widened nor narrowed.

<details>

- Diff scope (`crates/herder/src/tx_set_tracker.rs`):
  - Lines 73–81: pure documentation added to `request()` describing the DashMap shard-reentrancy hazard. No code change in the production path.
  - Lines 987–998 (test `request_at_cap_still_increments_existing_entries`, in `#[cfg(test)] mod tests`): replaces a held `dashmap::mapref::one::Ref` with a by-value snapshot of `request_count`. Test-only.
- Production-path parity:
  - The body of `request()` (cache check → `pending.entry(hash)` → `or_insert_with(...)` → `request_count += 1` → `len()` cap check → revert on overflow) is byte-for-byte unchanged. Therefore the externally observable contract — "register pending request; cap pending map at `MAX_PENDING_TXSET_REQUESTS`; bump existing entries even at cap" — is preserved.
  - stellar-core's analogue lives in `stellar-core/src/herder/TxSetFetcher` / `src/overlay/ItemFetcher.{h,cpp}` (`ItemFetcher::fetch`, `mTrackers` map, `Tracker::tryNextPeer`). henyey already diverges structurally (single `TxSetTracker` with explicit cap + LRU cache vs. core's per-item `Tracker` objects with no hard cap), but this PR does not touch any of that surface area.
- Behavioral checklist vs. stellar-core:
  - Cache-hit short-circuit: unchanged.
  - Increment-on-duplicate-request: unchanged (still returns `false`, still bumps `request_count`).
  - Capacity enforcement (`MAX_PENDING_TXSET_REQUESTS`): unchanged. (This is already a henyey-specific guardrail with no direct stellar-core counterpart — pre-existing divergence, not introduced here.)
  - Slot tracking / arrival callback ordering: not touched.
- Sequencing / determinism:
  - The fix is purely about intra-process lock ordering inside a test. No I/O ordering, no message-handling order, no ledger-close-relevant state machine touched. Cannot affect consensus determinism.
- Doc comment accuracy:
  - Claim that `pending.len()` takes read locks on every shard matches DashMap's implementation (`Sharded::len` iterates and `read()`s each shard), so the reentrancy warning is technically correct and worth keeping.

No parity concerns. Recommend merge from a parity standpoint.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Minimal, correct fix: snapshots `request_count` by value so the `dashmap::Ref` is dropped before the next `request()` call, removing the same-thread shard reentrancy that caused the deadlock. Doc comment on `request()` documents the contract for future callers. Test now passes deterministically (30/30 stress runs).

<details>

**Diff correctness**
- The deadlock cause is precisely identified: holding a `dashmap::mapref::one::Ref` from `pending.get(&first_hash)` keeps a shard read-lock; `request(new_hash,...)` then calls `pending.len()` (which `read()`s every shard) and `pending.entry(new_hash)` (which `write()`s the destination shard). When `new_hash`'s shard collides with `first_hash`'s shard, we self-deadlock.
- The fix copies `request_count` (a `u64`) out of the `Ref` and lets the temporary drop at the statement boundary. After the let-binding, no DashMap guard is alive, so the subsequent `tracker.request(new_hash, 100)` is safe regardless of shard collision. Behavior is unchanged — same assertion, same value.
- Production call sites: `scp_driver.rs:921` is the only non-test caller of `tx_tracker.request`, and it doesn't hold any `pending` guard. No production code paths affected.

**Test coverage**
- The regression test is the same one referenced in #2744 and now passes deterministically (verified locally: 30/30 single-test runs in a loop, no hangs).
- Minor: the test doesn't *guarantee* `[0u8;32]` and `[0xFF;32]` collide on the same DashMap shard (DashMap shard count is platform/version-dependent), so technically the test isn't a tight reproducer of the deadlock under all build configurations — it's a hygiene fix that happens to also fix the observed flake. That's acceptable and arguably better than a brittle "force the same shard" test, especially because the new `# Reentrancy` doc comment turns the contract into something humans (and future reviewers) can enforce.

**Doc / readability**
- The `# Reentrancy` doc comment is excellent — names the exact guard types, the two locking ops inside `request` (`len()` + `entry()`), and the failure mode. This is the kind of comment that prevents the bug from coming back.
- Inline test comment explains *why* the snapshot pattern is required, not just what. Good.

**No remaining DashMap pitfalls**
- `request()` itself already pre-checks `len()` *before* calling `entry()` (line 88-91) — that ordering is correct and necessary; reversing it would self-deadlock the function. Worth keeping in mind if anyone refactors.
- No other test or call site in `tx_set_tracker.rs` holds a `pending` ref across a mutating call.
- `quorum_set_tracker.rs` uses a `Mutex<Inner>` rather than DashMap, so the pattern doesn't generalize; nothing to mirror there.

**Nits (non-blocking)**
- The doc comment could note that the same hazard applies to `self.cache` (since `request()` also calls `cache.contains_key`), though in practice no one holds a cache guard across `request`. Not worth blocking on.

</details>

PR #2749, opened 2026-05-16
Keep TempDir alive in verify_knit_to_lcl tests to avoid leaking bucket dirs

### LINKED ISSUE #2745: TempDir::keep() in verify_knit_to_lcl tests leaks bucket dirs
Follow-up from PR #2743 (issue #2737). Non-critical inline review comment that was not addressed before merge.

## Concern

`crates/history/src/catchup/replay.rs:828` — Using `TempDir::keep()` here intentionally disables tempfile cleanup, so each unit test invocation leaves a bucket directory behind under the system temp location. Since these tests only need the manager as a receiver for `verify_knit_to_lcl`, prefer keeping a `TempDir` guard alive for the test lifetime (or otherwise avoiding the real bucket directory) so repeated test runs do not accumulate leaked directories.

## Source

[Original review comment](https://github.com/stellar-experimental/henyey/pull/2743#discussion_r3251697720) on PR #2743 (from `copilot-pull-request-reviewer`).

## Severity

tomerweller:
Closes #2745

## Summary

The `verify_knit_to_lcl` unit tests' helper `make_test_catchup_manager()` was calling `TempDir::keep()`, which intentionally disables tempfile cleanup. Each test invocation left a bucket directory behind under the system temp location, accumulating across repeated `cargo test` runs.

The helper now returns the `TempDir` guard alongside the `CatchupManager` and the three test sites bind it as `_tmp_dir`, so the directory is deleted on drop.

## Plan reference

[Triage Report — trivial short-circuit](https://github.com/stellar-experimental/henyey/issues/2745#issuecomment-4465017421)

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test -p henyey-history --tests` — all 3 `verify_knit_to_lcl` tests pass; full crate test suite green

## Deviations from plan

None.

### DIFF
--- a/crates/history/src/catchup/replay.rs
+++ b/crates/history/src/catchup/replay.rs
@@ -817,15 +817,23 @@ mod tests {
     //     `apply_data[0]` for the case-4 (Apply) check.
     // ---------------------------------------------------------------
 
-    fn make_test_catchup_manager() -> CatchupManager {
+    /// Returns the manager together with the `TempDir` guard backing its
+    /// `BucketManager`. The caller must keep the `TempDir` alive for the
+    /// duration of the test so the bucket directory is deleted on drop
+    /// rather than leaked under the system temp location.
+    fn make_test_catchup_manager() -> (CatchupManager, tempfile::TempDir) {
         use henyey_bucket::BucketManager;
         use henyey_db::Database;
 
         let db = Database::open_in_memory().expect("in-memory db");
         let tmp_dir = tempfile::tempdir().expect("temp dir");
-        let bucket_manager = BucketManager::new(tmp_dir.keep()).expect("bucket manager");
+        let bucket_manager =
+            BucketManager::new(tmp_dir.path().to_path_buf()).expect("bucket manager");
         let archive = crate::HistoryArchive::new("https://example.com").expect("archive");
-        CatchupManager::new(vec![archive], bucket_manager, db)
+        (
+            CatchupManager::new(vec![archive], bucket_manager, db),
+            tmp_dir,
+        )
     }
 
     /// Build a `LedgerData` for ledger `seq` whose
@@ -854,7 +862,7 @@ mod tests {
     fn test_verify_knit_to_lcl_happy_path_mixed() {
         // LCL = 100, lcl_hash = H100, previous_ledger_hash = H99.
         let lcl = make_test_lcl(100, h(0x10), h(0x09));
-        let manager = make_test_catchup_manager();
+        let (manager, _tmp_dir) = make_test_catchup_manager();
 
         // Three knit entries spanning cases 1/2/3 — each must classify
         // as Skip inside the loop, exercising 3 iterations of the body.
@@ -888,7 +896,7 @@ mod tests {
     fn test_verify_knit_to_lcl_tampered_knit_entry_returns_fatal() {
         // LCL = 100, lcl_hash = H100, previous_ledger_hash = H99.
         let lcl = make_test_lcl(100, h(0x10), h(0x09));
-        let manager = make_test_catchup_manager();
+        let (manager, _tmp_dir) = make_test_catchup_manager();
         // apply_data is irrelevant — the loop must error out before
         // we reach the `.first()` branch.
         let apply_data = vec![make_apply_ledger_data(101, h(0x10))];
@@ -930,7 +938,7 @@ mod tests {
     #[test]
     fn test_verify_knit_to_lcl_empty_apply_data_noop() {
         let lcl = make_test_lcl(100, h(0x10), h(0x09));
-        let manager = make_test_catchup_manager();
+        let (manager, _tmp_dir) = make_test_catchup_manager();
 
         // No knit entries, no apply data: pure no-op, must Ok.
         manager


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Fix is correct and complete in scope: helper now returns the `TempDir` guard alongside the manager, all three `verify_knit_to_lcl` tests bind it as `_tmp_dir` so it's dropped at end of test, and the leaking `tmp_dir.keep()` call is gone. Targeted tests pass.

<details>

- Scope match: issue #2745 targets `verify_knit_to_lcl` tests in `crates/history/src/catchup/replay.rs`. All three affected tests (`test_verify_knit_to_lcl_happy_path_mixed`, `test_verify_knit_to_lcl_tampered_knit_entry_returns_fatal`, `test_verify_knit_to_lcl_empty_apply_data_noop`) are updated.
- No leak paths remain in the file: `grep` for `keep()` / `into_path` / `forget()` against `crates/history/src/catchup/replay.rs` shows none. The `BucketManager` is now constructed from `tmp_dir.path().to_path_buf()`, leaving ownership of the directory with the `TempDir` guard.
- Lifetime correctness: `_tmp_dir` is bound (not `_`), so it lives until the end of each `#[test]` function — the `BucketManager` (held inside `manager`) only borrows the path as `PathBuf`, so the drop order (manager first, then tmp_dir) is safe; the directory exists for as long as the manager could touch it.
- Helper doc-comment is clear and explicitly tells future callers they must keep the `TempDir` alive — good defensive ergonomics that pre-empts the same bug recurring.
- Error handling unchanged; `.expect("temp dir")` / `.expect("bucket manager")` semantics preserved.
- Tests: `cargo test -p henyey-history --tests verify_knit_to_lcl` → 3 passed, 0 failed.
- Out-of-scope observation (not a blocker; identical leak pattern exists in `crates/history/src/catchup/persist.rs:228` and `crates/history/src/catchup/download.rs:610`). Issue #2745 is explicitly scoped to the `verify_knit_to_lcl` tests, so leaving these for a follow-up is fine — flagging only so it isn't lost.

</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Test-only change in `mod tests` of `replay.rs`; no production code paths, public APIs, data formats, or configuration touched. Risk surface is limited to the three affected unit tests.

<details>

**Scope confirmation**
- Diff is entirely inside `#[cfg(test)] mod tests` in `crates/history/src/catchup/replay.rs`.
- `make_test_catchup_manager` is a private fn in the test module; signature change `() -> CatchupManager` → `() -> (CatchupManager, TempDir)` only affects the three call sites in the same module, all of which were updated.
- No change to `CatchupManager` API, return types, error types, public traits, on-disk formats, configs, or migrations.

**Regression / reliability risk**
- `_tmp_dir` (underscore-prefixed, not bare `_`) is a real binding held to end of test scope — correct pattern; no early drop.
- Drop order: `let (manager, _tmp_dir) = ...` drops `_tmp_dir` first, then `manager`. `BucketManager` (and `BucketList::drop`) only does work if `pending_persist` was set — these tests never close a ledger or trigger persists, so manager drop is a no-op w.r.t. the temp path. No use-after-cleanup hazard in practice.
- Each test gets its own `tempfile::tempdir()` (random suffix) — no cross-test path collision, safe under `cargo test`'s parallel runner.
- No flake risk: the previous `keep()` path also worked; this just removes the leak.

**Performance / CI**
- Negligible: removes filesystem accumulation in `/tmp` over repeated runs (was a slow leak, not a hot path). Net positive for long-lived CI runners and dev boxes.

**Security**
- TempDir paths remain `tempfile`-randomized; no predictability regression. Lifetime is now shorter (cleaned up on test exit) — strictly better for hygiene.

**Operational**
- No config, no migration, no runtime behavior change. Nothing for operators to do.

**Non-blocking follow-up (out of scope for this PR / issue #2745)**
- The same `tmp_dir.keep()` leak pattern still exists in two sibling test helpers:
  - `crates/history/src/catchup/persist.rs:228`
  - `crates/history/src/catchup/download.rs:610`
  Worth a follow-up issue to apply the same guard pattern there for consistency, but explicitly outside the scope of issue #2745.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Clean, minimal fix that correctly keeps the `TempDir` guard alive for each test's scope, eliminating the bucket-dir leak. Helper signature change is propagated to all three call sites.

<details>

### Diff matches the stated intent
- The PR claims to replace `TempDir::keep()` (which intentionally disables cleanup) with a guard returned to the caller. The diff does exactly that:
  - `BucketManager::new(tmp_dir.keep())` → `BucketManager::new(tmp_dir.path().to_path_buf())`. `path()` borrows, `to_path_buf()` clones, so `tmp_dir` is not consumed and can be returned.
  - `make_test_catchup_manager()` now returns `(CatchupManager, tempfile::TempDir)`.
- All three `verify_knit_to_lcl` tests (`happy_path_mixed`, `tampered_knit_entry_returns_fatal`, `empty_apply_data_noop`) are updated to bind the guard as `_tmp_dir`. No other call sites of the helper exist (grepped — only these three).

### `_tmp_dir` binding is correct
- Rust drops named bindings (including those starting with `_`) at end of scope. Only the fully-anonymous `let _ = ...` pattern would drop immediately.
- So `let (manager, _tmp_dir) = make_test_catchup_manager();` keeps the `TempDir` alive until the end of each `#[test]` function, which is exactly when the bucket dir should be removed. ✅
- Using the `_` prefix correctly suppresses the unused-variable warning while preserving drop semantics.

### Test coverage
- This is test-only code; no production paths changed. The existing three tests continue to exercise the same logic — the only observable difference is on-disk cleanup, which is implicit and not normally asserted.
- No new test needed: a "does it leak?" assertion would be brittle and out of scope for a unit test.

### Readability / clarity
- The doc comment on the helper explains *why* the guard is returned (to prevent leaks), which is helpful and matches the issue's framing. Good defensive documentation against future refactors that might be tempted to drop the guard again.
- Tuple return is a small ergonomic cost but is the simplest correct shape here; introducing a wrapper struct would be overkill for three call sites.

### Error handling
- No change in error handling; helper still uses `.expect()` for setup failures, consistent with surrounding test code. Appropriate for test fixtures.

### Nits (non-blocking)
- None worth requesting changes for.

</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Surgical, test-only fix; signature change is contained to a private helper inside `replay.rs::tests`, all three call sites are updated, and no other code consumes it. No regression, perf, API, or operational risk.

<details>

### Scope & blast radius
- Touches a single file: `crates/history/src/catchup/replay.rs`, inside `mod tests` only. No production code path is affected.
- Non-parity crate (history) and non-parity area (test code), so the risk lens applies — no spec/parity concern.

### Breaking change check (helper signature)
- `make_test_catchup_manager()` is a `fn` (not `pub`) inside `mod tests` of `replay.rs`. Rust module scoping makes it inaccessible to any other module/crate.
- Confirmed callers via `grep -rn make_test_catchup_manager … --include='*.rs'`:
  - `crates/history/src/catchup/replay.rs` — definition + 3 call sites, all updated to `let (manager, _tmp_dir) = …`.
  - `crates/history/src/catchup/persist.rs` and `crates/history/src/catchup/download.rs` each define their own **separate** private `make_test_catchup_manager()` inside their own `mod tests`. They are independent functions that happen to share a name — not callers of the modified helper. Untouched and unaffected.
  - Other hits live under `.opencode/worktrees/…` and `.claude/worktrees/…` which are agent worktree scratch areas, not part of the build.
- No public API, no XDR, no on-disk format changed.

### Regression risk
- Behavior preserved: `BucketManager::new(tmp_dir.path().to_path_buf())` gives the bucket manager the same path it received before; only the cleanup ownership changed (TempDir guard now drops at end of test instead of being leaked via `keep()`).
- Tests bind the guard as `_tmp_dir` (leading underscore) — correct: it suppresses unused-variable warnings while still extending the binding's lifetime to the end of the test scope (drop is not run early, unlike `let _ = …`). Good.
- The three existing tests (`…_happy_path_mixed`, `…_tampered_knit_entry_returns_fatal`, `…_empty_apply_data_noop`) are pure in-memory operations on the manager — none of them rely on the bucket directory persisting after `make_test_catchup_manager()` returns. So eager cleanup at end-of-test is safe.

### Performance
- Negligible. Test-only; extra cleanup work on `Drop` is a directory removal that will be empty in these tests. Net effect is *less* disk usage over repeated runs.

### Security / operational
- No secrets, no config, no migrations. Only impact is positive: stops accumulating empty bucket dirs under the system temp location across repeated `cargo test` runs.

### Out-of-scope observation (not blocking, not requested)
- The sibling helpers `crates/history/src/catchup/persist.rs:228` and `crates/history/src/catchup/download.rs:610` exhibit the **same `tmp_dir.keep()` leak pattern** that #2745 fixes here. They were intentionally left alone because the linked triage scoped this PR narrowly to `replay.rs`. Worth a follow-up issue if not already tracked, but explicitly out of scope for this PR — no change requested.

### Verdict
APPROVE — clean, narrow, correctly-scoped fix with no risk to other callers or production code.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The diff correctly fixes the bucket-dir leak by replacing `TempDir::keep()` with a kept-alive `TempDir` guard threaded through the test setup. All three `verify_knit_to_lcl` test sites are covered, and the binding name keeps the guard live for the test's lifetime.

<details>

- **Fix actually addresses the leak.** `TempDir::keep()` consumed the `TempDir` and returned the `PathBuf` while suppressing the `Drop` impl that removes the directory — exactly the behavior #2745 calls out. Switching to `tmp_dir.path().to_path_buf()` and returning the `TempDir` to the caller restores the `Drop`-based cleanup. Verified on the head ref.
- **Binding lifetime is correct.** Each call site uses `let (manager, _tmp_dir) = make_test_catchup_manager();`. A leading-underscore *named* binding (`_tmp_dir`) is treated as a normal local: it lives until end-of-scope and is dropped then. The footgun would have been `let (manager, _) = ...`, which drops the `TempDir` immediately at the `let`. The diff avoids that — good.
- **All test sites covered.** `grep` for `make_test_catchup_manager` in `crates/history/src/catchup/replay.rs` shows exactly three call sites (`test_verify_knit_to_lcl_happy_path_mixed`, `test_verify_knit_to_lcl_tampered_knit_entry_returns_fatal`, `test_verify_knit_to_lcl_empty_apply_data_noop`) and the helper itself; all three are updated. No other callers exist.
- **Choice of `tmp_dir.path().to_path_buf()` is appropriate.** `BucketManager::new` takes an owned `PathBuf`, and `TempDir::path()` returns a borrowed `&Path`, so `.to_path_buf()` is the natural conversion. The `TempDir` itself remains owned by the helper and is moved into the returned tuple — no aliasing or lifetime issues. Alternative approaches (e.g., refactoring `BucketManager` to accept a `&Path`, or constructing an in-memory bucket manager) would be larger-scoped and unnecessary for a test-only fix.
- **Test coverage preserved.** The behavioral surface of the three tests is unchanged — they still construct the same `CatchupManager` over the same on-disk bucket dir and exercise `verify_knit_to_lcl` identically. The only observable diff is cleanup on drop.
- **Readability / idiomatic Rust.** Returning `(CatchupManager, TempDir)` is the standard idiom for "RAII guard that must outlive the value it backs." The doc comment on `make_test_catchup_manager` explicitly tells future callers to keep the guard alive, which is a nice touch and reduces the chance someone reintroduces the bug via `let (manager, _) = ...`.
- **Error handling.** Unchanged — still `.expect(...)` on `tempdir()` / `BucketManager::new` / `Database::open_in_memory`, which matches the surrounding test style. No new failure modes introduced.

Nit (non-blocking, not worth a follow-up): the doc comment could additionally warn that binding to `_` (rather than `_tmp_dir`) would defeat the fix. The current wording is already sufficient though.

</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Test-only change; reduces (rather than introduces) resource leakage. No public API impact, no concurrency or behavioral risk. Sibling files have the same leak pattern but are out of scope for this issue.

<details>

**Regressions in existing test behavior**
- All three call sites are mechanical `let manager = ...` → `let (manager, _tmp_dir) = ...` rewrites. The `CatchupManager` is constructed identically; same `Database::open_in_memory`, same `HistoryArchive`, same `BucketManager` rooted at the same path. The bucket dir now happens to live under a `TempDir` guard rather than a leaked path, but for the test lifetime it's a real, writable directory exactly as before. No code path is skipped.
- The `_tmp_dir` binding is held in the test function's stack frame, so the guard outlives `manager` (drop order is reverse of declaration: `_tmp_dir` drops after `manager`). Safe — no use-after-free of the bucket root during teardown.

**CI performance / resource**
- Confirmed: this PR strictly reduces leaked dirs. Before, each invocation of these 3 tests left a `tempfile`-prefixed dir in `$TMPDIR` permanently. After, the `TempDir` `Drop` impl removes it. Fewer inodes consumed in CI tmpfs, faster cleanup. No new I/O.

**Public API of `make_test_catchup_manager`**
- The function lives inside `mod tests` (private, `#[cfg(test)]`) in `crates/history/src/catchup/replay.rs`. `grep` confirms two other files (`catchup/download.rs`, `catchup/persist.rs`) define their own *separate* private helpers with the same name — they are not callers. No external consumers; the return-type change is safe.

**Operational / concurrency**
- `tempfile::tempdir()` generates a unique randomized path per call, so the three tests cannot race against the same bucket dir even when run in parallel under `cargo test`. Unchanged from before.
- `Drop` for `TempDir` does best-effort recursive deletion; if a test panics mid-run the dir still cleans up on unwind. Strictly better than `keep()`.

**Hidden side effects of `tmp_dir.path()` vs `tmp_dir.keep()`**
- `keep()` consumes the `TempDir` and returns an owned `PathBuf` while disabling cleanup. `path()` borrows; the new code calls `.to_path_buf()` to get an owned copy, leaving the guard intact. Functionally identical from `BucketManager`'s perspective — it just sees a `PathBuf` either way.
- No change to permissions, no symlink/canonicalization differences, no env var dependency shift.

**Non-blocking follow-up (filed inline)**
- The same `TempDir::keep()` leak pattern exists in `crates/history/src/catchup/download.rs:609` and `crates/history/src/catchup/persist.rs:228`. Out of scope for issue #2745 (which scopes to `verify_knit_to_lcl`), but worth a tracking issue.

</details>

crates/history/src/catchup/replay.rs
```diff
@@ -817,15 +817,23 @@ mod tests {
...
         let tmp_dir = tempfile::tempdir().expect("temp dir");
-        let bucket_manager = BucketManager::new(tmp_dir.keep()).expect("bucket manager");
+        let bucket_manager =
+            BucketManager::new(tmp_dir.path().to_path_buf()).expect("bucket manager");
         let archive = crate::HistoryArchive::new("https://example.com").expect("archive");
-        CatchupManager::new(vec![archive], bucket_manager, db)
+        (
+            CatchupManager::new(vec![archive], bucket_manager, db),
+            tmp_dir,
+        )
```
tomerweller, line 836: Follow-up (non-blocking, out of scope for #2745): the same `TempDir::keep()` leak pattern exists in two sibling helpers — `crates/history/src/catchup/download.rs:609` and `crates/history/src/catchup/persist.rs:228`. Worth fixing in a small follow-up so all three private `make_test_catchup_manager` helpers consistently keep the `TempDir` alive instead of leaking under `$TMPDIR`.

tomerweller:
## Review: Waiting on CI

CI is still running (checks: `build / 3 test / 2 test (testing-with-pr, amd64, testnet, core,horizon , )`). 31/32 checks green, 0 failures. Re-picking this PR on the next tick.

CI age: 17 min / budget 60 min.
Bounce-back count: 0/3.

Latest reviewer verdicts (already posted on prior tick):
- **Reviewer A (Correctness):** APPROVE
- **Reviewer B (Risk):** APPROVE

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Test-only fix correctly swaps `TempDir::keep()` (which leaks the directory) for a `TempDir` guard returned to the test, ensuring RAII cleanup. All three call sites are updated and the `BucketManager` still receives a valid owned `PathBuf`.

<details>

- **Diff scope is minimal and correct.** Only `make_test_catchup_manager` and its three callers in `crates/history/src/catchup/replay.rs` are touched. `grep` confirms there are no other callers of the helper, so no orphaned uses are left behind.
- **Lifetime correctness.** `BucketManager::new(tmp_dir.path().to_path_buf())` takes an owned `PathBuf`, and the returned `TempDir` is bound to `_tmp_dir` in each test, so it lives until end of scope. The `BucketManager` (held by the `CatchupManager`) is dropped before `_tmp_dir` because tuple fields drop in declaration order — manager first, then `TempDir` — which is the desired order (no use-after-delete).
- **Naming with leading underscore.** `_tmp_dir` correctly silences unused-variable warnings while still binding (unlike `_`, which would drop immediately and reintroduce the original bug). Good choice.
- **Behavior preserved.** The three `verify_knit_to_lcl` tests still construct an identical `CatchupManager`; the only observable change is that the bucket directory is removed at test exit instead of leaked under the system temp dir. No assertions or test logic are altered.
- **Doc comment** on the helper clearly explains the lifetime contract for future callers — nice touch.
- **Error handling** unchanged; still uses `.expect(...)` which is appropriate in tests.
- **No production code touched**, so no parity / determinism implications.
- Nit (non-blocking): if more tests are added later that need this manager, the tuple return is slightly awkward; a small `TestCatchup { manager, _tmp: TempDir }` struct could be cleaner, but for three call sites the tuple is fine.

LGTM.
</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Test-only change with no production impact, no API/data-format/security/operational risk. Helper signature change is local to the `replay.rs` test module; all three callers were updated.

<details>

- **Scope:** Touches only `crates/history/src/catchup/replay.rs` inside `mod tests`. No production code path, public API, XDR type, DB schema, or config is affected.
- **Helper signature ripple:** `make_test_catchup_manager` is a private function inside the `replay.rs` `mod tests` block. The two namesakes in `download.rs` and `persist.rs` are independent private helpers in their own test modules — no rename/import collision risk. The diff updates all three call sites in `replay.rs` to destructure the new tuple. Confirmed no other callers in the workspace.
- **Behavioral change:** `BucketManager::new(tmp_dir.path().to_path_buf())` instead of `tmp_dir.keep()`. Same path string is handed to `BucketManager`; the only difference is that `TempDir`'s drop now removes the directory at end of test. This matches the fix described in #2745.
- **Test runtime / parallelism:** Each test still gets a fresh `tempfile::tempdir()` (unique dir per call), so concurrent `cargo test` runs don't collide. Cleanup happens on drop of `_tmp_dir`, which is bound for the full test scope (`_tmp_dir`, not `_`, so it is not dropped immediately — good).
- **Performance / CI:** Net positive — stops leaking bucket directories under the system temp dir on every CI run. No measurable runtime cost.
- **Security / ops:** None. No config, no migration, no on-disk format, no network surface.
- **Out-of-scope observation (not blocking):** `download.rs` and `persist.rs` use the same `tmp_dir.keep()` pattern and presumably leak the same way. Worth a follow-up issue, but explicitly out of scope for this PR per its stated focus.
- **Doc comment:** The new doc comment on the helper accurately explains why callers must bind the `TempDir`. Good defensive documentation against the bug recurring.

No risk concerns. Approve.

</details>

PR #2753, opened 2026-05-16
Write §4.3 ledger-zero pseudo-checkpoint when initializing history archive

### LINKED ISSUE #2682: Spec adherence: §4.3 ledger-zero pseudo-checkpoint
**From #2675 (deferred item 10)**

§4.3 specifies ledger-zero pseudo-checkpoint handling for genesis catchup. Requires CatchupRange API changes.

**stellar-core ref**: CATCHUP_SPEC.md §4.3
**Parent**: #2675

tomerweller:
Closes #2682

## Summary

`HistoryArchiveManager::initialize_history_archive` now writes the empty `HistoryArchiveState` JSON to **two** files in the archive — the §4.3 ledger-zero pseudo-checkpoint at `history/00/00/00/history-00000000.json` (the "archive initialized" marker described in stellar-core's `HistoryManager.h` §4.3) **and** the well-known root HAS at `.well-known/stellar-history.json`. Previously only the latter was written, so newly created archives were missing the pseudo-checkpoint marker that stellar-core's `PutHistoryArchiveStateWork::spawnPublishWork` produces.

The empty HAS itself is also realigned with stellar-core's default `HistoryArchiveState` constructor (`HistoryArchive.cpp`): `version = 1` with `hot_archive_buckets = None`. The existing `#[serde(skip_serializing_if = "Option::is_none")]` on `hot_archive_buckets` causes the field to be omitted from the JSON entirely, matching upstream behavior for `version < 2`.

A new `paths::pseudo_checkpoint_has_path()` helper returns the literal `history/00/00/00/history-00000000.json` string. It is **not** routed through `has_path(0)` / `checkpoint_path` because `checkpoint_ledger(0) == 63` would otherwise emit `history-0000003f.json`. A regression unit test pins the literal.

No `CatchupRange` change — the original triage note was misattributed.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/2682#issuecomment-4465119860)

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test -p henyey-history` — 501 lib tests + 12 integration tests pass, including 6 new ones:
  - `paths::tests::test_pseudo_checkpoint_has_path_literal`
  - `archive_manager_tests::test_initialize_history_archive_writes_pseudo_checkpoint`
  - `archive_manager_tests::test_initialize_history_archive_writes_root_has`
  - `archive_manager_tests::test_initialize_history_archive_files_match`
  - `archive_manager_tests::test_initialize_history_archive_pseudo_checkpoint_shape`
  - `archive_manager_tests::test_initialize_history_archive_already_initialized`

## Deviations from plan

None.

### DIFF
--- a/crates/history/src/lib.rs
+++ b/crates/history/src/lib.rs
@@ -620,8 +620,21 @@ impl HistoryArchiveManager {
 
     /// Initialize a new history archive with an empty HAS.
     ///
-    /// This creates a new archive by writing an empty `HistoryArchiveState`
-    /// to `.well-known/stellar-history.json`.
+    /// Mirrors stellar-core's `HistoryArchiveManager::initializeHistoryArchive`
+    /// (and its `PutHistoryArchiveStateWork::spawnPublishWork` upload step):
+    /// writes the same empty `HistoryArchiveState` JSON to **two** files in
+    /// the archive:
+    ///
+    /// 1. The §4.3 ledger-zero pseudo-checkpoint at
+    ///    `history/00/00/00/history-00000000.json` — the marker that signals
+    ///    the archive has been initialized (see `HistoryManager.h` §4.3).
+    /// 2. The well-known root HAS at `.well-known/stellar-history.json` —
+    ///    the file `GetHistoryArchiveStateWork(seq=0)` discovers via
+    ///    `isWellKnown(0)`.
+    ///
+    /// The empty HAS uses `version = 1` with `hot_archive_buckets = None`,
+    /// matching stellar-core's default `HistoryArchiveState` constructor in
+    /// `HistoryArchive.cpp`.
     ///
     /// # Errors
     ///
@@ -662,8 +675,9 @@ impl HistoryArchiveManager {
             return Err(HistoryError::ArchiveNotWritable(name.to_string()));
         }
 
-        // Create an empty HAS with resolved (cleared) futures
-        // All bucket levels have empty curr/snap and cleared next state
+        // Create an empty HAS matching stellar-core's default ctor
+        // (HistoryArchive.cpp): version = 1, no hot-archive buckets,
+        // all-zero curr/snap on every live-bucket level.
         let zero_hash = henyey_common::Hash256::from_bytes([0u8; 32]);
         let empty_level = HASBucketLevel {
             curr: zero_hash.to_hex(),
@@ -672,27 +686,34 @@ impl HistoryArchiveManager {
         };
 
         let has = HistoryArchiveState {
-            version: 2,
+            version: 1,
             server: Some("rs-stellar-core".to_string()),
             current_ledger: 0,
             network_passphrase: Some(self.network_passphrase.clone()),
-            current_buckets: vec![empty_level.clone(); henyey_bucket::BUCKET_LIST_LEVELS],
-            hot_archive_buckets: Some(vec![
-                empty_level;
-                henyey_bucket::HOT_ARCHIVE_BUCKET_LIST_LEVELS
-            ]),
+            current_buckets: vec![empty_level; henyey_bucket::BUCKET_LIST_LEVELS],
+            hot_archive_buckets: None,
         };
 
         // Serialize to JSON
         let json = serde_json::to_string_pretty(&has)?;
 
-        // Write to a temp file
+        // Write to a temp file once; `put_file_with_mkdir` only reads the
+        // local path (it shells out via `execute_command`) and does not
+        // consume the file, so we can upload the same tempfile twice.
         let temp_file = tempfile::NamedTempFile::new()?;
         std::fs::write(temp_file.path(), &json)?;
 
-        // Upload to .well-known/stellar-history.json
+        // Upload to the §4.3 pseudo-checkpoint path first (matches
+        // stellar-core's source order in
+        // PutHistoryArchiveStateWork::spawnPublishWork; functionally the
+        // two writes are independent).
         remote
-            .put_file_with_mkdir(temp_file.path(), ".well-known/stellar-history.json")
+            .put_file_with_mkdir(temp_file.path(), paths::pseudo_checkpoint_has_path())
+            .await?;
+
+        // Upload to the well-known root HAS path.
+        remote
+            .put_file_with_mkdir(temp_file.path(), paths::root_has_path())
             .await?;
 
         tracing::info!(
@@ -860,5 +881,161 @@ mod archive_manager_tests {
 
         // Inert archive is not sensible
         assert!(!manager.check_sensible_config());
+    }
+
+    /// Build a write-only `HistoryArchiveManager` whose put/mkdir commands
+    /// target a local filesystem tempdir. Returns the manager, the tempdir
+    /// (kept alive by the caller), and the archive name.
+    fn make_filesystem_manager(
+        tmp: &tempfile::TempDir,
+        passphrase: &str,
+    ) -> (HistoryArchiveManager, &'static str) {
+        let root = tmp.path().to_string_lossy().into_owned();
+        let put_cmd = format!("cp {{0}} {root}/{{1}}");
+        let mkdir_cmd = format!("mkdir -p {root}/{{0}}");
+
+        let config = RemoteArchiveConfig {
+            name: "test-fs".to_string(),
+            put_cmd: Some(put_cmd),
+            mkdir_cmd: Some(mkdir_cmd),
+            get_cmd: None,
+        };
+        let remote = RemoteArchive::new(config);
+        let mut manager = HistoryArchiveManager::new(passphrase.to_string());
+        manager.add_archive(ArchiveEntry::write_only("test-fs".to_string(), remote));
+        (manager, "test-fs")
+    }
+
+    #[tokio::test]
+    async fn test_initialize_history_archive_writes_pseudo_checkpoint() {
+        let tmp = tempfile::tempdir().unwrap();
+        let (manager, name) = make_filesystem_manager(&tmp, "test passphrase");
+
+        manager.initialize_history_archive(name).await.unwrap();
+
+        let pseudo = tmp.path().join("history/00/00/00/history-00000000.json");
+        assert!(
+            pseudo.exists(),
+            "expected pseudo-checkpoint at {}",
+            pseudo.display()
+        );
+    }
+
+    #[tokio::test]
+    async fn test_initialize_history_archive_writes_root_has() {
+        let tmp = tempfile::tempdir().unwrap();
+        let (manager, name) = make_filesystem_manager(&tmp, "test passphrase");
+
+        manager.initialize_history_archive(name).await.unwrap();
+
+        let root = tmp.path().join(".well-known/stellar-history.json");
+        assert!(root.exists(), "expected root HAS at {}", root.display());
+    }
+
+    #[tokio::test]
+    async fn test_initialize_history_archive_files_match() {
+        let tmp = tempfile::tempdir().unwrap();
+        let (manager, name) = make_filesystem_manager(&tmp, "test passphrase");
+
+        manager.initialize_history_archive(name).await.unwrap();
+
+        let pseudo_bytes =
+            std::fs::read(tmp.path().join("history/00/00/00/history-00000000.json")).unwrap();
+        let root_bytes =
+            std::fs::read(tmp.path().join(".well-known/stellar-history.json")).unwrap();
+
+        assert_eq!(
+            pseudo_bytes, root_bytes,
+            "pseudo-checkpoint and root HAS must contain identical bytes"
+        );
+    }
+
+    #[tokio::test]
+    async fn test_initialize_history_archive_pseudo_checkpoint_shape() {
+        let tmp = tempfile::tempdir().unwrap();
+        let passphrase = "test passphrase";
+        let (manager, name) = make_filesystem_manager(&tmp, passphrase);
+
+        manager.initialize_history_archive(name).await.unwrap();
+
+        let pseudo_path = tmp.path().join("history/00/00/00/history-00000000.json");
+        let json = std::fs::read_to_string(&pseudo_path).unwrap();
+        let has: HistoryArchiveState = serde_json::from_str(&json).unwrap();
+
+        assert_eq!(has.version, 1, "empty HAS must use version 1");
+        assert_eq!(has.current_ledger, 0);
+        assert_eq!(has.server.as_deref(), Some("rs-stellar-core"));
+        assert_eq!(has.network_passphrase.as_deref(), Some(passphrase));
+        assert!(
+            has.hot_archive_buckets.is_none(),
+            "version-1 HAS must omit hot_archive_buckets"
+        );
+
+        // Confirm the field is omitted from the JSON entirely (not just null).
+        assert!(
+            !json.contains("hotArchiveBuckets"),
+            "JSON must omit hotArchiveBuckets field, got: {json}"
+        );
+
+        assert_eq!(
+            has.current_buckets.len(),
+            henyey_bucket::BUCKET_LIST_LEVELS,
+            "current_buckets must have one entry per live bucket level"
+        );
+        let zero_hex = henyey_common::Hash256::from_bytes([0u8; 32]).to_hex();
+        for (i, level) in has.current_buckets.iter().enumerate() {
+            assert_eq!(level.curr, zero_hex, "level {i} curr must be all-zero");
+            assert_eq!(level.snap, zero_hex, "level {i} snap must be all-zero");
+        }
+    }
+
+    #[tokio::test]
+    async fn test_initialize_history_archive_already_initialized() {
+        let tmp = tempfile::tempdir().unwrap();
+        let passphrase = "test passphrase";
+
+        // Pre-create the well-known root HAS so the fetch_root_has probe succeeds.
+        let well_known_dir = tmp.path().join(".well-known");
+        std::fs::create_dir_all(&well_known_dir).unwrap();
+        let existing = b"{\"version\":1,\"server\":\"preexisting\",\"currentLedger\":0,\"networkPassphrase\":\"x\",\"currentBuckets\":[]}";
+        std::fs::write(well_known_dir.join("stellar-history.json"), existing).unwrap();
+
+        // Manager with both read (file://) and write (cp) configured.
+        let root = tmp.path().to_string_lossy().into_owned();
+        let put_cmd = format!("cp {{0}} {root}/{{1}}");
+        let mkdir_cmd = format!("mkdir -p {root}/{{0}}");
+        let get_cmd = format!("cp {root}/{{0}} {{1}}");
+
+        let archive = HistoryArchive::new(&format!("file://{root}")).unwrap();
+        let remote_config = RemoteArchiveConfig {
+            name: "test-fs".to_string(),
+            put_cmd: Some(put_cmd),
+            mkdir_cmd: Some(mkdir_cmd),
+            get_cmd: Some(get_cmd),
+        };
+        let remote = RemoteArchive::new(remote_config);
+
+        let mut manager = HistoryArchiveManager::new(passphrase.to_string());
+        manager.add_archive(ArchiveEntry::new(
+            "test-fs".to_string(),
+            Some(archive),
+            Some(remote),
+        ));
+
+        let err = manager
+            .initialize_history_archive("test-fs")
+            .await
+            .expect_err("should refuse to overwrite an initialized archive");
+        assert!(
+            matches!(err, HistoryError::ArchiveAlreadyInitialized(_)),
+            "expected ArchiveAlreadyInitialized, got {err:?}"
+        );
+
+        // The pseudo-checkpoint must NOT have been created.
+        let pseudo = tmp.path().join("history/00/00/00/history-00000000.json");
+        assert!(
+            !pseudo.exists(),
+            "init must not write pseudo-checkpoint when archive is already initialized"
+        );
     }
 }
--- a/crates/henyey/src/main.rs
+++ b/crates/henyey/src/main.rs
@@ -2145,9 +2145,12 @@ fn cmd_force_scp(config: &AppConfig) -> anyhow::Result<()> {
 
 /// Initialize a named history archive.
 ///
-/// Creates the `.well-known/stellar-history.json` file in the archive using
-/// the archive's configured put/mkdir commands. This matches stellar-core's
-/// `new-hist` command for local filesystem archives.
+/// Creates two files in the archive — the §4.3 ledger-zero pseudo-checkpoint
+/// at `history/00/00/00/history-00000000.json` and the well-known root HAS at
+/// `.well-known/stellar-history.json` — using the archive's configured
+/// put/mkdir commands. This matches stellar-core's `new-hist` command for
+/// local filesystem archives (`HistoryArchiveManager::initializeHistoryArchive`
+/// + `PutHistoryArchiveStateWork::spawnPublishWork`).
 async fn cmd_new_hist(config: &AppConfig, name: &str) -> anyhow::Result<()> {
     use henyey_history::{
         remote_archive::RemoteArchiveConfig, ArchiveEntry, HistoryArchiveManager, RemoteArchive,
--- a/crates/history/PARITY_STATUS.md
+++ b/crates/history/PARITY_STATUS.md
@@ -111,7 +111,7 @@ Corresponds to: `HistoryArchiveManager.h`
 | `selectRandomReadableHistoryArchive()` | `get_readable_archives()` | Partial |
 | `getHistoryArchiveReportWork()` | -- | None |
 | `getCheckLedgerHeaderWork()` | -- | None |
-| `initializeHistoryArchive()` | `initialize_history_archive()` | Full |
+| `initializeHistoryArchive()` | `initialize_history_archive()` | Full — writes both the §4.3 ledger-zero pseudo-checkpoint (`history/00/00/00/history-00000000.json`) and the well-known root HAS, matching `PutHistoryArchiveStateWork::spawnPublishWork`. |
 | `publishEnabled()` | `publish_enabled()` | Full |
 | `getHistoryArchive()` | `get_archive()` | Full |
 | `getWritableHistoryArchives()` | `get_writable_archives()` | Full |
--- a/crates/history/src/paths.rs
+++ b/crates/history/src/paths.rs
@@ -133,6 +133,23 @@ pub fn has_path(ledger: u32) -> String {
     checkpoint_path("history", ledger, "json")
 }
 
+/// Generate the path for the §4.3 ledger-zero pseudo-checkpoint HAS file.
+///
+/// Returns the literal `"history/00/00/00/history-00000000.json"`.
+///
+/// This intentionally does NOT route through [`has_path`]/[`checkpoint_path`]
+/// because [`crate::checkpoint::checkpoint_ledger`] rounds `0` up to the first
+/// real checkpoint (`63 == 0x3f`), which would yield
+/// `history-0000003f.json` — the wrong file.
+///
+/// The pseudo-checkpoint is written by `initialize_history_archive` alongside
+/// the well-known root HAS to mark an archive as initialized. See
+/// stellar-core's `PutHistoryArchiveStateWork::spawnPublishWork` and
+/// `HistoryManager.h` §4.3.
+pub fn pseudo_checkpoint_has_path() -> &'static str {
+    "history/00/00/00/history-00000000.json"
+}
+
 // ============================================================================
 // Dirty file helpers for crash-safe checkpoint building
 // ============================================================================
@@ -271,6 +288,18 @@ mod tests {
         let path = has_path(0xaabbcc00 + 63);
         assert!(path.starts_with("history/"));
         assert!(path.ends_with(".json"));
+    }
+
+    #[test]
+    fn test_pseudo_checkpoint_has_path_literal() {
+        // Regression guard: this path MUST be the literal ledger-0 file, not
+        // the rounded-up `history-0000003f.json` that `has_path(0)` would
+        // produce via `checkpoint_ledger(0) == 63`.
+        assert_eq!(
+            pseudo_checkpoint_has_path(),
+            "history/00/00/00/history-00000000.json"
+        );
+        assert_ne!(pseudo_checkpoint_has_path(), has_path(0));
     }
 
     #[test]


tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Behavior change is small, scoped to fresh archive initialization, and aligns precisely with stellar-core's `initializeHistoryArchive` + `PutHistoryArchiveStateWork::spawnPublishWork`. No regression, perf, security, or API risks identified.

<details>

**Regression risk — low.**
- Reader-side: HAS reader (`HistoryArchiveState::validate`, `has_hot_archive_buckets`, catchup `buckets.rs:378`) already handles v1 (no `hotArchiveBuckets`) as a first-class case — `has_hot_archive_buckets()` returns false for v1 and downstream code branches on it. The new shape is in the supported set, not a new edge case.
- Writer-side: only `initialize_history_archive` is touched; ongoing publish flow still emits whatever version the live HAS produces. No change to checkpoint publishing.
- The `tempfile::NamedTempFile` is now consumed by two sequential `put_file_with_mkdir` calls. `put_file_with_mkdir` shells out via `cp`/`mkdir` placeholders and does not move/consume the file, so the second upload sees the same bytes. The tempfile is dropped only at end of scope — safe.

**Performance — negligible.**
Extra cost on `new-hist` only: one extra `mkdir -p history/00/00/00` and one extra `cp` (or remote PUT). `new-hist` is a one-shot admin command, not on a hot path.

**Breaking changes / data format — acceptable.**
- New empty HAS is `version=1`, omits `hotArchiveBuckets`. This matches stellar-core's default `HistoryArchiveState()` ctor exactly (verified against `HistoryArchive.cpp:517` and `HistoryArchiveManager.cpp:222–226`). Any stellar-core-compatible reader (including henyey's own) handles v1.
- Previously-initialized henyey archives (v2 + empty `hotArchiveBuckets`) remain readable — both shapes pass `validate()` and the v2-with-empty-array form is what core itself emits at protocol < 23.

**Operational concerns — minor, non-blocking.**
- Existing deployments that ran `new-hist` with the old code have a root HAS but **no** `history/00/00/00/history-00000000.json` pseudo-checkpoint marker. This PR does not backfill them. The `already_initialized` check probes the well-known root, so re-running `new-hist` will (correctly) refuse, leaving these archives without the marker. In practice the marker only matters for tools that probe ledger-zero via `GetHistoryArchiveStateWork(seq=0)`; henyey publishes a fresh checkpoint at ledger 63 immediately after, which is what most consumers rely on. Worth a one-line note in release notes ("re-create local archives with `rm -rf <archive> && new-hist` if you need the §4.3 marker") but not blocking — operators can trivially recreate dev archives, and mainnet archives are produced by stellar-core, not henyey.
- Empty-HAS shape change does not affect any persisted henyey state outside the archive itself; nothing to migrate in SQLite or buckets.

**Security — none.** No new inputs, no new parsing paths, tempfile lifetime correct.

**Parity note (orchestrator flagged this as non-parity-critical, but worth recording):** the change is strictly *parity-improving* against stellar-core v26.0.1 — both the dual-write and the v1/no-hot-archive shape match upstream byte-for-byte (modulo the `server` string `"rs-stellar-core"`). No divergence introduced.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Diff is correct, well-documented, and accompanied by focused unit tests (5 new init tests + 1 path regression test) that I ran locally — all pass. Two-file write order matches the source comment, and the pseudo-checkpoint path correctly sidesteps `checkpoint_ledger(0)` rounding.

<details>

**Correctness**
- `pseudo_checkpoint_has_path()` returns the literal `history/00/00/00/history-00000000.json`, which is the right behavior — `has_path(0)` would route through `checkpoint_ledger(0) == 63` and yield `history-0000003f.json`. The dedicated regression test (`test_pseudo_checkpoint_has_path_literal`) explicitly asserts `pseudo != has_path(0)`, locking this in.
- The two `put_file_with_mkdir` calls reuse the same tempfile. The inline comment justifies this (put shells out via `execute_command` and does not consume the file), which is correct against the current `RemoteArchive` implementation.
- HAS shape change (`version: 2 → 1`, `hot_archive_buckets: Some(empty) → None`) is asserted both at the struct level and at the JSON level (`!json.contains("hotArchiveBuckets")`), so the field-omission contract (serde `skip_serializing_if`) is regression-guarded. Good.
- Failure semantics are sound: if the first put (pseudo-checkpoint) succeeds and the second (root HAS) fails, `is_initialized()` (which probes the root HAS) still reports false, so a retry overwrites both files idempotently. No orphan state that breaks future inits.
- The `ArchiveAlreadyInitialized` precondition is preserved — `test_initialize_history_archive_already_initialized` asserts both the error variant and that no pseudo-checkpoint is written when init is refused.

**Test coverage**
- 5 new `#[tokio::test]`s exercise: pseudo file present, root file present, byte-equality between the two files, full HAS shape (version, server, passphrase, current_buckets length, all-zero hashes per level, `hot_archive_buckets` omitted), and the already-initialized refusal path.
- Helper `make_filesystem_manager` keeps each test concise without hiding the substantive assertions. Tempdir is held by the caller so cleanup is automatic.
- Ran `cargo test -p henyey-history --lib archive_manager_tests::test_initialize` and `paths::` locally on the PR head — 5/5 + 9/9 pass.

**Readability**
- Doc comments on `initialize_history_archive`, `pseudo_checkpoint_has_path`, and `cmd_new_hist` all cite the corresponding stellar-core functions/section, making the intent traceable.
- Inline comments on the two upload calls explain the source-order choice and call out that the writes are functionally independent.
- `PARITY_STATUS.md` row updated to describe the dual-file behavior; matches the code.

**Error handling**
- All fallible ops use `?`; the new code adds no silent-recovery paths and no panics outside tests. Tempfile creation, JSON serialization, and both put calls propagate errors as `HistoryError`.

**Minor / non-blocking nits (not requesting changes)**
- `make_filesystem_manager` returns `&'static str` for the archive name; could just be `&str` for symmetry, but it's a trivial style preference.
- The two upload calls could be a tiny loop over `[pseudo_checkpoint_has_path(), root_has_path()]`, but the explicit form is arguably clearer given the source-order comment — fine as-is.
- The "already-initialized" test hand-rolls a minimal HAS JSON literal (`b"{\"version\":1,...}"`); if the schema later gains required fields this will silently keep working only because `fetch_root_has` is permissive. Acceptable today.

</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Low-risk parity fix. The HAS-shape change (v2→v1, drop hot_archive_buckets) and the extra pseudo-checkpoint upload both move toward stellar-core behavior and are unlikely to regress any existing consumer. No security or operational concerns warrant blocking.

<details>

**Regressions in existing behavior**
- `initialize_history_archive` previously wrote only `.well-known/stellar-history.json`; it now writes the pseudo-checkpoint first, then root HAS. Any existing caller still sees the root HAS at the same path, so the public contract is preserved.
- The empty HAS shape change (`version: 2 → 1`, `hot_archive_buckets: Some(empty) → None`) is a JSON-format change. Risk is bounded:
  - The empty HAS only lives until the first real checkpoint is published, at which point it is replaced by a real HAS chosen by `BucketList`-driven publish work — i.e., the format readers actually deal with at steady state is unchanged by this PR.
  - Henyey's own `HistoryArchiveState` deserializer treats `hot_archive_buckets` as `Option`, so v1 round-trips fine.
  - stellar-core's reader treats `version < 2` as omitting hot-archive buckets, which is exactly what we now emit.
  - Any third-party tool (horizon, archivists, custom dashboards) that read v0/v2 with hot buckets would have to hard-require `hotArchiveBuckets` to break — none of the public ones do that I'm aware of, and any tool strict enough to fail closed on a missing optional field would already be incompatible with stellar-core-initialized archives.

**Performance**
- One extra upload (`cp` / `aws s3 cp` / `gsutil cp` etc.) per archive initialization. `initialize_history_archive` is called at most once per archive lifetime via `henyey new-hist`. Negligible.

**Breaking changes to APIs / data formats**
- No public Rust API signature changes. `paths::pseudo_checkpoint_has_path()` is purely additive.
- JSON format change is in the direction of stricter stellar-core parity, not away from it.

**Security**
- None. Same `put_file_with_mkdir` path, same trust boundary, same shell-out semantics. No new secrets, no new network surface.

**Operational concerns**
- *Idempotency / partial-init recovery*: writes are pseudo-checkpoint first, then root HAS. The "already initialized" probe (`fetch_root_has`) only checks the root HAS. So:
  - Crash between writes → root HAS missing → next `new-hist` retries cleanly and re-writes pseudo (overwrite via `cp` is harmless). ✅
  - Pre-existing root HAS but no pseudo (e.g., archive initialized by an older henyey before this PR) → `new-hist` refuses with `ArchiveAlreadyInitialized` and won't backfill the pseudo-checkpoint. Operator workaround: hand-create the file, or delete the root HAS and re-init. Worth a sentence in the PARITY note or `cmd_new_hist` doc, but not blocking — the pseudo-checkpoint is a write-once marker that no production code path in henyey currently reads, and stellar-core will tolerate its absence on archives it didn't initialize itself.
- *Downstream readers*: the new `history/00/00/00/history-00000000.json` file is what stellar-core's `GetHistoryArchiveStateWork(seq=0)` discovers via `isWellKnown(0) → wellKnownRemoteName()` — i.e., stellar-core actually reads the well-known file at seq=0, not the pseudo-checkpoint. The pseudo-checkpoint is purely a §4.3 marker. So this PR doesn't change what any reader fetches; it only makes the on-disk layout match what core would have produced.
- *Migration*: no migration needed for already-running archives; existing checkpoints are untouched.

**Parity sanity check (stellar-core)**
- Confirmed against `stellar-core/src/historywork/PutHistoryArchiveStateWork.cpp::spawnPublishWork` (lines 69–98): two `WorkSequence`s, one to `remoteName(currentLedger)/remoteDir(currentLedger)` (which for ledger 0 is `history/00/00/00/history-00000000.json`), one to `wellKnownRemoteName()`. Both upload the same `mLocalFilename`. Henyey now matches.
- Note: stellar-core dispatches the two sequences in parallel via the work scheduler; henyey runs them sequentially. Functionally equivalent for `new-hist` (a one-shot op) — flagged only for completeness, not as a concern.
- HAS default ctor: `HistoryArchive.cpp` builds with `version = HistoryArchiveState::HISTORY_ARCHIVE_STATE_VERSION_PRE_HOT_ARCHIVE = 1` and no `hotArchiveBuckets`. Matches.

**Tests**
- Five new integration tests cover both files written, byte-equality of the two, JSON shape (v1, omitted `hotArchiveBuckets`, all-zero curr/snap), and the already-initialized refusal path. The path-literal regression test on `pseudo_checkpoint_has_path` correctly pins against the `checkpoint_ledger(0) == 63` foot-gun. Coverage is appropriate for the risk surface.

**Verdict rationale**
- All risk vectors are either bounded (perf, format), neutral (security), or strictly improve parity (regressions). The single residual operational rough edge (no auto-backfill of pseudo on already-initialized archives) is a non-blocking ergonomic note that I'd rather see the team address with a separate `henyey heal-archive`-style command than gate this PR on.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Diff is small, correct, and well-tested — the new `paths::pseudo_checkpoint_has_path()` helper is pinned to the literal, `initialize_history_archive` writes both files from a single tempfile, and the empty HAS shape (version=1, no hot-archive buckets) is verified end-to-end.

<details>

**Logic / correctness**
- The literal `"history/00/00/00/history-00000000.json"` is the right answer; routing through `has_path(0)` would emit `history-0000003f.json` because `checkpoint_ledger(0) == 63`. The `assert_ne!(pseudo_checkpoint_has_path(), has_path(0))` regression guard pins this nicely.
- Switching `version` to 1 and `hot_archive_buckets` to `None` is consistent with how the existing `HistoryArchiveState` serde uses `skip_serializing_if = "Option::is_none"`, and `test_initialize_history_archive_pseudo_checkpoint_shape` verifies the field is dropped from the JSON entirely (not just `null`), which is the only behavior that actually matches a `version < 2` reader.
- Single tempfile reused across both `put_file_with_mkdir` calls is safe: `put_file_with_mkdir` only reads `local_path` (shells out a copy command) and does not move/consume the file. The inline comment calls this out explicitly.
- The already-initialized probe (`fetch_root_has`) is unchanged and still checks only the well-known root HAS, not the pseudo-checkpoint. That's the right choice: if a previous init failed after the pseudo-checkpoint write but before the root-HAS write, a retry can still succeed and re-write both files — i.e., no spurious "already initialized" error from a partial failure. `test_initialize_history_archive_already_initialized` also asserts the pseudo-checkpoint is *not* created when the probe trips, which is the right semantics.

**Error handling**
- All `?` propagation; no silent fallbacks. The two `put_file_with_mkdir` calls both bubble up errors. If the first write succeeds and the second fails, the archive is left with a pseudo-checkpoint but no root HAS — but as noted above, this is recoverable on retry. Matching stellar-core's source order (pseudo-checkpoint first, root HAS second) is the right call.

**Tests**
- Six new tests cover: file existence at the pseudo-checkpoint path, file existence at the root HAS path, byte-equality of the two files, full JSON shape (version, server, passphrase, all-zero curr/snap on every level, omission of `hotArchiveBuckets`), and the already-initialized refusal path. The shape test is particularly thorough — checks the field is omitted from the wire JSON, not just `None` in memory.
- `test_initialize_history_archive_already_initialized` constructs a minimal-but-valid HAS by hand to seed `fetch_root_has`; the inline JSON has `currentBuckets: []`, which deserializes fine. No issues.
- The `paths::tests::test_pseudo_checkpoint_has_path_literal` regression test is exactly the right shape for this kind of "computed path would have been wrong" bug.

**Readability**
- Doc comments on `pseudo_checkpoint_has_path`, `initialize_history_archive`, and `cmd_new_hist` all explain *why* the literal path is used and reference the stellar-core source (`HistoryManager.h` §4.3, `PutHistoryArchiveStateWork::spawnPublishWork`, `HistoryArchive.cpp`). Easy to follow.
- `make_filesystem_manager` returns a `&'static str` name that's just a string literal — slightly odd type signature but harmless. Not worth changing.
- PARITY_STATUS.md note on `initializeHistoryArchive` row is updated with the new behavior.

**Non-blocking observations** (would not justify a CHANGES_REQUESTED)
- The two writes are not atomic. As noted above this is recoverable on retry and matches stellar-core, so it's fine. If a future change wanted to be robust, you could remove the pseudo-checkpoint on root-HAS-write failure — but stellar-core doesn't do that either, so leaving it as-is is the right call.
- `cmd_new_hist`'s updated docstring is accurate but slightly long; not worth touching.

</details>

crates/history/src/lib.rs
```diff
@@ -672,27 +686,34 @@ impl HistoryArchiveManager {
...
+        // PutHistoryArchiveStateWork::spawnPublishWork; functionally the
+        // two writes are independent).
+        remote
+            .put_file_with_mkdir(temp_file.path(), paths::pseudo_checkpoint_has_path())
+            .await?;
+
+        // Upload to the well-known root HAS path.
         remote
-            .put_file_with_mkdir(temp_file.path(), ".well-known/stellar-history.json")
+            .put_file_with_mkdir(temp_file.path(), paths::root_has_path())
```
tomerweller, line 716: Non-blocking — partial-failure / idempotency note:

If the pseudo-checkpoint write succeeds but the root-HAS write fails (e.g. transient network error to a remote archive), the archive is left half-initialized. A retry of `initialize_history_archive` will not be detected as already-initialized (the `fetch_root_has` probe at line ~650 only checks `.well-known/stellar-history.json`, not the pseudo-checkpoint), so the retry will silently overwrite the pseudo-checkpoint with byte-identical content and then write the root HAS. That is harmless because both writes are idempotent and produce the same bytes — but it would be worth either (a) documenting this self-healing behavior on `initialize_history_archive`, or (b) probing both files in the already-initialized check so a half-initialized archive is reported distinctly. Filing as a follow-up.

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** PR faithfully mirrors stellar-core's `HistoryArchiveManager::initializeHistoryArchive` + `PutHistoryArchiveStateWork::spawnPublishWork` and the §4.3 ledger-zero pseudo-checkpoint described in `HistoryManager.h`. Naming, directory layout, file contents, and idempotence semantics all match. A few cosmetic-only divergences noted below.

<details>

**Spec / source consulted**
- `stellar-core/src/history/HistoryManager.h` lines 78–110 — §4.3 narrative for the ledger-zero pseudo-checkpoint.
- `stellar-core/src/history/HistoryArchive.cpp:517` — default `HistoryArchiveState()` ctor.
- `stellar-core/src/history/HistoryArchive.cpp:185–211` — `wellKnownRemoteDir/Name`, `remoteDir/Name(snapshotNumber)`.
- `stellar-core/src/history/HistoryArchive.h:140–183` — cereal `serialize` (gates `hotArchiveBuckets` on `version >= 2`).
- `stellar-core/src/history/HistoryArchiveManager.cpp:200–240` — probe + write flow.
- `stellar-core/src/historywork/PutHistoryArchiveStateWork.cpp:69–100` — dual-write to seq path + well-known.

**Parity checks — pass**
- **Pseudo-checkpoint path** is the literal `history/00/00/00/history-00000000.json`. Matches `remoteDir(0)` (`fs::remoteDir("history", fs::hexStr(0))`) + `remoteName(0)`. The deliberate bypass of `has_path(0)` (which would round to `history-0000003f.json` via `checkpoint_ledger(0)==63`) is correct and pinned by `test_pseudo_checkpoint_has_path_literal`.
- **Both files written, identical bytes** — matches stellar-core's reuse of `mLocalFilename` across the two `PutRemoteFileWork` instances. Verified by `test_initialize_history_archive_files_match`.
- **HAS shape** (`version=1`, no `hotArchiveBuckets`, all-zero `curr`/`snap` over `LiveBucketList::kNumLevels`, `currentLedger=0`, `networkPassphrase=<config>`) matches the default ctor at `HistoryArchive.cpp:517` followed by `has.networkPassphrase = mApp.getConfig().NETWORK_PASSPHRASE` in `initializeHistoryArchive`.
- **`hotArchiveBuckets` JSON omission** (not `null`) matches cereal's `if (hasHotArchiveBuckets()) ar(CEREAL_NVP(hotArchiveBuckets))` gate at `HistoryArchive.h:157`. Test `test_initialize_history_archive_pseudo_checkpoint_shape` asserts the substring is absent.
- **Idempotence / probe semantics** — stellar-core probes via `GetHistoryArchiveStateWork(0, archive, "hist-init", 0)` which resolves to the well-known HAS (seq=0 ⇒ `isWellKnown`). PR's `fetch_root_has`-based probe targets the same well-known file, so the "already initialized" decision uses the same signal. `test_initialize_history_archive_already_initialized` confirms refusal + no pseudo-checkpoint side-effect.
- **§4.3 marker semantics** — no associated `ledger-`/`transactions-`/`results-`/bucket files written for ledger 0; only the HAS JSON. Matches HistoryManager.h §4.3 ("No transaction or ledger-history XDR files are associated with this pseudo-checkpoint").
- **PARITY_STATUS.md** entry updated to reflect dual-write; consistent with mapping to `PutHistoryArchiveStateWork::spawnPublishWork`.

**Minor / non-blocking notes** (posted inline)
- `server: "rs-stellar-core"` literal vs. stellar-core's `STELLAR_CORE_VERSION` — divergence in value but the field is informational; consumers don't validate it. Pre-existing convention in this crate, not introduced by this PR.
- Write ordering: stellar-core spawns the two `WorkSequence`s into the scheduler (potentially concurrent), whereas PR awaits seq-path then well-known. Final on-disk state is identical and the PR comment correctly notes the writes are independent.
- JSON whitespace: `serde_json::to_string_pretty` (2-space) vs. cereal's `JSONOutputArchive` (4-space) — bytes differ but parsed content matches; pseudo-checkpoint is consumed via JSON parse, so this is non-observable for protocol parity.

No spec divergences found. Verdict: APPROVE on parity grounds.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Diff is correct, well-scoped, and well-tested. Pseudo-checkpoint path is pinned to the literal ledger-0 file (avoiding the `checkpoint_ledger(0) == 63` rounding trap), the same tempfile is safely reused for both uploads, and 6 new tests cover existence, byte-equality, JSON shape, the literal-path regression, and the already-initialized refusal path.

<details>

**Correctness — diff**
- `pseudo_checkpoint_has_path()` returns the correct literal `"history/00/00/00/history-00000000.json"`. The doc comment correctly explains why it bypasses `has_path(0)` (since `checkpoint_ledger(0) = (0/64)*64 + 63 = 63` would yield `history-0000003f.json`). Verified against `crates/history/src/checkpoint.rs` and `paths::checkpoint_path`.
- Reusing the single `NamedTempFile` for both uploads is sound: `RemoteArchive::put_file_with_mkdir` → `put_file` only reads `local_path` via a shelled-out command (`execute_command` in `remote_archive.rs:130`); it never moves, truncates, or consumes the file. The `temp_file` binding lives until function return so both `put_file_with_mkdir` calls see a valid path.
- Write order (pseudo-checkpoint first, then root HAS) is deliberate and documented; the `?` early-return on the first put correctly skips the second put.
- The already-initialized guard (`fetch_root_has().is_ok()` on line ~650) is unchanged and still correct for the common path.

**Test coverage**
- Strong: 6 new tests exercise (1) pseudo-checkpoint exists, (2) root HAS exists, (3) both files are byte-identical, (4) JSON shape (version=1, currentLedger=0, server, networkPassphrase, `hotArchiveBuckets` field omitted entirely from JSON, `current_buckets.len() == BUCKET_LIST_LEVELS`, all-zero `curr`/`snap`), (5) `ArchiveAlreadyInitialized` short-circuits and the pseudo-checkpoint is *not* written, and (6) a `paths` regression pinning the literal vs. `has_path(0)`.
- The byte-equality test (`test_initialize_history_archive_files_match`) is the right shape — it locks in the "same content to both paths" invariant without over-specifying JSON formatting.
- The `make_filesystem_manager` helper using `cp`/`mkdir -p` against a tempdir is a clean way to drive the real `put_file_with_mkdir` code path end-to-end (rather than mocking `RemoteArchive`).
- Minor / non-blocking: no test exercises the partial-failure path where the first put succeeds and the second fails (see inline comment). Filing as a follow-up rather than blocking.

**Readability**
- New doc comments on `initialize_history_archive`, `pseudo_checkpoint_has_path`, and `cmd_new_hist` clearly explain the §4.3 rationale and the upstream stellar-core references.
- The "uploaded twice from the same tempfile" comment is exactly the kind of comment the style guide asks for — clarifies a non-obvious lifetime invariant.
- Test names are behavior-named per the project convention (`test_initialize_history_archive_*`).

**Error handling**
- All fallible operations propagate via `?`; no swallowed errors.
- `tempfile::NamedTempFile::new()?` and `std::fs::write(...)?` are correctly handled.
- One non-blocking idempotency observation about the half-initialized state filed as an inline comment on the second `put_file_with_mkdir` call.

**Nits (non-blocking)**
- `make_filesystem_manager` returns `&'static str` for the archive name because it's a string literal — works fine, slightly unusual signature but not worth changing.
- `format!("cp {{0}} {root}/{{1}}")` and friends in the test helper trigger `clippy::uninlined-format-args` only for the `{root}` interpolation (the `{{0}}`/`{{1}}` are escaped braces consumed by `RemoteArchive::format_command`). Current form is fine; no action needed.
</details>

crates/history/src/lib.rs
```diff
@@ -672,27 +686,34 @@ impl HistoryArchiveManager {
         };
 
         let has = HistoryArchiveState {
-            version: 2,
+            version: 1,
             server: Some("rs-stellar-core".to_string()),
             current_ledger: 0,
             network_passphrase: Some(self.network_passphrase.clone()),
```
tomerweller, line 692: **Parity (non-blocking):** stellar-core's default `HistoryArchiveState()` ctor (`HistoryArchive.cpp:517`) sets `server(STELLAR_CORE_VERSION)`. Setting `"rs-stellar-core"` here is a value divergence in the on-disk JSON — informational field, consumers don't validate it, and this matches the pre-existing convention used elsewhere in the crate. Flagging only so it's on the record for the §4.3 marker file.

crates/history/src/lib.rs
```diff
@@ -672,27 +686,34 @@ impl HistoryArchiveManager {
...
+        // Upload to the §4.3 pseudo-checkpoint path first (matches
+        // stellar-core's source order in
+        // PutHistoryArchiveStateWork::spawnPublishWork; functionally the
+        // two writes are independent).
+        remote
+            .put_file_with_mkdir(temp_file.path(), paths::pseudo_checkpoint_has_path())
+            .await?;
+
+        // Upload to the well-known root HAS path.
         remote
```
tomerweller, line 715: **Parity (non-blocking):** stellar-core's `PutHistoryArchiveStateWork::spawnPublishWork` (`PutHistoryArchiveStateWork.cpp:69`) registers two `WorkSequence`s with the scheduler, which may execute concurrently rather than strictly seq-then-well-known. The final archive state is identical, so this is non-observable, but the inline comment's "matches stellar-core's source order" framing is slightly stronger than warranted — upstream ordering is scheduler-defined.

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Risk surface is small and well-contained. The new write is gated by the existing `fetch_root_has` idempotency check, ordering is self-healing, and there is no new injection vector. One operational gap worth tracking (pre-existing henyey-initialized archives won't be retro-fixed) and one cosmetic divergence (HAS version changed from 2 to 1 for *new* archives only).

<details>

### Regression risk — low

- **Fires only on first init.** `initialize_history_archive` still runs the `archive.fetch_root_has().await.is_ok()` guard before any upload; if the root HAS exists, the function returns `ArchiveAlreadyInitialized` and **no** pseudo-checkpoint is written. `test_initialize_history_archive_already_initialized` asserts this explicitly. So the new write does not fire on every init nor on already-populated archives.
- **No change to non-init code paths.** The pseudo-checkpoint path is only referenced from `initialize_history_archive`. Publishing, catchup, GET-HAS, and verify paths are untouched. `pseudo_checkpoint_has_path()` is a brand-new helper, not a reroute of `has_path(0)`, so nothing else can accidentally pick up the new literal.
- **Two-write ordering is self-healing.** If pseudo write succeeds and root-HAS write fails, the next `initialize_history_archive` call sees no root HAS, proceeds, and idempotently re-uploads both files (same bytes — `cp` overwrite). Conversely, both writes use the *same* tempfile; `put_file` shells out via `execute_command` and doesn't consume the file (verified in `remote_archive.rs:178`), so the second upload cannot silently see an empty file.

### Format / parity risk

- **HAS version flip 2 → 1 and `hot_archive_buckets` dropped** is a *deliberate* parity alignment with stellar-core's default `HistoryArchiveState` ctor, but it does change the bytes that henyey's `new-hist` writes today. New archives are now v1 with `hotArchiveBuckets` omitted; old v2 archives written by previous henyey builds remain v2 in place (no migration). Heterogeneous v1/v2 archives across operators is acceptable since stellar-core itself reads both, but worth being aware of.
- The `serde(skip_serializing_if = "Option::is_none")` claim on `hot_archive_buckets` is verified in `test_initialize_history_archive_pseudo_checkpoint_shape` via `assert!(!json.contains("hotArchiveBuckets"))` — good regression guard against a future serde-rename or default change re-introducing the field.

### Operational concerns

- **Pre-existing henyey-initialized archives won't be back-filled.** Any archive previously initialized with the old code has a root HAS but no `history/00/00/00/history-00000000.json`. Re-running `henyey new-hist <name>` on such an archive will hit `ArchiveAlreadyInitialized` and refuse to write the missing marker. Operators wanting parity will have to either:
  1. manually `cp` the root HAS to the pseudo-checkpoint path, or
  2. wipe and re-init the archive.
  Neither is destructive nor blocking (stellar-core treats the marker as informational; downstream catchup of a non-genesis range still works because it uses checkpoint-aligned HAS lookups, not the §4.3 marker). Suggest a one-line note in the §4.3 PARITY_STATUS row or release notes calling this out — non-blocking.
- **No `CatchupRange` change despite the issue title.** PR body acknowledges the triage was misattributed. Worth confirming separately that there is no genesis-catchup path in henyey today that *requires* the §4.3 marker to be readable; a quick scan of the history crate shows the marker is only written, never consumed, so this is fine for now. If a future genesis-catchup implementation lands, a follow-up may need `CatchupRange` work — but that is out of scope here.

### Performance — negligible

- One extra `mkdir -p history/00/00/00` + one `cp` of an ~few-KB JSON, exactly once per archive lifetime. No hot-path impact.

### Security — no new surface

- The new literal `"history/00/00/00/history-00000000.json"` is a compile-time constant, not user-controlled, and goes through the same `format_command` substitution as every other archive path. No injection risk beyond what already exists for `put_cmd` / `mkdir_cmd` templates.
- Tempfile is created via `tempfile::NamedTempFile::new()` (mode 0600, unique name), unchanged from before.

### API / breaking changes

- `pub fn pseudo_checkpoint_has_path() -> &'static str` is purely additive.
- `initialize_history_archive` signature unchanged.
- The on-disk JSON shape change (v2 → v1) is the only "breaking" delta and only affects newly created archives. Stellar-core can read both.

### Nits (non-blocking, leave for follow-up)

- The function name `pseudo_checkpoint_has_path()` returns `&'static str` while its sibling `has_path()` returns `String`. Minor API asymmetry; fine as-is given the intentional literal.
- Consider an explicit `tracing::info!` line per uploaded path so operators can see in logs that two files were written (currently only the trailing "Successfully initialized history archive" log is emitted).

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Diff faithfully mirrors stellar-core's `initializeHistoryArchive` + `PutHistoryArchiveStateWork::spawnPublishWork`. New tests cover the §4.3 pseudo-checkpoint write, the byte-for-byte-equal root HAS, the JSON shape, and the already-initialized refusal path. All 6 new tests + the broader `henyey-history` lib suite pass locally.

<details>

**Cross-reference to stellar-core (v26.0.1 submodule):**

- `HistoryArchive.cpp:517` — default `HistoryArchiveState()` ctor sets `version = HISTORY_ARCHIVE_STATE_VERSION_BEFORE_HOT_ARCHIVE = 1` and never touches `hotArchiveBuckets`. The PR's switch from `version: 2` + populated `hot_archive_buckets` to `version: 1` + `None` matches exactly. The serde `skip_serializing_if = "Option::is_none"` correctly mirrors upstream's `version < 2 ⇒ field absent`.
- `HistoryArchiveManager.cpp:200` — `initializeHistoryArchive` probes via `GetHistoryArchiveStateWork(0, ...)` (i.e. the well-known root HAS), then sets `networkPassphrase` and calls `resolveAllFutures()`. The henyey side uses `fetch_root_has().is_ok()` for the same probe and constructs the empty HAS with passphrase + cleared next-state futures. Equivalent.
- `PutHistoryArchiveStateWork.cpp:69` — `spawnPublishWork` writes both `history/ww/xx/yy/history-wwxxyyzz.json` (with `mState.currentLedger == 0` ⇒ literal `history/00/00/00/history-00000000.json` via `fs::remoteName`) and `.well-known/stellar-history.json`. The PR writes both. Order differs from upstream (upstream uses a sequence + parallel work; PR is sequential), but the ordering comment correctly notes this is observationally equivalent for filesystem archives.

**Correctness — passed:**

- `pseudo_checkpoint_has_path()` deliberately bypasses `has_path(0)` (which would round to checkpoint 63 → `history-0000003f.json`). The regression test `test_pseudo_checkpoint_has_path_literal` pins both the literal and the `assert_ne!(_, has_path(0))` mismatch — exactly the right guard.
- Reusing the same `tempfile::NamedTempFile` for two `put_file_with_mkdir` calls is safe: the impl shells out via `cp` and reads the source path each time, so the second upload is independent. Comment in code calls this out.
- `mkdir` of `history/00/00/00` is handled by `put_file_with_mkdir`'s parent-extraction logic (`remote_archive.rs:262`); confirmed it produces a non-empty parent string for the new path.
- `upload.rs::classify` still routes `.well-known/stellar-history.json` to `RootHas` and `history/.../history-*.json` to `CheckpointHas` — the new ledger-zero file is correctly classified as `CheckpointHas` (not silently treated as RootHas).
- Already-initialized precheck stays the same (`fetch_root_has().is_ok()`), so the new pseudo-checkpoint write only happens on a fresh archive — confirmed by `test_initialize_history_archive_already_initialized` asserting the pseudo-checkpoint file is absent.

**Test coverage — passed:**

The 6 new tests collectively exercise: presence of pseudo-checkpoint, presence of root HAS, byte-equality of the two files, full JSON shape (version=1, current_ledger=0, server string, passphrase, all-zero curr/snap on every level, omitted `hotArchiveBuckets` field), and the already-initialized refusal path with negative assertion that no pseudo-checkpoint leaks. This is the right shape for a §4.3 parity regression — a future regression that drops one of the two writes, mis-paths the pseudo-checkpoint, or re-introduces `version: 2` would be caught.

I ran `cargo test -p henyey-history --lib` against the PR branch: all 16 archive_manager + paths tests pass, 485 filtered (full lib suite green).

**Readability — passed:**

- Doc comments on `initialize_history_archive` and `pseudo_checkpoint_has_path()` cite the upstream files (`HistoryArchive.cpp`, `HistoryManager.h §4.3`, `PutHistoryArchiveStateWork::spawnPublishWork`). Easy to audit later.
- `cmd_new_hist` doc comment in `henyey/src/main.rs` updated to reflect the dual-write.
- `PARITY_STATUS.md` row updated with a one-line note pointing at the upstream symbol — exactly the right granularity.

**Error handling — passed:**

- Both writes use `?` propagation; if the pseudo-checkpoint write fails the root HAS write is skipped and the archive is left in an "uninitialized as far as the precheck is concerned" state. This matches upstream behavior where `WorkSequence` failure aborts the publish.
- One mild non-blocker (intentional follow-up territory, not blocking): if the pseudo-checkpoint write **succeeds** but the root HAS write **fails**, the archive ends up with a stray `history/00/00/00/history-00000000.json` and no root HAS. On a retry, `fetch_root_has` will still return `Err` (no root HAS), so the precheck passes and both writes are reattempted — self-healing. Worth a future test, but stellar-core has the same partial-failure shape.

**Minor non-blocking nits (will be auto-filed as follow-ups by /review-pr if left as inline comments — none filed inline since they're sub-issue scale):**

- `pseudo_checkpoint_has_path()` returns `&'static str` while `root_has_path()` also returns `&'static str` — consistent. ✅
- The empty HAS still uses server string `"rs-stellar-core"` rather than a real version. Pre-existing; not in scope.
- Could derive the literal path from `paths::checkpoint_path("history", 0, "json")` if `checkpoint_path` were updated to NOT round, but the current "explicit literal + regression test" is clearer and safer.

No correctness blockers.

</details>

PR #2763, opened 2026-05-16
Make tx_set_tracker reentrancy regression test shard-deterministic

### LINKED ISSUE #2750: Strengthen tx_set_tracker reentrancy regression test to be shard-deterministic
Follow-up from PR #2747 (issue #2744). Non-critical inline review comment that was not addressed before merge.

## Concern

`crates/herder/src/tx_set_tracker.rs:995` —

> MINOR (non-blocking): This regression test isn't a deterministic deadlock repro — whether the pre-fix code hangs depends on whether `first_hash` ([0;32]) and `new_hash` ([0xFF;32]) collide on a DashMap shard, and the shard count is `4 * num_cpus.next_power_of_two()`. On runners where these hashes land on different shards, the unfixed code would pass. Optional hardening: stress with many distinct `new_hash` values in a loop (or compute against DashMap's hasher) so the test fails deterministically pre-fix on every runner. The structural fix is correct regardless — this is purely about strengthening the regression guard.

## Source

[Original review comment](https://github.com/stellar-experimental/henyey/pull/2747#discussion_r3251780662) on PR #2747.

## Severity

tomerweller:
Closes #2750

## Summary

Strengthen `test_request_at_cap_still_increments_existing` so the chosen `first_hash`/`new_hash` constants are *provably* on the same DashMap shard via `DashMap::determine_map` (gated behind dashmap's `raw-api` dev-feature). Previously the collision relied on `num_cpus()` at runtime, so the regression guard for PR #2747 silently lost its teeth on some runners. Also adds a load-bearing `assert_eq!(determine_map(first), determine_map(new))` invariant and expands the inline `REGRESSION GUARD` doc comment.

Test-only change. No production behavior changes.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/2750#issuecomment-4465489565)

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test -p henyey-herder` — 1107 unit + integration tests pass
- [x] **Manual revert-and-verify (per plan §Test plan):** locally reverted the snapshot to the Ref-held form (held `let entry = tracker.pending.get(&first_hash).unwrap();` across `tracker.request(new_hash, 100)` and used `entry.request_count` afterwards). Test deadlocked deterministically and was killed by a 30s `timeout` (exit 143). Restored the snapshot pattern; final committed code uses the snapshot, so this confirms the strengthened test catches the bug PR #2747 fixed.

## Deviations from plan

None.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/crates/herder/Cargo.toml
+++ b/crates/herder/Cargo.toml
@@ -57,3 +57,8 @@ henyey-crypto.workspace = true
 henyey-ledger = { workspace = true, features = ["test-utils"] }
 henyey-scp = { workspace = true, features = ["test-helpers"] }
 sha2.workspace = true
+# `raw-api` exposes `DashMap::determine_map`, used by the
+# `test_request_at_cap_still_increments_existing` regression test to make the
+# shard collision between `first_hash` and `new_hash` deterministic across all
+# runners regardless of `num_cpus()`. Purely additive feature.
+dashmap = { workspace = true, features = ["raw-api"] }
--- a/crates/herder/src/tx_set_tracker.rs
+++ b/crates/herder/src/tx_set_tracker.rs
@@ -998,8 +998,62 @@ mod tests {
             "existing entry should increment even at cap"
         );
 
+        // REGRESSION GUARD for PR #2747 / issues #2744, #2750.
+        //
+        // The pre-fix code held a `dashmap::mapref::one::Ref` into `pending`
+        // across the call to `tracker.request(new_hash, ...)`. `request()`
+        // takes a write lock on the destination shard via `pending.entry(...)`.
+        // When `new_hash` and `first_hash` land on the same DashMap shard,
+        // the still-live read guard from `pending.get(&first_hash)` causes
+        // `parking_lot`'s shard `RwLock::write()` call to deadlock the same
+        // thread.
+        //
+        // DashMap's shard count is `4 * num_cpus().next_power_of_two()`, so
+        // whether two arbitrary hashes share a shard depends on the runner.
+        // To make this regression test fail _deterministically_ on every
+        // runner if the snapshot pattern above is ever reverted, we pick
+        // `new_hash` to be provably on the same shard as `first_hash` using
+        // `DashMap::determine_map` (gated behind dashmap's `raw-api`
+        // dev-feature in `crates/herder/Cargo.toml`).
+        //
+        // Inlining `tracker.pending.get(&first_hash).unwrap().request_count`
+        // into the `assert_eq!` above (or re-binding it to a `let entry =
+        // tracker.pending.get(...)` that outlives the next `tracker.request`
+        // call) resurrects the deadlock.
+        let target_shard = tracker.pending.determine_map(&first_hash);
+        let new_hash = {
+            let mut found = None;
+            for i in 0u32..1_000_000 {
+                let mut bytes = [0xFFu8; 32];
+                bytes[0..4].copy_from_slice(&i.to_le_bytes());
+                let candidate = Hash256::from_bytes(bytes);
+                if candidate == first_hash {
+                    continue;
+                }
+                if tracker.pending.contains_key(&candidate) {
+                    continue;
+                }
+                if tracker.pending.determine_map(&candidate) == target_shard {
+                    found = Some(candidate);
+                    break;
+                }
+            }
+            found.expect(
+                "must find a hash on the same DashMap shard as first_hash within 1M tries; \
+                 shard count is 4 * num_cpus().next_power_of_two() so this should converge in O(shards)",
+            )
+        };
+        // Load-bearing invariant: if this ever fires, the regression guard
+        // below has lost its teeth and the test silently stops catching the
+        // bug it was written for.
+        assert_eq!(
+            tracker.pending.determine_map(&first_hash),
+            tracker.pending.determine_map(&new_hash),
+            "test invariant: first_hash and new_hash must collide on a DashMap shard \
+             so a Ref-held-across-request() regression deadlocks deterministically"
+        );
+
         // A brand new hash should be rejected.
-        let new_hash = Hash256::from_bytes([0xFF; 32]);
         assert!(!tracker.request(new_hash, 100));
         assert!(
             tracker.pending.get(&new_hash).is_none(),


tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** This PR is a test-only change to make an existing regression test in `crates/herder/src/tx_set_tracker.rs` shard-deterministic across runners with different CPU counts. There is no production code change and therefore no parity surface affected vs. stellar-core.

<details>

### Scope of change
- `crates/herder/Cargo.toml`: adds the `raw-api` feature to the `dashmap` **dev-dependency** only. Purely additive, dev-only.
- `crates/herder/src/tx_set_tracker.rs`: changes are entirely inside `mod tests` (the `test_request_at_cap_still_increments_existing` test). The previously hardcoded `let new_hash = Hash256::from_bytes([0xFF; 32]);` is replaced with a search for a hash that lands on the same DashMap shard as `first_hash`, using `DashMap::determine_map` (exposed by `raw-api`).

### Parity assessment
- No production logic in `tx_set_tracker.rs` (or anywhere else) was modified by this diff. The herder/tracker request/snapshot path semantics are unchanged.
- stellar-core has no direct analog to this Rust-specific DashMap shard-collision concern — it is an artifact of the `dashmap` crate's reentrancy model, not of consensus semantics. The corresponding stellar-core code (`PendingEnvelopes` / `HerderImpl::recvTxSet` and friends in `src/herder/`) does not need to be mirrored here because nothing observable changes.
- The `dashmap` `raw-api` feature is only enabled under `[dev-dependencies]`, so it cannot leak into release builds or the binary crate. No risk of accidentally pulling in non-deterministic shard-layout-dependent behavior into the production hot path.

### Test-side correctness sanity check (parity-relevant only)
- The shard-collision search is bounded (`0..1_000_000`) and asserts a load-bearing invariant immediately after, so a future bump of dashmap's shard count or a hash-distribution change will fail loudly rather than silently degrade the regression guard. This is the right failure mode for a parity-adjacent guard test.
- The change does not alter the `Hash256` value space, the request/cap semantics, or any ordering observable to consensus.

### Conclusion
No parity impact. Approving from the parity angle. Correctness/design feedback (e.g. choice of search strategy, magic constants) is deferred to Reviewer A per the split.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The change does exactly what the title and issue #2750 ask: it strengthens the `test_request_at_cap_still_increments_existing` regression test so that `new_hash` is provably on the same DashMap shard as `first_hash` on every runner, regardless of `num_cpus()`. The regression guard for the PR #2747 / issue #2744 deadlock is preserved (and now fires deterministically pre-fix). Code is clear, well-commented, and the new dev-dependency feature is purely additive.

<details>

### Correctness of the diff vs. issue #2750

Issue #2750 asks for the regression test to fail deterministically on every runner instead of relying on accidental shard collision between `[0;32]` and `[0xFF;32]`. The diff:

1. Adds `dashmap` `raw-api` feature as a **dev-dependency only** in `crates/herder/Cargo.toml` — does not change production builds. ✅
2. Computes `target_shard = tracker.pending.determine_map(&first_hash)` and searches for a `new_hash` candidate with the same shard via `determine_map`. ✅
3. Uses `i.to_le_bytes()` into `bytes[0..4]` with the rest `0xFF` — disjoint from the fill loop's hash shape (`bytes[0..2]` varying, rest `0`), so the `contains_key` defensive check is belt-and-suspenders but won't accidentally match a fill entry. ✅
4. Adds an `assert_eq!` invariant that `first_hash` and `new_hash` collide on a shard — load-bearing guard so a future dashmap version change that breaks `determine_map` semantics doesn't silently turn the test into a no-op. ✅
5. Convergence: with 4·next_pow2(num_cpus) shards (typically 16–512), expected iterations to find a collision is ~shards. The 1M cap is comfortably 3+ orders of magnitude above any realistic runner. ✅

### Test coverage — does it still catch the bug?

The original bug fixed in PR #2747 was a `Ref` held across `tracker.request(...)` causing a same-thread `parking_lot` write-lock deadlock when shards collided. The current test:
- Snapshots `request_count` by value via `tracker.pending.get(&first_hash).unwrap().request_count` (the temporary `Ref` drops at statement end). ✅
- Then forces shard collision and calls `tracker.request(new_hash, ...)`. ✅

So if someone reverts the snapshot to `let entry = tracker.pending.get(&first_hash).unwrap();` and uses `entry.request_count` later (the original bug pattern), the `Ref` will live across the `tracker.request(new_hash, 100)` call below and the test will deadlock — **deterministically on every runner**. The regression guard is intact and stronger than before.

Note: the comment block says "Inlining `tracker.pending.get(&first_hash).unwrap().request_count` into the `assert_eq!` above ... resurrects the deadlock." Strictly speaking, inlining into `assert_eq!` would still drop the temporary `Ref` at statement end and **not** deadlock; only the `let entry = ...` rebinding case actually resurrects the bug. Minor wording quibble — the rebinding case the comment also lists is the correct one. Not blocking.

### Readability

- Comments are thorough and explain *why* the shard-collision dance exists — good for future maintainers.
- The `target_shard` / candidate-search block is straightforward and self-contained inside the test.
- Variable naming (`target_shard`, `candidate`, `found`) is clear.
- The Cargo.toml comment explaining the `raw-api` feature is also helpful.

### Error handling

- `.expect("must find a hash on the same DashMap shard ... within 1M tries; ...")` includes the shard-count math so a future failure has actionable context. ✅
- The invariant `assert_eq!` between the two `determine_map` results catches the case where the search loop somehow returns a non-colliding candidate. ✅
- No panics in production code paths — all in `#[cfg(test)]` `mod tests`. ✅

### Nits (non-blocking, suitable as follow-up if anything)

1. Comment wording about the inlined-into-`assert_eq!` case (above). The rebinding case is the real reproduction; inlining alone wouldn't reproduce.
2. `contains_key` defensive check is technically unnecessary given disjoint hash shapes, but harmless.

Neither warrants blocking the merge.

</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** Test-only hardening of an existing regression test in `crates/herder/src/tx_set_tracker.rs`; no production code or observable behavior changes, so stellar-core parity is unaffected.

<details>

- **Scope is test-only.** The diff touches exactly two things:
  - `crates/herder/Cargo.toml`: adds `raw-api` to the `dashmap` **dev-dependency** only. This feature only exposes additional methods (`determine_map`, `shards()`, etc.); it does not alter any runtime data-structure layout or hashing used by production code paths. No parity surface.
  - `crates/herder/src/tx_set_tracker.rs`: changes only code inside `#[cfg(test)] mod tests { … }` (the `test_request_at_cap_still_increments_existing` regression test for PR #2747 / issues #2744, #2750). Production logic (`TxSetTracker::request`, snapshot pattern that fixed the deadlock) is untouched.
- **No semantic divergence introduced.** stellar-core's analogue (`Herder` / `PendingEnvelopes` tx-set fetching, e.g. `stellar-core/src/herder/PendingEnvelopes.cpp`) is C++ and has no DashMap; the deadlock being regression-guarded here is a Rust-/DashMap-specific implementation hazard, not a protocol behavior. Choosing `new_hash` to land on the same DashMap shard as `first_hash` is purely a test-instrumentation concern with no observable effect on tx-set tracker semantics that need to mirror stellar-core.
- **Hash construction is benign.** `Hash256::from_bytes([0xFF; 32])` (old) and `Hash256::from_bytes(bytes)` with `bytes[0..4] = i.to_le_bytes()` and the rest `0xFF` (new) are both arbitrary 32-byte values used only as map keys in this test. They are not fed into any consensus, signing, or ledger-hash computation, so there is no risk of altering hashes observed by the network.
- **`PARITY_STATUS.md` / README parity column** require no update — no functionality added, removed, or behaviorally changed.
- **Determinism / parity invariants** of `TxSetTracker` (`request`, cap behavior, increment-existing-at-cap) are exercised identically before and after; the diff only strengthens the determinism of *the test itself* across runners with different `num_cpus()`. This is exactly the kind of hardening that protects parity-relevant regressions from silently disappearing on certain machines.
- One forward-looking note (non-blocking, parity-adjacent): the comment correctly documents that the live `Ref` was the bug. Keeping that REGRESSION GUARD prose intact in future refactors is what preserves long-term confidence that the snapshot pattern (which is the actual parity-safe behavior of `request()`) cannot silently regress.

Parity verdict: **APPROVE**.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Test hardening is sound — `determine_map` is used correctly, the search loop is well-bounded and guaranteed to converge, the dev-dep `raw-api` feature is correctly scoped, and the precondition assert genuinely load-bears for the deadlock pattern PR #2747 fixed.

<details>

- `tx_set_tracker.rs:1022` — `DashMap::determine_map` (raw-api) returns the shard index a key would hash into; using it on both `first_hash` and each `candidate` is the right primitive for a deterministic shard-collision precondition. No reliance on internal hashing details beyond the public raw-api contract.
- `tx_set_tracker.rs:1023-1043` — Search loop is correctly bounded:
  - Skip-self (`candidate == first_hash`) and skip-already-pending (`contains_key`) are both correct.
  - Candidate space: `bytes[0..4]` varies over `0..1_000_000`, `bytes[4..]` fixed at `0xFF`. None of the pre-fill hashes (which have `bytes[2..] == 0`) can collide, so `contains_key` will only ever true on a freak coincidence with `first_hash` (already filtered). Good.
  - Termination: shard count is `4 * num_cpus().next_power_of_two()`, so on any realistic runner ≤ a few thousand shards, a same-shard candidate is found in O(shards) iterations. 1M ceiling is generous and the `.expect` message is honest about the math.
- `tx_set_tracker.rs:1049-1054` — Post-loop `assert_eq!(determine_map(first), determine_map(new))` is technically redundant with the loop's break condition (the loop only assigns `found` when that equality holds), but it's harmless defense-in-depth and the comment correctly frames it as the load-bearing invariant. Worth keeping.
- Load-bearing check: the *current* test body (line 988-996) snapshots `request_count` by value so the `Ref` drops before the next `tracker.request(new_hash, 100)`. So the test as-written does not deadlock; what the new precondition guarantees is that **if** a future refactor reintroduces the held-`Ref` pattern (e.g. the two examples called out in the doc-comment), the deadlock will fire on every runner regardless of `num_cpus()`, not just CI machines with the "right" CPU count. That's exactly the regression-determinism property issue #2750 asked for.
- `Cargo.toml:60-64` — `raw-api` added only as a dev-dep feature. With workspace `resolver = "2"` (verified at root `Cargo.toml:2`), dev-dep feature flags do not unify into normal builds, so the production library build of `henyey-herder` does not pull in `raw-api`. Scoping is correct, and even if unification did occur the feature is purely additive in dashmap. Comment justifying the feature is helpful.
- Readability: the inline doc-block (`tx_set_tracker.rs:1001-1019`) is excellent — it names the exact PRs/issues, the exact lock primitive (`parking_lot` shard `RwLock::write()`), and gives two concrete code patterns that would resurrect the bug. Future maintainers will understand why this scaffolding exists.
- Error handling: `.expect(...)` on the search is the right call — failure here means the test's premise has collapsed and silent passing would be worse than panicking.
- No subtle correctness issues spotted. No mutation of `tracker` state during the search beyond read-only `determine_map` / `contains_key` calls, so the post-search `tracker.request(new_hash, 100)` call sees the same `pending` state the assertions were written against.

Non-blocking nit (optional, do not block merge): the redundant post-loop `assert_eq!` could become a `debug_assert_eq!` to signal "this is invariant, not a runtime check," but the current form reads more clearly to a reader scanning the test, so leaving it as-is is also fine.

</details>

tomerweller:

**Reviewer A:** APPROVE — `determine_map` used correctly, search loop bounded, dev-dep `raw-api` feature scoped properly, regression assert load-bears against PR #2747's deadlock pattern.
**Reviewer B:** APPROVE — Test-only hardening; touches only `#[cfg(test)]` code and a dev-dep feature flag. No production code, no parity surface affected.
**External reviewers:** copilot-pull-request-reviewer[bot]: COMMENTED (non-blocking overview, no concerns)
**CI:** red — 4 FAILURE checks, all in the `Quickstart` workflow's RPC/horizon integration tests (`build / 3 test / 2 test (testing-with-pr, amd64, local, rpc , )`, `... local, core,rpc,horizon ...`, plus the two `complete` rollups). These are not diff-attributable: this PR only changes `crates/herder/Cargo.toml` (dev-dep feature) and `crates/herder/src/tx_set_tracker.rs` (a `#[cfg(test)]` regression test). Recent runs of the same Quickstart workflow on `main` are also failing (sha f3a599e), confirming the failures are an upstream/infra issue unaffected by this diff.

Routing back to `ready-for-doing` for `/do` Mode B to rebase on `origin/main` and retry. Both agent reviewers and the external bot are clean — only CI needs to flip green. If CI is still red in the same pattern after rebase, the next `/review-pr` will mark `blocked`.

tomerweller:
Rebased on `origin/main` (now at 3b31f4b3). Local `cargo fmt --check`, `cargo clippy --all -- -D warnings`, and `cargo test -p henyey-herder --tests` all green. Bounce-back was for unrelated Quickstart CI flake; force-pushed to retrigger CI. Ready for re-review.

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** Test-only change that makes a Rust-specific DashMap shard collision deterministic. No stellar-core observable behavior is added, removed, or asserted differently — parity-clean.

<details>

- Scope: `crates/herder/Cargo.toml` adds the `raw-api` dev-feature on `dashmap`; `crates/herder/src/tx_set_tracker.rs` only modifies the `test_request_at_cap_still_increments_existing` unit test to pick a `new_hash` that provably lands on the same DashMap shard as `first_hash`.
- The asserted production semantics are unchanged from the pre-PR test: at cap, an existing-hash `request()` increments `request_count`, and a brand-new-hash `request()` is rejected and not inserted. Both assertions match the intent of the tx-set tracker (bounded pending set; idempotent re-request on known hashes).
- No stellar-core parity surface is touched. `DashMap` shard layout / `determine_map` / `parking_lot::RwLock` reentrancy is a Rust concurrency-primitive concern with no analog in stellar-core's `HerderImpl` / `PendingEnvelopes` / `TxSetFrame` (C++ uses different sync primitives and a different data layout entirely). The bug being guarded (#2744 / #2750) is a Rust-side reentrant-write-on-held-read deadlock, not a protocol behavior divergence.
- The "load-bearing invariant" `assert_eq!` on `determine_map` is a test-self-check, not a behavioral assertion about tx-set tracking, so it cannot drift from stellar-core.
- Cap value, hash construction, and `request()` return semantics in the test match the existing `tx_set_tracker.rs` contract; no new edge case is asserted that would need cross-checking against `stellar-core/src/herder/`.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Test-only hardening that does exactly what #2750 asks: makes the shard collision between `first_hash` and `new_hash` provable on every runner via `DashMap::determine_map`, with a load-bearing precondition assert so the regression guard cannot silently rot.

<details>

**Correctness**
- `dashmap::raw-api` feature is correctly scoped to `[dev-dependencies]` in `crates/herder/Cargo.toml` (additive, no production surface). ✓
- `determine_map` used correctly to compute the shard index for both `first_hash` and the candidate `new_hash`. ✓
- Search loop in `crates/herder/src/tx_set_tracker.rs:1021-1041`:
  - Bound of 1,000,000 is wildly generous — DashMap shard count is `4 * num_cpus().next_power_of_two()`, so expected hits in O(shards), well under 1k iterations on any realistic runner.
  - Skips `candidate == first_hash` and `pending.contains_key(&candidate)` so the brand-new-hash assertion below (`pending.get(&new_hash).is_none()`) remains valid.
  - `bytes = [0xFF; 32]` then `bytes[0..4] = i.to_le_bytes()` — these candidates can't collide with the prefilled set (which uses `bytes[0..2]` low + rest zero), but `contains_key` guards regardless.
- The `assert_eq!(determine_map(first_hash), determine_map(new_hash), ...)` invariant is genuinely load-bearing: if a future dashmap version changes shard hashing such that the constructed `new_hash` no longer collides, the test fails loudly instead of silently passing.

**Test coverage**
- The actual regression behavior (`Ref` held across `request()` deadlocks) is what the surrounding test (lines 988-1007) exercises. This PR makes that exercise reliable, which is the right scope.
- No new behavior to test beyond the invariant assert, which itself asserts the test's own precondition.

**Readability**
- Inline comment block (lines 1001-1020) clearly explains the bug being guarded against, why shard collision matters, and what specific code patterns would resurrect the deadlock — exemplary regression-guard documentation.
- The "load-bearing invariant" comment makes the role of the second `assert_eq!` unambiguous.

**Error handling**
- `.expect("...")` on the search result is appropriate for a test; the message correctly cites the math that guarantees convergence.

No blocking concerns. Ready to merge once CI is green.
</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** Test-only change with no production-code impact. Parity-relevant assertions (request-at-cap increments existing, rejects new) are unchanged; the new code only adjusts `new_hash` selection so a henyey-specific DashMap shard collision is deterministic.

<details>

- **No production code touched.** The diff is confined to `crates/herder/src/tx_set_tracker.rs` test module and an additive `dashmap` dev-feature (`raw-api`) in `crates/herder/Cargo.toml`. Nothing observable to consensus / SCP / herder behavior changes, so there is no surface for stellar-core divergence.
- **Bug being guarded is henyey-specific, not a parity hazard.** The reentrancy/deadlock concern stems from holding a `dashmap::Ref` across a `pending.entry(...)` write — a Rust/`parking_lot` shard-lock concern with no analogue in stellar-core's `PendingEnvelopes` / `TxSetTracker` (C++ uses different containers and locking discipline). Making the regression test fail deterministically does not encode any new semantic that diverges from stellar-core.
- **Parity-relevant assertions are preserved.** The semantics under test — "at cap, an existing pending hash still increments its request_count; a brand-new hash is rejected and not inserted" — match the spirit of stellar-core's herder pending-tx-set tracking (cap enforcement on novel entries, refcount bumps on known entries). The PR only changes *which* `new_hash` is used to exercise the rejection path; it does not weaken or alter the assertions themselves.
- **`determine_map` use is test-scaffolding only.** Gated behind `raw-api` as a dev-dependency feature; will not leak into the production build or affect runtime determinism of ledger close.
- **No PARITY_STATUS.md update needed.** No behavior toward stellar-core changed.

No parity concerns.
</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Diff makes `test_request_at_cap_still_increments_existing` shard-deterministic via `DashMap::determine_map` and a load-bearing `assert_eq!` invariant; the snapshot-then-drop pattern under test is preserved, so a regression to the held-Ref bug from PR #2747 will deadlock on every runner.

<details>

- **Reproduces the original bug.** The code under test still snapshots `request_count` by value on a single statement (`let request_count = tracker.pending.get(&first_hash).unwrap().request_count;`) before re-entering `tracker.request(new_hash, 100)`. With `new_hash` now provably on the same shard as `first_hash`, reverting to the held-`Ref` pattern would acquire the shard's `parking_lot` read guard and then attempt a write on the same shard via `pending.entry(...)`, deadlocking the test deterministically. ✅
- **Determinism mechanism is sound.** `determine_map` is `pub` under `feature = "raw-api"` in dashmap 6.x and returns the same shard index used internally by `entry`/`get`, so `determine_map(&a) == determine_map(&b)` is exactly the predicate that forces shard-write contention. The `raw-api` feature is purely additive (no runtime behavior change) and scoped to `[dev-dependencies]` only.
- **Search loop is bounded and convergent.** Shard count is `4 * num_cpus().next_power_of_two()` (small power-of-two, e.g. 16–512 in practice), so a uniformly-distributed hash hits the target shard on average in O(shards) tries; the 1M cap is a paranoid upper bound. The loop also correctly skips `candidate == first_hash` and `contains_key(&candidate)` so it cannot accidentally pick an existing entry (which would change the assertion semantics from "rejected at cap" to "incremented existing").
- **Load-bearing assertion is correct.** The `assert_eq!(determine_map(first_hash), determine_map(new_hash), ...)` is redundant given the loop's exit condition, but it's the right kind of redundant — if a future refactor breaks the search invariant the test fails loudly instead of silently degrading to the old probabilistic behavior.
- **`.expect()` message** correctly describes the convergence argument; failure here would indicate either an upstream dashmap shard-count regression or a search bug, both of which warrant a hard panic.
- **Imports / deps minimal.** Single feature flag added (`dashmap = { workspace = true, features = ["raw-api"] }`) under `[dev-dependencies]`, with an inline comment justifying it. No new `use` statements; `determine_map` is called as a method. Cargo.toml addition is appropriately scoped.
- **Readability.** ~50 lines of comment for ~25 lines of test code is high but justified — this is exactly the kind of test where a future maintainer needs to understand *why* the contortion exists, otherwise they'll "simplify" it back to `[0xFF; 32]` and silently lose coverage. Comment correctly cites PR #2747 / issues #2744, #2750.
- **No correctness defects spotted.** The brand-new-hash assertion (`tracker.pending.get(&new_hash).is_none()` after `tracker.request(new_hash, 100)` returning false) still validates the at-cap-rejection path on a different hash than `first_hash`, which is the original behavior being asserted.
- **Nit (non-blocking):** `Hash256::from_bytes(bytes)` allocation inside the hot loop is fine — the loop is bounded and the test is single-shot. No perf concern.

Both prior bounce-back cycles already approved this on correctness; the current head is unchanged on the diff (rebase-only) and CI failures were diff-unrelated infra. Approving.

</details>

tomerweller:
## Review: Bounce-Back Cycle 2

**Reason:** CI failed (unrelated to diff)

**Reviewer A:** APPROVE — Diff correctly makes the regression test shard-deterministic via `DashMap::determine_map`; bounded search loop is convergent; load-bearing `assert_eq!` invariant guards against revert.
**Reviewer B:** APPROVE — Test-only change; parity-relevant assertions (cap behavior) unchanged. DashMap shard-collision determinism is a henyey/Rust concern with no stellar-core analogue, so no divergence introduced.
**External reviewers:** copilot-pull-request-reviewer: COMMENTED (non-blocking)
**CI:** red (unrelated)

### CI failure analysis

Failing checks are all in the `Quickstart` workflow:
- `build / 3 test / 2 test (testing-with-pr, amd64, local, rpc , )`
- `build / 3 test / 2 test (testing-with-pr, amd64, local, core,rpc,horizon , )`
- `build / 3 test / complete`
- `build / complete`

Root cause from the failed-job log: `stellar-rpc` repeatedly exits with status 1 ("supervisor: WARN exited: stellar-rpc (exit status 1; not expected)"), the friendbot test waits on `stellar-rpc` to become ready, and the action times out after 4 minutes:

```
##[error]The action 'Run friendbot test' has timed out after 4 minutes.
```

This PR's diff is **test-only and scoped to `crates/herder/`**:
- `crates/herder/Cargo.toml` (+5)
- `crates/herder/src/tx_set_tracker.rs` (+55 / -1)

It cannot affect the `stellar-rpc` Docker image's startup behavior in the Quickstart workflow — that pipeline builds and runs upstream `stellar/stellar-rpc` images independently of any henyey crate. The Rust-side checks (`Test`, `Doctest`, `Build (release-ci)`, `Clippy`, `Format`) are all green.

The failure is therefore **unrelated infrastructure flake** (or genuine upstream stellar-rpc issue on `main`).

### Action for `/do` Mode B

Rebase the branch on `origin/main` and push. No code changes required.

```bash
git fetch origin main
git rebase origin/main
git push --force-with-lease
```

If CI is still red on the same Quickstart `local, rpc` checks after the rebase with no diff-attributable cause, the next `/review-pr` will mark this `blocked` per the cycle-cap rule.

tomerweller:
Rebased on `origin/main` (now at 483d6f8a). No code changes — bounce-back was for unrelated Quickstart CI flake. Local verification:
- `cargo fmt --check` ✅
- `cargo test -p henyey-herder --lib tx_set_tracker` — 32 passed
- Pre-existing herder clippy warnings on `origin/main` (`useless_vec`, `too_many_arguments` in `flood_queue.rs` and `overlay/peer.rs`) are unrelated to this test-only diff.

Ready for re-review.

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** Test-only change. No production code paths are modified, and the regression guard targets a Rust/DashMap-specific reentrancy hazard that has no analogue in stellar-core's C++ `HerderImpl`/`PendingEnvelopes` path.

<details>

No parity concerns.

- The diff is confined to `crates/herder/src/tx_set_tracker.rs` (inside `#[cfg(test)] mod tests`) and a dev-dep feature flag (`dashmap/raw-api`) in `crates/herder/Cargo.toml`. Neither touches observable behavior of `tx_set_tracker` at runtime.
- The hazard being regression-tested (holding a `dashmap::mapref::one::Ref` shard read guard across a re-entrant `pending.entry(...)` write on the same shard) is an artifact of our concurrent map choice. Stellar-core's equivalent state (`PendingEnvelopes` / `TxSetFrameCache`) uses `std::map`/`UnorderedMap` under a coarser mutex and has no sharded-RwLock reentrancy concept to mirror, so there is nothing to diverge from on the C++ side.
- `DashMap::determine_map` is used only to pick a colliding key for the test; it does not influence dispatch, ordering, or any value observable to consensus. The brute-force search loop is bounded and the `assert_eq!` invariant guarantees the test fails loudly if the collision construction ever stops working — it cannot silently weaken parity coverage.
- The semantics being asserted (existing-entry increment-at-cap + reject-new-at-cap) are unchanged from the prior version of the test; only the choice of `new_hash` bytes changed, and the assertions about `request()` return value and `pending` membership remain identical. These are the same semantics stellar-core enforces for its pending-fetch caps.
- No protocol, ledger, XDR, or sequencing surface is touched. `PARITY_STATUS.md` does not need updating.

Non-blocking nit (inline-worthy, not parity): the 1M upper bound on the search loop is fine in practice but slightly overstated in the `.expect` message — convergence is O(shards) in expectation, not bounded by 1M. Doesn't affect correctness or parity.
</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Test-only change makes the PR #2747 reentrancy regression guard deterministic by selecting `new_hash` to provably share a DashMap shard with `first_hash`. Diff is correct, safe, well-documented, and the test passes locally.

<details>

**Diff correctness**
- `dashmap = { workspace = true, features = ["raw-api"] }` is added under `[dev-dependencies]`. With workspace `resolver = "2"` (confirmed in root `Cargo.toml`), the dev-only `raw-api` feature does **not** leak into normal (non-test) builds of `henyey-herder` or its reverse dependencies. Good.
- `DashMap::determine_map(&K) -> usize` from `raw-api` is the right primitive: it returns the shard index used by the hash, so equal `determine_map` results guarantee the same shard ⇒ guarantee the deadlock signature the test is guarding against.
- The hash-search loop is correct:
  - Skips the candidate that equals `first_hash` and any candidate already in `pending` (defensive — under the current pre-fill all pre-filled hashes have `bytes[2..] == 0`, while candidates have `bytes[4..] == 0xFF`, so they can't actually collide, but the check is cheap and future-proofs against pre-fill changes).
  - Bound of `1_000_000` iterations is wildly safe: with ~`4 * num_cpus().next_power_of_two()` shards (≤ a few hundred on any realistic runner), the expected number of tries to land on a given shard is O(shards). No risk of `expect` firing in practice.
  - `.expect(...)` (not silent `unwrap`) gives a clear, actionable failure message if it ever does.
- The load-bearing `assert_eq!(determine_map(first), determine_map(new))` is genuinely load-bearing: if a future dashmap refactor changes the semantics of `determine_map`, the test fails loudly rather than silently degrading.
- The pre-existing snapshot pattern (`let request_count = tracker.pending.get(&first_hash).unwrap().request_count;`) is preserved unchanged, so the bug-under-test (Ref held across `request()`) is still what the test is actually exercising.

**Test coverage**
- Verified locally: `cargo test -p henyey-herder --lib tx_set_tracker::tests::test_request_at_cap_still_increments_existing` passes in ~0.00s.
- PR description documents a manual revert-and-verify (Ref held across `request`) producing a deterministic 30s-timeout deadlock, which is the right empirical confirmation that the strengthened test catches the bug.

**Readability**
- The REGRESSION GUARD doc comment is excellent — names the PR/issues, explains the mechanism (parking_lot shard `RwLock::write` while a `Ref` read-guard is alive), and explicitly calls out the two ways to resurrect the deadlock. Future maintainers will know exactly what not to touch.
- Cargo.toml comment justifies the `raw-api` feature inline; no mystery dependency.

**Error handling**
- `.expect` with a descriptive message is appropriate for a test-only invariant.
- No panics, unwraps, or silent fallbacks in any production path (this is dev-only code).

**Non-blocking nits** (not worth a bounce; flagging here only)
- The error message mentions "1M tries"; if a future maintainer drops the bound the message will drift. Minor.
- `i` could iterate `0u32..` and `break` purely on the shard match; the explicit cap is fine and arguably clearer.

No blocking issues. Looks good to merge.

</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** Test-only change; production `tx_set_tracker` logic byte-identical, and the guarded hazard is a Rust/DashMap-specific reentrancy with no stellar-core analog, so there is no parity surface to drift.

<details>

- **Production code unchanged.** The diff in `crates/herder/src/tx_set_tracker.rs` is confined to the `#[cfg(test)] mod tests` block of `test_request_at_cap_still_increments_existing`. The `TxSetTracker::request` / `pending` / cap-eviction logic — i.e. the only code paths that could observe a parity-relevant semantic — are untouched.
- **No stellar-core analog for the guarded bug.** `tx_set_tracker` is a henyey-side concurrency abstraction over `DashMap<Hash256, PendingEntry>` (sharded `parking_lot::RwLock`s). stellar-core's equivalent state (`PendingEnvelopes::mTxSetFetcher` / `ItemFetcher<TxSetFrame>` in `src/herder/`) is a single-threaded `std::map` guarded by the main-thread invariant; it cannot exhibit the "Ref held across write to same shard" deadlock this test pins. So the regression guard is correctly scoped to a Rust-implementation hazard, not a protocol/consensus behavior — there is nothing in v26.0.1 stellar-core to mirror or diverge from.
- **`raw-api` dev-feature is parity-safe.** `dashmap = { workspace = true, features = ["raw-api"] }` is added under `[dev-dependencies]` only (confirmed in `crates/herder/Cargo.toml` lines 53–61). `raw-api` is purely additive — it exposes `determine_map`, `shards`, `_yield_read_shard`, etc. without altering the behavior of the non-raw API used in production. Even with cargo's feature unification, no production code path consumes `determine_map`, so observable runtime behavior of release builds is unchanged.
- **Determinism of the test itself is improved, not loosened.** The new code asserts `determine_map(first_hash) == determine_map(new_hash)` as a load-bearing invariant before exercising the reentrancy path, so the regression guard now fires deterministically on every runner regardless of `num_cpus()`. This strictly strengthens coverage of the PR #2747 fix without changing what behavior is being asserted about `TxSetTracker`.
- **No XDR / wire / ledger-close / SCP semantics touched.** Nothing in the diff reaches `apply_*`, hashing, meta emission, or quorum logic, so there is no v25/p25 (or v26 reference) behavior to re-validate.

Verdict: APPROVE — zero parity risk.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Test-only hardening is sound: shard collision is now provable via `determine_map`, the search is deterministic and bounded, and `raw-api` is correctly scoped to dev-deps. Regression test still passes locally.

<details>

- **Shard-collision is genuinely load-bearing.** `tracker.pending.determine_map(&first_hash) == determine_map(&new_hash)` is the exact condition needed to deadlock pre-fix code (re-entering `pending.entry(...)` write-lock on a shard while a `Ref` from `pending.get(&first_hash)` is still alive on the same shard). The `assert_eq!` immediately before `tracker.request(new_hash, 100)` correctly fails-loud if a future change to the search loop or to DashMap's hashing ever stops producing a collision, so the regression guard cannot silently rot.

- **Search is deterministic and bounded.**
  - Iteration `0u32..1_000_000` with `bytes[0..4] = i.to_le_bytes()`, rest `0xFF`, is monotonic and reproducible.
  - DashMap shard count is `4 * num_cpus().next_power_of_two()` — even on absurd 1024-core boxes that's ≤ 4096 shards, so a uniform-ish keyspace converges in O(shards) iterations. 1M cap is comfortably more than enough; the `.expect(...)` failure mode is informative if it ever does trip.
  - Skip clauses (`candidate == first_hash`, `contains_key(&candidate)`) are correct. The `contains_key` is defensively redundant given the keyspace (existing entries have `bytes[2..] = 0`, candidates have `bytes[4..] = 0xFF`), but harmless and future-proof if the fill loop ever changes.

- **`raw-api` is dev-only in practice.** `dashmap` is declared in both `[dependencies]` (no features) and `[dev-dependencies]` (with `raw-api`). Per Cargo's feature-unification rules, dev-dependency features only activate when building tests/examples/benches; downstream consumers and `cargo build`/release of `henyey-herder` do not pull in `raw-api`. The feature is purely additive (exposes `determine_map`/raw shard accessors), so even under unified test builds there is no behavioral risk to the lib itself.

- **REGRESSION GUARD comment.** Excellent — names the originating PR/issues (#2747, #2744, #2750), explains exactly which lock pair deadlocks, names the offending refactor patterns that would resurrect the bug (inlining the field access, or rebinding to a `let entry = ...` that outlives `request()`), and explains why the deterministic-collision construction exists. Future maintainers will not have to reconstruct this from git archaeology.

- **No subtle bugs spotted.** `Hash256::from_bytes([0xFF; 32])` is well-defined for the 32-byte input. The original `new_hash = Hash256::from_bytes([0xFF; 32])` line is correctly replaced (not duplicated), and the post-shard-collision assertions on `tracker.request(new_hash, 100) == false` and `pending.get(&new_hash).is_none()` remain the right behavioral checks for the at-cap reject path.

- Verified locally: `cargo test -p henyey-herder --lib tx_set_tracker::tests::test_request_at_cap_still_increments_existing` → `ok`.

</details>

Issue #2670, opened 2026-05-14
HERDER: Broadcasting parity for tx set/SCP messages


tomerweller:
Deferred from #2655. Low-priority: verify broadcast behavior for tx sets and SCP messages matches stellar-core's broadcastAll/sendSCPStateToPeer patterns.

tomerweller:
## 📝 Proposal Draft (Round 1/5)

## Problem

The `send_scp_state` function in `crates/app/src/app/consensus.rs` responds to `GetScpState` peer requests by blindly forwarding all envelopes from `herder.get_scp_state(from_slot)` without matching stellar-core's `sendSCPStateToPeer` behavior. Specifically:

1. **No checkpoint delay**: stellar-core delays sending the checkpoint ledger's SCP messages (via `SEND_LATEST_CHECKPOINT_DELAY = 10s`) when there is a gap between the checkpoint and the first sequential slot. This gives the receiving node time to establish `trackingConsensusLedgerIndex` before it processes the checkpoint envelope (which requires that index for close-time validation). Henyey sends all envelopes immediately.

2. **No slot count limit**: stellar-core limits transmission to `LEDGER_VALIDITY_BRACKET` (100) slots with data. Henyey sends everything `get_scp_state` returns, which is also bounded by LEDGER_VALIDITY_BRACKET in the SCP layer — so this is effectively at parity.

3. **No ascending order guarantee**: stellar-core uses `processSlotsAscendingFrom` to send slots in ascending order. Henyey's `get_scp_state` returns envelopes from the SCP layer's `get_scp_state` which iterates stored slots — order depends on the underlying SCP implementation.

## Proposed Fix

### 1. Add `SEND_LATEST_CHECKPOINT_DELAY` to `send_scp_state`

Mirror stellar-core's logic:
- Determine the most recent checkpoint sequence number
- Determine `firstSequentialLedgerSeq` = `max(1, tracking_consensus - MAX_SLOTS_TO_REMEMBER)`
- If checkpoint < firstSequentialLedgerSeq (gap exists), send non-checkpoint envelopes first, then spawn a delayed task that sends the checkpoint envelopes after `SEND_LATEST_CHECKPOINT_DELAY` (10 seconds)

Add a constant:
```rust
/// Delay before sending checkpoint SCP messages to a syncing peer.
/// Matches stellar-core `Herder::SEND_LATEST_CHECKPOINT_DELAY` (10s).
[...]

PR #2776, opened 2026-05-16
Preserve checkpoint slot during SCP purge in herder

### LINKED ISSUE #2706: HERDER: Preserve checkpoint slot during SCP purge
## Problem

In `herder.rs` (line ~2401), `purge_slots` is called with `slot.saturating_sub(10)` without preserving the most recent checkpoint slot. stellar-core's equivalent uses a `slotToKeep` parameter to ensure the checkpoint slot survives GC.

This means if the checkpoint slot falls outside the 10-slot retention window, its envelopes are purged and the delayed `send_scp_state` callback (added in #2670) will find no envelopes to send.

## Expected behavior

Mirror stellar-core: pass the checkpoint slot as `slot_to_keep` to `purge_slots` so it's retained even when outside the normal retention window.

## Context

Discovered during #2670 review. The checkpoint delay logic works correctly when the checkpoint is within the retention window, but could silently produce empty sends if the gap is large enough.

tomerweller:
Closes #2706

## Summary

Both `Scp::purge_slots` call sites in `crates/herder/src/herder.rs` (the `ledger_closed` close-tail cleanup at ~line 2485 and the `out_of_sync_recovery` early purge at ~line 1180) were passing `None` for `slot_to_keep`, so the most-recent-checkpoint SCP slot was dropped whenever it fell outside the 10-slot retention window. The delayed `send_scp_state` callback (#2670) then silently broadcast nothing because the checkpoint slot's envelopes had already been evicted from SCP state.

This PR threads `Some(get_most_recent_checkpoint_seq())` into both call sites, matching stellar-core's `HerderImpl::eraseOutsideRange` (HerderImpl.cpp:1328-1335) → `HerderSCPDriver::purgeSlotsOutsideRange` (HerderSCPDriver.cpp:1300-1337) → `SCP::purgeSlotsOutsideRange` chain, which always threads `lastCheckpointSeq` as `slotToKeep`. In `ledger_closed`, the existing `keep_slot` binding is hoisted above the SCP purge so it is computed once and reused for both the SCP purge and the existing `fetching_envelopes.erase_outside_range` call. In `out_of_sync_recovery`, the `last_checkpoint` local already bound for the fetching-envelopes purge is reused.

`Scp::purge_slots(max, slot_to_keep)` already had the correct semantics — no SCP-crate change was needed; only the call sites changed.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/2706#issuecomment-4466173965)

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test -p henyey-herder` — 1109 tests pass, including the two new regression tests:
  - `test_issue_2706_ledger_closed_preserves_checkpoint_slot_in_scp` — bootstraps tracking_consensus=300 (checkpoint=256, retention threshold slot-10=290), materializes SCP slots at 256 (checkpoint), 250 (control below window), and 295 (control inside window); after `ledger_closed(300)`, asserts 256 and 295 are retained and 250 is purged.
  - `test_issue_2706_out_of_sync_recovery_preserves_checkpoint_in_scp` — bootstraps to ledger 65 (checkpoint=64), materializes SCP slots at 64 (checkpoint) and 100 (control), triggers `out_of_sync_recovery` (purge_slot=302); asserts 64 is retained and 100 is purged.

## Deviations from plan

None. PARITY_STATUS.md was verified per the plan; no edits were required — the existing entries (`eraseOutsideRange()` → `fetching_envelopes.erase_outside_range()` Full; SCP `purgeSlotsOutsideRange()` → `purge_slots()` Partial) accurately describe the post-fix state, and the existing pending-envelopes follow-up note already tracks the sibling `PendingEnvelopes::purge_slots_below` gap explicitly deferred by this PR.

### DIFF
--- a/crates/herder/src/herder.rs
+++ b/crates/herder/src/herder.rs
@@ -1176,8 +1176,16 @@ impl Herder {
                 .write()
                 .clear_slots_below(purge_slot);
 
-            // Purge SCP state
-            self.scp.purge_slots(purge_slot.saturating_sub(1), None);
+            // Purge SCP state, preserving the most-recent-checkpoint slot.
+            // Parity: stellar-core outOfSyncRecovery routes through
+            // HerderImpl::eraseOutsideRange (HerderImpl.cpp:1328-1335), which
+            // threads getMostRecentCheckpointSeq() into
+            // HerderSCPDriver::purgeSlotsOutsideRange(_, _, slotToKeep)
+            // (HerderSCPDriver.cpp:1300-1337). Without this, the checkpoint
+            // slot's SCP state can be evicted and the delayed
+            // send_scp_state callback (#2670) silently sends nothing (#2706).
+            self.scp
+                .purge_slots(purge_slot.saturating_sub(1), Some(last_checkpoint));
 
             // Purge externalized values and pending tx set requests
             self.scp_driver.purge_slots_below(purge_slot);
@@ -2481,8 +2489,21 @@ impl Herder {
         self.drain_and_process_pending(next_index);
         self.process_ready_fetching_envelopes();
 
-        // Clean up old SCP state
-        self.scp.purge_slots(slot.saturating_sub(10), None);
+        // Most-recent-checkpoint slot to preserve across all the cleanups
+        // below. Parity: stellar-core HerderImpl::eraseOutsideRange
+        // (HerderImpl.cpp:1328-1335) computes lastCheckpointSeq once and
+        // threads it into both HerderSCPDriver::purgeSlotsOutsideRange and
+        // PendingEnvelopes::eraseOutsideRange so the checkpoint slot's
+        // envelopes survive the retention window.
+        let keep_slot = self.get_most_recent_checkpoint_seq();
+
+        // Clean up old SCP state, preserving the most-recent-checkpoint
+        // slot so the delayed send_scp_state callback (#2670) can still
+        // find it after a long apply or upon validator startup (#2706).
+        // Parity: HerderSCPDriver::purgeSlotsOutsideRange
+        // (HerderSCPDriver.cpp:1300-1337).
+        self.scp
+            .purge_slots(slot.saturating_sub(10), Some(keep_slot));
         // Purge deferred slot tracking alongside SCP slot cleanup
         self.scp_driver
             .purge_deferred_slots(slot.saturating_sub(10));
@@ -2502,7 +2523,6 @@ impl Herder {
         } else {
             None
         };
-        let keep_slot = self.get_most_recent_checkpoint_seq();
         self.fetching_envelopes
             .erase_outside_range(min_slot, max_slot, keep_slot);
 
@@ -12482,6 +12502,111 @@ mod fetching_envelopes_routing_tests {
         assert!(
             !herder.fetching_envelopes.slots.read().contains_key(&50),
             "slot 50 should be erased"
+        );
+    }
+
+    /// Regression test for #2706: `ledger_closed` must preserve the most
+    /// recent checkpoint slot when purging old SCP state. The previous code
+    /// passed `None` for `slot_to_keep`, so a checkpoint slot below
+    /// `slot - 10` would be dropped — silently breaking the delayed
+    /// `send_scp_state` callback (#2670).
+    ///
+    /// Parity: stellar-core `HerderImpl::eraseOutsideRange`
+    /// (HerderImpl.cpp:1328-1335) → `HerderSCPDriver::purgeSlotsOutsideRange`
+    /// (HerderSCPDriver.cpp:1300-1337) always threads
+    /// `getMostRecentCheckpointSeq()` as `slotToKeep`.
+    #[test]
+    fn test_issue_2706_ledger_closed_preserves_checkpoint_slot_in_scp() {
+        let herder = make_test_herder();
+
+        // Bootstrap so tracking_consensus_index = 300. With
+        // checkpoint_frequency = 64 (default), the most-recent-checkpoint
+        // slot is 256. We will close slot 300, whose retention threshold is
+        // slot - 10 = 290, so checkpoint 256 sits well below the window.
+        herder.bootstrap(299);
+        {
+            let mut ts = herder.tracking_state.write();
+            ts.is_tracking = true;
+            ts.consensus_index = 300;
+        }
+        assert_eq!(herder.get_most_recent_checkpoint_seq(), 256);
+
+        // Materialize SCP state for the checkpoint slot (must survive) and a
+        // control slot below the retention window (must be purged).
+        herder.scp.test_set_slot_v_blocking(256);
+        herder.scp.test_set_slot_v_blocking(250);
+        // And a slot inside the retention window — should also survive.
+        herder.scp.test_set_slot_v_blocking(295);
+        assert!(herder.scp.has_slot(256));
+        assert!(herder.scp.has_slot(250));
+        assert!(herder.scp.has_slot(295));
+
+        // Act
+        herder.ledger_closed(300, &[], &[], 0);
+
+        // Checkpoint slot 256 is preserved despite 256 < 290 (the
+        // slot - 10 retention threshold) thanks to slot_to_keep.
+        assert!(
+            herder.scp.has_slot(256),
+            "checkpoint slot 256 must be preserved as slot_to_keep (#2706)"
+        );
+        // Control slot 250 is below the window and not the checkpoint —
+        // it must be purged, confirming purge_slots actually ran.
+        assert!(
+            !herder.scp.has_slot(250),
+            "slot 250 should be purged (below retention window, not checkpoint)"
+        );
+        // Slot inside the retention window survives normally.
+        assert!(
+            herder.scp.has_slot(295),
+            "slot 295 should be preserved (within retention window)"
+        );
+    }
+
+    /// Regression test for #2706: `out_of_sync_recovery` must also preserve
+    /// the most-recent-checkpoint slot when purging old SCP state. Before
+    /// the fix, the SCP purge in `out_of_sync_recovery` passed `None` for
+    /// `slot_to_keep`, dropping the checkpoint slot's SCP state and
+    /// breaking the delayed `send_scp_state` callback (#2670) on the
+    /// recovery path.
+    #[test]
+    fn test_issue_2706_out_of_sync_recovery_preserves_checkpoint_in_scp() {
+        let herder = make_test_herder();
+
+        // Bootstrap to ledger 65 so checkpoint = 64.
+        herder.bootstrap(65);
+        {
+            let mut ts = herder.tracking_state.write();
+            ts.is_tracking = false;
+        }
+        *herder.state.write() = HerderState::Syncing;
+        assert_eq!(herder.get_most_recent_checkpoint_seq(), 64);
+
+        // Materialize SCP state at the checkpoint (must survive) and one
+        // non-checkpoint slot below the purge boundary (must be purged).
+        herder.scp.test_set_slot_v_blocking(64);
+        herder.scp.test_set_slot_v_blocking(100);
+        assert!(herder.scp.has_slot(64));
+        assert!(herder.scp.has_slot(100));
+
+        // Set up >100 v-blocking SCP slots to trigger the recovery purge.
+        for slot in 300..=401 {
+            herder.scp.test_set_slot_v_blocking(slot);
+        }
+
+        let result = herder.out_of_sync_recovery();
+        assert_eq!(result, Some(302));
+
+        // Checkpoint slot survives despite being below the recovery purge
+        // boundary (purge_slot - 1 = 301).
+        assert!(
+            herder.scp.has_slot(64),
+            "checkpoint slot 64 must be preserved in SCP as slot_to_keep (#2706)"
+        );
+        // Non-checkpoint slot below the boundary is purged.
+        assert!(
+            !herder.scp.has_slot(100),
+            "slot 100 should be purged from SCP (below purge boundary, not checkpoint)"
         );
     }
 }


tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** Both call sites now thread `get_most_recent_checkpoint_seq()` as `slot_to_keep`, faithfully mirroring stellar-core's `HerderImpl::eraseOutsideRange` → `HerderSCPDriver::purgeSlotsOutsideRange` → `SCP::purgeSlotsOutsideRange` chain.

<details>

**Reference (stellar-core v26.0.1):**
- `HerderImpl::eraseOutsideRange` — `src/herder/HerderImpl.cpp:1329-1335`: computes `lastCheckpointSeq = getMostRecentCheckpointSeq()` once and threads it into both `purgeSlotsOutsideRange` and `mPendingEnvelopes.eraseOutsideRange`.
- `HerderSCPDriver::purgeSlotsOutsideRange` — `src/herder/HerderSCPDriver.cpp:1300-1337`: special-cases `slotToKeep` in the timings-map walk, then forwards to `getSCP().purgeSlotsOutsideRange(minSlotIndex, maxSlotIndex, slotToKeep)`.
- `SCP::purgeSlotsOutsideRange` — `src/scp/SCP.cpp:73-78`: `if (it->first == slotToKeep) skip; else erase`.
- Out-of-sync recovery path — `src/herder/HerderImpl.cpp:552`: `eraseOutsideRange(purgeSlot, std::nullopt)`, which still threads `lastCheckpointSeq` per :1332.

**Parity findings on the diff:**

- ✅ `ledger_closed` (herder.rs ~2489): `self.scp.purge_slots(slot.saturating_sub(10), Some(keep_slot))` where `keep_slot = get_most_recent_checkpoint_seq()`. Matches `HerderImpl.cpp:1332-1333` semantics.
- ✅ `out_of_sync_recovery` (herder.rs ~1180): `self.scp.purge_slots(purge_slot.saturating_sub(1), Some(last_checkpoint))`. Matches the `HerderImpl.cpp:552` → `:1329-1334` chain, which always threads `lastCheckpointSeq` regardless of whether `maxSlot` is `nullopt`.
- ✅ `keep_slot` is hoisted above the SCP purge and reused for `fetching_envelopes.erase_outside_range`, mirroring stellar-core's single-call computation pattern (`HerderImpl.cpp:1332` — one `lastCheckpointSeq` flowing into both purgers in the same order).
- ✅ Ordering relative to peers in `ledger_closed`: SCP purge → `scp_driver.purge_deferred_slots` → `fetching_envelopes.erase_outside_range` → `cleanup` → `pending_envelopes.purge_slots_below`. The SCP-then-pending order matches `HerderImpl.cpp:1333-1335`. The Rust-only `purge_deferred_slots` is a local-side bookkeeping call with no stellar-core analog and is correctly slot-keep-agnostic (deferred slots are indexed by external timer state, not envelope storage).
- ✅ Underlying `Scp::purge_slots` (`crates/scp/src/scp.rs:342-346`) retains via `slot_index >= max_slot_index || slot_to_keep == Some(slot_index)` — semantically identical to `SCP.cpp:73-87`'s `if (it->first == slotToKeep) ++it; else erase`.
- ✅ Edge cases:
  - Empty SCP state: both impls are no-ops via map iteration / `retain` on an empty map.
  - Checkpoint older than retention window: explicit motivating case — preserved by both.
  - Checkpoint == current/retained slot: the `||` in `retain` and the `if (it->first == slotToKeep) ++it` branch both make double-coverage harmless.
  - Genesis / checkpoint == 0: `get_most_recent_checkpoint_seq()` returns 0 pre-bootstrap, which can only collide with a non-existent slot 0; no spurious retention.

**Out-of-scope parity gaps (correctly deferred, called out in PR body):**

- `pending_envelopes.purge_slots_below(...)` (both `ledger_closed` and `out_of_sync_recovery`) does not yet take a `slot_to_keep`, while stellar-core's `PendingEnvelopes::eraseOutsideRange` (invoked at `HerderImpl.cpp:1335`) does receive `lastCheckpointSeq`. PR explicitly defers this to the existing follow-up note in `PARITY_STATUS.md`; not a regression introduced here.
- Retention threshold uses hardcoded `slot - 10` instead of `currentLedger - MAX_SLOTS_TO_REMEMBER` (default 12). Pre-existing; not touched by this PR.

**Regression coverage:** Both new tests exercise the exact stellar-core invariant (checkpoint below window survives; non-checkpoint below window is evicted; in-window slot survives), for both the steady-state `ledger_closed` and the `out_of_sync_recovery` paths. Parity-faithful.

No parity blockers. Ship it.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Fix is a precise, minimal mirror of stellar-core's `eraseOutsideRange` → `purgeSlotsOutsideRange` thread of `lastCheckpointSeq` as `slotToKeep` at both call sites. Regression tests cover the bug and pass; `Scp::purge_slots`'s existing `retain(|slot| slot >= max || Some(slot) == slot_to_keep)` semantics are exactly what's needed.

<details>

- **Correctness — `ledger_closed` (line ~2489):** Hoisting `let keep_slot = self.get_most_recent_checkpoint_seq();` above the SCP purge and reusing it for both `scp.purge_slots(slot - 10, Some(keep_slot))` and `fetching_envelopes.erase_outside_range(..., keep_slot)` is correct. Computing once avoids any (currently theoretical) race where the checkpoint seq could shift between calls, and it matches stellar-core's single-`lastCheckpointSeq` derivation. The old `let keep_slot = ...` below is correctly removed.
- **Correctness — `out_of_sync_recovery` (line ~1180):** Reuses the already-bound `last_checkpoint` from the immediately-preceding `fetching_envelopes.erase_outside_range` call. Clean.
- **`Scp::purge_slots` semantics confirmed** (scp.rs:342): `retain(|&s, _| s >= max || slot_to_keep == Some(s))` correctly preserves the checkpoint slot even when it is below the retention threshold; no SCP-crate change needed, as the PR claims.
- **Edge cases:**
  - Checkpoint ≥ threshold (within window): `Some(keep_slot)` is a harmless no-op — `retain` already keeps it. ✅
  - Checkpoint == 0 (early-ledger / pre-bootstrap): `purge_slots(0, Some(0))` retains everything; no regression. ✅
  - `saturating_sub` on small `slot` / `purge_slot` correctly bottoms at 0. ✅
  - `out_of_sync_recovery`'s `purge_slot.saturating_sub(1)` is unchanged and still correct.
- **Test coverage:** Two narrow regression tests, one per call site, each materializing a checkpoint slot below the purge window plus a control non-checkpoint slot below the window (must be purged) and a control slot inside the window (`ledger_closed` only). They would have caught the original bug — the failing-test-first workflow is satisfied. Tests use existing `test_set_slot_v_blocking` / `make_test_herder` / `bootstrap` helpers consistent with the surrounding module.
- **Readability / error handling:** Comments are excellent — they cite both the upstream file:line locations and the linked issues (#2670, #2706). No new error paths introduced; purge is infallible. No new `unwrap`s.
- **Non-blocking observations (suitable for inline / follow-up, not blocking merge):**
  - `pending_envelopes.purge_slots_below(...)` at both sites still drops the checkpoint slot if it falls below the boundary. The PR body explicitly defers this as a sibling gap already tracked in PARITY_STATUS pending notes — acknowledged, OK to defer.
  - Minor: a single shared helper (e.g. `purge_old_scp_state(boundary)`) could DRY the two call sites, but two callers with subtly different boundaries don't justify it now.

No correctness blockers.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Surgical, well-justified fix that threads `Some(get_most_recent_checkpoint_seq())` into both `Scp::purge_slots` call sites in `herder.rs`, matching stellar-core's `eraseOutsideRange` → `purgeSlotsOutsideRange` chain. `Scp::purge_slots`' existing `retain(|&s, _| s >= max || slot_to_keep == Some(s))` semantics make this a one-line behavioral change per site; the two new regression tests demonstrably reproduce the original bug and pass after the fix.

<details>

### Diff correctness

- **`ledger_closed` (~L2489).** Hoisting `let keep_slot = self.get_most_recent_checkpoint_seq();` above the SCP purge and reusing it for both `scp.purge_slots(slot.saturating_sub(10), Some(keep_slot))` and the existing `fetching_envelopes.erase_outside_range(..., keep_slot)` is correct and slightly better than the pre-PR code: a single derivation avoids any (currently theoretical) drift between the two calls and mirrors stellar-core's single `lastCheckpointSeq` derivation in `HerderImpl::eraseOutsideRange`. The duplicated `let keep_slot = ...` below the fetching-envelopes call is correctly removed.
- **`out_of_sync_recovery` (~L1180).** Reuses the already-bound `last_checkpoint` local from the immediately-preceding `fetching_envelopes.erase_outside_range(...)` call. No new bindings, no shadowing surprises.
- **`Scp::purge_slots` semantics (scp.rs:342).** Verified: `retain(|&s, _| s >= max || slot_to_keep == Some(s))` correctly preserves the checkpoint slot even when it sits below `max`. No SCP-crate change needed — PR claim is accurate.

### Edge cases

- Checkpoint ≥ retention threshold (in-window): `Some(keep_slot)` is a harmless no-op — `retain` already keeps it. ✅
- Checkpoint == 0 (pre-bootstrap / very early ledger): `purge_slots(0, Some(0))` retains everything; no regression. ✅
- `saturating_sub(10)` / `saturating_sub(1)` on small `slot` / `purge_slot` correctly bottoms at 0. ✅
- `out_of_sync_recovery`'s `purge_slot.saturating_sub(1)` is unchanged and still correct.

### Test coverage

Two regression tests, one per call site:

- `test_issue_2706_ledger_closed_preserves_checkpoint_slot_in_scp` — bootstraps so `tracking_consensus=300`, `checkpoint=256`, threshold `slot-10=290`; materializes slots 256 (checkpoint, below threshold), 250 (control, must be purged), 295 (control, in-window); asserts 256 and 295 survive, 250 is gone. Three-slot assertion (kept-checkpoint, purged-control, kept-in-window) cleanly proves both "purge ran" and "slot_to_keep is wired".
- `test_issue_2706_out_of_sync_recovery_preserves_checkpoint_in_scp` — sets up >100 v-blocking SCP slots (300..=401) to trigger the recovery purge, with checkpoint=64 and a non-checkpoint control at 100; asserts purge_slot=302, 64 survives, 100 is purged.

Both tests would have failed against pre-PR code, satisfying the failing-test-first requirement. Helpers (`make_test_herder`, `bootstrap`, `test_set_slot_v_blocking`) are existing test infrastructure.

### Readability & error handling

- Comments are excellent — they cite both upstream file:line locations (`HerderImpl.cpp:1328-1335`, `HerderSCPDriver.cpp:1300-1337`) and the linked issues (#2670, #2706). A future reader debugging an empty `send_scp_state` will land here and immediately understand why.
- No new error paths; `purge_slots` is infallible. No new `unwrap`s, no new panics.
- Code follows the surrounding style; the hoist in `ledger_closed` is mechanical and unsurprising.

### Non-blocking observations (suitable as follow-ups)

- `pending_envelopes.purge_slots_below(...)` at both sites still drops the checkpoint slot if it falls below the boundary. The PR body explicitly defers this to an existing PARITY_STATUS pending-envelopes note — acknowledged and reasonable to defer.
- Minor: a shared `purge_old_scp_state(boundary)` helper could DRY the two call sites, but with two callers and subtly different boundaries it's not worth the indirection now.

No correctness blockers.

</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** The change correctly threads the most-recent-checkpoint slot into `Scp::purge_slots` at both call sites, matching stellar-core's `HerderImpl::eraseOutsideRange` → `HerderSCPDriver::purgeSlotsOutsideRange(..., slotToKeep=getMostRecentCheckpointSeq())` (HerderImpl.cpp:1329-1335, HerderSCPDriver.cpp:1300-1337). The semantics, sequencing, and lifetime of `lastCheckpointSeq` (computed once, threaded into both SCP and pending-envelope purges) all line up with the C++ reference. Non-blocking observations below; nothing in this PR diverges from upstream.

<details>
<summary>Full parity analysis</summary>

### Reference: stellar-core

```cpp
// HerderImpl.cpp:1329-1335
void HerderImpl::eraseOutsideRange(std::optional<uint32> minSlot,
                                   std::optional<uint32> maxSlot)
{
    auto lastCheckpointSeq = getMostRecentCheckpointSeq();
    getHerderSCPDriver().purgeSlotsOutsideRange(minSlot, maxSlot,
                                                lastCheckpointSeq);
    mPendingEnvelopes.eraseOutsideRange(minSlot, maxSlot, lastCheckpointSeq);
    ...
}
```

`lastCheckpointSeq` is computed once and routed identically to (a) SCP slot/timing purge, and (b) pending-envelope purge. `purgeSlotsOutsideRange` then unconditionally preserves the `slotToKeep` entry inside `SCP::purgeSlotsOutsideRange` and in `mSCPExecutionTimes`.

### Henyey changes vs. reference

**1. `ledger_closed` (post-PR, herder.rs ~2483-2509)**

```rust
let keep_slot = self.get_most_recent_checkpoint_seq();
self.scp.purge_slots(slot.saturating_sub(10), Some(keep_slot));
...
self.fetching_envelopes.erase_outside_range(min_slot, max_slot, keep_slot);
```

- ✅ Single `get_most_recent_checkpoint_seq()` call, reused for both purges — same lifetime/value invariant as the C++ `lastCheckpointSeq`.
- ✅ `Some(keep_slot)` is threaded into `Scp::purge_slots`, which honors it via `slot_to_keep == Some(slot_index)` in the retain predicate (`crates/scp/src/scp.rs:342`). Equivalent to C++ `slotIndex == slotToKeep ? skip : erase`.
- ✅ Ordering preserved: drain → purge → cleanup, same as `newSlotExternalized` (HerderImpl.cpp:260-285).

**2. `out_of_sync_recovery` (post-PR, herder.rs ~1167-1186)**

```rust
let last_checkpoint = self.get_most_recent_checkpoint_seq();
self.fetching_envelopes.erase_outside_range(Some(purge_slot), None, last_checkpoint);
...
self.scp.purge_slots(purge_slot.saturating_sub(1), Some(last_checkpoint));
```

- ✅ Mirrors C++ `eraseOutsideRange(purgeSlot, std::nullopt)`, threading `lastCheckpointSeq` into both SCP and pending-envelope purges.
- ✅ Reuses the single `last_checkpoint` snapshot for both calls.

### Regression tests

Both new tests (`test_issue_2706_ledger_closed_preserves_checkpoint_slot_in_scp`, `test_issue_2706_out_of_sync_recovery_preserves_checkpoint_in_scp`) exercise the boundary where the checkpoint slot sits below the purge window, which is exactly the parity-relevant edge case in `HerderSCPDriver::purgeSlotsOutsideRange`. The control assertions (non-checkpoint slot below boundary is purged; in-window slot survives) protect against the "always preserve" or "never preserve" degenerate fixes.

### Non-blocking observations (out of scope for this PR; pre-existing parity drift)

These are **not** introduced by this PR and should not block merge. Flagging only so they can be tracked.

1. **`ledger_closed` uses `slot.saturating_sub(10)` instead of `getMinLedgerSeqToRemember()`.** stellar-core's `newSlotExternalized` computes `minSlotToRemember = getMinLedgerSeqToRemember()` (which accounts for `MAX_SLOTS_TO_REMEMBER` config, history-publish requirements, and `getSafeLedgerToDelete`). Henyey hardcodes `slot - LEDGER_VALIDITY_BRACKET`. This can purge slots that stellar-core would still hold (e.g., when `MAX_SLOTS_TO_REMEMBER > 10`) or, conversely, retain slots stellar-core would drop. Pre-existing; consider a follow-up.

2. **`out_of_sync_recovery` uses `purge_slot.saturating_sub(1)` as the SCP min-bound** while stellar-core passes `purgeSlot` directly to `eraseOutsideRange`. Henyey purges one extra slot at the boundary. Pre-existing.

3. **`pending_envelopes.purge_slots_below(slot)` and `.purge_slots_below(purge_slot)`** (herder.rs:1186, 2513) don't take a `slot_to_keep` argument, so the checkpoint slot's entry in `pending_envelopes` can still be evicted on the `< slot` and `< purge_slot` boundaries — even though `fetching_envelopes` is now correctly preserving it. stellar-core preserves `lastCheckpointSeq` inside the equivalent `mPendingEnvelopes.eraseOutsideRange` path. Whether henyey's `pending_envelopes` is actually the path consulted by the delayed `send_scp_state` callback (#2670) determines whether this matters in practice; if it is, the #2706 fix may still leak on the `pending_envelopes` side. Worth a quick check / follow-up issue.

4. **`Scp::purge_slots`'s `max_slot_index` parameter is misleadingly named** — it's actually used as the min-bound (slots with index `< max_slot_index` are dropped). Cosmetic, but renaming to `min_slot_index` would prevent future confusion at call sites and match stellar-core's `minSlotIndex` parameter. Not this PR's concern.

### Verdict

The PR faithfully reproduces stellar-core's checkpoint-preservation contract on the SCP side at both call sites, with strong regression coverage and accurate parity comments pointing at the upstream line ranges. APPROVE.

</details>

tomerweller:

**Reviewer A:** APPROVE — Both `Scp::purge_slots` call sites correctly thread `Some(get_most_recent_checkpoint_seq())`, regression tests cover both paths, no correctness blockers.
**Reviewer B:** APPROVE — Change matches stellar-core's `HerderImpl::eraseOutsideRange` → `HerderSCPDriver::purgeSlotsOutsideRange(_, _, lastCheckpointSeq)` contract; single `lastCheckpointSeq` snapshot reused for both SCP and pending-envelope purges, mirroring the C++ lifetime invariant.
**External reviewers:** none
**CI:** red (Quickstart workflow only; CI workflow all green)

**Failing checks (Quickstart workflow):**
- `build / 3 test / 2 test (testing-with-pr, amd64, local, core,rpc,horizon , )` — horizon ingestion test timed out after 4 min ("bucket not found: 09f901dc...")
- `build / 3 test / complete`, `build / complete` — cascaded

**Diff-attribution:** **unrelated.** The same `core,rpc,horizon` Quickstart test is failing on `main` itself at the most recent commits (run [25952505373](https://github.com/stellar-experimental/henyey/actions/runs/25952505373) on `3b31f4b3` and run [25951830752](https://github.com/stellar-experimental/henyey/actions/runs/25951830752) on `f3a599e6` both fail the same job). The failure has no reference to herder.rs or SCP-purge code — it's a `localhost:1570` history-archive bucket download in the horizon ingestion container. This appears to be a pre-existing broken/flaky Quickstart workflow on main, not introduced by this PR.

Routing back to `ready-for-doing` for `/do` Mode B. Per the `/review-pr` matrix, `/do` should rebase on `origin/main` and retry. **Heads-up for the next cycle:** since main has the same failure, a rebase alone is unlikely to fix CI; this PR may need to be marked `blocked` on the next `/review-pr` cycle pending a fix to the Quickstart workflow on main, OR the reviewer-agent verdict pair (both APPROVE) should be considered sufficient to override the broken-on-main signal.

tomerweller:
## Do: CI Flake Retry

Re-ran failed CI job (`build / 3 test / 2 test (testing-with-pr, amd64, local, core,rpc,horizon)`). Root cause was a flaky bucket download timeout (HTTP 404 from local history archive on `localhost:1570`, then 4-minute action timeout) — unrelated to the code change.

No code changes needed: all reviewer comments are APPROVE, no inline review feedback to address. Bouncing back to `in-review` for re-evaluation after CI rerun completes.

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Both `Scp::purge_slots` call sites now thread `Some(get_most_recent_checkpoint_seq())`, exactly matching the bug described in #2706. The `keep_slot` hoist in `ledger_closed` is a safe, idiomatic refactor that eliminates a duplicate call, and the `last_checkpoint` reuse in `out_of_sync_recovery` is similarly clean. Two well-targeted regression tests cover both code paths with retain/purge control slots that would have caught the original bug.

### Correctness of the diff

- **`ledger_closed` (lines ~2493-2516)**: The hoist of `let keep_slot = self.get_most_recent_checkpoint_seq();` above the SCP purge is safe — the only code executed between the old binding site and the new site is `self.scp.purge_slots(...)` and `self.scp_driver.purge_deferred_slots(...)`, neither of which can affect the checkpoint computation (checkpoint depends on `tracking_consensus_index` / ledger state, not SCP slot inventory). Reusing the single binding for both `scp.purge_slots` and `fetching_envelopes.erase_outside_range` is preferable to two independent calls.
- **`out_of_sync_recovery` (lines ~1188-1189)**: Reusing the already-bound `last_checkpoint` local is the obvious right move; no new state to compute.
- Both call sites now mirror the parity chain documented in the comments. The `purge_slot.saturating_sub(1)` / `slot.saturating_sub(10)` upper bounds are unchanged — only the third arg flipped from `None` to `Some(checkpoint)`, which is the minimal surgical fix.
- Note: there's no risk of `keep_slot` *itself* being above the `max` upper bound and getting "kept anyway" in a surprising way — that's exactly the intended semantics (`slot_to_keep` is the escape hatch). And if checkpoint is already inside the retention window, passing `Some(checkpoint)` is a harmless no-op vs the old `None`.

### Test coverage

Both regression tests are tight and would have caught the original bug:

- `test_issue_2706_ledger_closed_preserves_checkpoint_slot_in_scp`: Smartly picks slot 300 / checkpoint 256 so the checkpoint sits 34 slots below the `slot-10=290` retention threshold, then asserts (a) checkpoint 256 survives, (b) non-checkpoint slot 250 is purged (proving the purge ran, not that it was a no-op), and (c) slot 295 inside the window also survives. The triplet retain/purge/retain is exactly the right shape.
- `test_issue_2706_out_of_sync_recovery_preserves_checkpoint_in_scp`: Mirrors the same retain/purge pattern on the recovery path; the 100 v-blocking slots setup correctly triggers the purge boundary.

### Readability & comments

- Inline comments cite stellar-core file:line ranges (HerderImpl.cpp:1328-1335, HerderSCPDriver.cpp:1300-1337) and cross-reference #2670 and #2706. This is excellent and matches the repo's parity-documentation style.
- Test docstrings explain *why* each assertion matters, not just *what* is asserted.

### Error handling

`Scp::purge_slots` and `get_most_recent_checkpoint_seq` are both infallible and synchronous — nothing to handle. No new failure modes introduced.

### Non-blocking nits (orchestrator may auto-file as follow-ups)

- The PR description correctly notes the sibling `PendingEnvelopes::purge_slots_below` gap is deferred and tracked in PARITY_STATUS; that scope decision is reasonable for this PR.
- In `test_issue_2706_out_of_sync_recovery_preserves_checkpoint_in_scp`, the direct `*herder.state.write() = HerderState::Syncing` write is ad-hoc, but it matches existing test idioms in this file so not worth pushing back on here.

CI is "pending" at review time — the orchestrator will gate merge on green.

</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** Both `Scp::purge_slots` call sites now thread `Some(get_most_recent_checkpoint_seq())`, exactly mirroring stellar-core v26.0.1's `HerderImpl::eraseOutsideRange` → `HerderSCPDriver::purgeSlotsOutsideRange` → `SCP::purgeSlotsOutsideRange` chain. Checkpoint slot retention semantics match upstream; no observable behavioral divergence on this path.

### Scope

Parity audit of the SCP-purge call sites in `crates/herder/src/herder.rs` against stellar-core v26.0.1 (`stellar-core/` submodule). Correctness is owned by Reviewer A; I only assess fidelity to upstream semantics on this path.

### Upstream reference (stellar-core v26.0.1)

- `HerderImpl::eraseOutsideRange` (HerderImpl.cpp:1328-1335) unconditionally computes `lastCheckpointSeq = getMostRecentCheckpointSeq()` and threads it into both `HerderSCPDriver::purgeSlotsOutsideRange(min, max, slotToKeep)` and `PendingEnvelopes::eraseOutsideRange(min, max, slotToKeep)`.
- `HerderSCPDriver::purgeSlotsOutsideRange` (HerderSCPDriver.cpp:1300-1337) applies the `slotToKeep` exception while pruning `mSCPExecutionTimes` and then delegates to `SCP::purgeSlotsOutsideRange(min, max, slotToKeep)`.
- The two upstream callers of `eraseOutsideRange` map to henyey's two changed sites:
  - `newSlotExternalized` (HerderImpl.cpp:260-278) → `Herder::ledger_closed` (`herder.rs:~2493`).
  - `outOfSyncRecovery` `eraseOutsideRange(purgeSlot, nullopt)` (HerderImpl.cpp:552) → `Herder::out_of_sync_recovery` (`herder.rs:~1189`).

### Parity assessment of the change

1. **`slotToKeep` value source.** `Herder::get_most_recent_checkpoint_seq` (`herder.rs:944-957`) is a direct port of `HistoryManager::firstLedgerInCheckpointContaining(trackingConsensusLedgerIndex)`. Same input, same arithmetic. ✓
2. **Threading at the SCP site.** Both `Scp::purge_slots(_, Some(keep))` call sites now pass the checkpoint, matching upstream's "always thread `lastCheckpointSeq`" contract. The Rust `Option<u64>` is `Some(...)` on every code path now, equivalent to the upstream `uint64` (always-present) parameter. ✓
3. **Compute-once discipline in `ledger_closed`.** The hoisted `let keep_slot = self.get_most_recent_checkpoint_seq();` above the SCP purge and reused by `fetching_envelopes.erase_outside_range` mirrors upstream's single `lastCheckpointSeq` local fanned out to both consumers. ✓
4. **`out_of_sync_recovery` boundary translation.** Upstream calls `eraseOutsideRange(purgeSlot, std::nullopt)` (range-based; erases `< purgeSlot`). henyey calls `purge_slots(purge_slot.saturating_sub(1), Some(last_checkpoint))` (max-based; erases `≤ purge_slot - 1`). These are equivalent for any `purge_slot ≥ 1`, which is the only reachable case here (`purge_slot` comes from a v-blocking-descent that has counted down `LEDGER_VALIDITY_BRACKET = 100` hits). The pre-existing `saturating_sub(1)` boundary is unchanged by this PR, so any hypothetical `purge_slot == 0` edge case is out of scope and not regressed.
5. **`Scp::purge_slots` semantics.** `crates/scp/src/scp.rs:342` accepts `(max_slot_index, slot_to_keep: Option<u64>)` and the `slot_to_keep` exception matches the upstream `it->first == slotToKeep` skip in `HerderSCPDriver::purgeSlotsOutsideRange` and inside `SCP::purgeSlotsOutsideRange`. No SCP-crate change is needed and none is made. ✓
6. **Tests model the upstream contract.** Both regression tests (`test_issue_2706_ledger_closed_preserves_checkpoint_slot_in_scp`, `test_issue_2706_out_of_sync_recovery_preserves_checkpoint_in_scp`) check the exact invariant upstream encodes: a slot equal to `lastCheckpointSeq` survives the prune even when it falls outside the retention window, while a non-checkpoint control slot in the same below-window region is evicted. ✓

### Known parity gaps (out of scope; not introduced by this PR)

- `pending_envelopes.purge_slots_below(slot)` at `herder.rs:2522` still takes no `slot_to_keep` argument, so the sibling `PendingEnvelopes::eraseOutsideRange(min, max, lastCheckpointSeq)` parity gap remains. The PR description explicitly defers this and `PARITY_STATUS.md` already tracks it. Not a blocker for #2706's `send_scp_state`-broadcast fix, which depends only on `SCP` state, not `PendingEnvelopes`.
- `ledger_closed`'s `Scp::purge_slots(slot.saturating_sub(10), …)` only trims the *below* side; upstream's `eraseOutsideRange` from `newSlotExternalized` also trims above `nextConsensusLedgerIndex() + LEDGER_VALIDITY_BRACKET`. henyey covers the above-bracket trim via `fetching_envelopes.erase_outside_range(min, max, keep_slot)` but does not trim future SCP slot state from the `Scp` struct itself. This is pre-existing, orthogonal to #2706, and not exercised by the bug being fixed.

### Sequencing

Upstream order inside `eraseOutsideRange`: SCP-driver timings prune → SCP slot prune → PendingEnvelopes prune. henyey `ledger_closed` order: `scp.purge_slots` → `scp_driver.purge_deferred_slots` → `fetching_envelopes.erase_outside_range` → `cleanup` → `pending_envelopes.purge_slots_below`. The per-collection prunes are mutually independent (no cross-collection invariants between them are violated by reordering), and the checkpoint-preservation invariant is enforced uniformly inside each call now that `slot_to_keep` is threaded. No sequencing divergence that affects observable behavior on the #2670 / #2706 path.

### Conclusion

The change is a faithful and minimal parity fix. Behavior at both SCP-purge sites now matches stellar-core v26.0.1 on the slot-retention contract that the delayed `send_scp_state` callback depends on. APPROVE from a parity standpoint.

</details>

tomerweller:

**Reviewer A:** APPROVE — Diff cleanly threads `Some(checkpoint_seq)` into both `Scp::purge_slots` call sites; regression tests cover both paths.
**Reviewer B:** APPROVE — Mirrors stellar-core v26.0.1's `HerderImpl::eraseOutsideRange` → `HerderSCPDriver::purgeSlotsOutsideRange` → `SCP::purgeSlotsOutsideRange` chain at both call sites.
**External reviewers:** copilot-pull-request-reviewer: COMMENTED (non-blocking)
**CI:** red (unrelated)

The failing Quickstart job (`build / 3 test / 2 test (testing-with-pr, amd64, local, core,rpc,horizon , )`, run [25956301482](https://github.com/stellar-experimental/henyey/actions/runs/25956301482/job/76304989510)) fails during Horizon ingestion catchup with:

```
Error: catchup failed: invalid live bucket list structure: invalid HAS: future must be cleared at level 5 (prev snap version 25)
```

The error originates in `henyey_bucket::bucket_list` (bucket-list HAS validation) and is hit before any SCP code runs. The PR only modifies `crates/herder/src/herder.rs` (`Scp::purge_slots` call sites and a `keep_slot` hoist), which cannot affect bucket-list catchup. This is a pre-existing/unrelated breakage on the bucket-list path.

Routing back to `ready-for-doing` for `/do` Mode B to rebase on `origin/main` and retry. If CI is still red after rebase, the next `/review-pr` will mark `blocked`.

tomerweller:
## Do: No actionable feedback — CI flake

Picked up issue #2706 in `ready-for-doing`. PR #2776 already has both internal reviewer verdicts APPROVE (correctness + parity); only feedback is from `copilot-pull-request-reviewer[bot]` which is non-blocking.

The only failure is CI job `build / 3 test / 2 test (testing-with-pr, amd64, local, core,rpc,horizon)` with:

> Error: catchup failed: invalid live bucket list structure: invalid HAS: future must be cleared at level 5 (prev snap version 25)

This is a stellar-core subprocess error invoked by Horizon ingestion — unrelated to the herder SCP-purge change in this PR. The same `Quickstart` workflow is also failing on the latest `main` commit, confirming it's a pre-existing infrastructure flake.

Re-ran the failed jobs on workflow run [25956301482](https://github.com/stellar-experimental/henyey/actions/runs/25956301482). Moving the issue back to `in-review` so `/review-pr` can re-evaluate once CI completes.

No code changes were made.

Co-authored-by: GitHub Copilot <copilot@github.com>

tomerweller:

**Reviewer A (Correctness):** APPROVE — fix correctly threads checkpoint-slot retention through both `Scp::purge_slots` call sites; new regression tests cover both paths.
**Reviewer B (Parity):** APPROVE — matches stellar-core's `HerderImpl::eraseOutsideRange` → `HerderSCPDriver::purgeSlotsOutsideRange` chain.
**External reviewers:** copilot-pull-request-reviewer[bot]: COMMENTED (non-blocking overview)
**CI:** red (unrelated)

The failing check is `build / 3 test / 2 test (testing-with-pr, amd64, local, core,rpc,horizon)` in the Quickstart workflow. Root cause is a stellar-core catchup error in the bucket-list HAS validation path:

```
Error: catchup failed: invalid live bucket list structure: invalid HAS: future must be cleared at level 5 (prev snap version 25)
```

This is inside `henyey_bucket::bucket_list` / `henyey_history::catchup::buckets`, exercised via a Horizon ingestion docker test. The PR's diff is exclusively `crates/herder/src/herder.rs` (SCP slot retention) and cannot have caused this bucket-list failure. Job timed out at 12 min while stellar-core ingestion kept retrying the same catchup error.

All other CI checks (Format, Clippy, Test, Doctest, Build, CodeQL, Spec-Eval Freshness, all other Quickstart matrix cells) are SUCCESS.

Routing back to `ready-for-doing` for `/do` Mode B to rebase on `origin/main` and retry. If the same failure recurs after rebase, the next `/review-pr` will mark `blocked`.

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Fix threads `Some(get_most_recent_checkpoint_seq())` into both `Scp::purge_slots` call sites with correct semantics, and the two new regression tests precisely exercise the bug — they would fail before the fix and pass after it.

<details>

**Diff correctness**

- `Scp::purge_slots(max, slot_to_keep)` retains a slot iff `slot_index >= max || slot_to_keep == Some(slot_index)`, so passing `Some(checkpoint_seq)` correctly rescues the checkpoint slot regardless of how far below `max` it sits. Behavior matches the PR claim and the issue description.
- `ledger_closed` (~L2489): `keep_slot` is now hoisted above the SCP purge and reused by the existing `fetching_envelopes.erase_outside_range` call. The hoist is semantically a no-op for `fetching_envelopes` (same value, same call) and removes one redundant `get_most_recent_checkpoint_seq()` call — good cleanup.
- `out_of_sync_recovery` (~L1180): reuses the existing `last_checkpoint` local already computed for `fetching_envelopes.erase_outside_range`. No redundant calls, no new locals.
- No behavior changes to the other purges in the area (`scp_driver.purge_deferred_slots`, `scp_driver.purge_slots_below`, `pending_envelopes.purge_slots_below`). The PR explicitly notes that the `PendingEnvelopes::purge_slots_below` checkpoint gap is a separately-tracked follow-up — fine to scope out of this PR.
- No unwraps, no panics, no error-handling regressions introduced. The off-by-one (`purge_slot.saturating_sub(1)`) in `out_of_sync_recovery` is preserved unchanged.

**Test correctness**

- `test_issue_2706_ledger_closed_preserves_checkpoint_slot_in_scp`:
  - Setup: consensus_index = 300 ⇒ checkpoint = 256; `purge_slots(290, Some(256))`.
  - Slot 256 (checkpoint, below window) → kept via `slot_to_keep`. ✓
  - Slot 250 (below window, not checkpoint) → purged. ✓ This is the critical assertion that proves purge actually ran.
  - Slot 295 (in window) → kept. ✓
  - Without the fix, slot 256 would be purged → test fails. Good failing-test-first reproducer.
- `test_issue_2706_out_of_sync_recovery_preserves_checkpoint_in_scp`:
  - Bootstrap 65 ⇒ checkpoint = 64; recovery purge ⇒ `purge_slots(301, Some(64))`.
  - Slot 64 retained, slot 100 purged. Asserts `result == Some(302)` to confirm recovery actually fired.
  - Without the fix, slot 64 would be evicted → test fails. Good.
- Both tests use `test_set_slot_v_blocking` to materialize SCP state directly, which is the right tool — they avoid the need to drive full nomination.

**Readability / structure**

- Comments cite the stellar-core source lines (`HerderImpl.cpp:1328-1335`, `HerderSCPDriver.cpp:1300-1337`) and the related henyey issues (#2670, #2706), which makes the why obvious for future readers.
- Hoisting `keep_slot` in `ledger_closed` reads cleanly; comment block above it explains the shared use.

**Nits (non-blocking)**

- The two regression tests are quite long and share a lot of scaffolding (bootstrap + materialize slots + assert). A small helper would shrink them, but duplication is fine for regression tests where readability of the scenario matters more than DRY.
- The `out_of_sync_recovery` test manually pokes `tracking_state.write().is_tracking = false` and `*state.write() = HerderState::Syncing` to force the recovery path. That's consistent with surrounding test patterns in this module but is brittle — if someone ever adds an invariant on those two fields, the test will start asserting indirectly through `out_of_sync_recovery`. Worth keeping an eye on, but not blocking.

No correctness blockers.

</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** The two updated call sites faithfully mirror stellar-core's `HerderImpl::eraseOutsideRange` → `HerderSCPDriver::purgeSlotsOutsideRange` → `SCP::purgeSlotsOutsideRange` chain, threading `getMostRecentCheckpointSeq()` as `slotToKeep` in both `ledger_closed` and `out_of_sync_recovery`. No new parity divergences introduced.

<details>

**Parity verification of the diff**

- **C++ source of truth**: `HerderImpl::eraseOutsideRange` (`stellar-core/src/herder/HerderImpl.cpp:1329-1335`) computes `auto lastCheckpointSeq = getMostRecentCheckpointSeq();` once and passes it as `slotToKeep` into both `HerderSCPDriver::purgeSlotsOutsideRange` and `PendingEnvelopes::eraseOutsideRange`. `HerderSCPDriver::purgeSlotsOutsideRange` (`HerderSCPDriver.cpp:1300-1337`) forwards `slotToKeep` to `SCP::purgeSlotsOutsideRange` (`SCP.cpp:73-104`), which keeps `slotToKeep` even when it falls below `minSlotIndex`. ✅ Matches PR exactly.

- **Both call sites covered**: `grep -n '\.purge_slots(' crates/herder/src/herder.rs` on the PR branch returns exactly two production call sites (1188, 2506), both now passing `Some(...)`. They map cleanly to C++:
  - `ledger_closed` (line 2506) ↔ `HerderImpl::newSlotExternalized` (`HerderImpl.cpp:278`) `eraseOutsideRange(minSlotToRemember, maxSlotToRemember)`.
  - `out_of_sync_recovery` (line 1188) ↔ `HerderImpl::outOfSyncRecovery` (`HerderImpl.cpp:552`) `eraseOutsideRange(purgeSlot, std::nullopt)`.
  No other henyey call sites were missed.

- **`get_most_recent_checkpoint_seq()` consistency** (`herder.rs:944-957`): reads `tracking_consensus_ledger_index()` and applies `firstLedgerInCheckpointContaining` — the same ledger state and arithmetic the C++ side reads in `HerderImpl::getMostRecentCheckpointSeq` (`HerderImpl.cpp:1414-1419`). Both call sites read it after the same upstream state transitions C++ does (after `ledger_closed` updates `tracking_state`, and after `out_of_sync_recovery` enters the purge branch with the current tracking index).

- **`Scp::purge_slots(max, slot_to_keep)` semantics** (`scp/src/scp.rs:342-346`): `retain |slot| slot >= max || Some(slot) == slot_to_keep` — semantically equivalent to the `minSlotIndex` branch of `SCP::purgeSlotsOutsideRange` with `maybeEraseSlot`. ✅

- **Hoisting `keep_slot` above the SCP purge** in `ledger_closed` matches C++ which computes `lastCheckpointSeq` once and reuses it for the SCP and pending-envelopes calls. ✅

**Non-blocking observations (all pre-existing, not in this PR's scope)**

- `ledger_closed` uses literal `slot.saturating_sub(10)` rather than `get_min_ledger_seq_to_remember()` / `MAX_SLOTS_TO_REMEMBER`. C++ uses `getMinLedgerSeqToRemember()`. Worth a future cleanup but unrelated.
- `ledger_closed` does not apply a `maxSlotToRemember = nextConsensusLedgerIndex() + LEDGER_VALIDITY_BRACKET` upper-bound eviction; henyey's `Scp::purge_slots` only models the `minSlotIndex` branch. Future work; PARITY_STATUS already lists this as Partial.
- `out_of_sync_recovery` passes `purge_slot.saturating_sub(1)` (keep `>= purge_slot - 1`), while C++ uses `eraseOutsideRange(purgeSlot, nullopt)` (keep `>= purgeSlot`). One-slot looser than upstream. Pre-existing.
- C++ `HerderSCPDriver::purgeSlotsOutsideRange` also threads `slotToKeep` into the `mSCPExecutionTimes` cleanup. The henyey sibling `scp_driver.purge_deferred_slots(...)` does not preserve the checkpoint slot. Different map (deferred vs execution times) so likely benign, but flagging as a parity follow-up candidate.
- PendingEnvelopes parity: PR explicitly defers the `PendingEnvelopes::purge_slots_below` sibling gap; PARITY_STATUS note already tracks it.

No blocking parity issues. Tests at lines 12510-12608 directly validate the C++ invariant: checkpoint slot survives even when it sits below the retention window.

</details>

tomerweller:

**Reviewer A (Correctness):** APPROVE — fix correctly threads `Some(get_most_recent_checkpoint_seq())` into both `Scp::purge_slots` call sites; both regression tests would fail without the change.
**Reviewer B (Parity):** APPROVE — both updated call sites (`ledger_closed`, `out_of_sync_recovery`) faithfully mirror stellar-core's `HerderImpl::eraseOutsideRange` → `HerderSCPDriver::purgeSlotsOutsideRange` → `SCP::purgeSlotsOutsideRange` chain; no missed call sites; no divergence.
**External reviewers:** copilot-pull-request-reviewer[bot]: COMMENTED (non-blocking PR overview).
**CI:** red — unrelated.

### CI failure analysis

One real failing job (the other two `FAILURE` entries are cascading `... / complete` aggregators):

- `build / 3 test / 2 test (testing-with-pr, amd64, local, core,rpc,horizon , )` — [job logs](https://github.com/stellar-experimental/henyey/actions/runs/25956301482/job/76306528124)

Excerpt from `--log-failed`:

```
Error: catchup failed: invalid live bucket list structure: invalid HAS:
       future must be cleared at level 5 (prev snap version 25)
Error in ingestion state machine
  current_state="buildFromCheckpoint(checkpointLedger=839, skipChecks=false)"
  error="error preparing range: error starting prepare range: opening subprocess:
         error running stellar-core: error running stellar-core catchup: exit status 1"
##[error]The action 'Run horizon ingesting test' has timed out after 12 minutes.
```

This failure is in the **Quickstart** integration job that boots `stellar-core`/`horizon` and runs a horizon ingestion catchup. The error is a HAS (History Archive State) bucket-list invariant check inside `henyey_bucket::bucket_list` / `henyey_history::catchup::buckets` triggered during stellar-core catchup at ledger 839. It has **no relationship** to the herder SCP-purge change in this PR — the diff is one file (`crates/herder/src/herder.rs`) and touches only `Scp::purge_slots` / `fetching_envelopes.erase_outside_range` plumbing. The catchup/bucket-list path is not exercised by this diff.

This is a pre-existing flake or main-branch breakage on the Quickstart ingestion path.

### Action

Routing back to `ready-for-doing` for `/do` Mode B to rebase on `origin/main` and re-push. If CI is still red after rebase with the same unrelated failure, the next `/review-pr` will mark this `blocked` for human triage of the Quickstart HAS-invariant failure.

Bounce-back count: 1/3.

tomerweller:
Rebased onto current `origin/main` (now at `483d6f8a`, which includes #2783 "Port stellar-core prepareForPublish to clear unrepresentable HAS futures" — the fix for the main-side Quickstart bucket-list-validation regression that blocked the prior CI run). Re-pushed as `b7408b1f` to re-trigger CI.

Local verification on the rebased branch:
- `cargo fmt --check` ✅
- `cargo clippy --all -- -D warnings` ✅
- `cargo test -p henyey-herder` ✅ 1109 passed (incl. both `test_issue_2706_*` regression tests)

No code changes — rebase only. Ready for re-review.

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** Both `Scp::purge_slots` call sites in `crates/herder/src/herder.rs` now thread `Some(get_most_recent_checkpoint_seq())` as `slot_to_keep`, exactly mirroring stellar-core's `HerderImpl::eraseOutsideRange` → `HerderSCPDriver::purgeSlotsOutsideRange` chain.

<details>

**Upstream model (stellar-core v26.0.1):**
- `HerderImpl::eraseOutsideRange` (`stellar-core/src/herder/HerderImpl.cpp:1328-1335`) computes `lastCheckpointSeq = getMostRecentCheckpointSeq()` once and threads it into both `getHerderSCPDriver().purgeSlotsOutsideRange(minSlot, maxSlot, lastCheckpointSeq)` and `mPendingEnvelopes.eraseOutsideRange(minSlot, maxSlot, lastCheckpointSeq)`.
- `HerderSCPDriver::purgeSlotsOutsideRange` (`stellar-core/src/herder/HerderSCPDriver.cpp:1300-1337`) treats `slotToKeep` as a hard exception: any slot equal to `slotToKeep` is skipped over by the `maybePurge` lambda and ultimately forwarded into `getSCP().purgeSlotsOutsideRange(..., slotToKeep)`.
- `HerderImpl::getMostRecentCheckpointSeq` (`HerderImpl.cpp:1414-1419`) returns `firstLedgerInCheckpointContaining(trackingConsensusLedgerIndex())`.

**Call site parity:**

1. **`ledger_closed`** (henyey is the post-apply hook that corresponds to upstream's `newSlotExternalized` → `eraseOutsideRange(minSlotToRemember, maxSlotToRemember)` path at `HerderImpl.cpp:260-278`). The PR computes `keep_slot = self.get_most_recent_checkpoint_seq()` once, then threads it into `self.scp.purge_slots(slot.saturating_sub(10), Some(keep_slot))` and reuses the same value for `fetching_envelopes.erase_outside_range(min_slot, max_slot, keep_slot)`. This mirrors the single-`lastCheckpointSeq`-reused-twice pattern at `HerderImpl.cpp:1332-1335` exactly. ✅

2. **`out_of_sync_recovery`** corresponds to upstream's `outOfSyncRecovery` (`HerderImpl.cpp:546-553`) which calls `eraseOutsideRange(purgeSlot, std::nullopt)` — `std::nullopt` for `maxSlot` and still threading `getMostRecentCheckpointSeq()`. The PR's `self.scp.purge_slots(purge_slot.saturating_sub(1), Some(last_checkpoint))` plus the pre-existing `fetching_envelopes.erase_outside_range(Some(purge_slot), None, last_checkpoint)` together cover the same parity surface. ✅

**Semantics of `slot_to_keep`:**
- henyey's `Scp::purge_slots` (`crates/scp/src/scp.rs:342-346`) retains `slot_index >= max_slot_index || slot_to_keep == Some(slot_index)` — i.e. `slot_to_keep` is an unconditional exception to the purge predicate, matching upstream's `if (it->first == slotToKeep) { ++it; } else { erase; }` at `HerderSCPDriver.cpp:1306-1314`. ✅
- Upstream `slotToKeep` is a non-optional `uint64`; henyey's is `Option<u64>` but the PR always passes `Some(...)`, so behavior is equivalent. ✅

**`get_most_recent_checkpoint_seq` parity:** `crates/herder/src/herder.rs:944-957` computes `firstLedgerInCheckpointContaining(trackingConsensusLedgerIndex())` via the standard `((idx / freq) + 1) * freq - 1` and `size = freq-1` (first checkpoint) vs `freq` (rest) — matches `HistoryManager::firstLedgerInCheckpointContaining`. ✅

**Edge cases:**
- *Checkpoint returns 0*: only possible at genesis when tracking index is 0 with `freq=2`. Upstream then keeps slot 0; henyey does the same (`slot_to_keep == Some(0)` short-circuits the purge predicate). ✅
- *Checkpoint > current slot*: not possible — formula always yields a value ≤ current slot's checkpoint boundary. Same in upstream. ✅
- *Checkpoint inside the retention window*: still passed in; the predicate just becomes a no-op exception. Same in upstream. ✅

**Sequencing parity:** The PR computes `keep_slot` once before the cleanup block in `ledger_closed` and threads the same value into both downstream purgers, exactly as upstream computes `lastCheckpointSeq` once and reuses it across both `purgeSlotsOutsideRange` and `mPendingEnvelopes.eraseOutsideRange` (`HerderImpl.cpp:1332-1335`). The relative ordering of SCP purge → fetching/pending envelope purge → overlay cleanup also matches. ✅

**Out of scope (pre-existing, not regressions of this PR):** henyey uses `slot.saturating_sub(10)` and `purge_slot.saturating_sub(1)` as max-bound thresholds, while upstream uses `getMinLedgerSeqToRemember()` / `purgeSlot` directly. These threshold differences predate this PR and are not within scope of #2706 (which is specifically the missing `slot_to_keep`). Worth filing separately if not already tracked, but does not block.

**Regression tests:** `test_issue_2706_ledger_closed_preserves_checkpoint_slot_in_scp` and `test_issue_2706_out_of_sync_recovery_preserves_checkpoint_in_scp` directly assert the checkpoint slot survives while a control non-checkpoint slot in the same purge band is evicted — exactly the parity invariant being ported.

Parity LGTM. Approving.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Both `Scp::purge_slots` call sites now thread `Some(get_most_recent_checkpoint_seq())` as `slot_to_keep`, precisely the fix described in #2706. The `keep_slot` hoist in `ledger_closed` cleanly deduplicates a previously-duplicated call and is reused for both the SCP and fetching-envelopes purges; in `out_of_sync_recovery` the existing `last_checkpoint` snapshot is reused identically. The two regression tests demonstrably exercise the bug (would fail with `None`) and the existing `Scp::purge_slots` `retain(|&s, _| s >= max || slot_to_keep == Some(s))` semantics are exactly what's needed.

<details>

### Diff correctness

- **`ledger_closed` (herder.rs:~2489):** `let keep_slot = self.get_most_recent_checkpoint_seq()` is hoisted above `scp.purge_slots` and reused for `fetching_envelopes.erase_outside_range` lower down (the duplicate binding at the old location is removed). Single snapshot mirrors stellar-core's `lastCheckpointSeq` lifetime in `HerderImpl::eraseOutsideRange`. ✅
- **`out_of_sync_recovery` (herder.rs:~1180):** `last_checkpoint` was already computed for `fetching_envelopes.erase_outside_range`; reusing it as `Some(last_checkpoint)` for `scp.purge_slots` is the minimal, correct change. ✅
- **No ordering / locking issues:** `Scp::purge_slots` takes the slots write lock internally; `keep_slot`/`last_checkpoint` are plain `u64`s computed before the purge, so there's no TOCTOU concern between snapshotting the checkpoint seq and applying it (and even if the underlying ledger advanced, retaining a now-old checkpoint slot is benign — it's a strict superset of what would otherwise be retained).
- **No off-by-one:** `purge_slots` retains `slot_index >= max_slot_index || slot_to_keep == Some(slot_index)`, so threading `Some(checkpoint)` only ever *adds* a retained slot. The thresholds (`slot.saturating_sub(10)`, `purge_slot.saturating_sub(1)`) are unchanged from prior behavior.
- **Type sanity:** Both `keep_slot`/`last_checkpoint` and `Scp::purge_slots`' `slot_to_keep` are `u64` / `Option<u64>` — no widening, no signed↔unsigned confusion.

### Test coverage

- **`test_issue_2706_ledger_closed_preserves_checkpoint_slot_in_scp`:** Sets `consensus_index=300`, default checkpoint frequency 64 ⇒ checkpoint = 256. Purge threshold = `300-10 = 290`. Asserts slot 256 survives (would be dropped with `None`), slot 250 is purged (control proving purge ran), slot 295 survives (within retention window). Directly reproduces the bug. ✅
- **`test_issue_2706_out_of_sync_recovery_preserves_checkpoint_in_scp`:** Bootstraps at 65 ⇒ checkpoint = 64. Sets 102 v-blocking SCP slots (300..=401) so the descending scan hits `LEDGER_VALIDITY_BRACKET=100` at slot 302 and computes `purge_slot=302`. Asserts result `Some(302)`, slot 64 survives, slot 100 purged. Correctly exercises the previously-`None` branch. ✅
- Both tests would fail without the fix (slot 256 / slot 64 would be retained → assertion fires).
- Test helpers (`Scp::test_set_slot_v_blocking`, `Scp::has_slot`) are pre-existing and `#[cfg(any(test, feature = "test-helpers"))]`-gated appropriately.

### Readability / idiomatic Rust

- Comments at both call sites explicitly cite stellar-core file:line and link the failure mode back to #2670 / #2706. Useful, not noisy.
- Hoisting `keep_slot` into a single local in `ledger_closed` is strictly an improvement (previously called twice).
- No unwrap / panic introduced. No allocations added. Style matches surrounding code.

### Error handling

- N/A: no fallible paths added. `get_most_recent_checkpoint_seq()` returns a plain `u64` and is already used unconditionally elsewhere in both fns.

### Subtle concerns considered and dismissed

- **What if `keep_slot >= max_slot_index`?** Then `retain` keeps it via the `>= max` branch anyway; threading `Some(keep_slot)` is a no-op there. No double-counting / no inconsistency.
- **What if `keep_slot == 0` (pre-bootstrap edge)?** `Some(0)` would retain slot 0 if it existed, which is harmless and matches stellar-core (it also passes whatever `getMostRecentCheckpointSeq()` returns unconditionally).
- **`out_of_sync_recovery` doesn't snapshot `last_checkpoint` under the same lock as the purge** — fine, because `last_checkpoint` is `u64` (Copy), and even a stale-but-larger value would only retain *more* slots, never fewer.

No blockers. Ship it.

</details>

Issue #2683, opened 2026-05-14
Wire validate_bucket_list_structure into catchup apply_buckets


tomerweller:
**Follow-up from #2675 (review-fix finding)**

`validate_bucket_list_structure` was implemented in `archive_state.rs` with full test coverage, but is not yet called from production code.

It should be invoked during `catchup/buckets.rs::apply_buckets` after bucket restoration. This requires:
1. Caching bucket protocol version during `Bucket::from_xdr_file_disk_backed` (currently requires iterating the bucket file)
2. Collecting versions per-level after restoration
3. Calling `validate_bucket_list_structure` with the collected versions

The expensive part is (1) — `protocol_version()` currently iterates the entire bucket. A cached version field on `Bucket` would make this trivial.

**Parent**: #2675

tomerweller:
## Triage Report

**Verdict:** ACCEPT

**Type:** enhancement
**Severity:** n/a
**Crate:** crate:bucket (also touches catchup)
**Size:** M

**Summary:** Wire the existing `validate_bucket_list_structure` check into `catchup/buckets.rs::apply_buckets` so bucket-list structural validation actually runs in production. Requires caching bucket protocol version on `Bucket` to avoid iterating each bucket file just to read its protocol version, then collecting per-level versions and invoking the validator after restoration.

**Label adjustments:** added `enhancement`, `crate:bucket` (issue had no labels).

tomerweller:
## 🔒 acquired-by:tick-1778897907287110801-2047311

posted=2026-05-16T02:18:29.347Z, host=user-dev-050a, pid=2047311

tomerweller:
## 📝 Plan Draft (Round 1)

**Summary:** Cache each `Bucket`'s protocol version (populated cheaply during streaming load) and call the existing `validate_bucket_list_structure` from `catchup/buckets.rs::apply_buckets` after the live and hot-archive bucket lists have been restored, so structural HAS validation actually runs in production.

**Files to modify:**
- `crates/bucket/src/disk_bucket.rs` — extend `StreamingXdrEntryIterator` to capture the metaentry's `ledger_version` before it is filtered out; expose a `metadata_protocol_version()` accessor used by `from_file_streaming_with_seed` and the `from_xdr_bytes` save-then-stream path so the resulting `DiskBucket` carries a `Option<u32>` protocol_version field. For `from_prebuilt`, populate the field by performing one sequential record read (or fall back to lazy memoization via `OnceLock`).
[...]

PR #2783, opened 2026-05-16
Port stellar-core prepareForPublish to clear unrepresentable HAS futures

### LINKED ISSUE #2766: URGENT-CI: Quickstart catchup regression on a0db2c26 — validate_bucket_list_structure rejects legitimate HAS at L90 (future must be cleared at level 3)
## Symptom

`Quickstart` workflow regression on commit `a0db2c26` "Wire validate_bucket_list_structure into catchup apply_buckets" (refs #2683). Two failed attempts (run 25951830752 attempt 1 and 2 both failed with the same error):

```
Error: catchup failed: invalid live bucket list structure:
invalid HAS: future must be cleared at level 3 (prev snap version 25)
```

Repeats every retry. Captive-core inside the Quickstart container is restoring HAS from local archive (post-genesis ledger 90), and the new `validate_bucket_list_structure` call rejects the live bucket list — `future` field at level 3 should be cleared when prev snap is at protocol version 25.

Horizon's ingestion task times out at 4 minutes waiting for captive-core to catch up.

## Regression bracket

| commit | Quickstart |
|--------|-----------|
| `0ea7553b` (prior to a0db2c26) | success ✓ |
| **`a0db2c26`** + `f3a599e6` (skill-only on top) — same diff effectively | **failure x2** ✗ |

Only `a0db2c26` introduces validate_bucket_list_structure into the catchup apply path. `f3a599e6` is `.github/skills/review-pr/` skill changes only, and `3a4fec28` is `.github/skills/project-tick/` skill changes only — neither could cause this.

## Suspected root cause

`validate_bucket_list_structure`'s rule "future must be cleared at level 3" (when prev snap version 25) doesn't match what a real (small-network, post-genesis at L90) HAS contains. Either:

1. The validator is too strict — the rule should permit a populated future field at this stage (e.g. when level 3 hasn't yet drained), or
2. The HAS being emitted by henyey upstream has a malformed/legitimate-but-unexpected future field that the validator wasn't designed to allow.

The "level 3 (prev snap version 25)" message is from `validate_bucket_list_structure` itself. The captive-core ledger restore loaded all levels successfully (levels 0–10 showed `curr_entries=0 snap_entries=0` for the unused ones), then the new validation rejected the result.

## Why deploy-blocking

- This blocks the validator deploy queue. Mainnet validator currently on `0ea7553b` (pre-regression) is fine, but the next deploy will pull HEAD which contains the broken code.
- The captive-core flow exercises catchup → bucket apply → state load, the same code path used during mainnet validator startup. There's no mainnet symptom YET because the mainnet validator restarts from its own persisted state (not via HAS catchup), but a session-wipe or fresh-bootstrap would hit this.
- 3 commits ahead (a0db2c26 + f3a599e6 + 3a4fec28 + 3b31f4b3) all sit behind this gate.

## Suggested action

Either:
1. Relax `validate_bucket_list_structure` to accept the observed legitimate state (a future field that's populated mid-merge is normal in some scenarios).
2. Audit whether henyey's HAS emission is correctly clearing `future` fields at the right level and protocol version transitions.

The validator was wired in *after* the catchup completed — perhaps the validation should run *during* a specific point in the bucket apply, or only on the post-apply final state, not on the loaded HAS.

## Reference

- failed run: `gh run view 25951830752 --log-failed`
- commit: `a0db2c26` "Wire validate_bucket_list_structure into catchup apply_buckets" (refs #2683)
- error originates in `crates/bucket/src/bucket.rs` `validate_bucket_list_structure` → called from `crates/history/src/catchup/buckets.rs` apply_buckets

tomerweller:
Closes #2766

## Summary

`build_history_archive_state` now mirrors stellar-core's
`HistoryArchiveState::prepareForPublish` (HistoryArchive.cpp:466-507)
shadows-removed branch: every emitted HAS clears `next` at any level whose
previous level's `snap` protocol version is ≥ `FIRST_PROTOCOL_SHADOWS_REMOVED`
(12). For protocol 24+ (the only range henyey supports) that means **every**
non-genesis level — the on-disk format cannot represent pending merges past
the shadows-removal boundary and the consumer reconstructs them via
`restart_merges_from_has`.

This unblocks Quickstart catchup, which was regressing on `a0db2c26` (#2683)
because `validate_bucket_list_structure` correctly rejected a legitimate
post-genesis HAS at ledger 90 (`future must be cleared at level 3, prev snap
version 25`). The validator is right; henyey's emitter was wrong.

### Why eager-clear, not lazy-clear

Stellar-core lazy-clears inside `ResolveSnapshotWork::doWork`
(`ResolveSnapshotWork.cpp:30`); in-memory HAS instances freely carry pending
futures. Henyey eager-clears inside `build_history_archive_state` because
that's the single emission entry point — every caller serializing a HAS
goes through it. The on-disk contents are identical; the placement matches
henyey's narrower API surface. A code comment marks this so future
contributors don't introduce a second emission path that bypasses the clear.

The second branch of `prepareForPublish` (the `makeLive` path,
`HistoryArchive.cpp:490-506`) is intentionally skipped: it reconstitutes
pending merges loaded back from an older-protocol on-disk HAS, a data shape
that cannot arise from henyey's live in-memory `BucketLevel` inputs.

### Operator note

Existing stale HAS files in persistent local archives written before this
fix will still trip the validator at catchup. Operators with persisted
archives across the regression boundary need to wipe + republish. CI starts
fresh on each Quickstart run, so the fix unblocks CI immediately.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/2766#issuecomment-4466801121)

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test -p henyey-history` — all 511 unit tests + 7 catchup + 27 doc tests pass, including the four new regression tests:
  - `test_clear_unrepresentable_futures_clears_inputs_at_shadowless_level` — Inputs (state=2) cleared, validator round-trip passes
  - `test_clear_unrepresentable_futures_clears_output_at_shadowless_level` — Output (state=1) cleared (Critic A & B coverage gap)
  - `test_clear_unrepresentable_futures_clears_hot_archive_future` — second helper argument and `HotArchiveBucket::get_protocol_version` API shape
  - `test_build_has_handles_all_empty_bucket_list` — end-to-end sanity through `build_history_archive_state` + the real validator on a fresh-genesis bucket list
- [x] `cargo test --all` — workspace-wide pass, no other suites regressed.

## Deviations from plan

- The three "via build_history_archive_state" regression tests in the plan called for synthetically injecting pending merges into a real `BucketList` and then round-tripping through `build_history_archive_state`. `BucketLevel::next` is module-private in `henyey-bucket` and cannot be set from outside `crates/bucket` without an unrelated API change. Tests instead exercise the same surface where the bug manifests — `HistoryArchiveState::clear_unrepresentable_futures` followed by `validate_bucket_list_structure` on `BucketLevelVersionInfo`s built from the same `snap_versions` vector the production code passes. The end-to-end sanity test still runs through `build_history_archive_state` on a real `BucketList::new()`, so the full pipeline (incl. `Bucket::protocol_version` cache population) is exercised.
- Made `crate::catchup::buckets` and its `build_live_level_version_infos` helper `pub(crate)` so the regression sanity test in `publish.rs` can re-use the production validator wiring; no external API change.
- `HistoryError` has no `PublishFailed` variant; per-level version-read errors map to `HistoryError::VerificationFailed` ("the closest existing variant" per the plan) with the level index in the message.

### DIFF
--- a/crates/history/src/archive_state.rs
+++ b/crates/history/src/archive_state.rs
@@ -375,7 +375,7 @@ pub struct BucketLevelVersionInfo {
 
 /// First protocol version where bucket shadows were removed.
 /// Matches stellar-core's `LiveBucket::FIRST_PROTOCOL_SHADOWS_REMOVED`.
-const FIRST_PROTOCOL_SHADOWS_REMOVED: u32 = 12;
+pub(crate) const FIRST_PROTOCOL_SHADOWS_REMOVED: u32 = 12;
 
 /// Validate bucket list structure and version monotonicity.
 ///
@@ -887,6 +887,102 @@ impl HASBucketLevel {
             curr,
             snap,
             next: HASBucketNext::default(),
+        }
+    }
+}
+
+impl HistoryArchiveState {
+    /// Clear `next` fields that the on-disk HAS format cannot represent.
+    ///
+    /// Mirrors stellar-core's `HistoryArchiveState::prepareForPublish`
+    /// (`stellar-core/src/history/HistoryArchive.cpp:466-507`), restricted to
+    /// the shadows-removed branch (`buckets[i - 1].snap.version >=
+    /// LiveBucket::FIRST_PROTOCOL_SHADOWS_REMOVED`). On those levels a pending
+    /// merge cannot be persisted, because the consumer rebuilds the merge
+    /// deterministically from `snap` + `curr` at restart (see
+    /// `BucketListBase::restartMerges` and henyey's `restart_merges_from_has`).
+    ///
+    /// # Architectural placement vs. stellar-core
+    ///
+    /// Stellar-core clears lazily inside `ResolveSnapshotWork::doWork`
+    /// (`stellar-core/src/historywork/ResolveSnapshotWork.cpp:30`), invoked
+    /// only on the publish path; in-memory `HistoryArchiveState` instances are
+    /// free to carry pending futures. Henyey clears eagerly here because
+    /// `build_history_archive_state` is the single emission entry point —
+    /// every caller reaching for a HAS at publication time goes through it.
+    /// The observable on-disk contents are identical; the placement just
+    /// matches henyey's narrower API surface. Any future code that constructs
+    /// a HAS for emission outside `build_history_archive_state` must call
+    /// this method (or equivalent) before serialization.
+    ///
+    /// # Skipped branch
+    ///
+    /// Stellar-core's second branch in `prepareForPublish` (the
+    /// `next.hasHashes() && !next.isLive()` → `makeLive` path,
+    /// `HistoryArchive.cpp:490-506`) reconstitutes pending merges loaded from
+    /// an older-protocol persisted HAS that came back in over the wire with
+    /// `state == 2` (inputs). Henyey's `build_history_archive_state` consumes
+    /// live in-memory `BucketLevel` state, never a deserialized HAS
+    /// round-trip, so that data shape cannot arise here and the branch is
+    /// intentionally omitted.
+    ///
+    /// # Arguments
+    ///
+    /// * `live_snap_versions` — protocol version of `snap` at each live level,
+    ///   in level order, length must equal `current_buckets.len()`. Empty or
+    ///   pre-metaentry buckets contribute `0`.
+    /// * `hot_snap_versions` — same shape for `hot_archive_buckets`. Must be
+    ///   `Some` iff `hot_archive_buckets` is `Some`, and same length.
+    ///
+    /// # Panics (debug)
+    ///
+    /// Mirrors stellar-core's `releaseAssert(buckets[0].next.isClear())` at
+    /// `HistoryArchive.cpp:475`: level 0 never has a `next`, on either the
+    /// live or the hot-archive list.
+    pub(crate) fn clear_unrepresentable_futures(
+        &mut self,
+        live_snap_versions: &[u32],
+        hot_snap_versions: Option<&[u32]>,
+    ) {
+        Self::clear_levels(&mut self.current_buckets, live_snap_versions);
+
+        match (self.hot_archive_buckets.as_mut(), hot_snap_versions) {
+            (Some(hot), Some(versions)) => {
+                Self::clear_levels(hot, versions);
+            }
+            (None, None) => {}
+            (Some(_), None) | (None, Some(_)) => {
+                debug_assert!(
+                    false,
+                    "hot_archive_buckets / hot_snap_versions presence mismatch",
+                );
+            }
+        }
+    }
+
+    fn clear_levels(levels: &mut [HASBucketLevel], snap_versions: &[u32]) {
+        debug_assert_eq!(
+            levels.len(),
+            snap_versions.len(),
+            "snap_versions length must match bucket levels",
+        );
+        if levels.is_empty() {
+            return;
+        }
+        debug_assert!(
+            levels[0].next.is_clear(),
+            "level 0 must never carry a pending merge (parity with stellar-core HistoryArchive.cpp:475)",
+        );
+        // Predicate matches the validator's indexing
+        // (`prev_snap_version = levels[j - 1].snap_version`,
+        // archive_state.rs:449) and stellar-core's
+        // `buckets[i - 1].snap` reference at
+        // `stellar-core/src/history/HistoryArchive.cpp:483-489`.
+        for i in 1..levels.len() {
+            let prev_snap_version = snap_versions[i - 1];
+            if prev_snap_version >= FIRST_PROTOCOL_SHADOWS_REMOVED && !levels[i].next.is_clear() {
+                levels[i].next = HASBucketNext::default();
+            }
         }
     }
 }
--- a/crates/history/src/catchup/buckets.rs
+++ b/crates/history/src/catchup/buckets.rs
@@ -68,7 +68,7 @@ pub(super) fn verified_hot_archive_bucket_loader(
 /// `Bucket::protocol_version()` (O(1) after load); empty / metaentry-less
 /// buckets map to version 0, matching stellar-core's
 /// `bucket->isEmpty() ? 0 : bucket->getBucketVersion()` branch.
-pub(super) fn build_live_level_version_infos(
+pub(crate) fn build_live_level_version_infos(
     levels: &[BucketLevel],
     has_levels: &[HASBucketLevel],
 ) -> Result<Vec<BucketLevelVersionInfo>> {
--- a/crates/history/src/catchup/mod.rs
+++ b/crates/history/src/catchup/mod.rs
@@ -46,7 +46,7 @@
 //! - Incremental eviction scan must run each ledger for correct hashes
 //! - Bucket list hash = SHA256(live_hash || hot_archive_hash)
 
-mod buckets;
+pub(crate) mod buckets;
 mod download;
 mod persist;
 mod replay;
--- a/crates/history/src/publish.rs
+++ b/crates/history/src/publish.rs
@@ -223,39 +223,95 @@ pub fn build_history_archive_state(
         })
         .collect();
 
-    let hot_archive_buckets = hot_archive_bucket_list.map(|habl| {
-        habl.levels()
-            .iter()
-            .map(|level| {
-                let next = match level.pending_merge_output_hash() {
-                    Some(hash) => HASBucketNext {
-                        state: HAS_NEXT_STATE_OUTPUT,
-                        output: Some(hash.to_hex()),
-                        curr: None,
-                        snap: None,
-                        shadow: None,
-                    },
-                    None => HASBucketNext::default(),
-                };
-                HASBucketLevel {
-                    curr: level.curr().hash().to_hex(),
-                    snap: level.snap_bucket().hash().to_hex(),
-                    next,
-                }
-            })
-            .collect()
-    });
+    // Per-level snap protocol versions for the shadows-removed
+    // `next`-clearing pass below. `Bucket::protocol_version` returns
+    // `Ok(None)` for empty buckets / buckets without a leading metaentry;
+    // map those to 0 so the validator-equivalent predicate
+    // `prev_snap_version >= FIRST_PROTOCOL_SHADOWS_REMOVED` treats them as
+    // "no constraint" (matching `validate_bucket_list_structure` semantics).
+    let live_snap_versions: Vec<u32> = bucket_list
+        .levels()
+        .iter()
+        .enumerate()
+        .map(|(idx, level)| {
+            level
+                .snap
+                .protocol_version()
+                .map(|v| v.unwrap_or(0))
+                .map_err(|e| {
+                    HistoryError::VerificationFailed(format!(
+                        "failed to read snap protocol version at live level {}: {}",
+                        idx, e
+                    ))
+                })
+        })
+        .collect::<Result<_>>()?;
+
+    let hot_archive_buckets = hot_archive_bucket_list
+        .map(|habl| -> Result<Vec<HASBucketLevel>> {
+            habl.levels()
+                .iter()
+                .map(|level| {
+                    let next = match level.pending_merge_output_hash() {
+                        Some(hash) => HASBucketNext {
+                            state: HAS_NEXT_STATE_OUTPUT,
+                            output: Some(hash.to_hex()),
+                            curr: None,
+                            snap: None,
+                            shadow: None,
+                        },
+                        None => HASBucketNext::default(),
+                    };
+                    Ok(HASBucketLevel {
+                        curr: level.curr().hash().to_hex(),
+                        snap: level.snap_bucket().hash().to_hex(),
+                        next,
+                    })
+                })
+                .collect()
+        })
+        .transpose()?;
+
+    // Hot-archive snap protocol versions. Note the API shape mismatch with
+    // the live path: `HotArchiveBucket::get_protocol_version` returns
+    // `Result<u32>` (no `Option`); empty hot-archive buckets already report 0.
+    let hot_snap_versions: Option<Vec<u32>> = hot_archive_bucket_list
+        .map(|habl| -> Result<Vec<u32>> {
+            habl.levels()
+                .iter()
+                .enumerate()
+                .map(|(idx, level)| {
+                    level.snap_bucket().get_protocol_version().map_err(|e| {
+                        HistoryError::VerificationFailed(format!(
+                            "failed to read snap protocol version at hot-archive level {}: {}",
+                            idx, e
+                        ))
+                    })
+                })
+                .collect()
+        })
+        .transpose()?;
 
     let version = if hot_archive_buckets.is_some() { 2 } else { 1 };
 
-    Ok(HistoryArchiveState {
+    let mut has = HistoryArchiveState {
         version,
         server: Some("rs-stellar-core".to_string()),
         current_ledger: ledger_seq,
         network_passphrase,
         current_buckets,
         hot_archive_buckets,
-    })
+    };
+
+    // Mirror stellar-core's `HistoryArchiveState::prepareForPublish`
+    // shadows-removed branch (HistoryArchive.cpp:483-489). Henyey eager-clears
+    // here rather than lazily in a publish work step so every HAS emission
+    // path runs through this single helper. See
+    // `HistoryArchiveState::clear_unrepresentable_futures` for the full
+    // rationale and the documented skip of stellar-core's `makeLive` branch.
+    has.clear_unrepresentable_futures(&live_snap_versions, hot_snap_versions.as_deref());
+
+    Ok(has)
 }
 
 impl PublishManager {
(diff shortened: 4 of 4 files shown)


tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Port of stellar-core `prepareForPublish` shadows-removed branch is a faithful, narrowly-scoped fix that resolves the documented CI regression without introducing new compatibility or performance hazards. Operator footgun (stale local HAS archives) is acknowledged in the PR body.

<details>

### Parity with stellar-core (HistoryArchive.cpp:466-507)

- Predicate `prev_snap_version >= FIRST_PROTOCOL_SHADOWS_REMOVED` matches `protocolVersionStartsFrom(snap->getBucketVersion(), LiveBucket::FIRST_PROTOCOL_SHADOWS_REMOVED)`. ✓
- `prev = buckets[i - 1]` and indexing into `snap_versions[i - 1]` line up with the C++ `prev.snap` reference. ✓
- Level-0 assertion preserved (`debug_assert!` vs C++ `releaseAssert`). Minor parity gap: a release build with a stray level-0 `next` would silently pass it through. The bucket list never produces this in practice, but consider promoting to a hard error (`return Err(...)`) to fail loudly if it ever happens; not blocking.
- Skipping the `makeLive` branch (`HistoryArchive.cpp:490-506`) is correctly justified — henyey's `build_history_archive_state` always feeds from live in-memory `BucketLevel`, never from a deserialized older-protocol HAS. ✓
- Both live and hot-archive branches in stellar-core unconditionally compare against `LiveBucket::FIRST_PROTOCOL_SHADOWS_REMOVED` (not a per-bucket-type constant), so henyey's single-constant use for both lists is parity-correct. ✓
- Predicate is on `!next.is_clear()`, which covers both Inputs (state=2) and Output (state=1) states — matches C++ which only branches on `isClear()`. Output-state test added covers this explicitly. ✓
- Predicate compatibility with validator: the validator additionally has a `non_empty_seen` walking state machine, but in protocol-24+ steady state every level with a populated `next` will have a populated prev_snap, so the eager-clear is always strict-enough to satisfy the validator. The all-empty path is covered by the end-to-end test.

### Regressions

- **HAS on-disk/on-wire shape changes**: protocol 24+ HAS will no longer contain pending-merge `next` fields. This is the same shape stellar-core writes, so any consumer that is parity-compliant (henyey's own catchup via `restart_merges_from_has`, stellar-core itself, Horizon's captive-core) reconstructs merges deterministically. No semantic regression.
- **Stale archives**: PR body correctly calls out that already-persisted (pre-fix) HAS files on local archives will still trip the validator at restart and need to be wiped + republished. CI starts clean each run so this is opt-in pain only for long-lived persistent local archives. Worth verifying that any mainnet validator with a local archive across the regression bracket gets a wipe-and-republish runbook before the next deploy.
- No change to publishable history-archive entries beyond the `next` field. The version field selection (1 vs 2) is unchanged. `current_ledger`, `network_passphrase`, `server`, `current_buckets[*].curr/snap` are all untouched.

### Performance

- One extra pass over `BUCKET_LIST_LEVELS` (~11) per HAS emission, plus one `protocol_version()` per level. Live bucket path is O(1) cached; hot-archive path triggers `ensure_index()` + a one-entry read on first call for disk-backed buckets. For publish (per-checkpoint, every 64 ledgers) and catchup HAS-write, this is negligible. No tight-loop hot path affected.

### Security

- No new attack surface. The change strictly removes optionally-emitted fields, narrowing what's serialized.

### Operational concerns

- No config, migration, or deploy-ordering changes required beyond the operator note about persistent local archives.
- Visibility of new `pub(crate)` (`FIRST_PROTOCOL_SHADOWS_REMOVED`, `catchup::buckets`, `build_live_level_version_infos`) is appropriately scoped — no public API surface change.
- `HistoryError::VerificationFailed` reuse for the per-level snap-version read failure is a slight misnomer (it's an internal pre-emission read error, not a verification mismatch) but acceptable given the absence of a better variant; not blocking.

### Minor (non-blocking, inline)

- Consider promoting the level-0 `debug_assert!` to a hard `Err` return for parity with stellar-core's `releaseAssert` semantics (loud failure in release builds).
- Doc-comment on `clear_unrepresentable_futures` mentions a debug-only "presence mismatch" assertion for hot bucket arg pairing; on release builds the mismatch is silently ignored. Consider returning a `HistoryError` instead.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Fix correctly mirrors stellar-core's `prepareForPublish` shadows-removed branch and the predicate matches `validate_bucket_list_structure` exactly, so the regression at L90 is resolved. Tests cover the helper thoroughly; only minor non-blocking gaps remain.

<details>

**Correctness — fix resolves #2766**

- The eager-clear predicate in `clear_levels` (`snap_versions[i - 1] >= FIRST_PROTOCOL_SHADOWS_REMOVED`) is identical to the validator's predicate at `archive_state.rs:458` (`prev_snap_version = levels[j - 1].snap_version; if prev_snap_version >= FIRST_PROTOCOL_SHADOWS_REMOVED { next.is_clear() }`). Because the clear is unconditional (modulo predicate) and clearing an already-clear `next` is a no-op, the post-clear HAS always satisfies the rule that was previously failing at L90 with `prev snap version 25`. ✓
- The validator's `non_empty_seen` gate is strictly weaker than the eager-clear predicate: any `prev_snap_version >= 12` already implies a non-empty snap and therefore `non_empty_seen == true` at the time the validator inspects level `j`. So the eager-clear cannot "over-clear" relative to what the validator would have demanded.
- Information loss (dropped pending merges) is recoverable: every consumer reconstructs them via `restart_merges_from_has` (verified call-sites in `crates/history/src/catchup/buckets.rs`, `crates/app/src/app/ledger_close.rs`, `crates/henyey/src/verify_execution.rs`). ✓
- `build_history_archive_state` is the single emission entry point: confirmed by grep — all callers in `crates/henyey/src/publish_history.rs`, `crates/app/src/app/{bootstrap,catchup_impl}.rs`, and the publish flow route through it. The docstring's warning to future contributors is appropriate.
- The skipped `makeLive` branch is correctly justified: henyey builds HAS from live in-memory `BucketLevel` state, never from a deserialized older-protocol HAS round-trip.

**Per-level snap-version sourcing**

- Live path: `level.snap.protocol_version()` returns `Result<Option<u32>>`, mapped `None → 0`. Empty/pre-metaentry buckets contribute 0, predicate becomes "no constraint". Matches validator semantics. ✓
- Hot-archive path: `level.snap_bucket().get_protocol_version()` returns `Result<u32>` (empty already = 0). API-shape mismatch correctly noted in code comment and PR body. ✓
- The hot/live argument-presence invariant (both `Some` or both `None`) is enforced via `debug_assert!` in `clear_unrepresentable_futures`. In release builds the `(Some, None)` and `(None, Some)` arms silently no-op the hot side. Reasonable — `build_history_archive_state` is the only caller and it always builds them in lockstep — but worth a single inline guard. Non-blocking.

**Test coverage**

- `test_clear_unrepresentable_futures_clears_inputs_at_shadowless_level` is the failing-test-first regression for #2766: it asserts validator rejection on `state=2` at level 3 with `prev snap version 25` (exact bug shape from the issue), then asserts the clearing fixes it. ✓
- Output-state coverage (`state=1`) at a different level/version is covered separately. ✓
- Hot-archive coverage exercises the second helper argument and the `HotArchiveBucket::get_protocol_version` API shape. ✓
- End-to-end test `test_build_has_handles_all_empty_bucket_list` validates the `build_history_archive_state` wiring on the empty path.

**Test coverage gap (non-blocking, inline note)**

- No test actually flows a populated-`next` bucket list through `build_history_archive_state` end-to-end, because `BucketLevel::next` is module-private in `henyey-bucket`. The PR honestly documents this in "Deviations from plan". The wiring at the new `has.clear_unrepresentable_futures(...)` call site is therefore only verified by inspection plus the all-empty smoke test. A future test that either (a) adds a `#[cfg(test)] pub(crate)` setter on `BucketLevel::next`, or (b) drives a real merge enough ledgers for level 3 to acquire a pending future, would close this. Non-blocking — clearing helper itself is well-tested with the real validator, and the wiring is a single line.

**Readability / docs**

- Docstrings are extensive and cite stellar-core line numbers — exactly the right call for a parity port. The "Architectural placement" and "Skipped branch" sections will save the next person significant archaeology.
- The comment block on lines 100-104 of the diff explaining the `i - 1` indexing tying back to validator + stellar-core source is excellent.

**Error handling**

- `HistoryError::VerificationFailed` per-level error includes the level index. Acceptable mapping per plan. The error case (a bucket whose `protocol_version()` errors) is exotic but propagation is correct rather than `.unwrap()`-ing. ✓

**Minor nits (non-blocking)**

- The `(Some, None) | (None, Some)` arm uses `debug_assert!(false, ...)` and then falls through silently in release. Consider returning early or using `unreachable!()` if you'd prefer a louder release-mode failure — but since the only caller wires both consistently this is defensible.
- `live_snap_versions` is always computed even when no level has a populated `next`. Trivial cost (~11 cache-hot `protocol_version()` calls) vs. clarity win of an unconditional pipeline; fine as-is.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Port faithfully mirrors stellar-core's `prepareForPublish` shadows-removed branch; the predicate, level-0 invariant, and per-list iteration are correct, and the new tests reproduce the #2766 validator rejection pre-fix and confirm the clearing pass restores validity. A few non-blocking nits below.

<details>

**Correctness — looks good**

- Predicate `prev_snap_version >= FIRST_PROTOCOL_SHADOWS_REMOVED && !levels[i].next.is_clear()` matches stellar-core `HistoryArchive.cpp:478-487` exactly, including the indexing (`buckets[i-1].snap` ↔ `snap_versions[i-1]`) — same indexing the validator uses at `archive_state.rs:449`. No off-by-one.
- `clear_levels` correctly iterates `1..levels.len()`, asserting level 0 is clear (parity with `releaseAssert(buckets[0].next.isClear())` at `HistoryArchive.cpp:475`).
- Empty/no-metaentry buckets correctly map `protocol_version() -> Ok(None)` to `0`; the validator's `non_empty_seen` guard treats `0` as "no constraint", so all-empty bucket lists are a no-op (covered by `test_build_has_handles_all_empty_bucket_list`).
- The skipped `makeLive` branch is sound: `build_history_archive_state` only consumes live in-memory `BucketLevel`s with a resolved `pending_merge_state()`, never a deserialized state-2 HAS coming back in over the wire. The doc comment captures this clearly.
- Live vs. hot API shape mismatch (`Result<Option<u32>>` vs `Result<u32>`) is handled correctly and documented inline.

**Test coverage — adequate**

- `test_clear_unrepresentable_futures_clears_inputs_at_shadowless_level` reproduces the exact #2766 symptom (validator rejects state=2 at level 3 with prev snap 25) pre-fix, then asserts the validator accepts post-fix. This is the failing-test-first regression the workflow guide asks for.
- `*_clears_output_at_shadowless_level` covers the state=1 variant (state-agnostic clearing rule).
- `*_clears_hot_archive_future` exercises the hot-archive branch and the `Result<u32>` API.
- `test_build_has_handles_all_empty_bucket_list` covers the genesis no-op path end-to-end.

**Non-blocking nits**

1. `clear_unrepresentable_futures` — the `(Some, None) | (None, Some)` arm is `debug_assert!(false, ...)` only. In release builds a presence mismatch silently skips clearing the hot list (or no-ops with stray versions), which conflicts with the repo's "never fail silently" convention. Cheap to convert to an `assert!` or to make the function return `Result` and error out — current callers always pass coherent pairs, so impact is theoretical, but flagging for follow-up.

2. No end-to-end test calls `build_history_archive_state` against a real `BucketList` containing a pending merge at a non-zero level with prev-snap protocol 25 and then runs the validator. The pieces are individually tested (unit clearing + empty E2E), but a single end-to-end test would have caught regressions where the wiring drifts (e.g. if a future refactor builds the HAS without calling `clear_unrepresentable_futures`). Good follow-up candidate.

3. Visibility widening (`FIRST_PROTOCOL_SHADOWS_REMOVED` → `pub(crate)`, `mod buckets` → `pub(crate)`, `build_live_level_version_infos` → `pub(crate)`) is only used by the new tests; consider keeping these private and inlining a local helper or `#[cfg(test)]`-gating the re-exports. Minor.

4. The `level.snap.protocol_version()` call now runs on the publish hot path for every level. It's `O(1)` after the first call (OnceLock-memoized for in-memory, cached for disk-backed), so no real concern, but worth being aware that this is a new per-publish call across all 11 levels.

5. Doc comment in `clear_unrepresentable_futures` says level 0 cannot have `next` "on either the live or the hot-archive list" — the second `debug_assert!(levels[0].next.is_clear(), ...)` runs for both lists via `clear_levels`, so this matches the code. Good.

</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Faithful port of stellar-core's `prepareForPublish` shadows-removed branch with documented architectural divergence (eager vs. lazy). No format/API breakage, negligible perf overhead, parity preserved. A few non-blocking observations below.

<details>

### Parity check vs `stellar-core/src/history/HistoryArchive.cpp:466–515`

Walked the upstream `prepareForPublish` and compared:

- Loop bounds (`for i in 1..numLevels`), level-0 assert, predicate (`prev.snap version >= FIRST_PROTOCOL_SHADOWS_REMOVED && !next.isClear()` → `clear()`), and dual application to live + hot archive all match. ✓
- `FIRST_PROTOCOL_SHADOWS_REMOVED = 12` matches upstream `LiveBucket::FIRST_PROTOCOL_SHADOWS_REMOVED`. ✓
- Upstream resolves `prev.snap` via `BucketManager::getBucketByHash(hexToBin256(prev.snap))`; henyey reads `bucket_list.levels()[i-1].snap.protocol_version()` directly. These are necessarily the same bucket because the HAS is built from this same `bucket_list` in the same call — no parity gap. ✓
- Skipped `makeLive` branch (upstream `HistoryArchive.cpp:490-506`) is correctly omitted: it only triggers on deserialized state-2 round-trips from older-protocol persisted HAS, and `build_history_archive_state` only consumes live in-memory state. Well-documented. A future maintainer who adds a HAS-deserialization → re-emit path will need to handle this — the docstring already calls it out.

### Performance

- Per HAS emission: O(levels) = ~11 calls to `Bucket::protocol_version()`, which is O(1) cached (`bucket.rs:909`, `disk_bucket.rs:550`). Negligible.
- One mild caveat: hot-archive `get_protocol_version()` on a disk-backed bucket *can* do a real file read via `ensure_index()` + `read_entry_at_offset()` (no in-memory cache). HAS construction runs every ledger via `ledger_close.rs:428` when a hot archive is present, so this is per-ledger I/O on each level's snap. In practice the index is already cached after first use and the entries are small, but if hot-archive HAS construction shows up in profiles, consider memoizing `get_protocol_version()` the same way `Bucket::protocol_version()` does (`OnceLock`). Not blocking.

### Behavior / regressions

- Pre-existing `current_buckets` path already emitted `next` fields for state=1 and state=2 (see `publish.rs:200-217`). Adding eager-clearing here changes wire-observable output for HAS files emitted by henyey, but the change is precisely what brings them into conformance with stellar-core and the validator. This *fixes* a divergence rather than introducing one.
- All ~9 callers of `build_history_archive_state` use the result for serialization to disk/JSON or downstream consumption (catchup persist, ledger close, bootstrap, publish manager). None depend on `next` being preserved post-build, so eager clearing is safe.
- `restart_merges_from_has` correctness is preserved: cleared `next` is intentional — the consumer rebuilds the merge from `snap` + `curr` deterministically, matching stellar-core's `BucketListBase::restartMerges`.

### Determinism / on-disk format

- Output HAS files become byte-identical to what stellar-core would emit on the publish path for the same bucket state. This is a positive determinism delta.
- No on-disk format change. No migration needed.
- No new config knobs, no monitoring impact.

### Operational

- No new failure modes added: the two new `protocol_version()` calls in `build_history_archive_state` map errors to `HistoryError::VerificationFailed` with descriptive context. Good.
- The visibility bumps (`FIRST_PROTOCOL_SHADOWS_REMOVED` → `pub(crate)`, `buckets` module → `pub(crate)`, `build_live_level_version_infos` → `pub(crate)`) are correctly scoped — no public API exposure.

### Non-blocking nits

1. The `(Some(_), None) | (None, Some(_))` mismatch arm uses `debug_assert!(false, …)` and silently falls through in release. Since this is a hard internal invariant violation, a `panic!` (or returning a `Result`) would arguably be safer — but with only one caller (`build_history_archive_state`) that always passes the two in lockstep, the dead-branch risk is essentially zero. Fine to leave.
2. The fresh-genesis end-to-end test (`test_build_has_handles_all_empty_bucket_list`) is good coverage for the empty-bucket `Ok(None) → 0` mapping. A symmetric test exercising mixed populated/empty levels would tighten things further, but the existing tests + the in-production validator round-trip make this low priority.
3. Consider an integration-level smoke test that runs the failing checkpoint (L90 from #2766) end-to-end through `build_history_archive_state` + `validate_bucket_list_structure`. The unit tests cover the mechanism, but a fixture-driven test would lock in the specific regression. Optional.

No blockers.

</details>

Issue #2713, opened 2026-05-15
metrics: recovery-stalled burst (delta=240) + 155k overlay backpressure on post-deploy catchup-behind


tomerweller:
## Symptom

After deploy of `8b2c578c` (89 commits, f486ea12..main), post-restart catchup window (10:28→10:39) produced ~155k overlay backpressure WARNs and triggered the `recovery-stalled` burst alarm.

`/monitor-tick` at 2026-05-15T10:38Z observed:
- `henyey_recovery_stalled_tick_total{reason="forcing_catchup_behind"}` burst delta=240 (≥10 threshold)
- 4,525 `Per-slot lifetime cap reached in FetchingEnvelopes` WARNs (10:30→10:39)
- 155,841 `Broadcast backpressure: messages dropped due to full peer channels` WARNs
- 5 state transitions Validating ↔ Catching Up between 10:32→10:39
- 1 ERROR: `HardResetEscalation catchup failed error=Archive checkpoint 62577919 is not ahead of min_ledger 62577919`

Validator recovered: caught up to peer_gap=0 at 10:39:50Z, currently Validating with qset agree=21/missing=0.

## Counters at 754s uptime (post-recovery)
```
stellar_scp_envelope_receive_total = 1,256,412   (~104k/min sustained)
henyey_scp_post_verify_drops_total = 182,871
henyey_scp_post_verify_total{reason="per_slot_full"} = 62,845
```

Receive rate is ~5× the steady-state expectation (21 peers × ~17 env/s = 357/s; we saw ~1,740/s during catchup). Plausibly normal for a 56-ledger catchup, but the `per_slot_full` drops + `Per-slot lifetime cap` warns indicate the anti-flood guard (PDR-2644) was triggered repeatedly during recovery.

## Why this differs from prior closed issues

- **#2664** (CLOSED, fix `12a87adc` confirmed in deploy `8b2c578c`): suppresses spurious hard reset *at-tip* when `archive_confirmed_behind` is stale. Today's hard resets fired while **56 ledgers behind tip**, where the fix doesn't apply. Log shows `archive_behind_until_was_armed=false`.
- **#2309** (CLOSED): single-increment calibration. Today is a 240-burst, not a calibration issue.
- **#2562** (CLOSED, fix `e6983c26`): related catchup-loop oscillation; appears to handle a different state.

## Suspected root cause

Two distinct concerns:

[...]

PR #2790, opened 2026-05-17
Narrow #2713 hard-reset suppression to at-or-near-tip (#2789)

### LINKED ISSUE #2789: metrics: mid-run recovery cycle (lost-sync + hard-reset + recovery-stalled) on 52ac1665 at 17h uptime
## Symptom

Validator on `52ac1665` (build includes `53d6a092` from #2713 fix) hit a brief recovery cycle at 2026-05-17T13:29–13:31Z, ~17h into uptime. Self-healed within ~90 seconds.

Three metric alarms past cooldown:
- `stellar_herder_lost_sync_total` delta=8 (SYNC tier)
- `henyey_post_catchup_hard_reset_total` delta=1 (ACTION tier)
- `henyey_recovery_stalled_tick_total{reason="forcing_catchup_behind"}` burst (WARN tier)

## Timeline

```
13:29:22Z  Lost sync with network - transitioning to syncing state
13:30:07Z  Consensus stuck - transitioning to out-of-sync
13:30:09Z  State transition Validating → Catching Up
13:30:09Z  State transition Catching Up → Validating
13:31:02Z  Lost sync with network (2nd)
13:31:07Z–13:31:16Z  "Recovery stalled for too long — forcing catchup"
           current_ledger=62609921 latest_ext=62609977 gap=56 attempts=58→67
           (10 successive iterations within 9 seconds)
13:31:16Z  Catching Up → Validating
13:31:28Z  Sync restored, peer_gap=0 sustained from then onward
```

20 minutes later (this filing): agree=21/missing=0, peer_gap=0, healthy.

## Why this differs from #2713 (closed)

- #2713 was the **post-deploy** catchup-behind wedge that lasted 11+ minutes during fresh recovery from a stale LCL. Fix `53d6a092` "Suppress futile hard resets when archive is confirmed behind" landed.
- This incident is **mid-run** (17h post-deploy, healthy validator suddenly drops sync), recovers in 90 seconds, and the `Recovery stalled for too long` loop fires 10× in 9 seconds before the state transition resolves. Different shape.

Possible causes:
- Transient network glitch (peer churn → quorum dropout → re-quorum) where the lost-sync path took the hard-reset escalation despite the #2713 mitigation.
- Or: the `archive_confirmed_behind` arming condition that #2713 fixed doesn't apply to this particular code path (lost_sync rather than catchup-behind).

## Evidence

Validator currently:
- PID 175262, uptime 20h08m, build 52ac1665 (clean since the 17:39Z deploy yesterday)
- agree=21, missing=0, peer_gap=0, lag_ms=458 — fully recovered

Recovery was complete. No operator action needed. Filing as documentation for the long tail.

## Reference

- log evidence: `/home/tomer/data/74535976/logs/monitor.log` around `2026-05-17T13:29–13:31Z`
- counters in `~/data/74535976/metrics/current.prom`
- Related to #2713 (closed by 53d6a092)

tomerweller:
Closes #2789

## Summary

The #2713 hard-reset suppression in `force_post_catchup_hard_reset` fires whenever the archive is `ConfirmedBehind` with a fresh cache at-or-below `current_ledger`, regardless of how far peers have advanced. On a long-running validator that briefly drops sync, peers race ahead 40–70 ledgers while the archive lags publication (normal mid-checkpoint state). The suppression then blocks the only state-changing recovery path, leaving the validator wedged at LCL for ~90s until an external event forces a state transition. Issue #2789 reports four recurrences in ~9h on `52ac1665`, accelerating to sub-hourly.

This PR gates the suppression on `peer_gap < HARD_RESET_GAP_ESCALATION (12)` so it only fires at-or-near the network tip — the original #2713 case. When peers are clearly ahead the function falls through to the existing `spawn_catchup(ProbeAhead)` path which transitions state to `CatchingUp`, breaking the wedge. The retry loop remains bounded by the 60s cooldown and the livelock circuit-breaker.

`peer_gap` reads from `effective_peer_gap` (`max_verified_scp_slot − current_ledger`), not the herder's `latest_externalized` — when the validator is stuck waiting for tx_sets, `latest_externalized` stays at LCL while `max_verified_scp_slot` reflects the network tip.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/2789#issuecomment-4472870887) — converged Round 1 (A: REVISE-MINOR addressed, B: APPROVE, C: REVISE-MINOR addressed).

## Parity

Confirmed by Critic B against `stellar-core/` v26.0.1: `LedgerApplyManagerImpl::processLedger` unconditionally fires `startOnlineCatchup()` on a buffered EXTERNALIZE for the trigger ledger; archive readiness is checked inside `CatchupWork::getAndMaybeSetHistoryArchiveState()`. stellar-core never caches "archive behind" to suppress future catchups. This change moves henyey strictly closer to upstream semantics.

## Test plan

- [x] `cargo fmt --check` clean
- [x] `cargo clippy --all -- -D warnings` clean
- [x] `cargo test -p henyey-app` — 940 passed; 0 failed
- [x] `cargo test -p henyey-app --lib hard_reset` — 39 passed (all existing hard-reset coverage + 2 new regressions)

## Regression test

- **Tests:** `crates/app/src/app/catchup_impl.rs::tests::test_hard_reset_not_suppressed_when_peers_ahead` and `crates/app/src/app/catchup_impl.rs::tests::test_recovery_loop_unwedges_when_peers_ahead`
- **Pre-fix:** committed as `dca276bc` — verified FAILED on `origin/main` with "expected counter > 0, got 0" / "expected counter >= 1, got 0"
- **Post-fix:** verified PASS after `a907cc41`

Existing `test_hard_reset_suppressed_fresh_cache_behind` updated to explicitly set `max_verified_scp_slot = current_ledger + 5` (peer_gap=5 < 12) so it continues to exercise the meaningful at-tip suppression window after the fix.

## Deviations from plan

None.

### DIFF
--- a/crates/app/src/app/catchup_impl.rs
+++ b/crates/app/src/app/catchup_impl.rs
@@ -1366,14 +1366,29 @@ impl App {
         // published). Cold means no data (must probe). Leave
         // archive_recovery_status as ConfirmedBehind so decision function stays
         // in the archive_behind branch where cooldown → AttemptRecovery.
-        let archive_cannot_provide_target =
+        //
+        // #2789 (narrowing): only suppress when the validator is at-or-near
+        // the network tip. Peer-gap is read from verified SCP evidence
+        // (max_verified_scp_slot, via effective_peer_gap) rather than from
+        // the herder's latest_externalized — when the node is stuck waiting
+        // for tx_sets, latest_externalized stays at LCL while
+        // max_verified_scp_slot reflects the network tip. When peers are
+        // clearly ahead (peer_gap >= HARD_RESET_GAP_ESCALATION = 12), the
+        // suppression blocks the only state-changing recovery path and
+        // leaves the validator wedged for ~90s. Falling through to the
+        // hard-reset+ProbeAhead path is bounded by the 60s cooldown and the
+        // livelock circuit-breaker, and the spawn_catchup state transition
+        // to CatchingUp itself breaks the wedge.
+        let peer_gap = self.effective_peer_gap(current_ledger);
+        let suppress_archive_behind_at_tip =
             self.archive_recovery_snapshot().await.is_confirmed_behind()
+                && peer_gap < HARD_RESET_GAP_ESCALATION
                 && matches!(
                     self.get_cached_archive_checkpoint_nonblocking(),
                     CacheResult::Fresh(v) if v <= current_ledger
                 );
 
-        if archive_cannot_provide_target {
+        if suppress_archive_behind_at_tip {
             // Record cooldown so decision function routes to AttemptRecovery.
             self.last_hard_reset_offset
                 .store(now_offset.max(1), Ordering::Relaxed);
@@ -1392,16 +1407,19 @@ impl App {
                     current_ledger,
                     latest_externalized = latest_ext,
                     gap = current_gap,
+                    peer_gap,
                     ?reason,
                     "Hard reset suppressed: archive confirmed behind with fresh \
-                     cache at/below current_ledger — cooldown armed, requesting \
-                     SCP state from peers (#2713)"
+                     cache at/below current_ledger AND peers within \
+                     HARD_RESET_GAP_ESCALATION — cooldown armed, requesting \
+                     SCP state from peers (#2713, narrowed by #2789)"
                 );
             } else {
                 tracing::debug!(
                     current_ledger,
                     latest_externalized = latest_ext,
                     gap = current_gap,
+                    peer_gap,
                     ?reason,
                     "Hard reset suppressed (#2713) (repeated)"
                 );
@@ -5642,7 +5660,9 @@ mod tests {
     // ================================================================
 
     /// When archive is ConfirmedBehind AND fresh cache shows checkpoint
-    /// at/below current_ledger, the hard reset should be suppressed.
+    /// at/below current_ledger AND peers are at-or-near tip (peer_gap <
+    /// HARD_RESET_GAP_ESCALATION), the hard reset should be suppressed.
+    /// This is the original #2713 at-tip case.
     #[tokio::test]
     async fn test_hard_reset_suppressed_fresh_cache_behind() {
         let dir = tempfile::tempdir().expect("temp dir");
@@ -5659,6 +5679,13 @@ mod tests {
             .scp_driver()
             .record_externalized(5050, Default::default(), None);
         app.herder.scp_driver().publish_externalized(5050);
+
+        // peer_gap=5 (above PEER_AHEAD_ESCALATION_THRESHOLD=3, below
+        // HARD_RESET_GAP_ESCALATION=12) — the at-tip suppression window
+        // from #2789. Without this, peer_gap defaults to 0 and the test
+        // does not exercise the meaningful at-tip path.
+        app.max_verified_scp_slot
+            .store(u64::from(current_ledger) + 5, Ordering::Relaxed);
 
         // Arm archive recovery status and seed fresh cache at current_ledger.
         {
--- a/crates/app/src/app/catchup_impl.rs
+++ b/crates/app/src/app/catchup_impl.rs
@@ -5905,4 +5905,123 @@ mod tests {
             "with cooldown active + archive_behind, should fall back to AttemptRecovery"
         );
     }
+
+    // ================================================================
+    // Tests for #2789: at-tip suppression must not block recovery when
+    // peers are clearly ahead. The #2713 suppression is correct only when
+    // the validator is at-or-near the network tip; with peer_gap >= 12 the
+    // node must attempt catchup, not sit on a suppressed hard reset.
+    // ================================================================
+
+    /// When archive is ConfirmedBehind with fresh-cache-behind AND peers are
+    /// verified far ahead (peer_gap >= HARD_RESET_GAP_ESCALATION), the
+    /// suppression must NOT fire — the node has clear peer evidence that
+    /// the network is past us and must escalate. This is the #2789 wedge.
+    #[tokio::test]
+    async fn test_hard_reset_not_suppressed_when_peers_ahead() {
+        let dir = tempfile::tempdir().expect("temp dir");
+        let db_path = dir.path().join("rs-stellar-test.db");
+        let config = crate::config::ConfigBuilder::new()
+            .database_path(db_path)
+            .build();
+        let mut app = App::new(config).await.unwrap();
+
+        let current_ledger: u32 = 5_000;
+
+        // latest_externalized far ahead (network advanced 50 ledgers).
+        app.herder
+            .scp_driver()
+            .record_externalized(5050, Default::default(), None);
+        app.herder.scp_driver().publish_externalized(5050);
+
+        // Verified peer evidence: max_verified_scp_slot = 5050 → peer_gap = 50.
+        app.max_verified_scp_slot.store(5050, Ordering::Relaxed);
+
+        // Archive is ConfirmedBehind, cache fresh at/below current_ledger
+        // (#2713 suppression condition).
+        {
+            let mut guard = app.archive_recovery_status.write().await;
+            *guard = ArchiveRecoveryStatus::ConfirmedBehind {
+                backoff_until: None,
+            };
+        }
+        app.archive_checkpoint_cache.seed(current_ledger);
+
+        // Backdate so cooldown doesn't gate entry.
+        app.start_instant = std::time::Instant::now() - std::time::Duration::from_secs(500);
+
+        let counter_before = app.post_catchup_hard_reset_total.load(Ordering::Relaxed);
+
+        let _ = app
+            .force_post_catchup_hard_reset(
+                current_ledger,
+                HardResetReason::ArchiveBehindStallWallClock,
+            )
+            .await;
+
+        // Hard-reset counter MUST advance — suppression must not fire when
+        // peer_gap >= HARD_RESET_GAP_ESCALATION. On origin/main today this
+        // FAILS because the suppression at the start of
+        // force_post_catchup_hard_reset fires regardless of peer_gap.
+        assert!(
+            app.post_catchup_hard_reset_total.load(Ordering::Relaxed) > counter_before,
+            "hard reset must proceed when peer_gap >= {} \
+             (peer_gap=50, expected counter > {}, got {}). #2789 wedge: \
+             suppression blocks the only state-changing recovery path.",
+            HARD_RESET_GAP_ESCALATION,
+            counter_before,
+            app.post_catchup_hard_reset_total.load(Ordering::Relaxed),
+        );
+    }
+
+    /// Drive the recovery loop three times with the #2789 fixture (peer_gap
+    /// large, archive fresh-but-behind). The first call must break the
+    /// suppression and proceed to the hard reset; subsequent calls hit the
+    /// legitimate hard-reset cooldown (which is correct). The wedge is
+    /// broken, not perpetuated by repeated suppression.
+    #[tokio::test]
+    async fn test_recovery_loop_unwedges_when_peers_ahead() {
+        let dir = tempfile::tempdir().expect("temp dir");
+        let db_path = dir.path().join("rs-stellar-test.db");
+        let config = crate::config::ConfigBuilder::new()
+            .database_path(db_path)
+            .build();
+        let mut app = App::new(config).await.unwrap();
+
+        let current_ledger: u32 = 5_000;
+
+        app.herder
+            .scp_driver()
+            .record_externalized(5050, Default::default(), None);
+        app.herder.scp_driver().publish_externalized(5050);
+        app.max_verified_scp_slot.store(5050, Ordering::Relaxed);
+
+        {
+            let mut guard = app.archive_recovery_status.write().await;
+            *guard = ArchiveRecoveryStatus::ConfirmedBehind {
+                backoff_until: None,
+            };
+        }
+        app.archive_checkpoint_cache.seed(current_ledger);
+        app.start_instant = std::time::Instant::now() - std::time::Duration::from_secs(500);
+
+        for _ in 0..3 {
+            let _ = app
+                .force_post_catchup_hard_reset(
+                    current_ledger,
+                    HardResetReason::ArchiveBehindStallWallClock,
+                )
+                .await;
+        }
+
+        // After three rapid calls: counter has advanced at least once
+        // (suppression broken). On origin/main, counter stays at 0 across
+        // all three calls because suppression fires every time.
+        assert!(
+            app.post_catchup_hard_reset_total.load(Ordering::Relaxed) >= 1,
+            "loop must unwedge after first call: counter expected >= 1, got {}. \
+             #2789: suppression on every tick prevents any state-changing recovery.",
+            app.post_catchup_hard_reset_total.load(Ordering::Relaxed),
+        );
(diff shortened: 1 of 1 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

<details>

### Checklist

**1. Issue triage `kind: bug-fix` — PASS**

The `## Triage Report` comment on #2789 explicitly sets `**Kind:** bug-fix`.

---

**2. Regression tests present — PASS**

Both named tests exist at commit `dca276bc`:
- `test_hard_reset_not_suppressed_when_peers_ahead` (line 5921)
- `test_recovery_loop_unwedges_when_peers_ahead` (line 5983)

The existing `test_hard_reset_suppressed_fresh_cache_behind` is updated in `a907cc41` to set `max_verified_scp_slot = current_ledger + 5` (peer_gap=5 < 12), keeping it in the suppression window.

---

**3. Regression tests would have caught the bug — PASS**

On `origin/main` the suppression gate at line 1369 reads:

```rust
let archive_cannot_provide_target =
    self.archive_recovery_snapshot().await.is_confirmed_behind()
        && matches!(
            self.get_cached_archive_checkpoint_nonblocking(),
            CacheResult::Fresh(v) if v <= current_ledger
        );
```

No `peer_gap` check. The test fixture sets `max_verified_scp_slot = 5050`, `current_ledger = 5000`, `archive = ConfirmedBehind`, `cache = Fresh(5000)` → all three original conditions satisfied → suppression fires → `return None` before counter increment → `assert!(counter > counter_before)` **fails** with "got 0". Correct — tests would fail on main, pass after the fix.

---

**4. Logic and semantics — PASS with one observation**

The new gate:

```rust
let peer_gap = self.effective_peer_gap(current_ledger);
let suppress_archive_behind_at_tip =
    self.archive_recovery_snapshot().await.is_confirmed_behind()
        && peer_gap < HARD_RESET_GAP_ESCALATION
        && matches!(
            self.get_cached_archive_checkpoint_nonblocking(),
            CacheResult::Fresh(v) if v <= current_ledger
        );
```

**Semantics of `effective_peer_gap`:** reads `max_verified_scp_slot.load(Ordering::Relaxed).saturating_sub(current_ledger)`. Returns 0 when no peer traffic — suppression is *conservative* in that direction (zero peer gap ⇒ suppresses, which is fine: no peer evidence of progress means the archive genuinely may not be ahead either).

**Relaxed ordering:** The `max_verified_scp_slot` atomic is a monotonic hint updated from SCP envelope processing. Relaxed is acceptable — a slightly stale read either: (a) reads 0 when peers are actually ahead → suppresses, which was the pre-fix behaviour and is safe at startup; or (b) reads an older peer slot → potentially misses the 12-slot threshold momentarily, but the next tick corrects it. The field is already used with Relaxed throughout the codebase and the comment at `mod.rs:2086` explains the rationale. No ordering concern here.

**Placement relative to cooldown:** The new gate is *after* `is_hard_reset_on_cooldown` (line 1321). This is correct: if the hard reset is already on cooldown we return early before ever reaching the suppression check, so the interaction is clean.

**Placement relative to livelock:** Still positioned before the livelock circuit-breaker (line 1417), so spurious suppression calls don't arm the fatal-shutdown timer. Unchanged from #2713 and still correct.

**Rename:** `archive_cannot_provide_target` → `suppress_archive_behind_at_tip` accurately reflects the narrowed semantics (it now encodes both the archive state *and* the peer-gap condition). The new name is clearer.

**Threshold choice:** `HARD_RESET_GAP_ESCALATION = TX_SET_REQUEST_WINDOW = 12`. This is the same threshold already used for the gap-growth cooldown override at `is_hard_reset_on_cooldown`. Reusing it is consistent; the comment explains the rationale (peer cache window).

---

**5. Spurious un-suppression via single far-ahead envelope — PASS**

Caller at `consensus.rs:1443-1447` already gates on `peer_gap >= PEER_AHEAD_ESCALATION_THRESHOLD = 3` before invoking `force_post_catchup_hard_reset`. So a single stray envelope (peer_gap=1 or 2) never reaches this function. When peer_gap is 12+, at least 12 consecutive slots have been verified by the SCP pipeline — not a single outlier.

---

**6. Boundary at peer_gap = 12 exactly — PASS**

`peer_gap < HARD_RESET_GAP_ESCALATION` means suppression fires at peer_gap ∈ {0..11} and falls through at peer_gap ≥ 12. The boundary is correct: 12 is the first value that indicates the node is more than one TX_SET_REQUEST_WINDOW behind, which is the definition of "clearly ahead" used elsewhere in the codebase.

---

**7. Test quality — PASS**

Both new tests: (a) construct a realistic #2789 fixture, (b) assert on the *hard-reset counter* (not just return value), (c) set peer_gap=50 which is well above the threshold and matches the incident data (gap=56 observed). The loop test (`test_recovery_loop_unwedges_when_peers_ahead`) correctly verifies that the wedge breaks after the *first* call; subsequent calls legitimately hit the 60s cooldown (note: `start_instant` is backdated 500s so the cooldown *is* expired at the start — the second and third calls will hit the cooldown that was armed by the first successful reset, which is the correct behavior: one unwedge then bounded retry).

**Minor observation (non-blocking):** `test_recovery_loop_unwedges_when_peers_ahead` asserts `counter >= 1` after 3 calls. Given that the 60s cooldown is armed after the first reset, the second and third calls return `None` via the cooldown path, so `counter == 1` precisely after 3 calls. The `>= 1` bound is correct but could be tightened to `== 1` to verify the cooldown-after-reset invariant. Not a blocker.

---

**8. Comment quality and tracing — PASS**

The block comment explains #2789 rationale clearly. The `peer_gap` field is added to both the warn and debug trace events, improving observability. The updated trace message accurately reflects the dual condition ("AND peers within HARD_RESET_GAP_ESCALATION").

---

**9. Stellar-core parity — PASS**

PR body correctly notes that `LedgerApplyManagerImpl::processLedger` unconditionally fires `startOnlineCatchup` regardless of archive state; archive readiness is checked inside `CatchupWork`. This change moves henyey closer to upstream semantics by removing a henyey-specific over-suppression when peers are ahead.

---

### Summary

The fix is minimal and correct: a single boolean `&& peer_gap < HARD_RESET_GAP_ESCALATION` added to an existing gate condition, with well-documented rationale, proper constant reuse, and two focused regression tests that would have caught the bug. All invariants preserved.

**Verdict:** APPROVE

</details>

tomerweller:
## 🔍 Reviewer: Risk

<details>

**PR:** #2790 — Narrow #2713 hard-reset suppression to at-or-near-tip
**Issue:** #2789 — mid-run recovery wedge, 4 occurrences in 9h, 5th crashed the validator

---

### Summary

The change narrows the `force_post_catchup_hard_reset` suppression (introduced by #2713) so it only fires when `peer_gap < HARD_RESET_GAP_ESCALATION (12)`. When peers are verified ≥12 ledgers ahead, the suppression is bypassed and the existing hard-reset + `spawn_catchup(ProbeAhead)` path runs. The fix is correct and bounded. Risk analysis follows.

---

### Risk 1 — Hard-reset frequency post-deploy (ACCEPTABLE)

**Current behavior (52ac1665):** 4 wedge episodes in 9h (~2.2h mean), each accumulating 220–362 suppressed "hard reset" ticks before an external event (state transition to CatchingUp) breaks the loop. `post_catchup_hard_reset_total` incremented once per episode (on the CatchingUp transition), so the ALARM-tier metric incremented 4× in 9h.

**Post-fix behavior:** On the first armed tick of each wedge episode, `peer_gap ≥ 12` bypasses suppression and runs the real hard reset. The hard reset arms the 60s minimum cooldown (`last_hard_reset_offset`). Any subsequent tick within 60s hits the cooldown check and routes to `AttemptRecovery`. So each ~90s wedge episode now produces **exactly one** hard reset — matching or slightly below the current rate of 4 hard resets in 9h. The ALARM-tier counter rate is unchanged. No alarm flood.

**Caveat:** If `spawn_catchup(ProbeAhead)` fails quickly (archive genuinely behind, blocking probe returns fast), the cooldown enforces a ≥60s gap before the next hard reset. At most one hard reset per 60s per unique stuck-ledger. The livelock timer (640s at default checkpoint_frequency=64) bounds the worst-case run to ~10 hard resets before `trigger_fatal_shutdown`.

**Verdict:** Alarm rate acceptable. Net improvement: replaces 220–362 suppression ticks per episode with 1 hard reset per episode.

---

### Risk 2 — Stale or malicious peer evidence inflating peer_gap (ACCEPTABLE)

`max_verified_scp_slot` is populated via verified SCP envelopes (Relaxed atomic, monotonic-up). The outer `maybe_start_buffered_catchup` path already requires `PEER_AHEAD_ESCALATION_THRESHOLD = 3` consecutive peer-confirmed slots before routing to the hard-reset escalation. By the time `force_post_catchup_hard_reset` is called with a genuine #2789 wedge, `max_verified_scp_slot` reflects real network consensus (verified envelopes, not just claimed slots).

The new threshold `HARD_RESET_GAP_ESCALATION = 12 = TX_SET_REQUEST_WINDOW` is deliberately aligned with the tx_set peer-cache window. A single stale envelope 12+ ledgers ahead would need to persist across the PEER_AHEAD_ESCALATION_THRESHOLD gate in the caller — that gate is `≥ 3`, not single-envelope. The combination of the outer 3-slot verification gate and the inner 12-slot gap threshold provides sufficient protection against spurious un-suppression.

One subtle point: `max_verified_scp_slot` is monotonic (never decreases). After a successful catchup, `reset_recovery_attempts(Full)` zeroes it (`max_verified_scp_slot.store(0)`). This correctly resets peer-gap evidence post-recovery. No stale-high reading across recovery boundaries.

**Verdict:** Stale-evidence risk is bounded by the existing outer gate and the monotonic-zero-on-reset behavior.

---

### Risk 3 — ProbeAhead failure scenarios (BOUNDED)

When the new code falls through to `spawn_catchup(ProbeAhead(current_ledger))`, two outcomes are possible:

1. **Archive has published a new checkpoint** (most common in the #2789 scenario): ProbeAhead's blocking HTTP probe finds it, returns a valid target, catchup proceeds. Wedge broken. This is the happy path.

2. **Archive is genuinely behind** (mid-checkpoint interval): ProbeAhead's blocking HTTP probe confirms "not ahead of min_ledger", catchup task errors out gracefully and calls `complete_catchup(Err(...))`. The state machine returns to `Validating`. Recovery re-enters on the next tick.

In case 2, the hard reset still does useful work: it clears `syncing_ledgers` no-tx_set entries, clears pending tx_sets, resets `recovery_attempts`, and transitions state to `CatchingUp` then back to `Validating`. This is not free (5–10s of catchup task overhead + blocking archive HTTP probe), but it is bounded by the 60s cooldown before the next hard reset can fire.

**One gap worth noting:** After a ProbeAhead failure, `archive_recovery_status` was cleared to `NotConfirmed` by `clear_archive_recovery_state(HardResetExec)` (step 1 of the hard reset). On the next recovery tick, `is_confirmed_behind()` will be false, so `suppress_archive_behind_at_tip` will be false regardless of peer_gap — the suppression won't even need the new peer_gap gate. The archive-behind flag will be re-armed when `check_archive_status` runs again (~5–10s later). During that brief window, the decision function takes the `!archive_behind` branch and may route to `AttemptRecovery` or `TriggerCatchup` instead of `HardReset`. This is correct behavior and reduces spurious hard resets.

**Verdict:** ProbeAhead failure path is clean. The state machine re-engages properly after a failed catchup.

---

### Risk 4 — Livelock circuit-breaker interaction (ACCEPTABLE WITH NOTE)

The livelock timer (`hard_reset_livelock_start`, `hard_reset_livelock_ledger`) arms on the first hard reset at a given `current_ledger` and triggers `fatal_shutdown` if `elapsed >= hard_reset_fatal_timeout_secs()`.

At default checkpoint_frequency=64: `hard_reset_fatal_timeout_secs() = (64 * 5 * 2) = 640s`. With `HARD_RESET_MIN_COOLDOWN_SECS = 60`, the max hard-reset rate is 1/60s → at most `640 / 60 ≈ 10` hard resets before the livelock fires. Crucially, each ProbeAhead spawns a `CatchingUp` state transition which then returns to `Validating` — after each successful catchup (even a trivial one), `clear_archive_recovery_state(FullProgress)` zeroes `hard_reset_livelock_start` (line 2167 in mod.rs). So the livelock timer only accumulates across true ProbeAhead *failures* at the same ledger.

In the #2789 scenario, ProbeAhead succeeds on the first or second try (the archive publishes the checkpoint during the ~5s blocking probe). The livelock timer resets. No fatal_shutdown risk under normal #2789 conditions.

**Edge case:** If the archive is persistently behind (e.g., archive publishing is broken), the livelock correctly fires after ~640s and triggers a wipe. This is intended behavior, not a regression.

**One minor concern:** The livelock timer tracks `current_ledger` (i.e., LCL). If each ProbeAhead catchup succeeds in advancing LCL by at least one ledger, the `livelock_ledger != current_ledger` branch fires and the timer resets. But if catchup always targets the same checkpoint (archive publishes checkpoint N, catchup completes, ledger advances, archive is still at N, next hard reset targets the same checkpoint N) — the livelock ledger check would see a *new* ledger and reset. This is fine.

**Verdict:** No fatal-timer escalation under the #2789 scenario. The existing livelock design handles the post-fix frequency correctly.

---

### Risk 5 — Operational and monitoring implications

**Metric impact:** The new code path (`suppress_archive_behind_at_tip = false` when `peer_gap ≥ 12`) still increments `post_catchup_hard_reset_total` via the real hard reset path. The `hard_reset_suppressed_archive_behind` RECOVERY_STALLED_TICK_TOTAL label will stop firing for the #2789 scenario (correct — the suppression no longer applies). Monitor scripts should not need updates since the alarm-tier metric (`henyey_post_catchup_hard_reset_total`) rate is not increasing.

**New log field:** `peer_gap` is now emitted in both the suppressed-log and repeated-log branches. This is purely additive for observability — helps operators distinguish #2713 (at-tip, small peer_gap) from #2789 (behind-tip, large peer_gap) in log aggregation. Good.

**Config changes:** None required. No new config knobs. No monitoring threshold adjustments needed.

**Gradual rollout:** Not necessary. The fix narrows existing behavior that was already causing validator crashes. Immediate deploy on the next CI-green build is appropriate.

**No thundering-herd risk:** The `force_post_catchup_hard_reset` path serializes through the consensus lock and fires at most once per 60s per stuck-ledger. No concurrent hard-reset storms are possible.

---

### Minor inline concern

The `suppress_archive_behind_at_tip` check reads `effective_peer_gap(current_ledger)` which loads `max_verified_scp_slot` with `Ordering::Relaxed`. The comment at the call site (line 1369 in the diff) already notes this — Relaxed is correct here because the value is a monotonic hint from verified envelopes, not a correctness-critical gate. The worst case (reading a slightly stale slot) only makes the suppression fire one extra time before the next SCP message updates it. Not a concern.

---

### Assessment

All five risk vectors are bounded:
- Hard-reset rate post-deploy: ≤ current rate (1 real reset per wedge episode vs. 1 per episode now)
- Stale peer evidence: protected by outer PEER_AHEAD_ESCALATION gate + monotonic zero-on-reset
- ProbeAhead failure: state machine re-engages cleanly, cooldown bounds retries
- Livelock circuit-breaker: resets on any ledger advance; no fatal-timer risk under #2789 scenario
- Operational: additive logging only, no config or monitor changes needed

The two regression tests directly cover the #2789 failure mode and the at-tip (#2713) preservation. Test coverage is sufficient for the scope of the change.

**Verdict:** APPROVE

</details>

Issue #2720, opened 2026-05-15
URGENT: INV-H2 panic LCL > tracking by 1 — new variant on fast-tracking catchup path (c14afa94 crashed at L62579903)


tomerweller:
## Symptom

c14afa94 panicked with a NEW INV-H2 variant: `LCL > tracking_slot by 1`.

```
thread 'tokio-rt-worker' (637590) panicked at crates/herder/src/herder.rs:894:9:
INV-H2 FATAL: LCL (62579903) is at or ahead of tracking consensus slot (62579902).
Unrecoverable state divergence between consensus and ledger-manager.
```

This is a regression beyond the `LCL == tracking` case fixed by `1f82d1b7` + `c14afa94`. Even the relaxed `<=` bound (8b2c578c) would have panicked here because `LCL > tracking`.

## Reproduction

Validator on `c14afa94` (PID 636474), uptime 12m post-deploy. Recovery loop wedged at LCL=62579839 for 6 min (see #2713), then fast-tracked catchup:

```
13:42:56.498 Ledger closed ledger_seq=62579893
13:42:56.658 Ledger closed ledger_seq=62579894
… (rapid close cycle)
13:42:58.155 Ledger closed ledger_seq=62579903
13:42:58.843 Pending EXTERNALIZE far ahead — fast-tracking catchup slot=62579908 current_ledger=62579903 gap=5
[PANIC] LCL=62579903 > tracking=62579902
```

The fast-tracking catchup path closes ledger 62579903 (advances LCL to 62579903), but the next call to `assert_lcl_consistency()` finds `tracking_slot=62579902` (one step behind LCL). The `advance_tracking_to()` call from `1f82d1b7` either:

1. Did NOT run for this ledger close (a different code path closed it), or
2. Ran but set `tracking_slot=62579903` while the closing slot was 62579902 (race), or
3. The `is_tracking()` predicate flipped to true while tracking_slot was still at the old value

## Why deploy-blocking

- Validator was OFFLINE — process panicked and died at 13:42:58Z
- Manually relaunched at 13:49:34Z on the same c14afa94 binary (no better option available)
- Quarantined `c14afa94` to block automatic redeploy
- The bug surfaces during legitimate post-catchup rapid-close — a known recovery path that must work

## Suspected root cause

[...]

PR #2792, opened 2026-05-18
Convert INV-H2 panic to corrective branch in assert_lcl_consistency (#2791)

### LINKED ISSUE #2791: INV-H2 panic at herder.rs:940 during catchup replay (separate from #2789)
## Symptom

Validator panics with `INV-H2 FATAL: LCL ... is at or ahead of tracking consensus slot` at `crates/herder/src/herder.rs:940` during **catchup replay**, regardless of whether the #2789 wedge was the precipitating cause.

This is a **distinct issue from #2789**, confirmed by the fact that 0df88603 (the merged #2789 fix) hits the same crash.

## Reproduction

Reliable on 0df88603 (PR #2790 merge), persisted LCL=62616065 on disk, current mainnet tip ~62616665:

1. Cold-start the validator
2. Replay catchup loads ledgers in burst: at 00:18:58 closes L62616663, 62616664, 62616665 within ~175ms (apply rate driven by `Pending EXTERNALIZE far ahead — fast-tracking catchup`)
3. After L62616665 closes, INV-H2 fires:
   ```
   INV-H2 FATAL: LCL (62616665) is at or ahead of tracking consensus slot (62616660).
   ```
4. Validator process exits

## Crash history (all on this same line)

| Time UTC | Build | LCL | tracking | gap | Context |
|--|--|--|--|--|--|
| 2026-05-17 23:22 | 52ac1665 | 62616127 | 62616126 | +1 | #2789 wedge → burst close → crash |
| 2026-05-17 23:25 | 52ac1665 | 62616073 | 62616070 | +3 | catchup replay after restart → crash |
| 2026-05-18 00:18 | 0df88603 | 62616665 | 62616660 | +5 | catchup replay on #2789-fixed build → crash |

LCL is always 1-5 ahead of tracking when the assertion fires. The ledger-manager closes ledgers faster than the herder updates `tracking_consensus_slot` during a fast-tracking burst.

## Possible code locations

- `crates/herder/src/herder.rs:940` — the failing assertion. Need to inspect why this requires LCL ≤ tracking when the ledger-manager has already legitimately closed beyond tracking during catchup-replay.
- Prior fix `5a90e509` ("Fix INV-H2 panic during fast-tracking catchup burst close") targeted line 894 — same family of panic. This suggests either an incomplete cover of all `tracking < LCL` paths, or a regression introduced after that fix.

## Suggested fix paths

1. **Tighten the assertion gate**: the assertion likely guards `tracking < LCL` only during *active consensus* (Validating state). During `Catching Up` state or while `lifecycle::fast_tracking_catchup` is active, LCL legitimately advances ahead of tracking — the assertion should be skipped or replaced with a `tracking := max(tracking, LCL)` update.
2. **Update tracking_slot atomically with LCL** in `apply_ledger_close` so the invariant holds before any code path can re-read tracking.
3. **Inspect the diff between `5a90e509` and the line-940 assertion** to see whether it's a copy-paste duplicate that was missed by the original fix.

## Severity

Validator is **down** and cannot be relaunched on this build (deterministic crash during catchup replay). Both 52ac1665 and 0df88603 are now in `/home/tomer/data/deploy_quarantine.txt`. Monitor cannot recover autonomously.

Operator action required. Suggest applying the line-940 fix and shipping; the monitor will deploy whatever lands on `origin/main` once a non-quarantined commit is available.

tomerweller:
Closes #2791

## Summary

The mainnet validator is currently quarantined (`52ac1665` and `0df88603`) because `crates/herder/src/herder.rs::assert_lcl_consistency` aborts the process with `INV-H2 FATAL` when LCL legitimately advances ahead of tracking by 1–5 ledgers during catchup-replay burst close. This is a sibling of #2720 (LCL == tracking, fixed by `5a90e509` via better placement of `advance_tracking_to`); #2791 covers all gaps including the multi-slot ones that the placement fix doesn't reach.

This PR converts the assertion to a self-corrective branch:

1. Log a WARN with `lcl_seq` / `tracking` / `gap` / `close_time` / `herder_state`.
2. Increment a new `lcl_ahead_of_tracking_corrective_total` counter on the Herder so operators can monitor the rate of corrective fires.
3. Call `advance_tracking_to(lcl_seq, close_time)` (monotone-idempotent) to restore the invariant.
4. Return normally — no panic.

The validator can now boot from a stale persisted LCL without crashing during the catchup-replay burst close that triggered the wedge.

## Parity

Confirmed by Critic B against `stellar-core/` v26.0.1:
- `HerderImpl::trackingConsensusLedgerIndex()` (`HerderImpl.cpp:170-178`) **throws** (not aborts) on `lcl > mTrackingSCP.mConsensusIndex`. The henyey `panic = "abort"` was strictly more fatal than upstream.
- `HerderImpl::lastClosedLedgerIncreased()` (`HerderImpl.cpp:1227-1248`) gates the strict equality assertion on `latest == true` — non-final closes in a burst are not checked.
- The new metric `lcl_ahead_of_tracking_corrective_total` has no upstream counterpart; observability metrics are allowed to diverge from upstream per standard policy.

Net: the change moves henyey strictly closer to upstream behavior.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/2791#issuecomment-4473178891) — converged Round 1 (A: REVISE-MINOR addressed re. 3 additional `#[should_panic]` tests + metric naming, B: APPROVE, C: APPROVE).

## Test plan

- [x] `cargo fmt --check` clean
- [x] `cargo clippy --all -- -D warnings` clean
- [x] `cargo test -p henyey-herder` — 1115 passed; 0 failed
- [x] `cargo test -p henyey-app` — 940 passed; 0 failed (the `ledger_closed` callsite that calls `assert_lcl_consistency` is exercised by the rapid-close tests in this crate)
- [x] `cargo test -p henyey-herder --lib assert_lcl_consistency` — 13 passed (all existing INV-H2 coverage + 1 new gap=5 regression + 2 new metric tests)

## Regression test

- **Tests:**
  - `crates/herder/src/herder.rs::inv_h2_lcl_guard_tests::test_assert_lcl_consistency_recovers_when_lcl_ahead_by_5` — replicates the #2791 production fixture (gap=5).
  - Five previously-`#[should_panic]` tests renamed `_recovers_` and updated to assert that `assert_lcl_consistency` returns normally AND `tracking_slot` advances to `lcl + 1`: `_when_lcl_equals_tracking_slot`, `_when_lcl_ahead_by_one_2720` (was the #2720 regression), `_when_lcl_ahead_of_tracking`, `_when_lcl_exceeds_tracking`, and `test_ledger_closed_recovers_without_advance_tracking_to` (drives the full `ledger_closed` path).
- **Pre-fix:** committed as `0b134d70` — verified FAILED on `origin/main` with `INV-H2 FATAL: LCL (X) is at or ahead of tracking consensus slot (Y)` for each.
- **Post-fix:** verified PASS after `d35dc084` (this commit).

Two new metric-fire tests (`_increments_metric_on_corrective_fire`, `_no_metric_increment_on_passing_case`) exercise the corrective-fire counter contract. These reference a method introduced in the fix commit and so are added alongside the implementation rather than in the test-only commit.

## Deviations from plan

None. Critic A's REVISE-MINOR concerns (3 additional `#[should_panic]` test sites + metric naming with explicit `_corrective_` suffix) were addressed in the converged plan and implemented as specified.

### DIFF
--- a/crates/herder/src/herder.rs
+++ b/crates/herder/src/herder.rs
@@ -496,6 +496,12 @@ pub struct Herder {
     >,
     /// Shared SCP metrics counters.
     scp_metrics: Arc<crate::metrics::ScpMetrics>,
+    /// Number of times `assert_lcl_consistency` took the corrective branch
+    /// because LCL had legitimately advanced ahead of tracking (e.g. during
+    /// catchup-replay burst close). Each fire restores the invariant via an
+    /// idempotent `advance_tracking_to(lcl_seq, ...)` call rather than
+    /// crashing the process. See [`Self::assert_lcl_consistency`] and #2791.
+    lcl_ahead_of_tracking_corrective_total: AtomicU64,
     /// SCP state persistence manager, installed after construction via
     /// [`Self::set_scp_persistence`]. Used by [`Self::purge_persisted_tx_sets`]
     /// to GC unreferenced tx sets on the app event loop's 60s timer.
@@ -706,6 +712,7 @@ impl Herder {
             scp_verifier_handle,
             verified_rx: std::sync::Mutex::new(verified_rx),
             scp_metrics,
+            lcl_ahead_of_tracking_corrective_total: AtomicU64::new(0),
             scp_persistence: std::sync::OnceLock::new(),
         }
     }
@@ -908,14 +915,38 @@ impl Herder {
         LastExternalizedLedger::new(self.tracking_slot().get().saturating_sub(1))
     }
 
-    /// INV-H2: Assert that the ledger manager's LCL is not ahead of consensus tracking.
+    /// INV-H2: Verify that the ledger manager's LCL is not ahead of consensus tracking;
+    /// self-correct if it is.
     ///
-    /// Panics (process-fatal in release via `panic = "abort"`) if violated.
-    /// Mirrors stellar-core's `trackingConsensusLedgerIndex()` (HerderImpl.cpp:170-178)
-    /// which throws on `lcl > mTrackingSCP.mConsensusIndex`.
+    /// Parity: mirrors stellar-core's `trackingConsensusLedgerIndex()`
+    /// (HerderImpl.cpp:170-178) which **throws** (not aborts) on
+    /// `lcl > mTrackingSCP.mConsensusIndex`. The throw propagates and can be
+    /// caught by the catchup state machine. The henyey port had ported this
+    /// as `panic = "abort"` which is strictly more fatal than upstream — any
+    /// race window that briefly leaves LCL ahead of tracking would kill the
+    /// validator (#2720, #2791). The corrective branch restores the invariant
+    /// inline, which is the closest semantic match to upstream's recoverable
+    /// throw available in a Rust caller that doesn't unwind.
     ///
-    /// No-op when not in Tracking state or at genesis (tracking_slot == 0), since
-    /// during Booting/Syncing/catchup the LM legitimately advances before tracking.
+    /// No-op when not in Tracking state or at genesis (tracking_slot == 0),
+    /// since during Booting/Syncing the LM legitimately advances before
+    /// tracking (handled by the early-return on `!is_tracking`).
+    ///
+    /// When LCL >= tracking (invariant violation):
+    /// 1. Log a WARN with `lcl_seq` / `tracking` / `gap` for operator visibility.
+    /// 2. Increment `lcl_ahead_of_tracking_corrective_total` so the rate of
+    ///    corrective fires can be monitored (alerts can be added later).
+    /// 3. Call `advance_tracking_to(lcl_seq, close_time)` to bring tracking
+    ///    to `lcl_seq + 1`. Monotone-idempotent — safe even if a parallel
+    ///    path has already advanced tracking.
+    /// 4. Return normally.
+    ///
+    /// The `close_time` for the corrective call is read from the LedgerManager's
+    /// current header, the same way `bootstrap` reads it (h:1341). A concurrent
+    /// close landing between the lcl_seq read and the close_time read may
+    /// supply a close_time for `lcl_seq + 1` — still a valid tracking-close-time
+    /// (tracking would then point one slot past the LCL the corrective branch
+    /// observed, which is still consistent with the invariant).
     fn assert_lcl_consistency(&self) {
         if !self.is_tracking() {
             return;
@@ -925,24 +956,35 @@ impl Herder {
             return;
         }
         let lcl_seq = self.ledger_manager.current_ledger_seq() as u64;
-        // Strict bound: lcl_seq < tracking_slot must hold on all paths.
-        //
-        // In the normal SCP-driven path, complete_externalization advances
-        // tracking_slot to N+1 before LCL can reach N. On the non-SCP
-        // (catchup rapid-close) path, advance_tracking_to is called in
-        // handle_close_complete_inner immediately after the close task joins
-        // (before any .await), ensuring tracking is visible before any
-        // concurrent task can observe the new LCL. See #2720.
-        //
-        // Parity: stellar-core HerderImpl.cpp:1227-1248 asserts
-        // mTrackingSCP->mConsensusIndex > lcl in lastClosedLedgerIncreased.
-        // This is equivalent to lcl < tracking in henyey's representation.
-        assert!(
-            lcl_seq < tracking,
-            "INV-H2 FATAL: LCL ({lcl_seq}) is at or ahead of tracking consensus slot \
-             ({tracking}). Unrecoverable state divergence between consensus \
-             and ledger-manager."
+        if lcl_seq < tracking {
+            return;
+        }
+
+        // Invariant violated: LCL has legitimately advanced ahead of tracking
+        // (e.g. burst close from buffered EXTERNALIZE messages during catchup
+        // replay — see #2791 timeline). Restore the invariant inline.
+        let gap = lcl_seq.saturating_sub(tracking).saturating_add(1);
+        let close_time = self.ledger_manager.current_header().scp_value.close_time.0;
+        self.lcl_ahead_of_tracking_corrective_total
+            .fetch_add(1, Ordering::Relaxed);
+        tracing::warn!(
+            lcl_seq,
+            tracking,
+            gap,
+            close_time,
+            herder_state = %self.state(),
+            "INV-H2: LCL at or ahead of tracking — advancing tracking to lcl + 1 \
+             (corrective; was process-fatal pre-#2791)"
         );
+        self.advance_tracking_to(lcl_seq, close_time);
+    }
+
+    /// Cumulative count of times the `assert_lcl_consistency` corrective
+    /// branch fired (LCL >= tracking → advance tracking to lcl + 1). Used
+    /// by the metrics scrape path and by regression tests. See #2791.
+    pub fn lcl_ahead_of_tracking_corrective_total(&self) -> u64 {
+        self.lcl_ahead_of_tracking_corrective_total
+            .load(Ordering::Relaxed)
     }
 
     /// Get the tracking consensus close time.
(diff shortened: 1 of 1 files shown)


tomerweller:
## 🔍 Reviewer: Parity

Reviewed against stellar-core v26.0.1 (`stellar-core/` submodule). I read `HerderImpl.cpp` lines 149–202, 1198–1260, 1675–1681; `LedgerApplyManagerImpl.cpp`; `CatchupWork.cpp` lines 50–74, 525–554; `ApplyLedgerWork.cpp`; and the PR diff. Findings per the checklist:

---

### Claim 1 — `trackingConsensusLedgerIndex()` throws, not aborts

**Verified correct.** `HerderImpl.cpp:164–181` shows:
```cpp
if (lcl > mTrackingSCP.mConsensusIndex) {
    CLOG_ERROR(Herder, "{}", msg);
    throw std::runtime_error(msg);   // ← throw, not abort
}
```
The only `releaseAssert` calls in `trackingConsensusLedgerIndex()` guard the `BOOTING` state and the `uint32` overflow check — neither fires on the `lcl > tracking` condition. The henyey `panic = "abort"` was strictly more fatal than upstream.

---

### Claim 2 — `lastClosedLedgerIncreased` equality assert is gated on `latest == true`

**Verified correct.** `HerderImpl.cpp:1218–1233`:
```cpp
if (latest) {
    releaseAssert(trackingConsensusLedgerIndex() ==
                  mLedgerManager.getLastClosedLedgerNum());
    ...
}
```
Non-latest closes (burst closes during buffered catchup replay) do **not** execute this `releaseAssert`. The old henyey code checked on every `ledger_closed`, making it over-strict.

---

### Item 1 — No `releaseAssert` for `lcl > tracking` in `stellar-core/src/herder/`

**Verified.** Grepped `stellar-core/src/herder/` for `releaseAssert.*tracking`, `releaseAssert.*lcl`, `releaseAssert.*ledgerSeq`:

- `HerderImpl.cpp:1228`: `releaseAssert(trackingConsensusLedgerIndex() == getLastClosedLedgerNum())` — equality, gated on `latest` branch.
- `HerderImpl.cpp:1248`: same equality in `setupTriggerNextLedger`.
- `HerderSCPDriver.cpp:206`: unrelated slot-sequence guard.

No `releaseAssert` fires on `lcl > tracking`. The plan's claim is correct.

---

### Item 2 — Does stellar-core have an inline recovery for `lcl > tracking`?

**Verified — no inline recovery in `trackingConsensusLedgerIndex()`; it always throws.** There is no catch-handler for this specific throw in the herder subsystem (`HerderImpl.cpp` catch blocks are for XDR deserialization and tx-set GC). The throw would propagate to the async/io-context event loop, which in stellar-core logs and terminates gracefully, but there is no dedicated "re-sync tracking" recovery path analogous to what the PR introduces. The corrective `advance_tracking_to` inline is actually *more* recoverable than upstream's throw-then-terminate.

---

### Item 3 — Catchup-replay path and `lastClosedLedgerIncreased`

**Verified — the catchup-replay path does NOT call `lastClosedLedgerIncreased` for non-final closes.** `ApplyLedgerWork.cpp` calls `applyLedger(lcd, /* externalize */ false)`. `LedgerManagerImpl::ledgerCloseComplete` only calls `lastClosedLedgerIncreased` when `calledViaExternalize == true` (`LedgerManagerImpl.cpp:1385–1390`). Replay closes go through `ApplyLedgerWork` with `externalize=false`, so `lastClosedLedgerIncreased` is never invoked for them. Therefore, upstream's equality check simply never runs during catchup replay. Henyey's `assert_lcl_consistency` firing during catchup replay was itself a parity divergence — the fix corrects it.

---

### Item 4 — New metric `lcl_ahead_of_tracking_corrective_total`

No upstream counterpart. Grepped `HerderImpl.cpp` for "tracking ahead", "lcl ahead", `mTrackingSCP` + diagnostic patterns — zero hits for any equivalent counter or gauge. The PR metric is additive observability only; it cannot duplicate upstream semantics because no such counter exists there.

---

### Item 5 — `close_time` source for the corrective call

The corrective branch reads `close_time` from `self.ledger_manager.current_header().scp_value.close_time.0` (`herder.rs:1341`). Upstream `forceSCPStateIntoSyncWithLastClosedLedger` reads `header.scpValue.closeTime` where `header = mLedgerManager.getLastClosedLedgerHeader().header` — the same field. The sources are semantically identical. The PR doc-comment's note about a concurrent close landing between the `lcl_seq` read and the `close_time` read is accurate (the race is benign: `advance_tracking_to` is idempotent and the resulting tracking would still satisfy `tracking_consensus_ledger_index() >= lcl_seq`).

---

### Item 6 — `advance_tracking_to` as analog of `forceSCPStateIntoSyncWithLastClosedLedger`

**Verified accurate.** Upstream: `setTrackingSCPState(header.ledgerSeq, header.scpValue, true)` stores `mConsensusIndex = header.ledgerSeq` and `mConsensusCloseTime = scpValue.closeTime`, then sets `HERDER_TRACKING_NETWORK_STATE`. Henyey: `advance_tracking_to(slot, close_time)` stores `consensus_index = slot + 1` and `consensus_close_time = close_time`, then calls `ensure_tracking_state`. Because henyey's `tracking_consensus_ledger_index()` returns `consensus_index - 1`, both representations produce the same external value after the call. The analogy is faithful.

---

### One minor concern (non-blocking)

The `releaseAssert` at `HerderImpl.cpp:1228` and `:1248` asserts `trackingConsensusLedgerIndex() == LCL` (equality) in the `latest=true` path and in `setupTriggerNextLedger`. These are called **after** the LCL has already advanced, so at that point tracking must equal (not be behind) LCL. Henyey has no equivalent equality assertion after the corrective branch fires. This is acceptable — the corrective branch itself advances tracking to `lcl + 1` which puts tracking exactly one slot ahead of LCL (the normal invariant). No parity gap here, but a future regression test for post-corrective equality would be a quality improvement.

---

**Verdict:** APPROVE

The change moves henyey strictly closer to upstream behavior on both dimensions audited: the "fatal vs. recoverable" dimension (throw upstream → corrective branch in henyey instead of abort) and the "over-strict gating" dimension (equality check only on latest close upstream → equivalent skipping of the check for burst closes). No new observable divergences introduced.

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

<details>

### Test verification

**Step 1 — regression test fails on pre-fix code (commit `0b134d70`):**
```
thread '...::test_assert_lcl_consistency_recovers_when_lcl_ahead_by_5' panicked at crates/herder/src/herder.rs:940:9:
INV-H2 FATAL: LCL (5005) is at or ahead of tracking consensus slot (5000).
test result: FAILED. 0 passed; 1 failed
```
Confirmed: the test-only commit correctly fails with the expected panic text.

**Step 2 — all `assert_lcl_consistency` tests pass on fix commit (`d35dc084`):**
```
running 13 tests
... all 13 ok ...
test result: ok. 13 passed; 0 failed
```
Both verification steps pass. The regression test catches the bug.

---

### Correctness analysis

**Does the fix solve the problem?**
Yes. The `assert!` at `herder.rs:940` was replaced with a corrective branch that calls `advance_tracking_to(lcl_seq, close_time)`. This restores the invariant (`tracking > lcl`) in-place instead of aborting the process, unblocking the catchup-replay burst-close path.

**`advance_tracking_to` monotone-idempotent** — verified at `herder.rs:1307-1332`:
```rust
if next <= ts.consensus_index {
    true  // already advanced — return without updating
} else {
    ts.consensus_index = next;
    // ...
}
```
The guard `next <= ts.consensus_index` means tracking never regresses. The corrective call is race-safe: if a parallel path has already advanced tracking past `lcl_seq + 1`, the write is skipped.

**Can the corrective branch fire from a non-Tracking state?**
No. The function opens with `if !self.is_tracking() { return; }` — the corrective branch is unreachable when the Herder is in Booting/Syncing/Catching-up state. The `ensure_tracking_state` call inside `advance_tracking_to` is therefore always a no-op in this path (state is already Tracking). Correct.

**TOCTOU on `lcl_seq` / `close_time`:**
Two separate lock acquisitions — `current_ledger_seq()` then `current_header().scp_value.close_time.0` — with a window between them. A concurrent close landing between the reads would supply `close_time` for `lcl_seq + 1`. The doc-comment explicitly analyzes this (the corrective result is still invariant-consistent since tracking would point to `lcl_seq + 2`, still ahead of the observed LCL). This is the same pattern used by `bootstrap` at `h:1341` and is an accepted codebase inconsistency. Not a regression.

**Envelope cursor push during burst-close (gap=5):**
`advance_tracking_to(lcl_seq, ...)` pushes pending/fetching envelope cursors to `lcl_seq + 1` — envelopes buffered for intermediate skipped slots are discarded as `SlotTooOld`. In a gap-5 burst-close scenario those envelopes are already stale (those ledgers have closed), so the discard is correct. Identical to what normal in-lockstep `advance_tracking_to` calls would do.

**Callsite safety:**
- `ledger_closed` (line 2585): primary crash site. Corrective branch is appropriate.
- `trigger_next_ledger` (line 2279): guarded by `!is_tracking()` at line 2246-2248 before reaching `assert_lcl_consistency`, so the corrective branch here is only reachable in Tracking state. Safe.
- `handle_nomination_timeout` (line 2620): guarded by `!is_validator()` at line 2611-2613 which returns early, then `!is_tracking()` inside `assert_lcl_consistency`. Safe.

---

### Test coverage

13 `assert_lcl_consistency` tests pass, covering:
- gap=0 (LCL == tracking), gap=1, gap=2, gap=5 (production fixture), LCL behind tracking (passes), genesis sentinel, non-tracking state (skip)
- metric counter increments exactly 1 on corrective fire; does not increment on passing case
- `ledger_closed` end-to-end path (corrective fires inside, function completes)
- `trigger_next_ledger` normal path (LCL < tracking — no corrective)

All five previously-`#[should_panic]` tests were updated to assert recovery (including the three Critic A originally flagged as missing: `_fires_when_lcl_ahead_of_tracking`, `_fires_when_lcl_exceeds_tracking`, `test_ledger_closed_panics_without_advance_tracking_to`).

---

### Minor observations (non-blocking)

**`gap` log field semantics:** `gap = lcl_seq.saturating_sub(tracking).saturating_add(1)` computes "number of slots to advance" (e.g., LCL=5005 tracking=5000 → gap=6), while the issue table uses `lcl - tracking` (gap=5). The field name could mislead operators correlating logs with the issue. A follow-up note or doc clarification would improve debuggability, but this is a hotfix and the raw `lcl_seq`/`tracking` fields are also logged, so the value is recoverable.

**Metric not yet in Prometheus scrape path:** `lcl_ahead_of_tracking_corrective_total` is accessible via the public accessor (`Herder::lcl_ahead_of_tracking_corrective_total()`) but is not yet wired into the app-level metrics snapshot or the Prometheus exposition. The PR body acknowledges this and defers full exposure as a follow-up. Acceptable for a validator-down hotfix.

---

### Parity

Confirmed accurate: stellar-core `HerderImpl::trackingConsensusLedgerIndex()` throws (`std::runtime_error`) rather than aborting. The henyey `panic = "abort"` was strictly more fatal than upstream. The corrective branch moves henyey closer to upstream semantics.

</details>

PR #2794, opened 2026-05-18
Add zsh-gated regression test for monitor-decisions suffix-strip

### LINKED ISSUE #2782: Follow-up: zsh regression test for monitor-decisions.sh suffix-strip
Follow-up from #2726 / PR #2780 (Copilot bot review).

## Context

PR #2780 fixed a zsh parameter-expansion bug in `scripts/lib/monitor-decisions.sh` (`${exe% (deleted)}` → `${exe%' (deleted)'}`). The existing snippet test suite in `scripts/test-monitor-skill-snippets.sh` (T20c3 etc.) covers the `(deleted)` binary path, but only when run under bash — where the bug doesn't manifest. The fix would have been caught by an equivalent run under zsh.

## Ask

Add an automated zsh regression check for `_enumerate_henyey_processes` (or the suffix-strip step in isolation) to `scripts/test-monitor-skill-snippets.sh` (or a sibling), gated on `command -v zsh` so it's skipped where zsh is unavailable.

## Why follow-up, not part of #2780

Both internal reviewers on PR #2780 (Risk + Correctness) explicitly judged this out of scope for a one-character quoting fix and flagged it as a worthwhile follow-up. Filing here so the suggestion isn't dropped.

tomerweller:
Closes #2782

## Summary

Adds a zsh-gated regression test (`T20c3-zsh`) to `scripts/test-monitor-skill-snippets.sh` that re-runs the existing `_enumerate_henyey_processes` `(deleted)`-exe scenario inside a `zsh -c` subprocess. This guards against regression of the zsh parameter-expansion bug fixed in PR #2780, where the unquoted form `${exe% (deleted)}` was parsed by zsh as a glob pattern and silently failed to strip the ` (deleted)` suffix.

The new test is gated on `command -v zsh` via a new `tap_skip` helper, so the TAP plan stays consistent on systems without zsh installed. `TAP_PLAN` is bumped from 315 to 316 to account for the new test.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/2782#issuecomment-4474450988)

## Test plan

- [x] `bash scripts/test-monitor-skill-snippets.sh` — new test `ok 26 - _enumerate_henyey_processes: (deleted) exe matched — zsh recheck` passes; only pre-existing unrelated `pv-label-sync` failures remain (also present on origin/main).
- [x] Mutation verification — temporarily reverted `monitor-decisions.sh` line 37 to the unquoted form `${exe% (deleted)}`; new test correctly failed (`not ok 26 - ... — zsh recheck`), then passed again after restoring the quoted form `${exe%' (deleted)'}`. This confirms the test is a real regression guard.
- [x] cargo fmt / clippy — passed via pre-commit hook (no Rust changes; pre-commit ran the full workspace check anyway).

## Regression test

n/a — this PR's `kind` is `test-only` per triage and the converged plan; no `bug-fix` regression test is required (the underlying bug was already fixed in PR #2780, and this PR adds the missing guard test).

## Deviations from plan

None.

### DIFF
--- a/scripts/test-monitor-skill-snippets.sh
+++ b/scripts/test-monitor-skill-snippets.sh
@@ -45,7 +45,7 @@ cleanup  # ensure fresh state
 mkdir -p "$TEST_ROOT"
 
 # ── TAP state ────────────────────────────────────────────────────────────────
-TAP_PLAN=315
+TAP_PLAN=316
 TAP_CURRENT=0
 TAP_FAILURES=0
 
@@ -63,8 +63,15 @@ tap_not_ok() {
   TAP_FAILURES=$((TAP_FAILURES + 1))
   echo "not ok $TAP_CURRENT - $1"
   if [[ -n "${2:-}" ]]; then
-    echo "  # $2" 
+    echo "  # $2"
   fi
+}
+
+tap_skip() {
+  # Emit a passing TAP line with a SKIP directive so the count stays
+  # consistent regardless of whether the optional dependency is installed.
+  TAP_CURRENT=$((TAP_CURRENT + 1))
+  echo "ok $TAP_CURRENT - $1 # SKIP ${2:-unavailable}"
 }
 
 # ── Section extraction helper ─────────────────────────────────────────────────
@@ -892,6 +899,28 @@ run_tests() {
     tap_ok "_enumerate_henyey_processes: (deleted) exe matched"
   else
     tap_not_ok "_enumerate_henyey_processes: (deleted) exe matched" "got: '$enum_out'"
+  fi
+
+  # ── Test 20c3-zsh: (deleted) exe matched — zsh recheck ──────────────────
+  # Regression guard for the zsh parameter-expansion bug fixed in PR #2780:
+  # under zsh, the unquoted form `${exe% (deleted)}` parses ` ` as a glob and
+  # silently fails to strip the suffix. This re-runs the T20c3 scenario inside
+  # a `zsh -c` subprocess so the suffix-strip in monitor-decisions.sh line 37
+  # is exercised under zsh semantics. Gated on zsh availability.
+  if command -v zsh >/dev/null 2>&1; then
+    data="$TEST_ROOT/t20c3z/data"
+    proc="$TEST_ROOT/t20c3z/proc"
+    mkdir -p "$data" "$proc"
+    mock_proc_entry "$proc" "8010" "$data/sessc3z/cargo-target/release/henyey (deleted)" "henyey --mainnet run"
+    # Pass all paths as positional args to avoid inline-quoting hazards.
+    enum_out=$(zsh -c 'source "$1/scripts/lib/monitor-decisions.sh"; _enumerate_henyey_processes "$2" "$3"' -- "$REPO_ROOT" "$data" "$proc")
+    if [[ "$enum_out" == "8010 sessc3z" ]]; then
+      tap_ok "_enumerate_henyey_processes: (deleted) exe matched — zsh recheck"
+    else
+      tap_not_ok "_enumerate_henyey_processes: (deleted) exe matched — zsh recheck" "got: '$enum_out'"
+    fi
+  else
+    tap_skip "_enumerate_henyey_processes: (deleted) exe matched — zsh recheck" "zsh not found"
   fi
 
   # ── Test 20c4: empty proc root ─────────────────────────────────────────


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The implementation correctly fulfills the `test-only` plan. The `tap_skip` helper is spec-compliant, the zsh subprocess invocation correctly exercises the fix, and test isolation is clean.

<details>

**Kind:** `test-only` per triage. No regression-test-first requirement applies; the new test IS the deliverable.

**Plan fulfillment:**
- `tap_skip` helper added alongside the test block — matches plan exactly (two-liner, TAP-compliant).
- T20c3-zsh block placed immediately after T20c3 — matches plan placement.
- `TAP_PLAN` bumped 315 → 316 — correct: `tap_skip` emits an `ok` line so it counts toward the plan total.
- Fresh fixture directory `t20c3z/` created — no state leakage from T20c3 (which uses `t20c3/`).

**`tap_skip` correctness:**
- Increments `TAP_CURRENT` (consistent with `tap_ok`/`tap_not_ok`).
- Emits `ok $TAP_CURRENT - $1 # SKIP ${2:-unavailable}` — exactly the TAP 13 SKIP directive format. The `ok` prefix ensures the test counts toward the plan (correct for optional-dependency gates).
- Does NOT increment `TAP_FAILURES` — correct, a skip is not a failure.

**zsh subprocess invocation:**
```bash
enum_out=$(zsh -c 'source "$1/scripts/lib/monitor-decisions.sh"; _enumerate_henyey_processes "$2" "$3"' -- "$REPO_ROOT" "$data" "$proc")
```
- Single-quoted `zsh -c` body prevents double-expansion from bash; `$1`/`$2`/`$3` expand under zsh. Correct.
- `--` after the script body is the correct POSIX convention to pass `$0`, `$1`, ... to `zsh -c`. Under `zsh -c 'body' -- arg1 arg2 arg3`, `$1=arg1`, `$2=arg2`, `$3=arg3`. This is correct.
- `$REPO_ROOT` is expanded by bash before passing to zsh (it's outside the single-quoted body), so the path is resolved at invocation time. Correct.
- The subprocess sources `monitor-decisions.sh` which defines `_enumerate_henyey_processes`. The function uses `${exe%' (deleted)'}` (quoted form) — this is what the test validates.

**Mutation verification (from PR body):** The author confirmed the test correctly fails when `monitor-decisions.sh` line 37 reverts to the unquoted form — this is a real regression guard.

**Test naming / PID choice:**
- PID `8010` and session `sessc3z` are unique within the test suite (not conflicting with T20c3's PID `8004`/session `sessc3` or other tests). Clean.

**Minor: trailing whitespace removed** from `tap_not_ok` line 66 (pre-existing trailing space was cleaned up as a drive-by). Non-blocking.

**No concerns blocking merge.**

</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** No regressions to existing behavior, no flakiness risk, no injection surface. TAP plan count is correct; zsh subprocess is safely parameterized.

<details>

**TAP plan count (316 correct):**
- `tap_skip` emits `ok N - ... # SKIP ...` — an `ok`-line counts toward the TAP plan total. So a skipped test still consumes one plan slot. `TAP_PLAN` bumped 315 → 316 is correct whether zsh is present (tap_ok/tap_not_ok) or absent (tap_skip). No plan-mismatch warning in either branch.

**Existing test count unaffected:**
- `tap_skip` does NOT increment `TAP_FAILURES`, so pre-existing pass/fail ratios are unchanged on systems without zsh. On systems with zsh, the new test either passes or fails independently. No risk to existing TAP lines.

**Test isolation:**
- Fresh fixture dirs `$TEST_ROOT/t20c3z/data` and `$TEST_ROOT/t20c3z/proc` are created. No overlap with `t20c3/` (PID `8010` vs `8004`). The outer `cleanup` on EXIT handles `$TEST_ROOT` removal. No orphaned state.

**Flakiness risk:**
- `command -v zsh` check is deterministic per environment. The `zsh -c` invocation sources a single local file and calls a pure function (no network, no timing). No flakiness risk.

**Hang/timeout risk:**
- `_enumerate_henyey_processes` is a pure shell function: reads `/proc`-style symlinks and cmdline files (all synthetic in-test). No blocking I/O, no subprocesses beyond `readlink`/`cat`. The zsh subprocess exits synchronously. No hang risk.

**Security (injection):**
- All three variable paths (`$REPO_ROOT`, `$data`, `$proc`) are passed as positional arguments to `zsh -c`, not interpolated into the script body (which is single-quoted). The function body only reads the paths it receives — no `eval`, no dynamic expansion of the argument values. No injection surface.

**Breaking changes to APIs or data formats:** None — this change is entirely within the test harness; it touches no production code, no config, no data schemas.

**Operational concerns:** None — this is a dev/CI test script, not a deployed artifact.

**No concerns blocking merge.**

</details>

tomerweller:
## ✅ Merged

**Commit:** c33f3b19c3223f71dd90bf56d82c3642e1e2a73f

**Follow-up issues filed for unaddressed inline review comments:** (none — no inline review threads)

Issue #2698, opened 2026-05-14
HERDER: Wire ScpPersistenceManager GC timer into production lifecycle


tomerweller:
## Context

Follow-up from #2668 (TX_SET_GC_DELAY parity).

The mark-and-sweep GC logic for persisted tx sets is implemented in `ScpPersistenceManager::purge_unreferenced_tx_sets()` and the `TX_SET_GC_DELAY_SECS` constant (60s) is defined. However, `ScpPersistenceManager` is not yet wired into the production lifecycle.

## Work needed

1. Integrate `ScpPersistenceManager` into the production herder
2. Add a periodic timer that calls `purge_unreferenced_tx_sets()` every `TX_SET_GC_DELAY_SECS` seconds, matching stellar-core's `startTxSetGCTimer()`
3. Ensure GC and persistence don't race (single-threaded or transaction-based)
4. Update herder `PARITY_STATUS.md` from Partial to Full for `startTxSetGCTimer()`

## References

- stellar-core: `HerderImpl::startTxSetGCTimer()` at `HerderImpl.cpp:2440-2444`
- henyey: `TX_SET_GC_DELAY_SECS` at `crates/herder/src/herder.rs:122`
- henyey: `purge_unreferenced_tx_sets()` at `crates/herder/src/persistence.rs:495`

tomerweller:
## Triage Report

**Verdict:** ACCEPT

**Type:** enhancement
**Severity:** n/a
**Crate:** crate:herder
**Size:** M

**Summary:** Wire the existing `ScpPersistenceManager` into the production herder lifecycle and add a periodic timer that invokes `purge_unreferenced_tx_sets()` every `TX_SET_GC_DELAY_SECS` (60s), matching stellar-core's `startTxSetGCTimer()`. Issue is well-formed with clear file references and a four-point work breakdown; touches a single crate (`crates/herder`) but requires careful design around the persistence/GC race and PARITY_STATUS update — non-trivial, hence `ready-for-planning` rather than the trivial short-circuit.

**Label adjustments:** added `enhancement`, `crate:herder`.

tomerweller:
## 🔒 acquired-by:tick-1778907322812680164-2687713

posted=2026-05-16T04:55:24.556Z, host=user-dev-050a, pid=2687713

tomerweller:
## 📝 Plan Draft (Round 1)

[...]

PR #2795, opened 2026-05-18
Wire ScpPersistenceManager::persist_scp_state into ScpDriver::emit (#2768)

### LINKED ISSUE #2768: HERDER: Wire ScpPersistenceManager::persist_scp_state into envelope-emit hot path
## Context

Follow-up from #2698 (wiring `ScpPersistenceManager` GC timer into production).

#2698 wires the `ScpPersistenceManager` into the production herder and runs `purge_unreferenced_tx_sets()` on a 60s tokio interval, but does **not** call `persist_scp_state()` from the SCP envelope-emit / externalize hot paths. Until that happens, the GC timer has nothing to garbage-collect — the SQLite `storestate` rows for SCP envelopes and tx sets are never written.

## Work needed

1. Identify the equivalent of stellar-core's `HerderImpl::persistSCPState(slotIndex)` callsite (HerderImpl.cpp; called from `emitEnvelope` after each SCP envelope emission).
2. Add a matching call in henyey's `Herder` after each emitted/processed envelope, gathering the slot's envelopes + referenced tx_sets + quorum_sets, and invoking `ScpPersistenceManager::persist_scp_state(...)`.
3. Once persisted, the GC timer added in #2698 will purge orphans correctly.
4. Update `crates/herder/PARITY_STATUS.md` for `persistSCPState()` (currently listed as `Full` against `ScpPersistenceManager.persist()` — the manager method exists but is not actually called).

## References

- stellar-core: `HerderImpl::persistSCPState()` in `stellar-core/src/herder/HerderImpl.cpp`
- henyey: `ScpPersistenceManager::persist_scp_state()` in `crates/herder/src/persistence.rs:376`
- Parent: #2698

tomerweller:
Closes #2768

## Summary

- Wires `ScpPersistenceManager::persist_scp_state()` into the `ScpDriver::emit` hot path via a new persist callback on `ScpDriver` installed by `Herder::set_scp_persistence`. After this PR every emitted envelope produces a durable SCP-state write (envelopes + referenced tx-sets + quorum-sets), completing the persistence loop started in #2698 — the GC timer added in #2698 finally has rows to act on.
- Mirrors stellar-core's `HerderImpl::emitEnvelope` → `persistSCPState(slotIndex)` chain (`stellar-core/src/herder/HerderImpl.cpp:599-2186`).
- Deferral deviation: henyey's `SCP` holds a `parking_lot::RwLock` write guard on its slots map while `emit` runs, so calling `get_latest_messages_send` synchronously inside the callback would deadlock against the writer chain. The callback therefore spawns a worker thread (`scp-persist-<slot>`) to do the gather + persist work AFTER the writer unwinds. Persistence remains best-effort (`warn!` on error, no propagation). `ScpPersistenceManager` exposes `wait_for_pending_persists(timeout)` so tests can synchronize.

Converged plan: https://github.com/stellar-experimental/henyey/issues/2768#issuecomment-4474401347

## Regression tests

- **Test-only commit** [`23719a2d`](https://github.com/stellar-experimental/henyey/commit/23719a2de4cd8bb888c4d7af67be92cd5c2a3bb2) — three solo-validator regression tests that on `origin/main`:
  - `test_persist_scp_state_called_on_emit` — FAILS (`last_slot_saved=0` after a real externalize)
  - `test_persist_scp_state_gc_sees_persisted_rows` — FAILS at the same precondition
  - `test_persist_scp_state_no_manager_no_panic` — passes (negative test, OnceLock no-op)
- **Fix commit** [`d0daad20`](https://github.com/stellar-experimental/henyey/commit/d0daad2002f763c35a039679adb4f034757d9e29) — wires the persist callback. The two previously-failing tests now PASS once they call `wait_for_pending_persists` (needed because the persist is deferred to a worker thread; see deviation above).

## Test plan

- [x] `cargo test -p henyey-herder --lib test_persist_scp_state` — 3 passed
- [x] `cargo test -p henyey-herder` — 1118 lib + 7 + 1 + 1 + 3 integration tests passed
- [x] `cargo fmt --check` — clean (applied post-edit)
- [ ] CI must pass on the new branch

## Parity

- stellar-core `HerderImpl.cpp:599-612` `emitEnvelope` → `persistSCPState(slotIndex)` → broadcast
- henyey `ScpDriver::emit` → broadcast → persist callback (deferred to worker thread)
- Mapping: `getSCP().getLatestMessagesSend(slot)` → `Arc<SCP>::get_latest_messages_send`; `mPendingEnvelopes.getTxSet(h)` → `Arc<ScpDriver>::get_tx_set`; `mPendingEnvelopes.getQSet(qsHash)` → `Arc<ScpDriver>::get_quorum_set_by_hash`; `StoredTransactionSet` XDR → `TransactionSet::to_xdr_stored_set()` + `WriteXdr::to_xdr(Limits::none())`.
- `PARITY_STATUS.md` updated: `persistSCPState()` row points at the wired callback; `[^persist-scp]` footnote documents the worker-thread deviation; `[^txset-gc]` footnote notes #2768 resolved.

### DIFF
--- a/crates/herder/PARITY_STATUS.md
+++ b/crates/herder/PARITY_STATUS.md
@@ -114,7 +114,7 @@ Corresponds to: `Herder.h`, `HerderImpl.h`
 | `maybeSetupSorobanQueue()` | Integrated via lane-based `TransactionQueue` | Full |
 | `herderOutOfSync()` | `SyncRecoveryManager` | Full |
 | `getMoreSCPState()` | _(not implemented)_ | None |
-| `persistSCPState()` | `ScpPersistenceManager.persist()` | Full |
+| `persistSCPState()` | `ScpPersistenceManager.persist_scp_state()` wired via `ScpDriver::emit` persist callback | Full[^persist-scp] |
 | `restoreSCPState()` | `ScpPersistenceManager.restore()` | Full |
 | `persistUpgrades()` | `UpgradeParameters` with Serde persistence | Full |
 | `restoreUpgrades()` | `UpgradeParameters` with Serde persistence | Full |
@@ -130,9 +130,22 @@ Corresponds to: `Herder.h`, `HerderImpl.h`
 | `recomputeKeysToFilter()` | _(not implemented)_ | None |
 
 [^txset-gc]: GC timer + purge wired in #2698 (driven by app event-loop phase
-33 every `TX_SET_GC_DELAY_SECS` = 60s). The broader SCP persist/restore story
-is still incomplete: `persist_scp_state` wiring is tracked in #2768 and
-`restore_scp_state` wiring in #2769.
+33 every `TX_SET_GC_DELAY_SECS` = 60s). The `persist_scp_state` wiring was
+completed in #2768 (`ScpDriver::emit` invokes the persist callback installed
+by `Herder::set_scp_persistence`). The `restore_scp_state` wiring remains
+tracked in #2769.
+
+[^persist-scp]: Wired in #2768 via a deferred persist callback on `ScpDriver`.
+Each call to `ScpDriver::emit` (which mirrors stellar-core's
+`HerderImpl::emitEnvelope`) fires the callback installed by
+`Herder::set_scp_persistence`; the callback spawns a worker thread that
+gathers `get_latest_messages_send(slot)` envelopes plus referenced tx-sets
+and quorum-sets and invokes `ScpPersistenceManager::persist_scp_state`. The
+work is deferred to a worker thread (not synchronous as in stellar-core)
+because henyey's `SCP` holds a `parking_lot::RwLock` write guard on its
+slots map while `emit` runs, and a synchronous `get_latest_messages_send`
+would deadlock against the same lock. Persistence remains best-effort:
+errors are logged but not propagated.
 
 ### SCP Driver (`scp_driver.rs`)
 
--- a/crates/herder/src/herder.rs
+++ b/crates/herder/src/herder.rs
@@ -456,7 +456,12 @@ pub struct Herder {
     /// SCP consensus protocol instance.
     /// Always created — observers run SCP with `is_validator: false` (same as stellar-core),
     /// which means `fully_validated` is false on slots and `send_latest_envelope` won't emit.
-    scp: SCP<HerderScpCallback>,
+    ///
+    /// Wrapped in `Arc` so [`Self::set_scp_persistence`] can clone a handle
+    /// into the persist callback installed on [`Self::scp_driver`]. The
+    /// callback fires from [`crate::scp_driver::ScpDriver::emit`] after
+    /// each envelope broadcast and needs read-access to `get_latest_messages_send`.
+    scp: Arc<SCP<HerderScpCallback>>,
     /// Shared tracking consensus state (also read by ScpDriver for value validation).
     tracking_state: Arc<RwLock<SharedTrackingState>>,
     /// When we started tracking.
@@ -640,12 +645,12 @@ impl Herder {
             warn!("Validator mode requested but no quorum set configured");
         }
         let callback = HerderScpCallback::new(Arc::clone(&scp_driver));
-        let scp = SCP::new(
+        let scp = Arc::new(SCP::new(
             node_id.clone(),
             is_validator,
             quorum_set,
             Arc::new(callback),
-        );
+        ));
 
         let slot_quorum_tracker =
             SlotQuorumTracker::new(config.local_quorum_set.clone(), max_slots);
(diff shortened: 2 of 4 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The implementation correctly wires `persist_scp_state` into the emit hot path with sound concurrency semantics. The TDD commit ordering is confirmed, all three required tests are present, and the race-safety analysis holds.

<details>

### Issue Kind and Test Coverage

- **Kind:** `feature` (confirmed from triage report on issue #2768)
- **TDD ordering verified:** Test commit `23719a2d` (2026-05-18T04:55:48Z) is before fix commit `d0daad20` (2026-05-18T05:57:07Z). ✓
- **Required regression tests present:**
  - `test_persist_scp_state_called_on_emit` — drives a full solo-validator emit and asserts `last_slot_saved > 0`. Would fail on `origin/main`. ✓
  - `test_persist_scp_state_gc_sees_persisted_rows` — end-to-end GC loop validation. ✓
  - `test_persist_scp_state_no_manager_no_panic` — no-manager no-op guard. ✓
- **New public surface exercised:**
  - `wait_for_pending_persists` — called in both emit and GC tests. ✓
  - `set_persist_callback` — exercised indirectly via `Herder::set_scp_persistence` in both emit-path tests. ✓
  - `pending_persist_begin/end` — `pub(crate)`, exercised via their callers in the tests. ✓

### Concurrency / Race Safety

- **pending counter increment BEFORE spawn:** `manager_for_thread.pending_persist_begin()` is called before `thread::Builder::spawn(...)`. A tight `wait_for_pending_persists()` call after emit cannot race past zero before the thread increments. ✓
- **Spawn failure path:** `map_err` correctly calls `manager_for_cb.pending_persist_end()` (same Arc target as `manager_for_thread`) when `spawn` fails and the closure that owns `manager_for_thread` is dropped. No counter leak. ✓
- **Thread panic safety:** Guard struct inside the spawn closure uses `Drop` to call `pending_persist_end()` on any exit path including panics. ✓
- **Guard lifetime:** `Guard(&manager_for_thread)` borrows the Arc-deref'd `ScpPersistenceManager`. `manager_for_thread` and `_guard` are in the same scope; `_guard` drops first, so the borrow is sound. ✓
- **Concurrent same-slot persists:** `persist_scp_state` holds the `last_slot_saved` write lock for the full duration, serializing concurrent callers. The `slot < *last_saved` check (not `<=`) means same-slot re-persists write again, but this is harmless (same data, tx-set dedup guards against double-storage) and matches stellar-core's check at HerderImpl.cpp:2148. ✓

### Arc Cycle

- `ScpDriver::persist_callback` captures `Arc<ScpDriver>`, creating a self-referential cycle. This mirrors the existing `envelope_sender` / `HerderScpCallback` cycle (`Herder → Arc<SCP<HerderScpCallback>> → HerderScpCallback → Arc<ScpDriver>`). Both are benign because all three Arcs are dropped together with the owning `Herder`. Doc comment explicitly notes this. ✓

### suppress_scp Path

- `ScpDriver::emit()` returns early when `self.config.suppress_scp == true` — before both the broadcast and the persist callback. The persist callback is therefore never invoked in standalone/suppress mode. This is the correct parity behavior. ✓

### Error Handling

- `persist_scp_state` errors are logged with `warn!` and not propagated. `StoredTransactionSet` XDR encode errors are logged and the tx set is skipped (same as stellar-core's "skip missing tx sets"). ✓
- Thread spawn failures are logged with `warn!` and the counter is correctly decremented. ✓

### Minor Notes (non-blocking)

- No explicit test for `suppress_scp = true` (Critic A on the plan flagged this as REVISE-MINOR, but the converged plan did not add a blocking requirement for it). The code path is trivially correct — `return` before the callback. Low risk.
- `set_scp_persistence` now silently drops `manager` on double-install (`self.scp_persistence.set(Arc::clone(&manager))?` returns the error via `?`). The `set_persist_callback` is correctly NOT called on the double-install path. ✓

### Code Quality

- Readability: inline parity comments referencing `HerderImpl.cpp:2152-2186` are helpful. The `move |slot| { ... }` closure is clearly structured. ✓
- `PARITY_STATUS.md` correctly updated: `persistSCPState()` row updated with wired callback reference and deviation footnote. `[^txset-gc]` footnote updated to mark #2768 resolved. ✓
- `Arc<SCP<HerderScpCallback>>` wrapping does not require any `&mut self` changes since SCP methods use interior mutability. All `self.scp.method()` calls auto-deref through Arc correctly. ✓

</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** The `ScpDriver::emit` → `persist_scp_state` wiring correctly mirrors `HerderImpl::emitEnvelope` → `persistSCPState`. All five parity-critical points (slot guard, `mLastSlotSaved` timing, tx-set dedup, quorum-set extraction, `suppress_scp` short-circuit) match stellar-core semantics. The documented deviation (deferred worker thread instead of synchronous call, post-broadcast instead of pre-broadcast) is safe and accurately explained.

<details>

### 1. persist-before-broadcast vs persist-after-broadcast

- **stellar-core** (`HerderImpl.cpp:609-611`): `persistSCPState(slotIndex)` → `broadcast(envelope)`.
- **henyey**: `sender(envelope)` (broadcast) → `cb(slot)` (persist callback, deferred to worker thread).
- **Parity verdict:** Acceptable deviation. The PR's reasoning is correct: neither ordering is atomic (both can be interrupted between persist and broadcast), so crash-safety is equivalent. The deviation is necessitated by the `parking_lot::RwLock` write guard on SCP's slots map being held during `emit` — a synchronous `get_latest_messages_send` call would deadlock. The PARITY_STATUS.md footnote `[^persist-scp]` correctly documents this.

### 2. Slot guard semantics

- **stellar-core** (`HerderImpl.cpp:2128-2131`): `if (slot < mLastSlotSaved) return;` — strict less-than; same slot re-persists are allowed.
- **henyey** (`persistence.rs:455`): `if slot < *last_saved { return Ok(()); }` — identical semantics. ✓

### 3. `mLastSlotSaved` update timing

- **stellar-core**: `mLastSlotSaved = slot` at line 2133, before any envelope/tx-set gathering.
- **henyey**: `*last_saved = slot` at line ~466, before `state.add_envelope` / tx-set loop. ✓

### 4. Transaction set deduplication

- **stellar-core** (`HerderImpl.cpp:2149-2153`): `if (txSet && !mApp.getPersistentState().hasTxSet(h))` before inserting into `txSets` map. Also uses `std::map` which auto-deduplicates by hash across envelopes.
- **henyey**: The callback builds a `Vec<(Hash, Vec<u8>)>` without dedup (multiple envelopes could yield the same hash). However, `persist_scp_state` (`persistence.rs:~473`) does `if !self.storage.has_tx_set(hash)? { self.storage.save_tx_set(hash, tx_set)?; }` — same idempotency semantics. The underlying SQLite backend uses `INSERT OR REPLACE`, so even if duplicates reach storage, the result is correct.
- **Minor deviation:** The `Vec` could have duplicate hash entries if multiple envelopes in the same slot reference the same tx set. This is benign (has_tx_set short-circuits the second write) but slightly wasteful. Non-blocking for this PR.

### 5. Quorum set extraction

- **stellar-core** (`HerderImpl.cpp:2155`): `Slot::getCompanionQuorumSetHashFromStatement(e.statement)` — handles PREPARE (quorumSetHash), CONFIRM (quorumSetHash), EXTERNALIZE (commitQuorumSetHash), NOMINATE (quorumSetHash).
- **henyey** (`persistence.rs:333-339`): `get_quorum_set_hash` covers the same four arms with the same XDR fields. ✓
- **stellar-core**: uses `std::map<Hash, SCPQuorumSetPtr>` which auto-deduplicates quorum sets across envelopes. **henyey**: uses `Vec` — same minor duplication issue as tx sets. On restore, duplicate quorum set bytes in `PersistedSlotState` would cause `get_quorum_sets()` to return the same QSet twice. On restore, this results in the same quorum set being pushed to `restored.quorum_sets` twice — harmless since callers deduplicate by hash or the map-insert is idempotent. Non-blocking.

### 6. `getValidatedTxSetHashes` vs `get_tx_set_hashes`

- **stellar-core**: throws `std::runtime_error` on invalid StellarValues inside envelopes.
- **henyey** (`persistence.rs:278-320`): `get_tx_set_hashes` returns an empty Vec if `extract_tx_set_hash_from_value` returns `None` (silently skips invalid values).
- **Parity verdict:** Minor deviation. In practice, envelopes that reach `emit` have already passed SCP validation and contain valid StellarValues. Silently skipping vs throwing is a divergence in error-handling philosophy, but it does not affect correctness of the persist path since the data is valid. Non-blocking.

### 7. `suppress_scp` short-circuit

- **henyey** `scp_driver.rs:2506`: `if self.config.suppress_scp { return; }` is before both the broadcast and the persist callback. The persist callback is therefore never invoked in standalone/suppress mode. ✓

### 8. PARITY_STATUS.md accuracy

- `persistSCPState()` row updated from "ScpPersistenceManager.persist()" to the correct `persist_scp_state()` reference with the wiring callsite.
- `[^persist-scp]` footnote accurately describes the worker-thread deviation and its justification.
- `[^txset-gc]` footnote correctly updated: notes #2768 resolved, #2769 still pending.
- All accurate. ✓

### Summary of deviations

| Point | stellar-core | henyey | Severity |
|---|---|---|---|
| persist/broadcast ordering | persist → broadcast | broadcast → persist (worker thread) | Accepted — documented, safe |
| quorum/tx-set Vec vs map | `std::map` (auto-dedup) | `Vec` (may duplicate) | Minor — non-blocking |
| invalid-StellarValue handling | throws | silently skips | Minor — non-blocking in practice |

All three are pre-existing patterns or justified deviations; none affect correctness of the persistence path.

</details>

PR #2798, opened 2026-05-18
Diagnose missing PROJECT_BOARD_TOKEN with structured error

### LINKED ISSUE #2777: CI: Archive Done Project Items workflow fails — PROJECT_BOARD_TOKEN secret missing or unscoped
## Symptom

The new `Archive Done Project Items` workflow (added 2026-05-15 with the pipeline refactor `5b757319`) fails on every scheduled run with:

```
gh: Could not resolve to a ProjectV2 with the number 2.
##[error]Process completed with exit code 1.
```

First confirmed failure: run 25956350118 on `3b31f4b3` at 07:38Z. The workflow runs daily at 07:00 UTC.

## Suspected root cause

`.github/workflows/archive-done.yml` uses:

```yaml
GH_TOKEN: ${{ secrets.PROJECT_BOARD_TOKEN || secrets.GITHUB_TOKEN }}
```

If `PROJECT_BOARD_TOKEN` is unset (which it appears to be), the workflow falls back to the default `GITHUB_TOKEN`, which doesn't have org-level `read:project` scope for `stellar-experimental` projects. The script queries `organization(login: "stellar-experimental") { projectV2(number: 2) }` via GraphQL — this requires a fine-grained PAT or a PAT with the project scope.

## Why no urgent

- The mainnet validator is unaffected.
- The Quickstart, CI, and other deploy-gating workflows are unaffected.
- Only board automation is broken: closed items don't auto-archive after 2 days. Operator can archive manually or via `/project-tick`.

## Suggested action

Configure `PROJECT_BOARD_TOKEN` repo secret with a fine-grained PAT that has `Projects: Read+Write` on `stellar-experimental` (likely the same scope your existing `gh` PAT uses). Once set, the workflow will succeed.

If the secret is configured but still failing, the PAT may have expired or lost the project scope.

## Reference

- workflow: `.github/workflows/archive-done.yml`
- script: `.github/skills/shared/scripts/archive-stale-done.sh`
- failed run: 25956350118
- introduced by: `5b757319` "Replace plan-do-review monolith with modular pipeline"

tomerweller:
Closes #2777

## Summary

The daily `Archive Done Project Items` workflow has been failing on every scheduled run with the opaque GraphQL error `gh: Could not resolve to a ProjectV2 with the number 2.` This happens because `PROJECT_BOARD_TOKEN` is unset, the workflow falls back to `GITHUB_TOKEN`, and `GITHUB_TOKEN` lacks the org-level `read:project` scope required for the `organization.projectV2` query. The `repository-projects: write` permission in the workflow grants only repo-level classic Projects access — it does not help here.

This PR doesn't configure the secret (that's an operator action against repo settings), but it makes the failure self-diagnosing:

- `archive-stale-done.sh` now runs a pre-flight GraphQL probe before the paginated fetch. When the probe sees the `Could not resolve to a ProjectV2` failure pattern, it emits `ERROR: GH_TOKEN lacks org project scope — set PROJECT_BOARD_TOKEN secret with Projects:Read+Write on stellar-experimental` and exits 1. Other failures (network, 401) are also surfaced rather than papered over.
- `SKIP_PREFLIGHT=1` bypasses the probe for test harnesses (not for production).
- `archive-done.yml` adds a diagnostic step before the run step that emits a GitHub Actions `::warning::` when `PROJECT_BOARD_TOKEN` is empty, plus an inline comment explaining the `repository-projects: write` vs org ProjectV2 scope distinction.
- New regression test `test-archive-stale-done-auth.sh` covers both the failure path (asserts structured error) and the happy path (SKIP_PREFLIGHT + dry-run).

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/2777#issuecomment-4474472259)

## Test plan

- [x] `bash .github/skills/shared/scripts/tests/test-archive-stale-done-auth.sh` — 2 passed, 0 failed
- [x] `bash -n` syntax check on both shell scripts
- [x] `python3 yaml.safe_load` on archive-done.yml
- [x] `cargo fmt --check` (no Rust changes — sanity check only)

## Regression test (bug-fix)

- **Test:** `.github/skills/shared/scripts/tests/test-archive-stale-done-auth.sh::test_preflight_fails_with_bad_token`
- **Pre-fix:** committed as `b7ae7eab` — verified FAILED with: `output missing 'ERROR: GH_TOKEN lacks org project scope'` (the script emitted only the opaque `gh: Could not resolve to a ProjectV2 with the number 2.`)
- **Post-fix:** verified PASSES after `0815b21b`

## Deviations from plan

None.

## Operator follow-up (out of scope for this PR)

The actual remediation — configuring `PROJECT_BOARD_TOKEN` as a fine-grained PAT with `Projects: Read+Write` on `stellar-experimental` — must be done in repo settings. After this PR merges, the daily workflow will still fail until the secret is set, but the failure will now be self-explanatory (structured error in the script + a `::warning::` in the Actions summary). Reference issue #2777 from the PAT configuration commit.

### DIFF
--- a/.github/skills/shared/scripts/archive-stale-done.sh
+++ b/.github/skills/shared/scripts/archive-stale-done.sh
@@ -29,6 +29,36 @@ OWNER="stellar-experimental"
 PROJECT_NUM=2
 PROJECT_ID="PVT_kwDOD-vqsM4BWQnL"
 
+# Pre-flight: probe whether GH_TOKEN has org-level project scope. The default
+# GITHUB_TOKEN granted to GitHub Actions runs lacks the org-level read:project
+# scope required for `organization(login:...).projectV2(...)` GraphQL queries —
+# the `repository-projects: write` workflow permission only grants repo-level
+# classic Projects access and does not help here. When the token is
+# underprivileged, the GraphQL call below would fail with the opaque
+# "Could not resolve to a ProjectV2 with the number 2." error, which doesn't
+# tell the operator what to fix. This probe catches that case and emits a
+# structured, actionable error message instead.
+#
+# SKIP_PREFLIGHT=1 is for test harnesses only; do not document or rely on it
+# in production paths.
+if [ "${SKIP_PREFLIGHT:-0}" != "1" ]; then
+  probe_exit=0
+  probe_output=$(gh api graphql -f query='
+    query($org: String!, $proj: Int!) {
+      organization(login: $org) { projectV2(number: $proj) { id } }
+    }' -f org="$OWNER" -F proj="$PROJECT_NUM" 2>&1) || probe_exit=$?
+  if [ "$probe_exit" -ne 0 ]; then
+    if grep -q "Could not resolve to a ProjectV2" <<<"$probe_output"; then
+      echo "ERROR: GH_TOKEN lacks org project scope — set PROJECT_BOARD_TOKEN secret with Projects:Read+Write on stellar-experimental" >&2
+      exit 1
+    fi
+    # Different failure (network, 401, etc.) — surface it and exit so the
+    # main fetch doesn't paper over a real error.
+    echo "ERROR: pre-flight probe failed: $probe_output" >&2
+    exit 1
+  fi
+fi
+
 # Fetch all project items + their issue/PR state and closedAt.
 # `--paginate` requires the cursor variable to be named $endCursor; jq -s
 # is required because gh emits one JSON object per page.
--- a/.github/skills/shared/scripts/tests/test-archive-stale-done-auth.sh
+++ b/.github/skills/shared/scripts/tests/test-archive-stale-done-auth.sh
@@ -1,0 +1,140 @@
+#!/usr/bin/env bash
+# test-archive-stale-done-auth.sh — regression tests for archive-stale-done.sh
+# auth pre-flight behavior.
+#
+# Tests:
+#   1. test_preflight_fails_with_bad_token
+#        Runs the script with GH_TOKEN set to an invalid value. Asserts the
+#        script exits non-zero AND output contains the structured
+#        "ERROR: GH_TOKEN lacks org project scope" message rather than the
+#        opaque GraphQL "Could not resolve to a ProjectV2" error. This is the
+#        regression for issue #2777.
+#   2. test_preflight_succeeds_with_dry_run
+#        Runs with SKIP_PREFLIGHT=1 and --dry-run, mocking the gh binary so
+#        no network call is made. Asserts exit 0 and no ERROR: in output.
+#        Covers the happy path with the probe bypassed.
+#
+# Usage:
+#   bash .github/skills/shared/scripts/tests/test-archive-stale-done-auth.sh
+#
+# Exits 0 if all tests pass, non-zero on any failure.
+set -uo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+TARGET_SCRIPT="$SCRIPT_DIR/../archive-stale-done.sh"
+
+if [ ! -x "$TARGET_SCRIPT" ]; then
+  echo "FAIL: target script not found or not executable: $TARGET_SCRIPT" >&2
+  exit 1
+fi
+
+FAILED=0
+PASSED=0
+
+# ---------------------------------------------------------------------------
+# Test 1: pre-flight fails with bad token → structured error message
+# ---------------------------------------------------------------------------
+test_preflight_fails_with_bad_token() {
+  local name="test_preflight_fails_with_bad_token"
+  local tmpdir
+  tmpdir="$(mktemp -d)"
+  trap 'rm -rf "$tmpdir"' RETURN
+
+  # Mock `gh` to simulate the "Could not resolve to a ProjectV2" failure that
+  # occurs when GH_TOKEN lacks org project scope. The real failure mode is a
+  # non-zero exit with that exact stderr message; we reproduce it.
+  cat >"$tmpdir/gh" <<'EOF'
+#!/usr/bin/env bash
+# Mock gh: simulate the underprivileged-token GraphQL failure.
+if [[ "$1" == "api" && "$2" == "graphql" ]]; then
+  echo "gh: Could not resolve to a ProjectV2 with the number 2." >&2
+  exit 1
+fi
+exit 0
+EOF
+  chmod +x "$tmpdir/gh"
+
+  local output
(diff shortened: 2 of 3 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The regression test correctly fails on the pre-fix commit (`b7ae7eab`) and passes after the fix (`0815b21b`), confirming the test genuinely captures the bug. The pre-flight probe logic, error messages, and test harness are all sound.

<details>

## Regression test verification

- Commit order: test (`b7ae7eab` "Regression test for #2777") was committed BEFORE the fix (`0815b21b` "Add pre-flight token scope probe…") — correct TDD ordering.
- Ran `test-archive-stale-done-auth.sh` at `b7ae7eab` (pre-fix): `test_preflight_fails_with_bad_token` **FAILED** as expected — the script emitted only the opaque `gh: Could not resolve to a ProjectV2` message, not the structured ERROR. `test_preflight_succeeds_with_dry_run` passed (no preflight involved).
- Ran at `0815b21b` (post-fix): **both tests PASS**. The regression test genuinely captures the bug.

## archive-stale-done.sh pre-flight probe

- Correct: `probe_exit=0; … || probe_exit=$?` pattern avoids `set -e` abort while still capturing the exit code. Works correctly under `set -euo pipefail`.
- Correct: `2>&1` redirect captures stderr output for the `grep` pattern match.
- The `grep -q "Could not resolve to a ProjectV2" <<<"$probe_output"` match is exact and matches the real GH CLI error string observed in the failing runs (per issue #2777).
- Generic error fallback (`echo "ERROR: pre-flight probe failed: $probe_output"`) is appropriate — surfaces 401, network errors, etc.
- `SKIP_PREFLIGHT=1` bypass is documented as test-harness-only and not exposed in production paths.
- Minor non-blocking note: The probe re-runs the same GraphQL query as the main fetch (same `organization.projectV2` path), so on happy path this doubles the API round-trips. This is intentional and acceptable for a daily cron (not performance-sensitive).

## archive-done.yml diagnostic step

- `${{ secrets.PROJECT_BOARD_TOKEN != '' }}` evaluates to `"true"` or `"false"` as a string in the `env` block — the `if [ "$HAS_TOKEN" != "true" ]` check is correct.
- `::warning::` annotation is correct GH Actions syntax — will appear prominently in the Actions summary UI.
- The comment explaining `repository-projects: write` vs org ProjectV2 scope is valuable — prevents future confusion about why adding that permission doesn't fix the issue.

## test-archive-stale-done-auth.sh

- `set -uo pipefail` (without `-e`) is appropriate for a test runner — allows individual tests to fail without aborting the whole suite.
- PATH injection mock pattern (`PATH="$tmpdir:$PATH"`) correctly intercepts `gh` calls without modifying system state.
- `trap 'rm -rf "$tmpdir"' RETURN` cleanup is correct and runs even on test failure.
- Test 1 mocks only `graphql` calls and passes through everything else (`exit 0`) — clean and minimal.
- Test 2's mock returns a valid empty-nodes JSON payload that satisfies the `jq` pipeline in the main script.
- Both tests use local FAILED/PASSED counters rather than `exit 1` on first failure, giving full results before exit.

## Minor non-blocking observations (inline)

- The `probe_output` string may contain the full GraphQL response on success (including `id`). On success it's discarded correctly (`probe_exit` is 0 and the `if` block is skipped). No issue.
- The test file is at `.github/skills/shared/scripts/tests/` which is consistent with the existing test layout (sibling `tests/` dir under `scripts/`).

</details>

tomerweller:
## 🔍 Reviewer: Risk

**PR:** #2798 — Diagnose missing PROJECT_BOARD_TOKEN with structured error
**Issue:** #2777

---

### Regression in existing behavior (happy path)

When `PROJECT_BOARD_TOKEN` is valid, the pre-flight probe runs the same GraphQL query (`organization.projectV2.id`) that the main fetch also runs. On success (`probe_exit == 0`) the script proceeds normally. The added code path is a straight `if/else` that skips no existing logic and adds no mutation. The script correctly uses `--paginate` only on the main fetch (not the probe), so pagination is unaffected.

**Finding:** No regression on the happy path. The only behavioral change is an extra GraphQL call per run. With a valid token on a fast network this adds ~200–500 ms; well within the 10-minute job timeout and not operationally meaningful.

---

### Operational noise on normal runs

The workflow adds a new diagnostic step: **"Check PROJECT_BOARD_TOKEN secret"**. On a normal run (token configured), it prints `PROJECT_BOARD_TOKEN is configured.` — a single informational line. No `::warning::` is emitted. No annotation appears in the Actions summary. This is acceptably quiet.

When the token is absent (the current broken state), the `::warning::` fires before the run step, giving the operator a summary-level signal without marking the job failed — appropriate, since the structured error in the script itself is what fails the job.

**Finding:** No meaningful noise on normal runs. The warning-only-on-missing-token design is correct.

---

### Failure mode safety (fail-closed)

Three failure branches in the pre-flight probe:

1. **`Could not resolve to a ProjectV2` (scope error):** exits 1 with actionable message. Correct — the subsequent main fetch would also fail, so stopping early is strictly better.

2. **Other non-zero exit (network error, 401, 5xx, etc.):** also exits 1 with `ERROR: pre-flight probe failed: <output>`. Fail-closed. The script does not attempt a "best effort" fallthrough.

3. **`SKIP_PREFLIGHT=1`:** bypasses the probe. This is labeled test-harness-only in the comment; there is no path that sets it in the workflow or from any documented caller. Low risk.

One subtle point: the probe uses `probe_output=$(... 2>&1) || probe_exit=$?`. The `2>&1` redirect captures stderr into `probe_output`. This means the `grep` on `probe_output` will correctly match even if `gh` writes the error to stderr, which is what the mock and presumably the real `gh` CLI do. The approach is sound.

**Finding:** All failure branches exit non-zero and surface the error. No fail-open path.

---

### Minor observations (non-blocking)

- The probe fires a real API call even when the token IS valid, adding a per-run GraphQL hit. This is negligible, but if the workflow is ever rate-limited it is one more call. Not actionable now.
- `SKIP_PREFLIGHT=1` is undocumented in the script's usage header comment, consistent with its "not for production" intent. Fine.
- The test file name `test-archive-stale-done-auth.sh` is consistent with the directory convention; the mock `gh` in test 2 handles only `api graphql` calls, so if the main script ever adds a non-GraphQL `gh` invocation this mock would need updating — but that is a future concern, not a defect today.

---

**Verdict:** APPROVE

The changes are narrowly scoped, fail-closed on all token failure modes, silent on normal runs, and covered by regression tests. No blocking risk concerns.

PR #2799, opened 2026-05-18
Make SCP tx-set purge transactional

### LINKED ISSUE #2770: HERDER: Make ScpPersistenceManager::purge_unreferenced_tx_sets transactional
## Context

Follow-up from #2698 (wiring `ScpPersistenceManager` GC timer into production).

`ScpPersistenceManager::purge_unreferenced_tx_sets()` (`crates/herder/src/persistence.rs:495`) performs three separate operations against `SqliteScpPersistence`:

1. `get_all_tx_set_hashes()` — reads all persisted tx-set hashes
2. `load_all_scp_states()` (via the per-state iteration inside `purge_unreferenced_tx_sets`) — reads all persisted SCP states to determine still-referenced hashes
3. `delete_tx_sets_by_hashes(orphans)` — deletes the orphans

Each of these checks out a connection from the pool independently (`db.with_connection(...)`), so they are **not** wrapped in a single SQLite transaction. Today this is harmless because nothing else writes to the persisted SCP state tables. Once `persist_scp_state()` is wired (companion follow-up), a concurrent persist of a brand-new envelope between steps 2 and 3 could see its referenced tx-set deleted as an orphan.

Stellar-core has no equivalent race because all herder work runs on the main thread (HerderImpl.cpp serializes persist and purge naturally).

## Work needed

1. Add a transactional API to `SqliteScpPersistence` (e.g. `purge_unreferenced_tx_sets_atomic(&self) -> Result<()>`) that holds a single connection across read-hashes, read-states, delete-orphans, wrapped in `BEGIN ... COMMIT`.
2. Have `ScpPersistenceManager::purge_unreferenced_tx_sets()` use the new API when the backend supports it (extend the `ScpStatePersistence` trait, or expose a new trait method with a default implementation that falls back to the current non-atomic sequence).
3. Add a regression test: spawn a thread that races `persist_scp_state` against `purge_unreferenced_tx_sets`, assert the persisted tx-set survives.

## References

- Race window noted in TODO comment landed by #2698 (`crates/herder/src/herder.rs::purge_persisted_tx_sets`)
- Parent: #2698
- Blocked on the persist-wiring follow-up (no race exists until persist is wired).

tomerweller:
Closes #2770

## Summary

`ScpPersistenceManager::purge_unreferenced_tx_sets()` historically performed three separate `with_connection` checkouts against `SqliteScpPersistence` (`get_all_tx_set_hashes`, `load_all_scp_states`, `delete_tx_sets_by_hashes`) without a single SQLite transaction. Once `persist_scp_state()` is wired in production (#2768), a concurrent persist between the read-states and delete-orphans steps could have its freshly-inserted tx-set deleted as an orphan.

This change makes the purge atomic at two levels:
1. **SQLite level** — a new `Database::transaction_immediate` helper wraps the three-step purge in a `BEGIN IMMEDIATE` transaction, holding a RESERVED write lock that blocks concurrent writers on other pool connections for the purge's full duration.
2. **Manager level** — `ScpPersistenceManager::purge_unreferenced_tx_sets` now also acquires the same `last_slot_saved` write lock as `persist_scp_state`. This is needed because `persist_scp_state` itself performs two separate storage calls (`save_tx_set` then `save_scp_state`), so a purge interleaved between them would still be able to delete a tx-set whose referencing state has not yet landed. Matches stellar-core, which serializes persist and purge on the main thread (`HerderImpl.cpp`).

The in-memory backend uses the default trait implementation (sequential calls), which is safe because its per-method `parking_lot::RwLock` already serializes.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/2770#issuecomment-4474384283)

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test --all` passes (all crates, including new tests)
- [x] `cargo test -p henyey-db` (109 passed) — includes 4 new `test_purge_unreferenced_tx_sets_atomic_*` tests against the `Connection` impl
- [x] `cargo test -p henyey-herder` (1116 passed) — includes new `sqlite_tests::test_purge_atomic_via_manager_sqlite`
- [x] `cargo test -p henyey-herder --test persistence_race` (2 passed) — new integration test file with structural + race-regression tests

## Deviations from plan

- The plan called for a single SQLite transaction (`BEGIN ... COMMIT`) without specifying the lock mode. Implementation uses `BEGIN IMMEDIATE` (via a new `Database::transaction_immediate` helper) so the purge holds a RESERVED write lock immediately — necessary to block concurrent `save_tx_set` writes on other pool connections.
- The plan focused on purge-side atomicity only. Implementation additionally takes the manager-level `last_slot_saved` write lock inside `purge_unreferenced_tx_sets` to serialize purge with the non-atomic two-step `persist_scp_state`. Without this, the race regression test in the issue is non-deterministic; with it, the test passes consistently. This is a Minor deviation that keeps the plan's intent (matching stellar-core's serialization) while satisfying the issue's explicit "assert the persisted tx-set survives" requirement.
- Added `serde_json` as a `henyey-db` dependency so the db-layer atomic implementation can parse JSON-encoded `PersistedSlotState` without depending on the herder crate.

### DIFF
--- a/crates/db/Cargo.toml
+++ b/crates/db/Cargo.toml
@@ -23,5 +23,12 @@ tracing.workspace = true
 hex.workspace = true
 base64.workspace = true
 
+# JSON decoding of persisted SCP slot states for atomic purge
+# (PersistedSlotState is JSON-encoded by the herder; the atomic purge
+# parses it on the same connection inside a transaction). Only `Value`
+# is used so the herder's `PersistedSlotState` shape doesn't have to be
+# duplicated here.
+serde_json.workspace = true
+
 [dev-dependencies]
 tempfile.workspace = true
--- a/crates/db/src/pool.rs
+++ b/crates/db/src/pool.rs
@@ -100,6 +100,30 @@ impl Database {
         Ok(result)
     }
 
+    /// Executes a closure within a `BEGIN IMMEDIATE` SQLite transaction.
+    ///
+    /// Unlike [`transaction`](Self::transaction) which uses `BEGIN DEFERRED`
+    /// (acquiring a write lock lazily on first write), this method acquires
+    /// the RESERVED write lock up-front. This means concurrent writers on
+    /// other pool connections are blocked for the entire duration of the
+    /// closure — useful when read-then-write sequences must be serialized
+    /// w.r.t. all other writers.
+    ///
+    /// Used by the atomic SCP-tx-set purge (#2770): without an immediate
+    /// lock, a concurrent `save_tx_set` on another connection can interleave
+    /// between the purge's reads and lead to the freshly-inserted tx-set
+    /// being deleted as an orphan.
+    pub fn transaction_immediate<T, F>(&self, f: F) -> Result<T, DbError>
+    where
+        F: FnOnce(&rusqlite::Transaction) -> Result<T, DbError>,
+    {
+        let mut conn = self.connection()?;
+        let tx = conn.transaction_with_behavior(rusqlite::TransactionBehavior::Immediate)?;
+        let result = f(&tx)?;
+        tx.commit()?;
+        Ok(result)
+    }
+
     /// Executes a closure with a database connection.
     ///
     /// This is useful for read operations or simple writes that don't
--- a/crates/db/src/queries/scp.rs
+++ b/crates/db/src/queries/scp.rs
@@ -647,6 +647,117 @@ mod tests {
         assert!(conn.has_tx_set_data(&hash1).unwrap());
     }
 
+    /// Builds a JSON-encoded `PersistedSlotState` whose single envelope
+    /// references `tx_set_hash`. Mirrors the herder-side
+    /// `PersistedSlotState::to_json` output, so the on-disk encoding here
+    /// matches what `ScpPersistenceManager` would write at runtime.
+    fn persisted_state_json_referencing(tx_set_hash: &Hash) -> String {
+        // Construct a NOMINATE envelope referencing `tx_set_hash` via a
+        // StellarValue.
+        let stellar_value = stellar_xdr::curr::StellarValue {
+            tx_set_hash: tx_set_hash.clone(),
+            close_time: stellar_xdr::curr::TimePoint(0),
+            upgrades: vec![].try_into().unwrap(),
+            ext: stellar_xdr::curr::StellarValueExt::Basic,
+        };
+        let value_xdr = stellar_value.to_xdr(Limits::none()).unwrap();
+        let envelope = ScpEnvelope {
+            statement: stellar_xdr::curr::ScpStatement {
+                node_id: NodeId(PublicKey::PublicKeyTypeEd25519(Uint256([0u8; 32]))),
+                slot_index: 100,
+                pledges: stellar_xdr::curr::ScpStatementPledges::Nominate(
+                    stellar_xdr::curr::ScpNomination {
+                        quorum_set_hash: Hash([0u8; 32]),
+                        votes: vec![stellar_xdr::curr::Value(value_xdr.try_into().unwrap())]
+                            .try_into()
+                            .unwrap(),
+                        accepted: vec![].try_into().unwrap(),
+                    },
+                ),
+            },
+            signature: stellar_xdr::curr::Signature::default(),
+        };
+        let env_xdr: Vec<u8> = envelope.to_xdr(Limits::none()).unwrap();
+        // Match the herder's PersistedSlotState JSON shape: serde_json renders
+        // `Vec<u8>` as a JSON array of numbers. Build that shape manually so
+        // we don't need a serde_json dev-dep on the db crate.
+        let env_array = env_xdr
+            .iter()
+            .map(|b| b.to_string())
+            .collect::<Vec<_>>()
+            .join(",");
+        format!(
+            r#"{{"version":1,"envelopes":[[{}]],"quorum_sets":[]}}"#,
+            env_array
+        )
+    }
+
+    #[test]
+    fn test_purge_unreferenced_tx_sets_atomic_basic() {
+        let conn = setup_db();
+
+        let referenced = Hash([0xAA; 32]);
+        let orphan = Hash([0xBB; 32]);
+
+        // Two tx sets persisted: one referenced, one orphan.
+        conn.save_tx_set_data(&referenced, &[1, 2, 3]).unwrap();
+        conn.save_tx_set_data(&orphan, &[4, 5, 6]).unwrap();
+
+        // One SCP state references `referenced`.
+        let state_json = persisted_state_json_referencing(&referenced);
+        conn.save_scp_slot_state(100, &state_json).unwrap();
+
+        // Atomic purge.
+        conn.purge_unreferenced_tx_sets_atomic().unwrap();
+
+        assert!(conn.has_tx_set_data(&referenced).unwrap());
+        assert!(!conn.has_tx_set_data(&orphan).unwrap());
+    }
+
+    #[test]
+    fn test_purge_unreferenced_tx_sets_atomic_empty_is_noop() {
+        let conn = setup_db();
+        // Empty DB — should be a no-op (and not error).
+        conn.purge_unreferenced_tx_sets_atomic().unwrap();
+    }
+
+    #[test]
+    fn test_purge_unreferenced_tx_sets_atomic_all_referenced() {
+        let conn = setup_db();
+
+        let hash = Hash([0xCC; 32]);
+        conn.save_tx_set_data(&hash, &[1]).unwrap();
+        let state_json = persisted_state_json_referencing(&hash);
+        conn.save_scp_slot_state(100, &state_json).unwrap();
+
+        conn.purge_unreferenced_tx_sets_atomic().unwrap();
+        assert!(conn.has_tx_set_data(&hash).unwrap());
+    }
+
+    #[test]
+    fn test_purge_unreferenced_tx_sets_atomic_skips_corrupt_state() {
+        let conn = setup_db();
+
+        let valid_hash = Hash([0xAA; 32]);
+        let orphan_from_corrupt = Hash([0xBB; 32]);
+        conn.save_tx_set_data(&valid_hash, &[1]).unwrap();
+        conn.save_tx_set_data(&orphan_from_corrupt, &[2]).unwrap();
+
+        // Valid state references valid_hash.
+        let valid_state = persisted_state_json_referencing(&valid_hash);
+        conn.save_scp_slot_state(100, &valid_state).unwrap();
+
+        // Corrupt JSON for slot 101 — invalid bytes that won't deserialize.
+        conn.save_scp_slot_state(101, "{not valid json").unwrap();
+
+        // Should NOT panic; corrupt state is logged + skipped. valid_hash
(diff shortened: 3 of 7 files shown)


tomerweller:
## 🔍 Reviewer: Parity

**PR:** #2799 — Make SCP tx-set purge transactional
**Issue:** #2770
**Scope:** `crates/herder/src/persistence.rs`, `crates/db/src/queries/scp.rs`, `crates/db/src/pool.rs`, `crates/db/src/scp_persistence.rs`, `crates/herder/tests/persistence_race.rs`
**Reference:** `stellar-core/src/herder/HerderImpl.cpp` + `stellar-core/src/main/PersistentState.cpp` (v26.0.1)

---

### Summary

The PR correctly identifies and fixes a race that doesn't exist in stellar-core because all herder work runs on the main thread. The two-level fix (manager-level `last_slot_saved` write lock + `BEGIN IMMEDIATE` SQLite transaction) is a sound adaptation of stellar-core's implicit single-thread serialization to a multi-threaded Rust context.

---

### Parity Analysis

#### 1. Purge logic — `HerderImpl::purgeOldPersistedTxSets()` (HerderImpl.cpp:2448–2487)

Stellar-core semantics:
1. Read all tx-set hashes into `hashesToDelete`.
2. For each slot state: decode, extract envelope hashes, erase from `hashesToDelete`.
3. Per-state catch: on error, `CLOG_ERROR` and continue (the state's hashes are **not** erased, so they get deleted as orphans).
4. Call `deleteTxSets(hashesToDelete)` — inside its own `soci::transaction`.

Henyey's new `purge_unreferenced_tx_sets_atomic`:
- Steps 1–3 mirror this exactly.
- On corrupt state: `tracing::warn!` + skip (state's tx-sets treated as orphans and deleted). **Matches upstream.**
- Note: the old henyey code docstring said "preserving [corrupt state's] tx set hashes as potentially referenced." The new code correctly changes this to "may be deleted as orphans." This is the right parity fix — the old docstring was wrong.

✅ Purge semantics match stellar-core.

#### 2. `setSCPStateV1ForSlot` atomicity (PersistentState.cpp:234–248)

Stellar-core wraps tx_set + scp_state writes in a **single** `soci::transaction`, so persist is atomic and a purge can never see a tx-set without its referencing state. Henyey's `persist_scp_state` still issues `save_tx_set` then `save_scp_state` as two separate calls. This pre-existing non-atomicity is the exact reason the manager-level `last_slot_saved` write lock is required in this PR.

The manager-level lock is a correct substitute for stellar-core's main-thread serialization: `persist_scp_state` holds the write lock for its entire duration (including both storage calls), and `purge_unreferenced_tx_sets` now also acquires that same lock. This means a purge can never interleave between the two persist storage calls.

✅ Serialization semantics match stellar-core's effective behavior.

#### 3. `BEGIN IMMEDIATE` vs stellar-core's session-level transactions

Stellar-core uses a single session for `getTxSetHashesForAllSlots`, `getSCPStateAllSlots`, and `deleteTxSets` in `purgeOldPersistedTxSets`. These calls all hit the same SOCI `MiscSession`, which in practice means no cross-connection interleaving. Henyey's `transaction_immediate` (RESERVED lock held from first statement) is the correct SQLite analogue for preventing concurrent writers on other pool connections from interleaving between the read and delete steps.

✅ Lock semantics are equivalent.

#### 4. Envelope hash extraction — `getValidatedTxSetHashes` vs `get_tx_set_hashes`

Stellar-core's `BallotProtocol::getStatementValues` (BallotProtocol.cpp:1930–1946) has a guard:

```cpp
if (b.counter != 0) {
    values.insert(prep.ballot.value);
}
```

Henyey's `get_tx_set_hashes` (persistence.rs:290–303) and the new db-layer `extract_tx_set_hashes_from_envelope` (scp.rs:194–220 of the diff) both unconditionally process `prep.ballot.value` without checking `counter != 0`. **This is a pre-existing parity gap, not introduced by this PR.** In real SCP operation, a Prepare message with counter=0 is not valid and would not be stored, so the gap is unlikely to cause observable divergence in practice — but it should be tracked.

⚠️ Pre-existing parity gap (not introduced here): missing `counter != 0` guard in Prepare branch of both `get_tx_set_hashes` and the new `extract_tx_set_hashes_from_envelope`. The new db-layer function duplicates the gap. Recommend filing a follow-up issue.

#### 5. Nominate deduplication

Stellar-core's `NominationProtocol::applyAll` deduplicates: it only processes `accepted[i]` if it doesn't appear in `votes`. Henyey uses `.chain()` without deduplication, but does a `sort+dedup` at the end of `get_tx_set_hashes`. The db-layer `extract_tx_set_hashes_from_envelope` doesn't deduplicate, but deduplication at the hash-set level (`HashSet`) happens implicitly in the caller. Either way, the final set of referenced hashes is correct.

✅ No functional divergence on Nominate.

#### 6. `serde_json` dependency in `henyey-db`

The db crate now owns JSON parsing logic that duplicates the herder's `PersistedSlotState` encoding shape. The comment acknowledges this and the test fixture validates the round-trip. The coupling is limited to `serde_json::Value` (untyped), which is acceptable. The regression tests in `crates/db/src/queries/scp.rs` use `persisted_state_json_referencing` which constructs the exact same JSON as the herder — good coverage of the interface contract.

✅ Acceptable; the encoding contract is tested.

---

### Minor Observations (non-blocking)

- The `test_purge_atomic_survives_concurrent_persist` race test uses `thread::sleep(Duration::from_micros(50))` in the persister loop, which gives the race test a probabilistic rather than deterministic character. This is documented in the test comment and the structural test (`test_purge_atomic_on_sqlite_backend`) provides the deterministic guard. Acceptable.
- The db-layer `purge_unreferenced_tx_sets_atomic` is called on a `Connection` (within a transaction closure) and then the herder's `SqliteScpPersistence::purge_unreferenced_tx_sets_atomic` wraps it in `transaction_immediate`. The two-level abstraction is sound; the inner call works correctly on either a bare `Connection` or a `Transaction` since both implement the same query methods.

---

### Verdict

**Verdict:** APPROVE

The core fix is correct and matches stellar-core's serialization intent. The one gap (missing `counter != 0` guard in Prepare) is pre-existing and not introduced by this PR. I recommend filing a follow-up issue for it, but it is not a blocker for merge.

tomerweller:
## 🔍 Reviewer: Correctness

Reviewing PR #2799: Make SCP tx-set purge transactional (closes #2770).

---

### Test coverage — regression test commit fails on origin/main, fix commit passes

**Verified.** The regression-test commit (`b989ce62`) adds `persistence_race.rs` calling `purge_unreferenced_tx_sets_atomic()` on `ScpStatePersistence`. On `origin/main` this fails to compile:

```
error[E0599]: no method named `purge_unreferenced_tx_sets_atomic` found for struct
`henyey_herder::SqliteScpPersistence` in the current scope
```

The fix commit (`276a0987`) makes all tests pass: `105 passed` in `henyey-db`, `1115 passed` in `henyey-herder` (including `test_purge_atomic_via_manager_sqlite`), and both integration tests in `persistence_race.rs` pass:

- `test_purge_atomic_on_sqlite_backend` — structural correctness of the new API
- `test_purge_atomic_survives_concurrent_persist` — race regression guard

---

### `BEGIN IMMEDIATE` transaction semantics

**Correct.** `transaction_immediate` calls `conn.transaction_with_behavior(TransactionBehavior::Immediate)` which issues `BEGIN IMMEDIATE`, acquiring SQLite's RESERVED write lock immediately. In WAL mode (confirmed configured via `PRAGMA journal_mode = WAL`):

- The purge holds RESERVED for its entire read-then-delete sequence.
- A concurrent `save_tx_set` (auto-commit write) will try to acquire its own RESERVED lock and block until the purge commits — eliminating the race window.
- Readers are not blocked (WAL allows concurrent reads under RESERVED).

The rollback-on-error path is correct: if the closure returns `Err`, the `?` propagates before `tx.commit()`, so `tx` drops and rolls back automatically (same pattern as the existing `transaction` method).

One minor comment inaccuracy: the early-return check `if all_hashes_vec.is_empty()` has a comment "avoid a needless write lock when caller wraps this in a transaction." This is misleading because `transaction_immediate` already acquired the RESERVED lock before calling the closure. The lock cannot be avoided by this early return. Not a correctness issue, just a documentation nit.

---

### Hash extraction from envelope XDR

**Correct.** `extract_tx_set_hashes_from_envelope` in the db-layer is a faithful duplicate of `get_tx_set_hashes` in the herder layer. Both handle all four SCP statement pledge types:

- `Nominate`: `votes` + `accepted`
- `Prepare`: `ballot.value`, `prepared.value` (optional), `prepared_prime.value` (optional)
- `Confirm`: `ballot.value`
- `Externalize`: `commit.value`

The JSON-to-bytes path is correct: `PersistedSlotState` uses `#[derive(Serialize, Deserialize)]` with `envelopes: Vec<Vec<u8>>`. `serde_json` serializes `Vec<u8>` as a JSON number array `[n1, n2, ...]` by default (no `serde_bytes` attribute). The db-layer parser reads the array the same way. The test helper `persisted_state_json_referencing` manually builds `[[n1, n2, ...]]` which matches exactly. Roundtrip is correct.

---

### Default `purge_unreferenced_tx_sets_atomic` impl for `InMemoryScpPersistence`

**Correct with caveat noted and accepted.** The trait default calls three sequential methods (get hashes, load states, delete orphans) on the `InMemoryScpPersistence` backend. Each method acquires its own per-field `RwLock` independently, so the default is not transactional across all three calls.

The PR doc acknowledges this: "safe for in-process backends whose individual calls already serialize on a single lock." Strictly, `InMemoryScpPersistence` uses *two* separate locks (`states` and `tx_sets`), not one, so the safety claim is only valid for single-threaded use. Since `InMemoryScpPersistence` is used exclusively in unit tests, and unit tests don't run `persist_scp_state` concurrently against `purge_unreferenced_tx_sets`, this is acceptable. No production safety risk.

---

### Manager-level `last_slot_saved` lock serialization

**Correct.** `purge_unreferenced_tx_sets` acquires `self.last_slot_saved.write()` before delegating to the storage backend. `persist_scp_state` also acquires `self.last_slot_saved.write()` at entry. This serializes purge and persist at the manager level, closing the second race window: `persist_scp_state` does two separate storage calls (`save_tx_set` then `save_scp_state`), so a purge interleaved between them could see the tx-set but not the referencing state. The write lock ensures this cannot happen.

No deadlock risk: `cleanup()` calls `purge_unreferenced_tx_sets()` but does not hold `last_slot_saved` when it does so.

---

### Error handling behavior (corrected docstring)

The old `purge_unreferenced_tx_sets` docstring said: "if a state fails to decode, we skip it (preserving its tx set hashes as potentially referenced)." This was incorrect — the old code, like the new code, does *not* add hashes from corrupt states to `referenced`, so those tx-sets are deleted. The new docstring is accurate ("its tx-sets may be deleted as orphans"), and matches stellar-core's `purgeOldPersistedTxSets` behavior (confirmed in `HerderImpl.cpp:2456-2478`).

---

**Verdict:** APPROVE

The implementation is correct. `BEGIN IMMEDIATE` semantics properly block concurrent writers. Hash extraction is a faithful duplicate with all statement types covered. The manager-level lock closes the two-step persist race. Tests follow the regression-first pattern and are structurally definitive. The one nit (misleading comment about early-return avoiding the write lock) is documentation-only and does not affect behavior.

PR #2832, opened 2026-05-19
Fix /review-pr silent exit when reviewer sub-agents skip verdict post

### LINKED ISSUE #2813: /review-pr reviewer sub-agents skip the structured verdict post
## Symptom
On PR #2812 (issue #2810), /review-pr has been dispatched 5+ times over 70+ minutes. Each tick the reviewer sub-agents do substantial analysis (reading code, running adjacent tests) but exit with traces like:

> I've reached the final verdict. I'm marking the remaining review steps done internally, then I'll return the fix review report.

…without actually posting the `## 🔍 Reviewer: Correctness` / `Parity` structured comment that /review-pr's Step 6 matrix consumes. Result: 0 bounce-back cycles recorded, issue stays in `in-review` forever, both loops keep re-picking and re-spawning expensive reviewer pairs.

## Impact
- High token cost (each cycle is ~2.5M tokens / 12 min wall-clock)
- Blocks the head-scoped bounce-cap from firing (the cap reads bounce comments)
- Blocks the lifetime cap from firing for the same reason
- Two-loop deployments compound the waste

## Reproduction
Watching tick log /home/tomer/data/project-tick-loop-b/20260519-015141-tick.log — the reviewer sub-agent ran for 12 min, marked TODOs complete, never posted a comment to PR #2812.

## Likely root cause
Reviewer brief in .github/skills/review-pr/SKILL.md Step 4 ("Reviewer A — Correctness", "Reviewer B — …") instructs to post the structured verdict, but the agent paraphrases the instruction during sub-agent dispatch and treats the verdict as an internal artifact instead of a side-effecting GitHub comment. Same paraphrase pattern as #2787.

## Suggested fix
Either:
1. Have /review-pr verify (post-sub-agent) that the structured verdict comment exists; if not, treat as a CHANGES_REQUESTED bounce with reason "reviewer sub-agent skipped verdict post". This converts the silent failure into an explicit bounce that the lifetime cap can catch.
2. Move the verdict-post out of the sub-agent's responsibility — have /review-pr's parent post the structured comment based on the sub-agent's free-form return.

Option 1 is cheaper to implement.

tomerweller:
Closes #2813

## Summary

Makes /review-pr deterministically verify that each reviewer sub-agent posts a fresh structured verdict comment after dispatch, retries one missing or malformed reviewer once, and emits explicit bounce artifacts (## Review: Bounce-Back Cycle or ## Review: Malformed Verdict) on persistent failure instead of silently exiting.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/2813#issuecomment-4486911909)

## Test plan

- [x] `scripts/test-review-pr-skill-snippets.sh` — all 13 TAP tests pass
- [x] `scripts/test-monitor-skill-snippets.sh` — pre-existing tests unaffected (4 pre-existing label-sync failures, not related)
- [x] cargo fmt --check

## Regression test

- **Test:** `scripts/test-review-pr-skill-snippets.sh` — TAP cases 1-5
- **Pre-fix:** library doesn't exist on `origin/main` → test script fails to source
- **Post-fix:** all classification and cutoff-filtering tests pass after `1e2d52ec`

## Deviations from plan

- Step 2b.1 cutoff concern: The bounce-cap-check.sh script only counts bounce comments (not reviewer verdicts), so no separate cutoff wiring needed there. The fix in Step 6.1 (using REVIEW_START_ISO as cutoff) is sufficient to prevent stale verdict reuse in the decision matrix.

### DIFF
--- a/.github/skills/review-pr/SKILL.md
+++ b/.github/skills/review-pr/SKILL.md
@@ -146,6 +146,12 @@ Otherwise, the PR is **non-parity** — Reviewer B uses risk lens.
 
 Launch both as `general-purpose` foreground sub-agents. Do not wait between them. **Each reviewer must be spawned with `--model gpt-5.4`** (or equivalent model parameter) explicitly — do not inherit from the parent. Cross-model diversity catches issues a same-model pipeline would miss.
 
+**Record the review start time** before dispatching reviewers — this is the freshness cutoff for verdict verification:
+
+```bash
+REVIEW_START_ISO=$(date -u +%Y-%m-%dT%H:%M:%SZ)
+```
+
 **Why structured comments, not `gh pr review --approve`:** the authenticated GH user is the PR author (the same user opened the PR via `/do` and now reviews it). GitHub disallows author self-approval, so `gh pr review --approve` is silently downgraded to a comment by `gh`. Instead, each reviewer posts a structured comment with a verdict marker that `/review-pr` parses in Step 6.
 
 **Verdict comment format** — each reviewer MUST post exactly one comment with this exact shape:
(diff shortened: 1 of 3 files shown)


tomerweller:
I found 2 blocking issues:

1. `scripts/lib/review-pr-verdicts.sh:35-55` — `gh api ... --paginate` emits one JSON document per page, but `fetch_reviewer_verdict_comments` pipes the concatenated stream straight into `jq`. On PRs with more than one page of issue comments, that returns one filtered array per page instead of a single merged array, and `latest_reviewer_verdict_state` / `classify_reviewer` then operate page-by-page. In practice we can reuse an older verdict from page 1 even when a newer page 2 verdict exists, or concatenate multiple bodies together. Long-lived review cycles can easily cross 30 comments, so this reintroduces stale-verdict reuse right on the path this PR is trying to harden. Please `--slurp` or otherwise flatten the paginated responses before filtering, and add a regression fixture with two pages.

2. `.github/skills/review-pr/SKILL.md:274-314` — the malformed-verdict branch says it should “emit an explicit audit artifact and bounce”, but the snippet only posts `## Review: Malformed Verdict`. Unlike the missing-verdict branch, it never actually moves the issue back to `ready-for-doing`, removes the assignee, or exits. An agent following this literally can still leave a malformed-verdict run stuck in `in-review`, which is the same state-machine problem this PR is meant to eliminate. The malformed path needs the same bounce/unassign/exit sequence after posting the audit note.

PR #2835, opened 2026-05-19
Implement ctValidityOffset trigger-time and far-ahead abort semantics

### LINKED ISSUE #2816: HERDER spec adherence: Align ctValidityOffset trigger-time and abort semantics
## Context

Narrowed out of parent issue #2803 during `/plan` round 2.

#2803 is being reduced to the future-slot timer parity fix only. This follow-up
carries the close-time parity work around `ctValidityOffset`, trigger-time
adjustment, and the far-ahead nomination abort path.

Related prior work: #2702 covers broader event-driven trigger scheduling and
timer ownership changes; this issue is the spec-adherence slice that must
converge HERDER_SPEC §5.1-2 / §5.2-4 with stellar-core.

## Work needed

1. Map stellar-core's `HerderImpl::ctValidityOffset`,
   `setupTriggerNextLedger`, and `triggerNextLedger` onto henyey's current
   `app` / `herder` split.
2. Implement the trigger-time adjustment equivalent to
   `ctValidityOffset(minCandidateCt, triggerOffset)` so the timer is delayed
   before nomination when the candidate close time would otherwise be too far
   ahead.
3. Implement the far-ahead abort / skip path for nomination when the chosen
   `nextCloseTime` is still invalid after adjustment, rather than relying only
   on the monotonic clamp in `build_nomination_value`.
4. Add focused tests for both the trigger-time adjustment and the far-ahead
   abort behavior.
5. Update `crates/herder/PARITY_STATUS.md` and any nearby parity comments once
   the behavior is settled.

## References

- Parent: #2803
- Related broader scheduling work: #2702
- stellar-core: `stellar-core/src/herder/HerderImpl.cpp`
- henyey: `crates/herder/src/herder.rs`, `crates/app/src/app/consensus.rs`

tomerweller:
Closes #2816

## Summary

Implements spec-compliant trigger-time adjustment and far-ahead nomination abort behavior matching stellar-core's `HerderImpl::ctValidityOffset` and `triggerNextLedger` abort path (HERDER_SPEC §5.1-2 / §5.2-4). The app-side polling loop now delays triggers until `prepareStart + expectedClose + ctValidityOffset` says the proposed close time is valid, and the herder refuses to build/nominate when the close time exceeds `now + MAX_TIME_SLIP_SECONDS`.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/2816#issuecomment-4487803489)

## Test plan

- [x] cargo fmt --check
- [x] cargo test -p henyey-herder --lib (1128 tests pass)
- [x] cargo test -p henyey-app --lib (941 tests pass)

## New tests

- `test_trigger_next_ledger_skips_far_ahead_close_time` — verifies `SkippedInvalidCloseTime` when LCL close time is far ahead
- `test_handle_nomination_timeout_far_ahead_cache_miss_returns_noop_without_state` — verifies cache-miss rebuild abort
- `test_ct_validity_offset_zero_at_validity_boundary` — boundary behavior
- `test_ct_validity_offset_positive_one_second_past_boundary` — one-past boundary
- `test_ct_validity_offset_trigger_offset_widens_window` — trigger offset expands window
- `test_herder_prepare_start_forwards_driver_timing` — public API forwarding
- `test_prepare_start_returns_none_before_ballot_start` — None shape
- `test_prepare_start_returns_recorded_ballot_start` — Some shape
- `test_prepare_start_purged_with_slot_trim` — purge removes entry

## Deviations from plan

- Tests are in a single commit with the implementation rather than a separate failing-test-first commit (feature kind: the new public API doesn't exist yet on main, so tests can't meaningfully fail differently than "unresolved function")
- Used `ScpDriver::set_test_clock()` (`pub(crate)`, `#[cfg(test)]`) rather than `Herder::set_test_clock_seconds()` (which requires `feature = "test-support"` unavailable in same-crate unit tests)

### DIFF
--- a/crates/herder/src/herder.rs
+++ b/crates/herder/src/herder.rs
@@ -2630,8 +2630,21 @@ impl Herder {
             // Parity: stellar-core HerderImpl.cpp:1546-1625 — observers build
             // the candidate tx set, self-validate, and publish it into the
             // pending/fetch cache, then stop before `makeStellarValue`/`nominate`.
-            self.build_and_cache_candidate_tx_set(next_close_time)
-                .ok_or_else(|| HerderError::Internal("Failed to build candidate tx set".into()))?;
+            //
+            // A None result here means the candidate build aborted on a benign
+            // retryable condition (LCL advanced concurrently, or the defensive
+            // close-time gate rejected). Map it to SkippedStale rather than a
+            // hard error so callers retry on the next tick (review feedback).
+            if self
+                .build_and_cache_candidate_tx_set(next_close_time)
+                .is_none()
+            {
+                tracing::warn!(
+                    slot,
+                    "trigger_next_ledger: candidate tx-set build returned None (stale race)"
+                );
+                return Ok(TriggerOutcome::SkippedStale);
+            }
             let build_value_ms = t0.elapsed().as_millis();
 
             // Drain ready fetching envelopes after cache publication.
@@ -2659,9 +2672,19 @@ impl Herder {
         }
 
         // --- Validator path ---
-        let value = self
-            .build_nomination_value(next_close_time)
-            .ok_or_else(|| HerderError::Internal("Failed to build nomination value".into()))?;
+        let value = match self.build_nomination_value(next_close_time) {
+            Some(v) => v,
+            None => {
+                // build_nomination_value returns None when LCL advanced
+                // concurrently (stale close_time race). This is benign and
+                // retryable — map to SkippedStale rather than a hard error.
+                tracing::warn!(
+                    slot,
+                    "trigger_next_ledger: build_nomination_value returned None (stale race)"
+                );
+                return Ok(TriggerOutcome::SkippedStale);
+            }
+        };
         let build_value_ms = t0.elapsed().as_millis();
 
         // build_nomination_value() caches the tx set but no longer drains
@@ -6864,6 +6887,89 @@ mod tests {
         assert!(
             result.is_none(),
             "build_nomination_value must abort when close_time <= snapshot lcl_close_time"
+        );
+    }
+
+    /// Regression test for #2816 Correctness review round 2:
+    /// `trigger_next_ledger()` must return `Ok(SkippedStale)` (not a hard
+    /// `Err(Internal)`) when `build_nomination_value()` returns None.
+    /// Simulates the race by resetting the LedgerManager (making
+    /// `create_snapshot()` fail with NotInitialized) after the entry guards
+    /// pass but before the snapshot is taken inside `build_nomination_value`.
+    #[test]
+    fn test_trigger_next_ledger_returns_skipped_stale_on_build_none() {
+        let now = 5000u64;
+        let seed = [70u8; 32];
+        let secret = SecretKey::from_seed(&seed);
+        let public = secret.public_key();
+        let local_node_id =
+            stellar_xdr::curr::NodeId(stellar_xdr::curr::PublicKey::PublicKeyTypeEd25519(
+                stellar_xdr::curr::Uint256(*public.as_bytes()),
+            ));
+        let quorum_set = ScpQuorumSet {
+            threshold: 1,
+            validators: vec![local_node_id].try_into().unwrap(),
+            inner_sets: vec![].try_into().unwrap(),
+        };
+        let config = HerderConfig {
+            is_validator: true,
+            node_public_key: public,
+            local_quorum_set: Some(quorum_set),
+            ..HerderConfig::default()
+        };
+
+        let lm = make_lm_with_close_time(0, now);
+        let herder =
+            Herder::with_secret_key(config, secret, lm.clone(), TimerManagerHandle::no_op());
+        herder.scp_driver.set_test_clock(now);
+        herder.bootstrap(0);
+
+        // Reset the LedgerManager so `create_snapshot()` returns Err(NotInitialized).
+        // Entry guards still pass: lcl_matches_slot(1) checks current_ledger_seq()
+        // which reads from the reset genesis header (seq=0), so 0+1=1 matches.
+        // But build_nomination_value → create_snapshot → Err → returns None.
+        lm.reset();
+
+        let result = herder.trigger_next_ledger(1);
+        assert_eq!(
+            result.expect(
+                "must not be a hard error — the old code erroneously returned Err(Internal)"
+            ),
+            TriggerOutcome::SkippedStale,
+            "trigger_next_ledger must return SkippedStale when build_nomination_value returns None"
+        );
+    }
+
+    /// Regression test for #2816 Correctness review round 2:
+    /// `ScpDriver::slot_trigger_time()` returns the first_seen instant for
+    /// a slot after `record_slot_activity` is called.
+    #[test]
+    fn test_slot_trigger_time_returns_recorded_instant() {
+        let (herder, _) = make_validator_herder();
+        herder.bootstrap(0);
+
+        // Before recording activity, slot_trigger_time should be None.
+        assert!(
+            herder.scp_driver.slot_trigger_time(1).is_none(),
+            "slot_trigger_time must be None before record_slot_activity"
+        );
+
+        // Record activity for slot 1.
+        herder.scp_driver.record_slot_activity(1);
+
+        // After recording, slot_trigger_time should be Some.
+        let trigger_time = herder.scp_driver.slot_trigger_time(1);
+        assert!(
+            trigger_time.is_some(),
+            "slot_trigger_time must be Some after record_slot_activity"
+        );
+
+        // The recorded time should be very recent (within 1 second of now).
+        let elapsed = trigger_time.unwrap().elapsed();
+        assert!(
+            elapsed.as_secs() < 1,
+            "slot_trigger_time should be very recent, got {:?} ago",
+            elapsed
         );
     }
 
--- a/crates/herder/src/herder.rs
+++ b/crates/herder/src/herder.rs
@@ -1324,6 +1324,10 @@ impl Herder {
     /// Returns the number of seconds we must wait before `close_time` falls
     /// within the `now + MAX_TIME_SLIP_SECONDS` validity window.
     /// Returns 0 when `close_time` is already valid.
+    ///
+    /// Stellar-core adds +1ms to the result as a boundary guard (ensuring the
+    /// timer fires strictly after the boundary). In henyey's second-granularity
+    /// polling architecture, we add +1 second to be conservative.
     pub fn ct_validity_offset(&self, close_time: u64, max_ct_offset: u64) -> u64 {
         let now = self.scp_driver.now_seconds();
         let boundary = now
@@ -1332,7 +1336,8 @@ impl Herder {
         if close_time <= boundary {
             0
         } else {
-            close_time - boundary
+            // +1 second boundary guard (parity: stellar-core adds +1ms)
+            (close_time - boundary).saturating_add(1)
         }
     }
 
@@ -3348,7 +3353,23 @@ impl Herder {
 
         // 2. Caller supplied a validated close time. Use its offset so tx-set
         // trimming still sees the proposed close time (#1192).
-        let close_time_offset = close_time - lcl_close_time;
+        // Use checked subtraction: if LCL advanced concurrently (race between
+        // the caller's lcl_close_time() read and our snapshot), abort nomination
+        // rather than wrapping to a huge offset.
+        let close_time_offset = match close_time.checked_sub(lcl_close_time) {
+            Some(0) | None => {
+                // close_time <= lcl_close_time from snapshot: LCL advanced
+                // between caller's read and our snapshot, making the proposed
+                // close time stale. Abort — caller will retry next tick.
+                tracing::warn!(
+                    close_time,
+                    lcl_close_time,
+                    "build_nomination_value: close_time stale vs snapshot LCL; aborting"
+                );
+                return None;
+            }
+            Some(offset) => offset,
+        };
 
         // 2.5. Close-time validity gate (parity: stellar-core HerderImpl.cpp:1524).
         // stellar-core calls `ctValidityOffset(nextCloseTime)` which checks
@@ -6727,7 +6748,7 @@ mod tests {
     }
 
     /// Unit test for #2816: `ct_validity_offset` returns positive offset
-    /// one second past the boundary.
+    /// one second past the boundary (includes +1 boundary guard).
     #[test]
     fn test_ct_validity_offset_positive_one_second_past_boundary() {
         let (herder, _) = make_validator_herder();
@@ -6737,7 +6758,11 @@ mod tests {
         // close_time = now + MAX_TIME_SLIP_SECONDS + 1 = one past boundary with offset=0
         let close_time = now + MAX_TIME_SLIP_SECONDS + 1;
         let offset = herder.ct_validity_offset(close_time, 0);
-        assert_eq!(offset, 1, "one second past boundary → need 1 second delay");
+        // Parity: stellar-core returns (1s + 1ms), in whole seconds = 2
+        assert_eq!(
+            offset, 2,
+            "one second past boundary + 1s guard → need 2 seconds delay"
+        );
     }
 
     /// Unit test for #2816: `ct_validity_offset` accounts for max_ct_offset
@@ -6754,9 +6779,9 @@ mod tests {
         let offset = herder.ct_validity_offset(close_time, 5);
         assert_eq!(offset, 0, "trigger offset should widen validity window");
 
-        // close_time = now + 66, trigger_offset = 5 → boundary = now+65, need 1s
+        // close_time = now + 66, trigger_offset = 5 → boundary = now+65, need 1s + 1 guard = 2
         let offset2 = herder.ct_validity_offset(now + 66, 5);
-        assert_eq!(offset2, 1);
+        assert_eq!(offset2, 2);
     }
 
     /// Unit test for #2816: `Herder::prepare_start` forwards to the driver.
@@ -6785,6 +6810,79 @@ mod tests {
             herder.prepare_start(1).is_none(),
             "prepare_start must return None after slot is purged"
         );
+    }
+
+    /// Regression test for #2816 Parity review: when `prepareStart + expectedClose <= now`
+    /// (trigger_time is clamped to `now`), the triggerOffset passed to `ct_validity_offset`
+    /// must be 0 (remaining time until trigger = 0), NOT `expectedClose` (elapsed time).
+    /// This ensures we don't widen the validity window by `expectedClose` seconds.
+    ///
+    /// Concrete case: now=1000, expectedClose=5s, lcl.close_time=1070.
+    /// minCandidateCt=1071. With triggerOffset=0: boundary=1060, offset=12 (blocks).
+    /// With the OLD (wrong) triggerOffset=5: boundary=1065, offset=7 (would fire too early).
+    #[test]
+    fn test_ct_validity_offset_clamped_trigger_time_gives_zero_offset() {
+        let (herder, _) = make_validator_herder();
+        let now = 1000u64;
+        herder.scp_driver.set_test_clock(now);
+
+        let lcl_close_time = 1070u64;
+        let min_candidate_ct = lcl_close_time + 1; // 1071
+
+        // Correct behavior: triggerOffset = 0 (trigger_time clamped to now)
+        let offset = herder.ct_validity_offset(min_candidate_ct, 0);
+        // boundary = 1000 + 60 + 0 = 1060, excess = 1071 - 1060 = 11, +1 guard = 12
+        assert_eq!(offset, 12, "clamped trigger must not widen validity window");
+
+        // Wrong behavior would pass expectedClose=5 as offset:
+        let wrong_offset = herder.ct_validity_offset(min_candidate_ct, 5);
+        // boundary = 1000 + 60 + 5 = 1065, excess = 1071 - 1065 = 6, +1 guard = 7
+        assert_eq!(
+            wrong_offset, 7,
+            "wider window from elapsed time = incorrect"
+        );
+
+        // The correct offset (12) must be larger than the wrong one (7),
+        // proving we'd wait longer and not nominate too early.
+        assert!(offset > wrong_offset);
+    }
+
+    /// Regression test for #2816 Correctness review: `build_nomination_value`
+    /// must not underflow when the snapshot's LCL has advanced past the
+    /// caller-provided close_time. Instead it returns None (abort).
+    #[test]
+    fn test_build_nomination_value_aborts_on_stale_close_time() {
+        let (herder, _) = make_validator_herder();
+        herder.bootstrap(0);
+        let now = 5000u64;
+        herder.scp_driver.set_test_clock(now);
+
+        // The default LM has lcl_close_time = 0.
+        // Pass close_time = 0 to trigger the checked_sub guard (offset = 0 → abort).
+        let lcl_ct = herder.lcl_close_time();
+        let result = herder.build_nomination_value(lcl_ct);
+        assert!(
+            result.is_none(),
+            "build_nomination_value must abort when close_time <= snapshot lcl_close_time"
+        );
+    }
+
+    /// Regression test for #2816 Correctness review: `lcl_close_time()` is
+    /// a public accessor that reads the current LCL close time.
+    #[test]
+    fn test_lcl_close_time_accessor() {
+        let now = 3000u64;
+        let herder = Herder::with_secret_key(
+            HerderConfig {
+                is_validator: true,
+                ..HerderConfig::default()
+            },
+            SecretKey::from_seed(&[60u8; 32]),
+            make_lm_with_close_time(0, now),
+            TimerManagerHandle::no_op(),
+        );
+        herder.bootstrap(0);
+        assert_eq!(herder.lcl_close_time(), now);
     }
 
     /// Direct regression test for the `lcl_matches_slot` helper used by the
--- a/crates/herder/src/herder.rs
+++ b/crates/herder/src/herder.rs
@@ -181,6 +181,10 @@ pub enum TriggerOutcome {
     /// `HerderImpl::triggerNextLedger` lines 1546-1625 — observers stop
     /// after `addTxSet` / cache publication and before `makeStellarValue`.
     ObserverCached,
+    /// The proposed close time is too far ahead of the wall clock; nomination
+    /// skipped to avoid building an invalid StellarValue.
+    /// Parity: stellar-core triggerNextLedger far-ahead abort.
+    SkippedInvalidCloseTime,
 }
 
 /// Intermediate result from [`Herder::build_and_cache_candidate_tx_set`]
@@ -206,6 +210,8 @@ pub enum TimeoutOutcome {
     /// LCL advanced during build/drain; the value is for an already-closed
     /// slot and was not broadcast.
     SkippedStale,
+    /// Close time still invalid on cache-miss rebuild; nomination skipped.
+    SkippedInvalidCloseTime,
     /// No-op: build returned None, or `scp.nominate_timeout` was a no-op
     /// (e.g., slot already externalized).
     NoOp,
@@ -1299,6 +1305,35 @@ impl Herder {
     /// to `u64` is lossless.
     pub fn ledger_close_duration(&self) -> Duration {
         self.ledger_manager.expected_ledger_close_duration()
+    }
+
+    /// Get the ballot-start instant for the given slot.
+    /// Parity: stellar-core `mLastTrigger` (HerderImpl.cpp), exposed as "prepare start".
+    pub fn prepare_start(&self, slot: SlotIndex) -> Option<std::time::Instant> {
+        self.scp_driver.prepare_start(slot)
+    }
+
+    /// Get the LCL close time from the current ledger header.
+    pub fn lcl_close_time(&self) -> u64 {
+        self.ledger_manager.current_header().scp_value.close_time.0
+    }
+
+    /// Compute the close-time validity offset.
+    ///
+    /// Parity: stellar-core `HerderImpl::ctValidityOffset(lastCtTrigger, ct)`.
+    /// Returns the number of seconds we must wait before `close_time` falls
+    /// within the `now + MAX_TIME_SLIP_SECONDS` validity window.
+    /// Returns 0 when `close_time` is already valid.
+    pub fn ct_validity_offset(&self, close_time: u64, max_ct_offset: u64) -> u64 {
+        let now = self.scp_driver.now_seconds();
+        let boundary = now
+            .saturating_add(MAX_TIME_SLIP_SECONDS)
+            .saturating_add(max_ct_offset);
+        if close_time <= boundary {
+            0
+        } else {
+            close_time - boundary
+        }
     }
 
     /// Get the maximum size of a transaction set (ops).
@@ -2564,6 +2599,20 @@ impl Herder {
             return Ok(TriggerOutcome::SkippedStale);
         }
 
+        let now_secs = self.scp_driver.now_seconds();
+        let lcl_close_time = self.lcl_close_time();
+        let next_close_time = now_secs.max(lcl_close_time.saturating_add(1));
+        if next_close_time > now_secs.saturating_add(MAX_TIME_SLIP_SECONDS) {
+            tracing::info!(
+                slot,
+                next_close_time,
+                now = now_secs,
+                max_slip = MAX_TIME_SLIP_SECONDS,
+                "Skipping nomination: proposed close time too far ahead"
+            );
+            return Ok(TriggerOutcome::SkippedInvalidCloseTime);
+        }
+
         tracing::debug!("Triggering consensus for ledger {}", ledger_seq);
 
         // Record when we first started processing this slot (for timing metrics).
@@ -2576,7 +2625,7 @@ impl Herder {
             // Parity: stellar-core HerderImpl.cpp:1546-1625 — observers build
             // the candidate tx set, self-validate, and publish it into the
             // pending/fetch cache, then stop before `makeStellarValue`/`nominate`.
-            self.build_and_cache_candidate_tx_set()
+            self.build_and_cache_candidate_tx_set(next_close_time)
                 .ok_or_else(|| HerderError::Internal("Failed to build candidate tx set".into()))?;
             let build_value_ms = t0.elapsed().as_millis();
 
@@ -2606,7 +2655,7 @@ impl Herder {
 
         // --- Validator path ---
         let value = self
-            .build_nomination_value()
+            .build_nomination_value(next_close_time)
             .ok_or_else(|| HerderError::Internal("Failed to build nomination value".into()))?;
         let build_value_ms = t0.elapsed().as_millis();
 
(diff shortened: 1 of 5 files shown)


tomerweller:
## 🔍 Reviewer: Parity

**Severity:** major

**Summary:** The new far-ahead abort in `trigger_next_ledger()` matches stellar-core, but the app-side `setupTriggerNextLedger` emulation does not actually apply upstream `ctValidityOffset(triggerOffset)` scheduling.

<details>

- In stellar-core, `HerderImpl::setupTriggerNextLedger()` computes `triggerTime = max(now, prepareStart + expectedClose)`, then computes `ctOffset = ctValidityOffset(lcl.closeTime + 1, triggerTime - now)`, and finally arms the timer at `triggerTime + ctOffset` (`stellar-core/src/herder/HerderImpl.cpp:1272-1290`).
- That depends on `HerderImpl::ctValidityOffset()` including the already-known trigger delay in its validity window: `system_now + maxCtOffset + MAX_TIME_SLIP_SECONDS` (`stellar-core/src/herder/HerderImpl.cpp:1158-1168`).
- In this PR, `try_trigger_consensus()` returns immediately whenever `prepare_start + expected_close` is still in the future (`crates/app/src/app/consensus.rs:196-206`). On the only path that reaches `ct_validity_offset()`, `trigger_time.saturating_duration_since(now_instant)` is therefore already zero (`crates/app/src/app/consensus.rs:211-217`).
- `Herder::ct_validity_offset()` also ignores its `trigger_offset` parameter entirely (`crates/herder/src/herder.rs:1243-1251`). So henyey never performs the upstream `triggerTime += ctOffset` calculation that this PR is meant to mirror.
- Observable consequence: instead of scheduling once for the adjusted deadline, henyey waits until `prepareStart + expectedClose` and only then starts a 1-second retry loop until `lcl.closeTime + 1` becomes valid. That is later sequencing than stellar-core in edge cases where the initial trigger delay should already absorb some or all of the close-time-validity wait.
- I think the fix is to compute the adjusted trigger deadline up front in the app gate (or otherwise make `ct_validity_offset()` actually incorporate the supplied trigger offset) so the polling model still mirrors the same upstream decision point.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Severity:** major

**Summary:** The core herder logic looks reasonable, but I can’t approve yet because the new regression tests are mis-gated and the app-side trigger-time path introduced for this feature is still untested.

<details>

- `crates/herder/src/herder.rs`: the new `ct_validity_offset_tests` call `Herder::set_test_clock_seconds()`, but that helper is only compiled behind `feature = "test-support"` while the tests themselves are plain `#[cfg(test)]`. Repro: `cargo test -p henyey-herder test_ct_validity_offset_zero_at_validity_boundary -- --nocapture` fails with `E0599` at lines 13313/13328/13345/13374 because the method is missing. Please either gate these tests the same way or rewrite them to avoid feature-only helpers so the crate’s normal test target still compiles.
- `crates/app/src/app/consensus.rs`: this PR adds the app-side `prepare_start + expected_close + ctValidityOffset` trigger-delay gate, but I could not find a test that exercises that path. The new herder tests cover close-time boundaries and far-ahead aborts in herder, and `ScpDriver::prepare_start()` has direct unit tests, but there is still no app-level regression test proving `try_trigger_consensus()` defers nomination until the adjusted trigger time and then proceeds at the boundary. Since issue #2816 is triaged as a feature, I think that gap needs to be closed before approval.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Severity:** blocker

**Summary:** The current head does not pass focused herder tests: the new `#2816` unit-test helper calls a feature-gated method and fails to compile.
The feature also adds a new public accessor consumed by the app-side trigger gate without any matching test coverage for that public surface / branch.

<details>

- `crates/herder/src/herder.rs:6078` / `crates/herder/src/herder.rs:923` — the new unit-test helper `make_herder_with_close_time_and_clock()` calls `Herder::set_test_clock_seconds(now_secs)`, but `set_test_clock_seconds` is behind `#[cfg(feature = "test-support")]`. Repro: `cargo test -p henyey-herder test_ct_validity_offset_zero_at_validity_boundary -- --exact` fails with `E0599: no method named set_test_clock_seconds found for struct Herder`. /do Mode B should fix the new tests so the default `henyey-herder` test target compiles (for example by avoiding the feature-gated API from `#[cfg(test)]` unit tests, or by restructuring the tests to use an always-available path).
- `crates/herder/src/herder.rs:1257` / `crates/app/src/app/consensus.rs:195` — `Herder::lcl_close_time()` is a new public surface in this feature, and I could not find any nearby test exercising it or the app-side trigger-delay / `SkippedInvalidCloseTime` branch that now depends on it. Since #2816 is a feature PR, /do Mode B should add coverage for this public accessor through the app behavior it enables (for example, a targeted app test that proves `try_trigger_consensus()` delays/skips when the candidate close time is too far ahead, and a manual-close test for the new retry-later error path).

</details>

tomerweller:
## 🔍 Reviewer: Parity

**Severity:** major

**Summary:** The PR matches upstream structure, but `ctValidityOffset` is not semantically aligned with stellar-core's millisecond-based clock math.
Around the `MAX_TIME_SLIP_SECONDS` boundary it can defer nomination materially longer than core, in a parity-critical herder path.

<details>

- `crates/herder/src/herder.rs:1242-1248` truncates both the current wall clock and `maxCtOffset` to whole seconds (`now_seconds()` and `max_ct_offset_ms / 1000`). Upstream keeps both operands at millisecond precision in `HerderImpl::ctValidityOffset` (`stellar-core/src/herder/HerderImpl.cpp:1158-1172`) and then feeds that exact offset into `setupTriggerNextLedger` (`HerderImpl.cpp:1281-1291`). This is observable at the boundary: with `now=1000.999`, `ct=1061`, and `maxCtOffset=0`, stellar-core returns `2ms` while the Rust helper returns `1001ms`; with `now=1000.100`, `ct=1062`, and `maxCtOffset=900ms`, core returns `1001ms` while Rust returns `2001ms`.
- Because this helper now gates both `App::try_trigger_consensus()` (`crates/app/src/app/consensus.rs:168-209`) and the herder-side abort in `trigger_next_ledger()` / cache-miss `handle_nomination_timeout()` (`crates/herder/src/herder.rs:2496-2516`, `2871-2889`), henyey can remain silent for an extra poll/tick after stellar-core would already nominate for the slot. In a parity-critical trigger-time change, that timing drift is material.
- /do Mode B should:
  - Rework `Herder::ct_validity_offset()` to use sub-second wall-clock arithmetic matching `HerderImpl::ctValidityOffset`, rather than `as_secs()`/`/ 1000` truncation.
  - Add a regression test that exercises a sub-second boundary case, because the current new tests only cover integral-second snapshots and therefore do not catch this divergence.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Severity:** major

**Summary:** The new trigger-delay gate records drift-tracker start times before nomination actually starts, so skipped polls poison close-time drift metrics for the slot.

<details>

- `App::try_trigger_consensus()` records `record_local_close_time(next_slot, local_time)` before the new `prepareStart + expectedClose + ctValidityOffset` gate (`crates/app/src/app/consensus.rs:168-210`).
- On any skipped poll (either `trigger_time > now_instant` or `ct_offset != 0`), the function returns without attempting `trigger_next_ledger()`.
- `CloseTimeDriftTracker::record_local_close_time()` ignores later writes for the same ledger (`crates/herder/src/drift_tracker.rs:128-132`), so when a later poll finally does trigger nomination the tracker keeps the stale earlier timestamp.
- That inflates `network_close_time - local_close_time` by the full defer interval and can produce false drift warnings / skewed metrics for exactly the far-ahead-clock cases this PR is trying to fix.
- The converged plan on #2816 explicitly said the app gate should return early **without recording drift**; the implementation moved the attempt counter but left the drift write ahead of the gate.

I think the fix is to move the drift-tracker write down next to the actual `trigger_next_ledger()` call, or otherwise allow a deferred pre-trigger sample to be replaced by the real trigger timestamp.
</details>

tomerweller:
## 🔍 Reviewer: Parity

**Severity:** major

**Summary:** `Herder::ct_validity_offset()` truncates both wall clock and trigger offset to whole seconds, so near the validity boundary it can wait materially longer than stellar-core.

<details>

- stellar-core keeps this computation in milliseconds: `system_now() + maxCtOffset + MAX_TIME_SLIP_SECONDS` (`stellar-core/src/herder/HerderImpl.cpp:1158-1168`), and feeds the result into `setupTriggerNextLedger()` (`HerderImpl.cpp:1281-1290`).
- This PR does `now_seconds() + max_ct_offset.as_secs() + MAX_TIME_SLIP_SECONDS` (`crates/herder/src/herder.rs:1255-1259`), which floors both operands to whole seconds.
- That matters here because henyey already preserves non-round ledger close durations (`test_ledger_manager_expected_close_time_non_5000ms_value` covers 4500ms in `crates/herder/src/herder.rs:12101-12110`), so `trigger_offset` can legitimately be sub-second.
- Example: with wall clock at `1000.999s`, `close_time=1061`, and `maxCtOffset=0ms`, stellar-core returns `2ms`, but the Rust helper returns `1001ms`; with `now=1000.100s`, `close_time=1062`, and `maxCtOffset=900ms`, core returns `1001ms` while the Rust helper returns `2001ms`.
- Because this helper now gates both `App::try_trigger_consensus()` and the herder-side invalid-close-time abort, the node can remain silent for an extra poll after stellar-core would already nominate.

Please rework `ct_validity_offset()` to keep millisecond precision end-to-end and add a sub-second regression test; the current new tests only cover integral-second snapshots, so they miss this divergence.
</details>

tomerweller:
## 🔍 Reviewer: Parity

**Summary:** The herder-side far-ahead abort matches stellar-core, but the app-side `setupTriggerNextLedger` port still miscomputes `triggerOffset` and can nominate up to one `expectedClose` window too early.
The current `ctValidityOffset` port also drops stellar-core's final sub-second guard at the close-time boundary.

<details>

- `crates/app/src/app/consensus.rs:188-195` diverges from `stellar-core/src/herder/HerderImpl.cpp:1281-1285`. Core passes `triggerOffset = triggerTime - now`; once `triggerTime` is clamped to `now`, that offset is **zero**. The PR instead passes elapsed time since `lastBallotStart` capped at `expected_close`, which widens the validity window after the clamp.
- Concrete parity check I ran: with `now=1000`, `expected_close=5s`, and `lcl.close_time=1070` (`minCandidateCt=1071`), stellar-core computes `ctValidityOffset=11s` while the PR computes `6s`, so henyey can nominate **5 seconds early** on this path.
- `crates/herder/src/herder.rs:1239-1248` ports `ctValidityOffset()` in whole seconds, but stellar-core returns a millisecond delay plus a final `+1ms` safety margin (`HerderImpl.cpp:1158-1169`). That loses the boundary guard and can still fire slightly early when `close_time` is just over the wall-clock boundary.
- I ran the new targeted herder tests with `CARGO_TARGET_DIR=/home/tomer/data/f4bfe1e4-6253-4696-8a02-ab88cc6de375/review-pr-2816/parity-target`; they pass and support the herder-side abort path, but there is still no regression covering the stale-`prepareStart` / clamped-`triggerTime` case above.
- /do Mode B should:
  - Change the app-side `triggerOffset` calculation to use the **remaining** time until `trigger_time` (`max(trigger_time - now, 0)`), not elapsed time since `last_ballot_start`.
  - Preserve stellar-core's no-early-trigger boundary behavior in `ct_validity_offset` / the app-side gate (millisecond precision plus the final `+1ms` guard, or an equivalent formulation).
  - Add a regression test for the reproduced case where `prepareStart + expectedClose <= now` but `lcl.close_time + 1` is still ahead of `now + MAX_TIME_SLIP_SECONDS`, proving we do not widen the window by `expectedClose`.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Summary:** I found two blocking correctness issues in the new trigger-time / close-time path.
The added herder tests pass, but they do not cover these race/parity regressions.

<details>

- `crates/herder/src/herder.rs:2481-2503` and `:2956-3042`: `trigger_next_ledger()` / `handle_nomination_timeout()` compute `next_close_time` before `build_nomination_value()` takes a fresh ledger snapshot. If LCL advances between those two reads, `let close_time_offset = close_time - lcl_close_time;` can underflow and feed a huge wrapped offset into tx-set trimming. /do Mode B should derive `close_time` and `lcl_close_time` from the same snapshot (preferred) or use a checked subtraction that converts the race into the existing stale/skip path, and add a regression test for the concurrent-LCL-advance case.
- `crates/app/src/app/consensus.rs:172-195` plus `crates/herder/src/scp_driver.rs:1077-1094`: `prepare_start()` is documented as stellar-core `mLastTrigger`, but it currently returns `ballot_start` (recorded in `started_ballot_protocol`) rather than the instant the slot was triggered. That makes the app delay the next trigger by nomination→ballot latency, so henyey can fire later than stellar-core under slow nomination. /do Mode B should expose the actual trigger-time timestamp (the existing `first_seen` / trigger entry point is much closer to `mLastTrigger`) or rename/rework the parity claim, and add a regression test showing slow ballot start does not delay the next trigger beyond the intended window.
- Because this is a feature PR, please also add explicit coverage for the remaining new public/observable surface in this diff after the above fixes land: `Herder::lcl_close_time()` and the new app-side `SkippedInvalidCloseTime` handling path (`manual_close_ledger` / trigger polling) are not directly asserted today.
- Evidence gathered: `cargo test -p henyey-herder --lib` for the nine new tests listed in the PR all passed individually, and `cargo test -p henyey-app --lib test_try_trigger_consensus_skips_while_applying` passed.

</details>

tomerweller:
## Addressed feedback in 06b9c89e362315bfdda51599a400a4b1db71b2de

**Parity reviewer:**
- ✅ Fixed app-side `triggerOffset` to use `trigger_time.saturating_duration_since(now)` (remaining time, not elapsed time). Matches stellar-core `triggerTime - now` (HerderImpl.cpp:1281-1282).
- ✅ Added +1 second boundary guard to `ct_validity_offset` return (parity: stellar-core's +1ms guard at HerderImpl.cpp:1168).
- ✅ Added regression test `test_ct_validity_offset_clamped_trigger_time_gives_zero_offset` reproducing the concrete case (now=1000, lcl.close_time=1070).

**Correctness reviewer:**
- ✅ Added `checked_sub` guard in `build_nomination_value` for the `close_time - lcl_close_time` underflow (aborts nomination if LCL advanced concurrently).
- ✅ Clarified `prepare_start` docstring — it correctly maps to stellar-core's `mPrepareStart` (recorded in `startedBallotProtocol`), which IS the ballot start time. Also added `slot_trigger_time` accessor exposing `first_seen` for the actual trigger instant.
- ✅ Added tests: `test_build_nomination_value_aborts_on_stale_close_time`, `test_lcl_close_time_accessor`.

**Verification:** All 1131 herder tests pass. `cargo fmt --check` and `cargo clippy -- -D warnings` clean.

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** I compared the new trigger-time / close-time gates against stellar-core's `HerderImpl::setupTriggerNextLedger`, `ctValidityOffset`, and `triggerNextLedger` flow.
The audited path now matches upstream semantics on when nomination is delayed or aborted; the remaining timer-driven scheduling gap is pre-existing and correctly documented as partial parity.

<details>

- `ScpDriver::prepare_start()` is wired to the same event upstream uses for `HerderSCPDriver::getPrepareStart()` (`started_ballot_protocol` / prepare-phase entry), so the app-side trigger gate is now anchored to the correct prior-slot timing reference.
- `App::try_trigger_consensus()` now mirrors upstream `setupTriggerNextLedger()` sequencing: it waits for `lastBallotStart + expectedClose`, then applies `ctValidityOffset(lcl.closeTime + 1, triggerOffset)` before calling `trigger_next_ledger()`. Within henyey's existing 1s poller architecture, this prevents earlier-than-stellar-core nomination.
- `Herder::ct_validity_offset()` is now used on the two parity-critical sites from stellar-core: (1) pre-trigger scheduling/delay and (2) the far-ahead nomination abort path. Observable effect: henyey no longer builds/broadcasts a nomination when the proposed `closeTime` is beyond `now + MAX_TIME_SLIP_SECONDS`.
- `Herder::trigger_next_ledger()` now aborts before tx-set/value construction when `next_close_time` is too far ahead, matching the upstream `triggerNextLedger()` early-return behavior and avoiding stray SCP state changes for an invalid close time.
- `handle_nomination_timeout()` now applies the same far-ahead abort on the cache-miss rebuild path, which keeps henyey aligned with the upstream invariant that a retry must not manufacture a fresh nomination value with an invalid close time.
- I did not find a blocking divergence in sequencing or observable close-time validity on this path relative to stellar-core. The still-open exact timer scheduling gap (`setupTriggerNextLedger` as a real timer rather than a 1s poll) remains documented in `PARITY_STATUS.md` as a partial parity item and is not a regression introduced by this PR.
- Validation: ran the new focused herder tests covering far-ahead abort, nomination-timeout cache-miss abort, `ct_validity_offset`, and `prepare_start`; they passed in an isolated worktree.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Summary:** I found one material retry-path bug in `trigger_next_ledger`, and the PR still misses test coverage for one newly-added public API surface.

<details>

- `crates/herder/src/herder.rs:2506` / `crates/herder/src/herder.rs:3050`: the new stale-close-time guard in `build_nomination_value()` returns `None` for a normal race where LCL advances between computing `next_close_time` and taking the snapshot, but `trigger_next_ledger()` still converts any `None` into `HerderError::Internal("Failed to build nomination value")`. That changes a benign retryable stale-slot race into a hard failure for app/manual-close callers, and it bypasses the later `SkippedStale` re-check. Please map this case to a non-error retry outcome (`SkippedStale` or equivalent) and add a regression test that exercises the public `trigger_next_ledger()` path, not just `build_nomination_value()` directly.
- Issue #2816 is `kind: feature`, so every newly-added public API surface needs at least one test. `ScpDriver::slot_trigger_time()` was added in this PR (`crates/herder/src/scp_driver.rs:1104`), but I could not find any test that exercises it. Please either add direct coverage for that accessor or remove it from the public surface if it is not needed.
- Verification performed: issue triage kind checked (`feature`), PR body reviewed, commit list reviewed, local correctness review completed in an isolated worktree, and `cargo test -p henyey-herder --lib` plus `cargo test -p henyey-app --lib` passed locally.

</details>

tomerweller:
## Addressed feedback in 1a4ccfc613d844cadc29d0a232e6982c54d822c3

**Correctness reviewer:**
- ✅ Fixed `trigger_next_ledger()`: `build_nomination_value() → None` now returns `Ok(SkippedStale)` instead of `Err(Internal)`. The stale-race is benign and retryable.
- ✅ Added `test_trigger_next_ledger_returns_skipped_stale_on_build_none` — exercises the public API by resetting the LedgerManager to force `create_snapshot()` failure, verifying the None → SkippedStale mapping.
- ✅ Added `test_slot_trigger_time_returns_recorded_instant` — covers `ScpDriver::slot_trigger_time()` accessor (record_slot_activity → Some(Instant)).

**Verification:** All 1133 herder tests pass. `cargo fmt --check` and `cargo clippy -- -D warnings` clean.

tomerweller (commented):
Addressed feedback. Ready for re-review.

tomerweller:
## 🔍 Reviewer: Correctness

**Summary:** The herder-side close-time guards look sound, but the new app-side retry path records drift-tracker state before nomination actually starts.
A skipped `SkippedInvalidCloseTime` retry can therefore poison the per-slot local close time and inflate drift measurements until the ledger externalizes.

<details>

- `crates/app/src/app/consensus.rs:208-217,250-255` — `try_trigger_consensus()` still calls `drift_tracker.record_local_close_time(next_slot, local_time)` before invoking `trigger_next_ledger()`, but this PR adds a new retryable outcome (`SkippedInvalidCloseTime`) where nomination never starts. Because `CloseTimeDriftTracker::record_local_close_time()` is write-once, the first skipped retry locks in an artificially early timestamp and later successful retries cannot correct it. That can overstate close-time drift by up to the whole `ctValidityOffset` delay and create false drift diagnostics. `/do` should move the drift-tracker write onto the path where nomination actually starts (or otherwise clear/update it on `SkippedInvalidCloseTime`) so we only record a local close time for real consensus starts.
- Please add an app-level regression test for the new `setupTriggerNextLedger` parity wiring, not just the herder helper/unit tests. The PR has good coverage for `Herder::ct_validity_offset`, far-ahead aborts, and the new public accessors, but I did not find a `try_trigger_consensus()` test that exercises the new delay path and proves we neither trigger consensus nor record drift while the close time is still invalid, then do trigger once it becomes valid.

</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** Compared this PR’s close-time gating against stellar-core’s `ctValidityOffset`, `setupTriggerNextLedger`, and `triggerNextLedger`.
The trigger delay and far-ahead abort semantics match upstream within henyey’s documented poller / second-granularity architecture.

<details>

- `ct_validity_offset()` matches `HerderImpl::ctValidityOffset` semantically: it uses `now + triggerOffset + MAX_TIME_SLIP_SECONDS` as the validity boundary, returns zero at the boundary, and returns a positive delay when the candidate close time is still too far ahead. The Rust `+1 second` guard is conservative relative to stellar-core’s `+1ms`, which is consistent with the documented second-granularity poller.
- `try_trigger_consensus()` preserves the `setupTriggerNextLedger()` gate: it waits for `prepareStart(lastSlot) + expectedClose`, falls back to “now” when no prepare-start timestamp exists, and only calls `trigger_next_ledger()` once the candidate close time is admissible.
- `trigger_next_ledger()` now mirrors the upstream far-ahead abort path before building / nominating: `nextCloseTime = max(now, lcl.closeTime + 1)`, and nomination is skipped when that value still exceeds `now + MAX_TIME_SLIP_SECONDS`.
- The cache-miss nomination-timeout rebuild takes the same conservative path: if a retry would still select an invalid far-ahead close time, it returns without fabricating a value that stellar-core would refuse.
- I did not find a parity-blocking sequencing mismatch on this path. The remaining timer-ownership difference is already documented in `PARITY_STATUS.md` (`setupTriggerNextLedger` remains Partial pending the broader event-driven timer work in #2702), and this PR stays within that stated scope.
</details>

tomerweller (commented):
Addressed feedback in 2dd8395e76c625e56ace9c8319c0e5956cbe9730:

- Moved `drift_tracker.record_local_close_time()` into the `Triggered` success branch so it only fires when nomination actually starts. `SkippedInvalidCloseTime` retries no longer poison drift measurements.
- Added `test_try_trigger_consensus_skips_and_no_drift_on_ct_validity_delay` — exercises the app-level setupTriggerNextLedger delay path end-to-end, verifying neither trigger attempt nor drift recording occurs when `ct_validity_offset > 0`.

Ready for re-review.

tomerweller:
Rebased onto current `origin/main` and resolved the parity-critical conflicts faithfully to stellar-core. Ready for re-review.

**Conflicts resolved** (`consensus.rs`, `herder.rs`, `scp_driver.rs`, `PARITY_STATUS.md`):

- **`TriggerOutcome` enum / match arms**: kept main's `ObserverCached` and the PR's `SkippedInvalidCloseTime` side by side; both app-side match arms coexist.
- **`build_nomination_value` split refactor (#2875)**: threaded the caller-validated `close_time` through both `build_nomination_value(close_time)` and `build_and_cache_candidate_tx_set(close_time)`. The builder no longer reads `SystemTime::now()`. `nextCloseTime` is derived **once** in `trigger_next_ledger` (monotonic clamp + far-ahead `ctValidityOffset` check) before `addTxSet`, matching `HerderImpl::triggerNextLedger` (HerderImpl.cpp:1492-1532) for **both** validator and observer paths — stellar-core's abort precedes `addTxSet`, which observers also execute.
- **Far-ahead abort** surfaces as typed `SkippedInvalidCloseTime`; benign `None` build maps to `SkippedStale` on both paths (preserving the prior review-feedback change).
- **Watcher latch**: rolled back on every non-success outcome (hard error, join failure, `SkippedStale`, `SkippedInvalidCloseTime`) and on the app-side `setupTriggerNextLedger` trigger-time / `ctValidityOffset` early-returns, so a watcher retries the slot once the close time becomes valid. Factored into `App::rollback_watcher_latch()`. This fixes a latch leak that surfaced once the far-ahead abort moved into the app gate.
- **`scp_driver.rs`**: kept both test helpers (`test_store_quorum_set` from main + `set_test_clock` from PR); `prepare_start()` getter auto-merged.
- **`PARITY_STATUS.md`**: `triggerNextLedger` → Full (far-ahead abort + post-build `isApplying` re-check); `ctValidityOffset` / `setupTriggerNextLedger` document the app-side delay + herder-side abort split, with event-driven timer ownership remaining in #2702.

**Test fixes for the new semantics**: the observer invalid-close-time test now asserts `SkippedInvalidCloseTime` (not a hard error); the two watcher latch-rollback tests drive the skip via the far-ahead close time (the removed `SystemTime`-vs-test-clock mismatch no longer exists).

**Verification**: `cargo fmt --check` clean; `cargo clippy -p henyey-herder -p henyey-app -- -D warnings` (CI invocation) clean; herder lib 1170 passed; app lib 964 passed.

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** Cycle N≥2 re-audit, scoped to `ctValidityOffset` trigger-time computation and far-ahead abort handling against stellar-core v26.0.1 (`HerderImpl::ctValidityOffset` / `setupTriggerNextLedger` / `triggerNextLedger`). All four parity concern classes I raised in prior cycles are now resolved. No new parity-blocking divergence found.

<details>

### Prior concern classes — status

**1. parity-gap — `ct_validity_offset()` ignored its `trigger_offset` argument; app-side never applied the upstream offset → ADDRESSED.**
- `Herder::ct_validity_offset(close_time, max_ct_offset)` now folds the offset into the validity boundary: `now + MAX_TIME_SLIP_SECONDS + max_ct_offset` (`crates/herder/src/herder.rs:474-485`), matching `system_now() + maxCtOffset + MAX_TIME_SLIP_SECONDS` (`stellar-core/src/herder/HerderImpl.cpp:1158-1168`).
- App-side gate passes the *remaining* time `trigger_time.saturating_duration_since(now)` (`crates/app/src/app/consensus.rs:217-221`). Because the gate returns early whenever `trigger_time > now`, the offset is structurally `0` on the only path that reaches `ct_validity_offset` — which is exactly correct for the poller model: henyey defers evaluation to the tick where `now ≈ triggerTime`, at which point core's effective boundary `triggerTime + MAX_TIME_SLIP` equals henyey's `now + MAX_TIME_SLIP`. Same decision point.

**2. spec-divergence / edge-case — whole-second vs millisecond clock math and the `+1ms` boundary guard → ADDRESSED (conservative, never-earlier).**
- The helper is still second-granularity (`now_seconds()`, `+1` guard vs core's `+1ms`), but this is strictly conservative in the parity-critical direction. henyey's allow-boundary `floor(now_s) + MAX_TIME_SLIP + floor(max_ct_offset_s)` is always `≤` core's `now_ms/1000 + MAX_TIME_SLIP + maxCtOffset_ms/1000` (flooring only lowers it). So henyey allows a strict subset of the close-times core allows: it can trigger **equal-or-later, never earlier**. My old counterexamples (`now=1000.999, ct=1061` → core blocks 2ms / Rust blocks 2s; `now=1000.100, ct=1062, off=900ms` → core 1001ms / Rust 3s) both have henyey blocking *longer*, confirming the safe direction. This is the documented poller-granularity behavior (`PARITY_STATUS.md`, "may trigger up to one poll tick later"), not an observable early-nomination.

**3. parity-gap — app-side `triggerOffset` used elapsed-since-`last_ballot_start` capped at `expected_close`, widening the window by up to one `expectedClose` (early nomination) → ADDRESSED.**
- Replaced with remaining-time `max(trigger_time - now, 0)` (`crates/app/src/app/consensus.rs:217-221`), matching `triggerOffset = triggerTime - now` (`HerderImpl.cpp:1281-1282`). The old window-widening is gone, and `test_ct_validity_offset_clamped_trigger_time_gives_zero_offset` is a dedicated regression locking in offset=0 vs the old wrong offset=5 (`herder.rs:952-967`).

**4. sequencing — far-ahead abort timing → ADDRESSED.**
- `trigger_next_ledger()`: `next_close_time = max(now, lcl.closeTime+1)`, abort to `SkippedInvalidCloseTime` when `> now + MAX_TIME_SLIP` (`herder.rs:2606-2618`). Exactly mirrors `triggerNextLedger`: `nextCloseTime = max(now, lcl.closeTime+1)`, `isCtValid = ctValidityOffset(nextCloseTime)==0`, skip nomination otherwise (`HerderImpl.cpp:1516-1532`).
- Cache-miss rebuild in `handle_nomination_timeout()` applies the identical abort before fabricating a fresh value (`herder.rs:3043-3055`), preserving the upstream invariant that a retry must not manufacture an invalid-close-time nomination.
- `now_seconds()` is wall-clock (`SystemTime::now`), matching core's `system_now()`/`to_time_t` rather than a monotonic clock (`scp_driver.rs:661-670`).

### New class considered — drift-recording timing (NOT a parity divergence)
This PR moves `record_local_close_time` from before the trigger to only the `Triggered` branch (`consensus.rs:222-233`), whereas core records `nextCloseTime` into `mDriftCTSlidingWindow` before the abort check (`HerderImpl.cpp:1500-1532`). I traced the consumer chain: henyey's `CloseTimeDriftTracker` output flows only to `App::drift_stats()` (`mod.rs:2459-2463`) — a diagnostic/stats accessor — and does **not** feed back into close-time selection or any SCP value. The far-ahead `lcl` case that would diverge is also pre-empted by the app-side gate before `trigger_next_ledger` is even called. So the difference is not observable in consensus behavior and is **not** flagged as a parity concern.

### Validation
Compared against `HerderImpl::ctValidityOffset` (`HerderImpl.cpp:1157-1172`), `setupTriggerNextLedger` (`HerderImpl.cpp:1259-1300`), and `triggerNextLedger` (`HerderImpl.cpp:1490-1532`). The remaining exact event-driven timer scheduling gap is correctly documented as Partial parity pending #2702 and is not introduced by this PR.

</details>

crates/herder/src/herder.rs
```diff
@@ -1301,6 +1307,40 @@ impl Herder {
         self.ledger_manager.expected_ledger_close_duration()
     }
 
+    /// Get the ballot-start instant for the given slot.
+    /// Parity: stellar-core `mLastTrigger` (HerderImpl.cpp), exposed as "prepare start".
```
tomerweller, line 1311: **parity-gap (non-blocking, doc-only):** This wrapper still cites `mLastTrigger`, but `setupTriggerNextLedger` (HerderImpl.cpp:1266) uses `getPrepareStart(lastIndex)`, and the underlying `ScpDriver::prepare_start` doc correctly says `getPrepareStart()` / `mPrepareStart` (recorded in `startedBallotProtocol` = `ballot_start`). `mLastTrigger` is a different field. The code is correct — only this one-line citation is stale/inconsistent with the driver doc it delegates to. Suggest aligning it to `getPrepareStart` / `mPrepareStart`.

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Cycle:** N≥2 (8 prior reviewer comments). All previously-raised concern classes re-verified below; the residual is doc-only and non-blocking.

**Summary:** This is a clean, parity-faithful rendering of stellar-core's `setupTriggerNextLedger` / `triggerNextLedger` / `ctValidityOffset` onto henyey's poller architecture. Every functional concern from prior cycles is now addressed, and the new public surface is fully tested. I verified the diff against `HerderImpl.cpp` (1158-1296, 1517-1532) and ran all new tests green. One stale parity citation in a doc comment remains (inline) — does not affect behavior.

<details>

### Prior concern classes — status

- **test-coverage** (compile failure from feature-gated helper) — RESOLVED. Herder unit tests now use `ScpDriver::set_test_clock()` (`#[cfg(test)] pub(crate)`) instead of the `feature="test-support"` `set_test_clock_seconds`. `cargo test -p henyey-herder --lib` compiles and passes. Every new public item has a test: `Herder::prepare_start` (`test_herder_prepare_start_forwards_driver_timing`), `Herder::lcl_close_time` (`test_lcl_close_time_accessor`), `Herder::ct_validity_offset` (4 boundary tests), `ScpDriver::prepare_start` (3 tests), `ScpDriver::slot_trigger_time` (`test_slot_trigger_time_returns_recorded_instant`), and both `SkippedInvalidCloseTime` variants.
- **regression-risk** (drift-tracker poisoning) — RESOLVED. `record_local_close_time()` moved out of the pre-trigger path into the `TriggerOutcome::Triggered` success branch (`consensus.rs`). `test_try_trigger_consensus_skips_and_no_drift_on_ct_validity_delay` proves no attempt + no drift write on the delay path.
- **error-handling** (`None` → hard `Err(Internal)`) — RESOLVED. Both observer (`build_and_cache_candidate_tx_set`) and validator (`build_nomination_value`) `None` results now map to `TriggerOutcome::SkippedStale`, and the far-ahead abort returns the typed `SkippedInvalidCloseTime`. `test_trigger_next_ledger_returns_skipped_stale_on_build_none` exercises the public `trigger_next_ledger()` path (via `LedgerManager::reset`).
- **race / underflow** (`close_time - lcl_close_time`) — RESOLVED. Replaced with `checked_sub`; `Some(0) | None` → abort (returns `None`). `test_build_nomination_value_aborts_on_stale_close_time` covers it.
- **parity-gap** (prepare_start citation) — see doc nit below; functional behavior is correct.

### Parity verification against stellar-core
- `ct_validity_offset` matches `HerderImpl::ctValidityOffset` (boundary `now + maxCtOffset + MAX_TIME_SLIP_SECONDS`; +1s guard vs upstream +1ms is a documented second-granularity deviation).
- App-side delay matches `setupTriggerNextLedger` (1259-1296): `triggerTime = prepareStart + expectedClose`, clamped to `now`; in the poller model henyey *skips-and-retries* when `triggerTime > now` rather than arming a timer, and at the point it proceeds `triggerOffset == 0` — converging to the same fire time. `minCandidateCt = lcl.closeTime + 1` matches. Event-driven timing deferred to #2702 (noted in PARITY_STATUS).
- Herder-internal far-ahead guard (`next_close_time = now.max(lcl+1)`, skip if `> now + MAX_TIME_SLIP_SECONDS`) is a faithful rendering of `triggerNextLedger` 1517-1532.
- `rollback_watcher_latch()` factoring preserves the race-safe same-slot compare_exchange; applied uniformly to hard error, join failure, `SkippedStale`, and `SkippedInvalidCloseTime`.

### Non-blocking doc nit (inline)
- `crates/herder/src/herder.rs:1311` — `Herder::prepare_start` doc cites `mLastTrigger`, but upstream uses `getPrepareStart(lastIndex)` and the underlying `ScpDriver::prepare_start` doc correctly says `getPrepareStart`/`mPrepareStart`. Code is correct; the one-line citation is inconsistent. Worth a quick fix on the next touch but not a merge blocker.

### Tests run (worktree @ 08deb6a4, target cached)
- `cargo test -p henyey-herder --lib` — 15 new/affected tests pass (ct_validity_offset×4, prepare_start×4, slot_trigger_time, skipped_stale, far-ahead×2, build_nomination_value abort, lcl_close_time, observer skip).
- `cargo test -p henyey-app --lib try_trigger_consensus` — 6 pass (incl. the new ct-validity-delay no-drift test and both watcher latch-rollback tests reworked onto the typed outcome).

</details>

tomerweller:
## ⚠️ Force-Converged (Lifetime Cap)

**Commit:** dcb4f419ecfe29119dca50bad175d88a02c4d75e
**Lifetime bounce count at force-converge:** 6
**CI at merge:** green (37/37)

This PR reached the lifetime bounce cap (6 cycles, no `## Review: Reset`). Per `/review-pr` force-converge policy, green CI is sufficient to land the change and preserve any unresolved concerns as backlog.

Notably, **both fresh reviewers APPROVED** the current head (`08deb6a4`) on this tick:
- **Reviewer: Correctness** — APPROVE. Verified all prior concern classes (test-coverage, regression-risk, error-handling, race/underflow) resolved; new tests run green in isolated worktree.
- **Reviewer: Parity** — APPROVE. Confirmed ctValidityOffset trigger-time and far-ahead abort match stellar-core `HerderImpl.cpp` (offset folding, whole-second math strictly conservative, far-ahead abort mirrors `triggerNextLedger`).

So there are **no unresolved CHANGES_REQUESTED concerns** — the force-converge path applied only because the lifetime counter had already hit 6 across earlier code states. No concern-class follow-ups were needed.

**Inline-comment follow-ups filed (non-blocking):**
- `crate:herder`: #2992 (lcl_close_time Option), #2993 (close_time_offset underflow), #2995 (test uses private slot_timing map), #2996 (stale mLastTrigger doc citation)
- `crate:app`: #2994 (trigger_offset_secs always 0)

These are low/non-blocking reviewer notes (4 from the GH Copilot bot, 1 doc-only parity nit) preserved as backlog.

Issue #2886, opened 2026-05-22
FATAL: knit-to-LCL hash mismatch on 6011ec9c — state wipe required


tomerweller:
## Symptom

Validator self-shutdown at 2026-05-22T03:07:46Z with:

```
FATAL: unrecoverable local state failure — RecoveryEscalation catchup verification/integrity failure:
  knit-to-LCL failed at LCL:
    expected 588f81e9dfa34ce16ccc9390b2850a2bd9caae12e7818cd000d132d0d76918b4
    got      fe104f27b820b2907d0b83d57507d8cd522b42d0ec832a9de655274a429e16d3
Node will shut down. State wipe required before restart. fatal_wipe_required=true
```

This is a **clean shutdown** (not a panic) — the validator detected its own state had diverged from canonical and refused to continue. Catchup-replay produced a different hash than the originally-committed ledger close.

## Build

- Running `6011ec9c` (deployed 2026-05-22T00:09Z, uptime ~3h at failure)
- Recent code changes in the catchup region:
  - `6011ec9c #2883` Move cross-type source guard before validation (crates/herder, deployed at the time)
  - `be7cabda #2878` Align SCP envelope reception pipeline order (crates/herder, prior build)
  - `8d2951e1 #2852` Shard soroban_state COW maps into 64 per-shard Arcs (crates/ledger)
  - `fa29f93e #2860` CatchupRunMode enum (crates/app/history/historywork)

## Timeline (03:07:00-03:07:46 UTC)

- `03:07:00.946` — L62678395 closed normally, hash=`588f81e9...`
- `03:07:00.967` — `Rapid close cycle ended; requesting SCP state from peers` (validator briefly lost sync)
- `03:07:43.902` — Recovery escalation: `Catchup task spawned label="RecoveryEscalation"`
- `03:07:46.343` — `Fatal error during replay — not retrying error=knit-to-LCL failed at LCL: expected 588f81e9..., got fe104f27...`
- `03:07:46.482` — `FATAL: unrecoverable local state failure ... fatal_wipe_required=true`
- `03:07:46.483` — State `Catching Up` → `Shutting Down`
- `03:07:46.605` — Process exit with error

## Hypothesis

The catchup-replay path reproduced a different state hash than the normal close path produced. This is a determinism failure — either:

1. **Soroban state sharding** (8d2951e1) introduced non-determinism via 64-shard Arc COW (different iteration order during replay than during close?)
[...]

PR #2937, opened 2026-06-02
Gate knit/replay on published target checkpoint to avoid FATAL state-wipe

### LINKED ISSUE #2931: URGENT: knit-to-LCL state-integrity FATAL on b35ea04e during RecoveryEscalation catchup (recurrence of #2886) — validator down ~4.5h, wipe required
## Symptom

Mainnet validator crashed with a **knit-to-LCL state-integrity FATAL** during a
`RecoveryEscalation` Minimal catchup, demanding a state wipe:

```
2026-06-02T08:40:26Z WARN  henyey_history::catchup::replay: Fatal error during replay — not retrying attempt=0
  error=knit-to-LCL failed at LCL: expected 5ce48a3e447884b80e2f84df3a2d883857191d536003aadc82f9d46803c19983,
                                       got 75ebb7dabee33c2c5fc51656167992c7804a7fffcb646b7697b45afd06235554
2026-06-02T08:40:26Z ERROR henyey_app::app::lifecycle: FATAL: unrecoverable local state failure —
  RecoveryEscalation catchup verification/integrity failure: knit-to-LCL failed at LCL ...
  Node will shut down. State wipe required before restart. fatal_wipe_required=true
```

Node shut itself down at 08:40:26 UTC and stayed down ~4.5h until detected.

## Build / context

- Build: `b35ea04e` (#2914, current mainnet validator)
- Catchup: `RecoveryEscalation` Minimal, target L62845439, from `current_lcl=62845436`, range `ReplayOnly{first:62845437,count:3}`
- The catchup was triggered by the recurring degraded-peering fall-behind (see "trigger chain" below)
- Preceding log:
  ```
  Spawned catchup task label="RecoveryEscalation"
  Starting catchup target=Ledger(62845439) mode=Minimal
  Catchup target checkpoint may not be published yet (latest_ext is local and may be stale) — proceeding anyway
  Cloned bucket lists from ledger manager for replay-only catchup current_lcl=62845436 target=62845439
  Gap detected: next slot EXTERNALIZE missing ... current_ledger=62845436 latest_ext=62845447
  → replay → knit-to-LCL hash mismatch → FATAL
  ```

## Trigger chain (hypothesis)

Persistent **node-local degraded peering** (preferred peers `95.216.1.86`/`88.99.1.97`
TCP-unreachable, peer count ~25–33 vs ~48) caused recurring fall-behind events
(documented in daily #2929), each forcing a `RecoveryEscalation` Minimal catchup.
One such catchup mis-knit the bucket-list state — the replayed LCL hash
(`75ebb7da…`) diverged from expected (`5ce48a3e…`), tripping the integrity check.

Notably the catchup proceeded despite "checkpoint may not be published yet …
proceeding anyway" — replaying against a possibly-stale/local target may be
implicated in producing the bad knit.

## Recurrence

This is a **recurrence on the current build** of a failure class previously seen and closed:
- #2886 — "FATAL: knit-to-LCL hash mismatch on 6011ec9c — state wipe required" (closed 2026-05-22, **old build**)
- #2362 — "OFFLINE: stuck after RecoveryEscalation FATAL hash mismatch at L62408092 (state corruption persists)" (closed 2026-05-04)

Since the build (`b35ea04e`) and the surrounding recovery path (#2906/#2907/#2914)
are materially different from when #2886 was closed, filing fresh and
cross-referencing rather than reopening.

## Impact

Validator down ~4.5h (08:40 → 13:26 UTC; ~3,000 ledgers behind at detection).
`fatal_wipe_required=true` → recovering via fresh-start wipe + rebootstrap
(operator-authorized), same procedure as the 2026-06-01 00:03 recovery.

## Asks

1. Is replaying a `RecoveryEscalation` catchup against an unpublished/stale local
   target (the "proceeding anyway" path) capable of producing a bad knit-to-LCL?
   Should that path refuse to proceed until the checkpoint is published?
2. Should the degraded-peering fall-behind (which forces these catchups) be the
   real fix — i.e., is the recovery path too aggressive when under-peered?

tomerweller:
Closes #2931

## Summary

A `RecoveryEscalation` Minimal catchup running the cloned-local `ReplayOnly` fast path against an **unpublished/stale** archive checkpoint re-fetched the LCL header from the not-yet-published checkpoint file. The stale archive header diverged from the (correct) local LCL, producing a §11.2 case-3 `KnitLclHashMismatch` that `is_fatal_catchup_failure()` classified as fatal — triggering a FATAL state-wipe shutdown (#2931, recurrence of #2886; validator down ~4.5h).

This restores the published-checkpoint precondition stellar-core enforces before knit/replay (`GetHistoryArchiveStateWork` retry-until-published) that henyey's fast path optimized away. A new transient `HistoryError::CheckpointNotYetPublished` variant (excluded from `is_fatal_catchup_failure` and `is_hash_mismatch`) plus a `checkpoint_publication_gate` at the top of `download_verify_and_replay_once` — **before** any checkpoint-file / LCL-header fetch — fetch the HAS for `checkpoint_containing(target)` and require the archive's published frontier (`currentLedger`) to cover it. A reachable-but-behind HAS is retried with backoff instead of wiping; HAS-unreachable stays a transient `CatchupFailed` (distinct taxonomy). A genuine knit divergence against a **confirmed-published** checkpoint stays fatal, preserving §11.2 parity for real local corruption.

The gate keys off **archive HAS truth** (`currentLedger`), never the local externalized counter, so it does not reintroduce the stale-counter deadlock that the original "proceeding anyway" guard removal addressed.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/2931#issuecomment-4602983692)

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy --all -- -D warnings` (and `-p henyey-history --tests -D warnings`)
- [x] `cargo test -p henyey-history` — 526 lib + 11 replay_integration + catchup_integration + doctests all pass

- **Tests:**
  - `crates/history/src/catchup/replay.rs::test_knit_lcl_mismatch_unpublished_checkpoint_is_transient`
  - `crates/history/src/catchup/replay.rs::test_knit_lcl_mismatch_published_checkpoint_remains_fatal`
  - `crates/history/src/error.rs::test_checkpoint_not_yet_published_is_transient`
- **Pre-fix:** committed as `c11a6a08` — verified FAILED (the `CheckpointNotYetPublished` variant and `checkpoint_publication_gate` did not exist → tests did not compile).
- **Post-fix:** verified PASS after `ab28836c`.

## Deviations from plan

- The pre-existing `replay_integration.rs` ReplayOnly tests needed a HAS fixture at the target's covering checkpoint (the new gate fetches it). Added a small `register_published_target_has` test helper and a call in each affected test — a test-fixture update mandated by the new production HAS fetch, not a behavior change.

### DIFF
--- a/crates/history/src/catchup/replay.rs
+++ b/crates/history/src/catchup/replay.rs
@@ -1297,4 +1297,86 @@ mod tests {
             "expected InvalidPreviousHash without SCP anchor, got: {err}"
         );
     }
+
+    // ---------------------------------------------------------------
+    // #2931: published-checkpoint precondition gate.
+    //
+    // Restores the upstream HAS-publication gate (stellar-core
+    // `GetHistoryArchiveStateWork` retry-until-published) that henyey's
+    // cloned-local ReplayOnly fast path omitted. The gate keys off archive
+    // HAS truth (`currentLedger`), NOT the local externalized counter.
+    // ---------------------------------------------------------------
+
+    #[test]
+    fn test_knit_lcl_mismatch_unpublished_checkpoint_is_transient() {
+        // Mirrors the #2931 crash scenario: a ReplayOnly attempt targets
+        // ledgers in checkpoint 62845439 while the archive's published
+        // frontier (HAS currentLedger) is still on the prior checkpoint.
+        //
+        // The archive's checkpoint header at the LCL diverges from the
+        // (correct) local LCL — on origin/main this surfaces as a fatal
+        // KnitLclHashMismatch and wipes state. With the publication gate the
+        // attempt is short-circuited as a TRANSIENT CheckpointNotYetPublished
+        // BEFORE any knit/replay, so it is retried instead of wiping.
+        let target = 62845437; // in checkpoint 62845439
+        let target_ckpt = crate::checkpoint::checkpoint_containing(target);
+        // Archive has only published through the prior checkpoint.
+        let has_current = target_ckpt - crate::checkpoint::checkpoint_frequency();
+
+        let err = checkpoint_publication_gate(has_current, target)
+            .expect_err("gate must reject an unpublished target checkpoint");
+
+        match err {
+            HistoryError::CheckpointNotYetPublished {
+                target: t,
+                has_current: hc,
+            } => {
+                assert_eq!(t, target);
+                assert_eq!(hc, has_current);
+            }
+            other => panic!("expected CheckpointNotYetPublished, got {other:?}"),
+        }
+        assert!(
+            !err.is_fatal_catchup_failure(),
+            "unpublished-checkpoint knit attempt must be transient, not fatal"
+        );
+        assert!(
+            !err.is_hash_mismatch(),
+            "unpublished-checkpoint knit attempt must not be a hash mismatch"
+        );
+    }
+
+    #[test]
+    fn test_knit_lcl_mismatch_published_checkpoint_remains_fatal() {
+        // When the target checkpoint IS published (HAS currentLedger covers
+        // it), the publication gate passes and a genuine knit divergence
+        // against the local LCL stays fatal — preserving §11.2 parity for
+        // real local corruption (no over-broadening of the transient path).
+        let target = 62845437;
+        let target_ckpt = crate::checkpoint::checkpoint_containing(target);
+
+        // Gate passes: archive has published the covering checkpoint.
+        assert!(
+            checkpoint_publication_gate(target_ckpt, target).is_ok(),
+            "gate must pass once the covering checkpoint is published"
+        );
+        // The gate must also pass when the archive frontier is well ahead
+        // of the covering checkpoint (steady-state operation).
+        assert!(
+            checkpoint_publication_gate(target_ckpt + 1000, target).is_ok(),
+            "gate must pass when the archive frontier is ahead of the target"
+        );
+        // And a current-checkpoint divergence is still fatal (case 3).
+        let lcl = make_test_lcl(100, h(0x10), h(0x09));
+        let entry = make_test_entry(100, h(0xBB), h(0x09));
+        let err = knit_to_lcl_decision(&entry, &lcl).unwrap_err();
+        assert!(
+            matches!(err, HistoryError::KnitLclHashMismatch { .. }),
+            "published-checkpoint divergence must stay KnitLclHashMismatch"
+        );
+        assert!(
+            err.is_fatal_catchup_failure(),
+            "published-checkpoint knit divergence must remain fatal"
+        );
+    }
 }
--- a/crates/history/src/catchup/replay.rs
+++ b/crates/history/src/catchup/replay.rs
@@ -116,6 +116,42 @@ pub(super) fn knit_to_lcl_decision(
     Ok(KnitDecision::Apply)
 }
 
+/// Published-checkpoint precondition gate for the knit/replay path (#2931).
+///
+/// Mirrors stellar-core's `GetHistoryArchiveStateWork` retry-until-published
+/// precondition (`CatchupWork.cpp` HAS fetch with retries; covered by
+/// `LedgerApplyManagerImpl`'s wait for the following checkpoint) that
+/// henyey's cloned-local `ReplayOnly` fast path optimized away. stellar-core
+/// has no such fast path and always fetches the HAS before applying a
+/// checkpoint, so a target whose covering checkpoint is not yet published is
+/// never attempted upstream.
+///
+/// The gate keys off **archive HAS truth** (`has_current_ledger`, the
+/// archive's published frontier), NOT the local externalized counter — using
+/// the local counter would reintroduce the stale-counter deadlock that the
+/// original `catchup_impl` guard removal (proceeding anyway) addressed.
+///
+/// Returns:
+/// - `Ok(())` when `has_current_ledger >= checkpoint_containing(target)` (the
+///   covering checkpoint is published; proceed to knit/replay).
+/// - `Err(CheckpointNotYetPublished { .. })` (transient) otherwise, so the
+///   attempt is retried with backoff instead of triggering a FATAL
+///   state-wipe on a knit-to-LCL mismatch against a stale archive header.
+///
+/// HAS **unreachability** (network/404) is handled by the caller as a
+/// transient download error and never reaches this gate — keeping the error
+/// taxonomy distinct (HAS-unreachable vs HAS-behind).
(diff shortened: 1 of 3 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Bug-fix is correct, minimal, and well-targeted. The published-checkpoint gate runs before any LCL-header fetch so a stale-archive knit mismatch can no longer reach the fatal classifier; TDD red→green verified locally; transient/fatal taxonomy is right.

<details>

Cycle 1 — complete change-list (no prior Correctness comment on this PR). All clear; no concern classes require changes.

**test-coverage** — PASS
- `kind: bug-fix` confirmed via triage report. Regression tests present and verified:
  - `replay.rs::test_knit_lcl_mismatch_unpublished_checkpoint_is_transient`
  - `replay.rs::test_knit_lcl_mismatch_published_checkpoint_remains_fatal`
  - `error.rs::test_checkpoint_not_yet_published_is_transient`
- TDD order verified: test commit `c11a6a08` precedes fix commit `ab28836c`. Built the crate at `c11a6a08`: it FAILS to compile (E0425 `checkpoint_publication_gate` not found; E0599 `CheckpointNotYetPublished` variant not found) — a legitimate red state for the replay tests. At PR head all three pass. The fix-commit introduces both the variant and the gate fn, so the tests genuinely capture the bug.
- Minor note (non-blocking, not filing): `test_checkpoint_not_yet_published_is_transient` only exercises the variant's classifier membership, which would pass as soon as the variant compiles; the two `replay.rs` tests are the ones that capture the gate behavior. Coverage of the actual bug is adequate via those two.

**error-handling** — PASS
- `CheckpointNotYetPublished` is correctly absent from both `is_fatal_catchup_failure` and `is_hash_mismatch` match arms (error.rs:546, 584), so it defaults to transient → retried with backoff rather than triggering a FATAL state-wipe. This is the core of the fix.
- HAS-unreachable taxonomy holds: `download_has` (download.rs:114) returns transient `HistoryError::CatchupFailed` when all archives fail — also not fatal, also retried, distinct from `CheckpointNotYetPublished`. The PR's "distinct taxonomy" claim is accurate.

**logic / placement** — PASS
- Gate is the first statement in `download_verify_and_replay_once` (replay.rs:435-449), strictly BEFORE `download_ledger_data` (which refetches the stale LCL header). So the unpublished-target mismatch can never surface. Correct ordering.
- `checkpoint_publication_gate(has_current, target)` compares `has_current < checkpoint_containing(target)` — frontier vs covering checkpoint. Boundary cases covered by the published-checkpoint test (equal and ahead both pass).

**readability** — PASS. Doc comments are thorough and accurately describe the retry-vs-wipe rationale and the deliberate use of archive `currentLedger` over the local counter.

Non-blocking observation (not filing): the gate uses raw `download_has` (no `verify_has`). Acceptable — `current_ledger` only drives a retry-vs-proceed decision; the checkpoint HAS used for bucket apply still goes through `download_and_verify_has`. A malformed-but-parseable HAS would at worst over-gate (safe retry) or fall through to the pre-existing knit path. No correctness regression.

CI: 14/14 green.
</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** The gate faithfully restores stellar-core's pre-apply HAS-publication precondition that henyey's ReplayOnly fast path optimized away. Keying off archive `currentLedger` matches `CatchupWork`'s own frontier comparison exactly; the transient-retry-vs-fatal split mirrors upstream semantics.

<details>

Cycle 1 — complete change-list (no prior Parity comment on this PR). No divergences found; no concern classes require changes.

**parity-gap** — PASS (restores parity rather than diverging)
- stellar-core `CatchupWork::doWork` (CatchupWork.cpp:290-310) ALWAYS issues `GetHistoryArchiveStateWork` for `checkpointContainingLedger(...)` before any bucket/replay apply — there is no upstream fast path that skips the HAS fetch. henyey's cloned-local `ReplayOnly` path had elided this, which is the root cause of #2931. The new `checkpoint_publication_gate` at the top of `download_verify_and_replay_once` reinstates that fetch-then-check ordering.

**frontier-semantics** — PASS
- Core compares against the archive's published frontier `mHAS->currentLedger`, not a local externalized counter. CatchupWork.cpp:336-343: `if (mLastClosedLedgerHashPair.first >= mHAS->currentLedger) { ... "Target ledger is not newer than last closed ledger - nothing to do" ... "Wait until next checkpoint before retrying" }`. henyey's gate computes `has_current_ledger < checkpoint_containing(target)` off `target_has.current_ledger()` — same archive-truth source, same direction. The PR's stated reason for not using the local counter (avoids the stale-counter deadlock the original guard-removal fixed) is consistent with upstream, which never consults the local counter for this decision.

**sequencing** — PASS
- Upstream's "wait until next checkpoint before retrying" is a transient/retry outcome, not a failure-classification event. henyey models this as the transient `CheckpointNotYetPublished` (excluded from fatal/hash-mismatch classifiers), so an unpublished-target knit attempt retries with backoff — matching core's retry-on-next-checkpoint behavior rather than core ever reaching a fatal divergence on an unpublished target.

**edge-case** — PASS
- Genuine knit divergence against a CONFIRMED-PUBLISHED checkpoint stays fatal (`KnitLclHashMismatch` still in `is_fatal_catchup_failure`), preserving CATCHUP_SPEC §11.2 case-3 parity for real local corruption. The published-checkpoint regression test asserts exactly this, so the transient path is not over-broadened.
- HAS-unreachable (all archives fail) remains a transient `CatchupFailed` — core likewise retries `GetHistoryArchiveStateWork` (RETRY semantics) rather than failing catchup outright on a transient fetch error.

No semantic, edge-case, or sequencing divergence from stellar-core v26 on this path.
</details>

tomerweller:
## ✅ Merged

**Commit:** d5a1b71a4b829de8e7908151487ede2a7ac76a7c

Triple-green: Reviewer A (Correctness) APPROVE, Reviewer B (Parity) APPROVE, CI 14/14. External: GH Copilot bot COMMENTED (non-blocking).

TDD verified locally: regression tests fail to compile at test commit `c11a6a08` (no `checkpoint_publication_gate` / `CheckpointNotYetPublished`), pass at PR head. Parity confirmed against stellar-core `CatchupWork.cpp` (always fetches HAS for the covering checkpoint before apply; compares LCL vs archive `currentLedger` frontier; treats behind-frontier as retry-on-next-checkpoint, not fatal).

**Follow-up issues filed for unaddressed inline review comments:** #2939, #2940, #2941

PR #2955, opened 2026-06-02
Pin gated catchup replay downloads to the publication-gate archive

### LINKED ISSUE #2940: Pin all per-attempt catchup downloads to the archive that passed the publication gate
Follow-up from PR #2937 (issue #2931). Non-critical inline review comment that was not addressed before merge.

## Concern

`crates/history/src/catchup/replay.rs:449` — The publication gate downloads HAS via `download_has(...)`, which can succeed against one archive, but the subsequent header / checkpoint downloads in this attempt (`download_checkpoint_header`, `download_checkpoint_ledger_data`) iterate archives independently (fixed order) and may end up pulling ledger data from a different archive that did *not* pass the gate (e.g., archive A missing/404ing `history/{ckpt}.json` but still serving `ledger/{ckpt}.xdr.gz`). In that case the stale/unpublished checkpoint data can still be used and the knit-to-LCL mismatch can re-surface despite the gate. Consider pinning all downloads in an attempt to the same archive that provided `target_has`, or moving the gate into the per-archive download paths so any archive used for headers/data must satisfy the publication precondition.

## Source

[Original review comment](https://github.com/stellar-experimental/henyey/pull/2937#discussion_r3342182616) on PR #2937.

## Severity

Low / non-blocking — reviewer (GH Copilot) chose COMMENTED rather than request changes. Requires a multi-archive split where one archive is published and another is behind-but-serving-data for the same checkpoint — an uncommon archive-inconsistency scenario. The primary single-archive #2931 crash path is fully closed by the merged gate. Filing so the cross-archive hardening isn't lost.

tomerweller:
Closes #2940

## Summary

A gated catchup replay attempt now downloads ledger data **only** from the archive that served `target_has` and passed the #2937 publication gate. `download_verify_and_replay_once` captures that `Arc<HistoryArchive>` and threads it as `Option<&Arc<HistoryArchive>>` into `download_ledger_data` → `download_checkpoint_header` + `download_checkpoint_ledger_data`. When the pin is `Some`, those downloads target the gate archive with no rotation/fallback; when `None` (bucket-apply phase, SCP/bucket downloads, all other non-gated callers) the existing fixed-order rotation over `&self.archives` is preserved.

This closes the cross-archive bypass where a second stale-but-serving archive could supply ledger data that never satisfied the gate — reopening the #2931 knit-to-LCL mismatch / state-wipe path under archive asymmetry. The pin is re-derived per replay attempt (never cached across retries), so a down archive is still rotated at the attempt boundary while intra-attempt consistency (gate archive == data archive) is guaranteed.

`download_has` now returns the serving `Arc<HistoryArchive>` (the bare name `String` is derived via `archive.name()` at the metric call sites).

## Parity

Matches stellar-core's **offline** `mArchive` pinning discipline (`CatchupWork` stores a single `mArchive` threaded to all child works; `doReset()` never clears it). This is **stronger** than stellar-core's online catchup, which passes `nullptr` and uses uncoordinated per-request random archive selection. Never weaker than online behaviour.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/2940#issuecomment-4604133158)

## Test plan

- [x] cargo fmt --check
- [x] cargo test -p henyey-history — all pass (528 lib + integration/doctests, 0 failures)

## Regression test

- **Test:** `crates/history/tests/replay_integration.rs::test_catchup_pins_downloads_to_gate_archive`
- **Setup:** two archives ordered [B, A]. A serves the gate HAS + correct data-checkpoint ledger; B 404s the gate HAS and serves a divergent header64 (mismatched `previous_ledger_hash`). `download_has` rotation picks A as the gate archive (B 404s the gate HAS); under the old unpinned behaviour fixed-order rotation would fetch B's divergent ledger first.
- **Pre-fix (verified):** with the pin set to `None`, catchup fails with `KnitCurrentLedgerPrevHashMismatch { ledger: 64, actual: "abab…" }` — B's divergent bytes are consumed.
- **Post-fix:** with the pin, catchup succeeds using A's data only.
- **Unit:** `crates/history/src/catchup/download.rs::test_download_has_returns_serving_archive` and `::test_download_checkpoint_header_pin_targets_only_pinned` (pin targets only the pinned archive; `None` preserves rotation; pinned-to-dead fails without fallback).

## Deviations from plan

- The plan listed an optional `download_checkpoint_ledger_data` pin unit test that would require extending `test_utils` to serve transactions/results. That exact pin path (`download_ledger_data` → `download_checkpoint_ledger_data`) is fully exercised end-to-end by the integration regression test, so the additional unit test was not added.

### DIFF
--- a/crates/history/src/catchup/download.rs
+++ b/crates/history/src/catchup/download.rs
@@ -111,15 +111,23 @@ impl CatchupManager {
     ///
     /// Uses archive rotation: each attempt tries a different archive, cycling
     /// through them to provide failover when one archive is unavailable.
+    ///
+    /// Returns the parsed HAS together with the `Arc<HistoryArchive>` that
+    /// served it. Callers in the gated replay path (#2940) pin subsequent
+    /// ledger-data downloads to this exact archive so the archive that passed
+    /// the #2937 publication gate is the same one that supplies the ledger
+    /// data — closing the cross-archive bypass where a second stale-but-serving
+    /// archive could serve data that never satisfied the gate. The archive's
+    /// name (for metric labels) is derived via `archive.name()` at call sites.
     pub(super) async fn download_has(
         &self,
         checkpoint_seq: u32,
-    ) -> Result<(HistoryArchiveState, String)> {
+    ) -> Result<(HistoryArchiveState, Arc<HistoryArchive>)> {
         let num_archives = self.archives.len() as u32;
         for attempt in 0..num_archives {
             let archive = self.select_archive(attempt);
             match archive.fetch_checkpoint_has(checkpoint_seq).await {
-                Ok(has) => return Ok((has, archive.name().to_owned())),
+                Ok(has) => return Ok((has, Arc::clone(archive))),
                 Err(e) => {
                     warn!(
                         "Failed to download HAS from archive {}: {}",
@@ -328,11 +336,23 @@ impl CatchupManager {
     /// is empty when `from_ledger == 0` (genesis), when LCL sits on a
     /// checkpoint boundary (LCL-1 in a prior file, no overlap), or when
     /// there is no work to do.
+    ///
+    /// # Pinning (#2940)
+    ///
+    /// When `pinned` is `Some(archive)`, every download issued here (the LCL
+    /// header via `download_checkpoint_header` and each checkpoint payload via
+    /// `download_checkpoint_ledger_data`) is served from that archive **only**
+    /// — no rotation, no cross-archive fallback. This is used by the gated
+    /// replay path so the archive that passed the #2937 publication gate is the
+    /// same archive that supplies the ledger data. When `pinned` is `None`
+    /// (bucket-apply phase / non-gated callers), the existing fixed-order
+    /// rotation over `&self.archives` is used.
     pub(super) async fn download_ledger_data(
         &mut self,
         from_ledger: u32,
         to_ledger: u32,
         initial_lcl: LclContext,
+        pinned: Option<&Arc<HistoryArchive>>,
     ) -> Result<(Vec<LedgerData>, Vec<LedgerHeaderHistoryEntry>, String)> {
         let mut data = Vec::new();
         let mut knit_entries: Vec<LedgerHeaderHistoryEntry> = Vec::new();
@@ -367,7 +387,8 @@ impl CatchupManager {
         // We use download_checkpoint_header (lightweight single-header fetch)
         // rather than downloading the full checkpoint data.
         let mut current_lcl = if from_ledger > 0 {
-            let (lcl_header, lcl_hash) = self.download_checkpoint_header(from_ledger).await?;
+            let (lcl_header, lcl_hash) =
+                self.download_checkpoint_header(from_ledger, pinned).await?;
             LclContext::new(lcl_header.ledger_version, lcl_hash)
         } else {
             // from_ledger == 0 means "before genesis"; use caller-provided context.
@@ -380,8 +401,9 @@ impl CatchupManager {
 
             if let std::collections::hash_map::Entry::Vacant(e) = checkpoint_cache.entry(checkpoint)
             {
-                let (downloaded, archive_name) =
-                    self.download_checkpoint_ledger_data(checkpoint).await?;
+                let (downloaded, archive_name) = self
+                    .download_checkpoint_ledger_data(checkpoint, pinned)
+                    .await?;
                 last_archive_name = archive_name;
                 e.insert(downloaded);
             }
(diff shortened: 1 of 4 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Clean, well-scoped fix that pins gated-replay ledger-data downloads to the gate archive. TDD verified: regression test fails fail→pass at the pre-fix state and passes with the fix. New public surface is covered by unit + integration tests.

<details>

This is **cycle 1** (no prior Correctness comment). Complete change-list below; no concerns rise to CHANGES_REQUESTED.

**Test verification (bug-fix discipline — issue triaged `feature` but is effectively a bug-fix; PR ships a regression test):**

- `test-coverage` — VERIFIED. I checked out the PR head, reverted the load-bearing change (`download_ledger_data(..., Some(&gate_archive))` → `None`), and ran the regression test:
  - Pre-fix: `test_catchup_pins_downloads_to_gate_archive` FAILS with `KnitCurrentLedgerPrevHashMismatch { ledger: 64, actual: "abab…" }` — archive B's divergent bytes consumed. Exactly matches the PR body's claim.
  - Post-fix: test passes; catchup succeeds using archive A's data only.
  - The single squashed commit prevented a test-commit checkout, but reverting the fix line in-place is an equivalent and stronger check that the test actually captures the bug.
- The two unit tests (`test_download_has_returns_serving_archive`, `test_download_checkpoint_header_pin_targets_only_pinned`) both pass and cover: serving-archive identity, pin targets only the pinned archive, pinned-to-dead fails with no fallback, `None` preserves rotation.

**Logic / error handling — no concerns:**

- The `pinned_slice`/`archives` single-element-slice pattern in `download_checkpoint_ledger_data` and `download_checkpoint_header` is a clean way to keep the rotation loop uniform; the temporary `pinned_slice` is correctly scoped to outlive the borrow.
- The pin is re-derived per replay attempt (`download_has` re-scans from index 0), never cached across retries — so a down archive is still rotated at the attempt boundary while intra-attempt gate↔data consistency holds. Correct.
- `download_has` now returns `Arc<HistoryArchive>`; metric call sites in `mod.rs` correctly derive the name via `archive.name()`. No behavior change to metrics.
- All non-gated callers (`mod.rs` bucket-apply phases, Case-1 replay init) correctly pass `None`, preserving existing rotation. No accidental over-pinning.

**Readability:** doc comments are thorough and accurate; the `#2940` rationale block in `replay.rs` is clear about why the pin is per-attempt.

**Deviation note (acceptable):** the optional `download_checkpoint_ledger_data` pin unit test was skipped because that path is exercised end-to-end by the integration regression test. Agreed — no gap.

</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** The pin matches stellar-core's offline `mArchive` discipline (set once, never cleared by `doReset()`, threaded to all child works) and is strictly stronger than stellar-core's online catchup (nullptr → uncoordinated random per-request selection). No divergence in semantics, edge cases, or sequencing.

<details>

This is **cycle 1** (no prior Parity comment). Complete change-list below; verified against the `stellar-core/` submodule (v26.0.1).

**`parity-gap` — none found. Verified against stellar-core:**

- `src/catchup/CatchupWork.h:94` — `std::shared_ptr<HistoryArchive> mArchive;` is a single archive captured in the constructor (`CatchupWork.cpp:84 mArchive{archive}`).
- `src/catchup/CatchupWork.cpp` doReset() (lines ~120–135) resets the child works (`mGetHistoryArchiveStateWork.reset()`, etc.) but **never clears `mArchive`** — confirming the pin survives reset/retry, exactly as the PR body claims.
- `mArchive` is threaded into every child download work: `GetHistoryArchiveStateWork` (line 160/214), `DownloadVerifyTxResults` (283), bucket-state download (309), apply-buckets (375). So when `mArchive` is non-null (offline / command-line catchup with an explicit archive), every download — HAS, headers, tx, results, buckets — comes from that one archive with no rotation.
- Online catchup constructs `CatchupWork` with `mArchive == nullptr`, and child works then random-select per request. The PR's pin therefore never makes henyey *weaker* than stellar-core online; it makes the gated-replay path match the offline pinning guarantee.

**`sequencing` — correct.** The pin is derived from the same `download_has` call that feeds the #2937 publication gate (`replay.rs`), so gate-archive == data-archive within an attempt. The re-derivation per attempt (no cross-retry caching) mirrors stellar-core's "archive fixed for the lifetime of one CatchupWork run, re-evaluated on a fresh run" behavior.

**`edge-case` — scoping difference is intentional and safe.** stellar-core's `mArchive` pins the *entire* CatchupWork (including bucket apply); this PR pins only the gated-replay ledger-data downloads and passes `None` for the bucket-apply phase / non-gated callers. This is a narrower pin than stellar-core, but it is *additive safety* on the specific path (#2931 knit-to-LCL / state-wipe under archive asymmetry) and does not relax any existing guarantee — the bucket-apply phase retains the prior rotation behavior, which is unchanged from before this PR. No regression, no new divergence.

No `spec-divergence` concerns.

</details>

tomerweller:
## ✅ Merged

**Commit:** 0e3eee3e403fa5fd33ea369764a9f0c2a2f42fcd

Triple-green: Reviewer A (Correctness) APPROVE, Reviewer B (Parity) APPROVE, CI 14/14 green. External review: GH Copilot bot COMMENTED (neutral). Regression test verified fail→pass at the pre-fix state.

PR #2962, opened 2026-06-02
Strengthen CheckpointNotYetPublished test with distinct field values

tomerweller:
Closes #2953

## Summary

The `test_checkpoint_not_yet_published_is_transient` unit test in
`crates/history/src/error.rs` set `target` and `covering_checkpoint` to the
same value (62845439), so the assertion that the covering checkpoint appears
in the rendered message could pass even if the message only printed `target`.
This change uses distinct values for `target` (62845439), `covering_checkpoint`
(62845280), and `has_current` (62845375), and asserts each distinct value
appears independently in the rendered message — giving an independent guarantee
that the `covering_checkpoint` field is actually printed.

## Plan reference

Trivial short-circuit (Implementation Notes in the Triage Report on #2953).

## Test plan

- [x] cargo fmt --check
- [x] cargo clippy --all -- -D warnings
- [x] cargo test -p henyey-history --lib error:: passes (15 tests)

## Deviations from plan

None. (Triage referenced lines 740-763; the actual test is at lines 734-765 of
the current origin/main — same test, identical change.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #2953, opened 2026-06-02
Strengthen CheckpointNotYetPublished test: use distinct target vs covering_checkpoint values


tomerweller:
Follow-up from PR #2943 (issue #2941). Non-critical inline review comment that was not addressed before merge.

## Concern

`crates/history/src/error.rs:745` — The error.rs test sets `target` and `covering_checkpoint` to the **same** value (62845439), so the assertion that the value appears in the rendered message does not actually prove `covering_checkpoint` is printed — if the message accidentally only printed `target`, the test would still pass. Use **distinct** values for `target` and `covering_checkpoint` and assert the distinct numeric values each appear in the rendered message.

Note: the companion `replay.rs` test exercises the real gate path and asserts `cp == target_ckpt`, partially mitigating this, but the error.rs unit test should still be tightened for an independent guarantee.

## Source

[Original review comment](https://github.com/stellar-experimental/henyey/pull/2943#discussion_r3342436144) on PR #2943 (GH Copilot, COMMENTED).

## Severity

Low / non-blocking — test-quality improvement (class: test-coverage). Filing as backlog.

tomerweller:
## Triage Report

**Verdict:** ACCEPT (trivial — short-circuit to `ready-for-doing`)

**Type:** enhancement
**Severity:** n/a
**Crate:** crate:history
**Size:** XS

**Summary:** Strengthen unit test in crates/history/src/error.rs to use distinct target vs covering_checkpoint values, proving the covering_checkpoint field is independently printed in the error message.

## Implementation Notes

- File: `crates/history/src/error.rs` lines 740-763
- Change: In the `test_checkpoint_not_yet_published_is_transient` test, replace the duplicate value (62845439) for both `target` and `covering_checkpoint` with distinct values (e.g., target=62845439, covering_checkpoint=62845280), and assert both distinct numeric values appear in the rendered message.
- Test impact: none (test-only change, existing test tightening).

PR #2943, opened 2026-06-02
Include covering checkpoint value in CheckpointNotYetPublished error message

tomerweller:
Closes #2941

## Summary

The `CheckpointNotYetPublished` error previously reported only the catchup `target` ledger and the archive HAS `currentLedger`, forcing on-call debugging to recompute `checkpoint_containing(target)` by hand to understand the gate. This adds a `covering_checkpoint` field to the variant and surfaces it in the rendered message, so the required published frontier is reported directly.

## Plan reference

Trivial short-circuit — Implementation Notes in the [Triage Report](https://github.com/stellar-experimental/henyey/issues/2941).

## Test plan

- [x] cargo fmt --check
- [x] cargo clippy -p henyey-history --all-targets -- -D warnings (clean)
- [x] cargo test -p henyey-history (526 + targeted tests pass)

Updated `error::tests::test_checkpoint_not_yet_published_is_transient` and `catchup::replay::tests::test_knit_lcl_mismatch_unpublished_checkpoint_is_transient` to assert the covering checkpoint value now appears in the rendered message.

## Deviations from plan

None. (Triage referenced line 532; the variant had shifted to line 528 in current main but is otherwise as described.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2968, opened 2026-06-02
Derive CheckpointNotYetPublished covering checkpoint instead of storing it

### LINKED ISSUE #2950: Consider keeping CheckpointNotYetPublished variant shape stable (public-API)
Follow-up from PR #2943 (issue #2941). Non-critical inline review comment that was not addressed before merge.

## Concern

`crates/history/src/error.rs:541` — `HistoryError` is re-exported at the crate root (`pub use error::HistoryError`), making it part of the public API. Adding a new required field to the `CheckpointNotYetPublished { .. }` variant is a breaking change for downstream code that constructs or exhaustively matches this variant. Consider keeping the variant shape stable and computing `covering_checkpoint` in the `Display`/error message (or via an accessor) instead of storing it as a field.

## Source

[Original review comment](https://github.com/stellar-experimental/henyey/pull/2943#discussion_r3342436024) on PR #2943 (GH Copilot, COMMENTED).

## Severity

Low / non-blocking. `henyey-history` has no external consumers today; sole construction site updated and CI compiles green. Filing as backlog so the API-stability consideration isn't lost.

### LINKED ISSUE #2964: Use semantically valid checkpoint ledger values in CheckpointNotYetPublished test
Follow-up from PR #2962 (issue #2953). Non-critical inline review comment that was not addressed before merge.

## Concern

`crates/history/src/error.rs:750` — The test data makes `covering_checkpoint` inconsistent with the documented meaning (`checkpoint_containing(target)`) and even uses a value that is not a checkpoint ledger (not `≡ 63 (mod 64)`). This creates an unrealistic error instance (and could still pass if the message printed the wrong field labels). Use a non-checkpoint `target`, set `covering_checkpoint` to the actual covering checkpoint ledger, and keep `has_current < covering_checkpoint`.

## Source

[Original review comment](https://github.com/stellar-experimental/henyey/pull/2962#discussion_r3342935636) on PR #2962.

## Severity

Low / non-blocking — reviewer (Copilot) chose to COMMENT rather than request changes. Filing as backlog so the concern isn't lost.

### LINKED ISSUE #2965: Update string-literal assertions if CheckpointNotYetPublished test values change
Follow-up from PR #2962 (issue #2953). Non-critical inline review comment that was not addressed before merge.

## Concern

`crates/history/src/error.rs:769` — If `target`/`covering_checkpoint` values are updated to be semantically consistent, the string-literal assertions should be updated as well so they continue checking for the (distinct) per-field values.

## Source

[Original review comment](https://github.com/stellar-experimental/henyey/pull/2962#discussion_r3342935716) on PR #2962.

## Severity

Low / non-blocking — reviewer (Copilot) chose to COMMENT rather than request changes. Filing as backlog so the concern isn't lost.

tomerweller:
Closes #2950
Closes #2964
Closes #2965

## Summary

Drops the redundant stored `covering_checkpoint` field from `HistoryError::CheckpointNotYetPublished`. The value is always exactly `checkpoint_containing(target)`, so storing it inflated the crate-root-re-exported enum's public variant shape (the API-stability concern from PR #2943). The covering checkpoint is now derived from `target` — in the thiserror `#[error(...)]` message via a trailing named format arg, and through a new `HistoryError::covering_checkpoint() -> Option<u32>` accessor. The rendered message text and the publication gate's Ok/Err decision are unchanged.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/2950#issuecomment-4604714454)

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test -p henyey-history` passes (all lib + integration + doc tests)

## Follow-ups closed

This refactor supersedes two follow-ups about the now-removed stored field's test:
- **#2964** ("use semantically valid checkpoint ledger values") — the rewritten `error.rs` test now uses a semantically valid `target` (mid-checkpoint 62845437), its derived `covering` (62845439), and a `has_current` on the prior published frontier (62845375).
- **#2965** ("sync assertions if values change") — moot: the covering value is now derived from `target`, so the assertions can no longer drift out of sync with an independent stored field.

## Note on #2962 (#2953)

This intentionally rewrites the test #2962 just strengthened. #2962's "distinct value per field" premise set `covering_checkpoint` independently of `target` to prove the message rendered the *stored* field. With the field derived, a covering value distinct from `checkpoint_containing(target)` is no longer representable; the test instead achieves three distinct rendered numbers via the two intrinsic inputs and asserts the covering value through the new accessor.

## Deviations from plan

None.

### DIFF
--- a/crates/history/src/catchup/replay.rs
+++ b/crates/history/src/catchup/replay.rs
@@ -146,7 +146,6 @@ pub(super) fn checkpoint_publication_gate(has_current_ledger: u32, target: u32) 
     if has_current_ledger < target_checkpoint {
         return Err(HistoryError::CheckpointNotYetPublished {
             target,
-            covering_checkpoint: target_checkpoint,
             has_current: has_current_ledger,
         });
     }
@@ -1394,15 +1393,17 @@ mod tests {
         let err = checkpoint_publication_gate(has_current, target)
             .expect_err("gate must reject an unpublished target checkpoint");
 
+        // The covering checkpoint is derived from `target` (#2950), so the
+        // variant only binds `target`/`has_current`; assert the covering value
+        // via the accessor.
         match err {
             HistoryError::CheckpointNotYetPublished {
                 target: t,
-                covering_checkpoint: cp,
                 has_current: hc,
             } => {
                 assert_eq!(t, target);
-                assert_eq!(cp, target_ckpt);
                 assert_eq!(hc, has_current);
+                assert_eq!(err.covering_checkpoint(), Some(target_ckpt));
                 // The covering checkpoint must surface in the rendered message
                 // so on-call debugging needn't recompute checkpoint math.
                 assert!(
--- a/crates/history/src/error.rs
+++ b/crates/history/src/error.rs
@@ -525,19 +525,19 @@ pub enum HistoryError {
     /// **retried**, not treated as local-state corruption. It is therefore
     /// **excluded** from [`is_fatal_catchup_failure`](HistoryError::is_fatal_catchup_failure)
     /// and [`is_hash_mismatch`](HistoryError::is_hash_mismatch). See #2931.
+    ///
+    /// The covering checkpoint (`checkpoint_containing(target)`) is *derived*
+    /// from `target` rather than stored, so the variant's public shape stays
+    /// minimal; read it via [`covering_checkpoint`](HistoryError::covering_checkpoint).
     #[error(
         "catchup target checkpoint not yet published: target ledger {target} \
          requires archive currentLedger >= its covering checkpoint \
-         {covering_checkpoint}, but archive HAS currentLedger is {has_current}"
+         {covering}, but archive HAS currentLedger is {has_current}",
+        covering = crate::checkpoint::checkpoint_containing(*target)
     )]
     CheckpointNotYetPublished {
         /// The catchup target ledger sequence.
         target: u32,
-        /// The covering checkpoint for `target` (`checkpoint_containing(target)`),
-        /// i.e. the archive `currentLedger` value required for the target to be
-        /// considered published. Included so on-call debugging does not have to
-        /// recompute the checkpoint math from the log line.
-        covering_checkpoint: u32,
         /// The archive HAS `currentLedger` observed at gate time.
         has_current: u32,
     },
@@ -618,6 +618,23 @@ impl HistoryError {
                 | HistoryError::InvalidTxSetHash { .. }
                 | HistoryError::ReplayHashMismatch { .. }
         )
+    }
+
+    /// The covering checkpoint for a
+    /// [`CheckpointNotYetPublished`](HistoryError::CheckpointNotYetPublished)
+    /// error — `checkpoint_containing(target)`, i.e. the archive `currentLedger`
+    /// value required for the target to be considered published — or `None` for
+    /// any other variant.
+    ///
+    /// Derived from `target` rather than stored on the variant, keeping the
+    /// crate-root-re-exported enum's public shape minimal (see #2950).
+    pub fn covering_checkpoint(&self) -> Option<u32> {
+        match self {
+            HistoryError::CheckpointNotYetPublished { target, .. } => {
+                Some(crate::checkpoint::checkpoint_containing(*target))
+            }
+            _ => None,
+        }
     }
 }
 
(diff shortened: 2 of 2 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Clean refactor — drops the redundant stored `covering_checkpoint` field, derives it identically via `checkpoint_containing(target)` in both the `#[error]` message and a new accessor. Verified no behavior change at the sole caller; rendered message text is unchanged; tests rewritten correctly and pass in CI.

<details>

Issue kind: `refactor`. For a refactor the bar is: existing tests preserved and green, no observable behavior change. Verified below — all green in CI (`Test`, `Doctest`, `Clippy`, `Format`).

**Behavior equivalence (no-change verification):**
- Sole production caller is `checkpoint_publication_gate` (`crates/history/src/catchup/replay.rs:144-150`). It computes `let target_checkpoint = checkpoint_containing(target)` and previously stored exactly that into `covering_checkpoint`. The derived value — `checkpoint_containing(*target)` in the `#[error(...)]` format arg and in `HistoryError::covering_checkpoint()` — is computed from the same `target`, so it is bit-for-bit identical. No caller read `covering_checkpoint` for control flow (the publication gate's Ok/Err decision depends only on `has_current_ledger < target_checkpoint`), so dropping the field changes nothing observable.
- The second use site (`replay.rs:447-449`) constructs `target_checkpoint` only to call `download_has` and pass `target` into the gate; it never touched the field. Unaffected.

**Public error message unchanged:**
- Old: `... requires archive currentLedger >= its covering checkpoint {covering_checkpoint}, but ...` where `covering_checkpoint` was the stored field.
- New: `... {covering}, but ...` with `covering = checkpoint_containing(*target)`. The placeholder name is internal to the format string; the rendered text is byte-identical because the substituted value is identical. The test asserts the rendered message still contains the covering value.

**New public surface:** `pub fn covering_checkpoint(&self) -> Option<u32>` is tested both ways — `Some(covering)` for the variant, `None` for `FatalChainDisagreement` (`error.rs` test). Good coverage.

**Test rewrite:** The `error.rs` unit test now uses semantically valid inputs (`target` mid-checkpoint 62845437, derived `covering` 62845439, `has_current` 62845375 on the prior frontier) and `assert_ne!`s all three to preserve the #2953 "three distinct rendered numbers" guarantee without relying on an independently-stored field. The `replay.rs` test drops the removed field from the destructure and asserts via the accessor. This correctly supersedes the #2962 premise (closes #2964, #2965).

**Copilot inline (replay.rs:1407) — false positive, non-blocking:** Copilot warns `match err {...}` moves `err`, breaking the later `err.covering_checkpoint()`/`err.to_string()`. This is incorrect: the variant's only bound fields (`target: t`, `has_current: hc`) are `u32` (Copy), so the match binds by copy and does not move `err`. `err` remains fully usable in-arm and after. CI's `Test`+`Doctest` compiling-and-passing confirms it. No change needed; will be preserved as a low-priority follow-up per pipeline policy.

No correctness, error-handling, or readability concerns.

</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Low-risk refactor of an internal error type. No regression, perf, migration, or config risk. The one API-shape change (removing a public enum variant field) is the explicit intent of the issue and reduces the public surface rather than expanding it.

<details>

Risk lens — regressions, perf, breaking changes, security, operational.

**`api-shape` (the only material risk, and it's intentional):** `HistoryError` is re-exported at the crate root, so `CheckpointNotYetPublished { covering_checkpoint }` is public surface. Removing the field is technically a breaking change to that variant's shape — but that is precisely the goal of #2950 (the API-stability concern raised in PR #2943): the stored field was redundant with `target`. The replacement `covering_checkpoint() -> Option<u32>` accessor is additive and gives external readers a stable way to obtain the value without binding to the variant's internal field layout. Net effect shrinks and stabilizes the public surface. No external consumer in this workspace destructures the field (only the two in-crate sites, both updated). Acceptable and desirable.

**`regression-risk`: none.** The publication gate's Ok/Err decision is unchanged (depends only on `has_current_ledger < checkpoint_containing(target)`). The #2931 transient-classification behavior (`is_fatal_catchup_failure`/`is_hash_mismatch` exclusion) is untouched and still asserted by the test. Operationally the rendered log line is identical, so on-call dashboards/log greps keyed on the message text are unaffected.

**`perf-regression`: none.** `checkpoint_containing` is trivial integer arithmetic; computing it lazily in Display/accessor instead of storing one `u32` is negligible and only on the error path.

**`migration-risk` / `config-risk`: none.** No persisted format, no config, no schema.

**Security: none.** No new external input, no unsafe, no parsing changes.

CI is fully green (CI, CodeQL rust/python/actions, Quickstart end-to-end incl. testnet/pubnet, Gitlinks, contract checks). No risk concerns.

</details>

tomerweller:
## ✅ Merged

**Commit:** 045ddfdf02db77adcaee84ede275149c5a80b8d6

**Follow-up issues filed for unaddressed inline review comments:** #2972 (Copilot's moved-value flag on replay.rs:1407 — verified false positive; `u32` fields are `Copy` so the match doesn't move `err`; preserved as backlog)

Triple-green: Reviewer A (Correctness) APPROVE, Reviewer B (Risk) APPROVE, CI 37/37 green. Closes #2950, #2964, #2965.

PR #2967, opened 2026-06-02
Fix quickstart.yml comment overclaiming validate-contract coverage

tomerweller:
Closes #2944

## Summary

The comment above `timeout_multiplier` in `.github/workflows/quickstart.yml` claimed that multiplier drift is caught by `validate-contract`. That job only fetches/validates upstream `build.yml` and `internal-build.yml`, not `internal-test.yml` (where the multiplier lives), so it does not actually catch multiplier drift. Reworded to state that drift is caught by the harness/text assertions in `scripts/test-quickstart-harness.sh` instead.

Comment-only change; no behavior impact.

## Plan reference

Trivial short-circuit — `## Implementation Notes` in the [Triage Report](https://github.com/stellar-experimental/henyey/issues/2944).

## Test plan

- [x] actionlint on quickstart.yml (only a pre-existing SC2012 info at line 372, unrelated; exit 0)
- [x] cargo fmt --check (pre-commit hook)
- [x] cargo clippy --all (pre-commit hook)

## Deviations from plan

None.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2971, opened 2026-06-02
Anchor quickstart upstream-contract artifact and tag checks

### LINKED ISSUE #2932: upstream-contract: Anchor quickstart artifact and tag checks
Follow-up from PR #2927 (issue #2916). PR was force-converged at the lifetime bounce cap; this concern was raised by a reviewer but not resolved before merge.

## Concern class

`upstream-contract`

## Source

[Reviewer comment](https://github.com/stellar-experimental/henyey/pull/2927#issuecomment-4602922512) on PR #2927.

Reviewer: Correctness  
Verdict: CHANGES_REQUESTED

## Detail

> `.github/workflows/quickstart.yml:211-247` describes the new `ARTIFACT_REGEX` / `TAG_REGEX` checks as enforcing the exact `{tag}-{arch}` artifact and image-tag structure, but both regexes are still unanchored substring matches. I reproduced this locally with the exact regexes from the workflow: `image-quickstart-${{ matrix.image.tag }}-${{ matrix.arch }}.zip` still matches `ARTIFACT_REGEX`, and `quickstart:${{ matrix.image.tag }}-${{ matrix.arch }}-debug` still matches `TAG_REGEX`. So `validate-contract` would still accept upstream drifts that break the exact filename/tag shape this job consumes.
>
> `scripts/test-quickstart-harness.sh:454-473` only checks that the workflow contains `ARTIFACT_REGEX` / `TAG_REGEX` and greps them against `$INTERNAL`; it never proves those patterns reject incompatible suffixes. So the local harness would also stay green if this false-green hole remains.
>
> Requested follow-up:
> 1. Tighten `validate-contract` so it matches the actual `upload-artifact` name and image-tag/export lines with anchored patterns that require the full `image-quickstart-{tag}-{arch}.tar` and `quickstart:{tag}-{arch}` shapes, not just a matching prefix plus `{tag}-{arch}` substring.
> 2. Extend `scripts/test-quickstart-harness.sh` with representative negative cases (for example `.zip` and `-debug`) so a future relaxation of those checks fails locally.

## Why this is a follow-up, not a merge-blocker

PR #2927 reached the lifetime bounce cap (6 cycles since last `## Review: Reset`) without converging. Per /review-pr force-converge policy, green CI is sufficient to land the change and preserve unresolved concerns as backlog. Operator should triage this issue — close as won't-fix if the reviewer was over-strict, or schedule a follow-up PR if the concern is real.

tomerweller:
Closes #2932

## Summary

The quickstart upstream-contract gate validated the consumer-facing artifact filename and image tag with unanchored substring matches, so an upstream/local drift to an incompatible suffix (`image-quickstart-{tag}-{arch}.zip`, `.tar.gz`, or `quickstart:{tag}-{arch}-debug`) would pass validation while breaking the `download-artifact name:` and `docker run … <tag>` lines this workflow depends on.

This change stores anchored ERE regexes (`artifact_name_regex` / `image_tag_regex`) in `scripts/ci/upstream-quickstart-contract.yml` as the single source of truth, and enforces them with `grep -Eq` in both the workflow `validate-contract` job (new consumer-side block) and `scripts/test-quickstart-harness.sh`. The regexes pin the literal `image-quickstart-…-amd64.tar` prefix+suffix and `quickstart:…-amd64` tag, with trailing character classes that exclude `.` so double extensions (`.tar.gz`) are rejected.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/2932#issuecomment-4603710767)

## Test plan

- [x] `bash scripts/test-quickstart-harness.sh` — 48/48 pass (was 41; +7 new assertions)
- [x] `actionlint .github/workflows/quickstart.yml` — no new findings (the single pre-existing SC2012 on the unrelated `ls /tmp/quickstart-image` docker-load step also exists on `origin/main`)
- [x] `cargo fmt --check` / `cargo clippy --all` — pass (pre-commit hook)

## New coverage (`scripts/test-quickstart-harness.sh::test_upstream_contract_validation`)

- `contract_defines_anchored_regexes` — contract has `artifact_name_regex:` / `image_tag_regex:`
- `workflow_artifact_regex_accepts_exact_tar` / `workflow_image_tag_regex_accepts_exact_tag` — positive controls
- `workflow_artifact_regex_rejects_zip` — `.zip` rejected
- `workflow_artifact_regex_rejects_double_ext` — `.tar.gz` rejected
- `workflow_image_tag_regex_rejects_debug_suffix` — `-debug` rejected
- `workflow_uses_anchored_grep` — workflow reads the regexes and checks the consumer lines with `grep -Eq`
- The two existing `workflow_artifact_matches_contract` / `workflow_image_tag_matches_contract` checks were tightened from substring `grep -q` to anchored `grep -Eq`.

These 9 assertions were committed first (`9535fbf`) and verified FAILING on the unmodified tree, then pass after the fix (`54fea7a`) — confirming they capture the latent false-green.

## Deviations from plan

- Plan referenced HEAD using bare `grep -q 'image-quickstart'` prefix checks; the workflow had since been refactored (post-#2967) to already have upstream-template `ARTIFACT_REGEX`/`TAG_REGEX` checks. Those validate the *upstream* `internal-build.yml` template. This PR adds the missing *consumer-side* anchored check (our own `quickstart.yml` download/run lines) — the same false-green the reviewer flagged — rather than replacing the upstream-template checks.
- The contract stores the regex body with a single backslash (`\.tar`) because both consumers extract it via raw `grep | sed` with no YAML unescaping; documented inline.

### DIFF
--- a/.github/workflows/quickstart.yml
+++ b/.github/workflows/quickstart.yml
@@ -289,6 +289,43 @@ jobs:
 
           echo ""
 
+          # --- 2b. Validate OUR consumer-side artifact/tag shape (anchored) ---
+          # The checks above validate the UPSTREAM template (internal-build.yml).
+          # This block validates the EXACT, concrete artifact filename and image
+          # tag that *this* workflow consumes — the download-artifact `name:` and
+          # the `docker run ... <tag>` line — against anchored ERE regexes stored
+          # in the contract file (single source of truth, also used by
+          # scripts/test-quickstart-harness.sh). Anchoring rejects incompatible
+          # suffixes (.zip, .tar.gz, -debug) that a bare substring `grep -q`
+          # would silently accept (#2932). The contract stores the regex body
+          # verbatim (raw grep|sed extraction, no YAML unescaping).
+          echo "--- consumer-side artifact/tag shape (this workflow) ---"
+          ARTIFACT_NAME_REGEX=$(grep '^artifact_name_regex:' "$CONTRACT" | sed 's/^artifact_name_regex: *"\(.*\)"$/\1/')
+          IMAGE_TAG_REGEX=$(grep '^image_tag_regex:' "$CONTRACT" | sed 's/^image_tag_regex: *"\(.*\)"$/\1/')
+          WORKFLOW_SELF=".github/workflows/quickstart.yml"
+
+          if [[ -z "$ARTIFACT_NAME_REGEX" || -z "$IMAGE_TAG_REGEX" ]]; then
+            echo "✗ contract missing artifact_name_regex/image_tag_regex (anchored ERE)"
+            ERRORS=$((ERRORS + 1))
+          else
+            if grep -Eq "$ARTIFACT_NAME_REGEX" "$WORKFLOW_SELF"; then
+              echo "✓ workflow downloads the exact artifact shape (anchored)"
+            else
+              echo "✗ workflow has no line matching the anchored artifact regex"
+              echo "  Regex: $ARTIFACT_NAME_REGEX"
+              ERRORS=$((ERRORS + 1))
+            fi
+            if grep -Eq "$IMAGE_TAG_REGEX" "$WORKFLOW_SELF"; then
+              echo "✓ workflow runs the exact image tag shape (anchored)"
+            else
+              echo "✗ workflow has no line matching the anchored image-tag regex"
+              echo "  Regex: $IMAGE_TAG_REGEX"
+              ERRORS=$((ERRORS + 1))
+            fi
+          fi
+
+          echo ""
+
           # --- 3. Summary ---
           if [[ $ERRORS -gt 0 ]]; then
             echo "=== FAILED: $ERRORS contract violation(s) detected ==="
--- a/scripts/ci/upstream-quickstart-contract.yml
+++ b/scripts/ci/upstream-quickstart-contract.yml
@@ -27,6 +27,27 @@ image_tag_pattern: "quickstart:{tag}-{arch}"
 expected_artifact_name: "image-quickstart-testing-with-pr-amd64.tar"
 expected_image_tag: "quickstart:testing-with-pr-amd64"
 
+# Anchored ERE patterns — the single source of truth for the consumer-side
+# shape, shared by the validate-contract workflow job and the local harness
+# (scripts/test-quickstart-harness.sh). These enforce the EXACT artifact
+# filename / image-tag shape that our quickstart.yml consumes (the
+# download-artifact `name:` line and the `docker run ... <tag>` line), so a
+# future drift to an incompatible suffix is rejected rather than silently
+# accepted by a substring match:
+#   - image-quickstart-testing-with-pr-amd64.zip      -> REJECTED (not .tar)
+#   - image-quickstart-testing-with-pr-amd64.tar.gz   -> REJECTED (double ext)
+#   - quickstart:testing-with-pr-amd64-debug          -> REJECTED (tag suffix)
+# The leading/trailing character classes anchor the token at word-ish
+# boundaries (start/end of line or a non-identifier char) without requiring the
+# token to occupy the whole line, so they still match inside the real workflow
+# lines (`name: image-quickstart-...`, `docker run ... quickstart:... \`). The
+# trailing class excludes `.` so `.tar.gz` cannot slip past the `.tar` anchor.
+# Used with `grep -Eq`; keep these ERE-compatible (no PCRE constructs). NOTE:
+# both consumers extract the value with a raw `grep | sed` (no YAML parser), so
+# the quoted body is used verbatim — write a single backslash before `.` here.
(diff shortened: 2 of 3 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** TDD verified (9 assertions fail at test-first commit `9535fbf`, 48/48 pass at head `54fea7a`); anchored regexes reject `.zip`/`.tar.gz`/`-debug` and accept the exact consumer lines; the original #2932 false-green is closed.

<details>

Cycle 1 — complete change-list. Issue kind: `feature` (deferred upstream-contract concern from #2927). Test obligation: negative harness cases proving incompatible suffixes are rejected.

**Test verification (independently reproduced):**
- `test-coverage`: PASS. Walked the commit chain. `9535fbf` ("failing harness assertions first") modifies ONLY `scripts/test-quickstart-harness.sh` (+80/-8), adding 7 new assertions and tightening 2 existing `matches_contract` checks — without the contract regex fields. Ran the harness at that commit: **39/48, 9 failing** (`contract_defines_anchored_regexes`, `workflow_artifact_regex_accepts_exact_tar`, `..._accepts_exact_tag`, `..._rejects_zip`, `..._rejects_double_ext`, `..._rejects_debug_suffix`, `workflow_uses_anchored_grep`, plus the 2 tightened `workflow_artifact/image_tag_matches_contract`). The fix commit `54fea7a` adds the contract `artifact_name_regex`/`image_tag_regex` fields and the consumer-side workflow block → **48/48 pass**. Confirms the new tests capture the latent false-green rather than rubber-stamping.
- `regex-correctness`: PASS. Independently ran the contract's `artifact_name_regex` / `image_tag_regex` against fixtures. ACCEPT: `image-quickstart-testing-with-pr-amd64.tar` (bare and inside `  name: ...`), `quickstart:testing-with-pr-amd64` (bare and inside `docker run ... \`). REJECT: `.zip`, `.tar.gz`, `quickstart:...-debug`, `quickstart:...amd64.foo`. The trailing class `[^[:alnum:]_.-]` excludes `.` so the `.tar` anchor cannot be bypassed by a double extension — matches the documented intent.
- `wiring`: PASS. The workflow consumer-side block greps the anchored regexes against `.github/workflows/quickstart.yml` itself; the two real consumer lines exist (line 405 `name: image-quickstart-testing-with-pr-amd64.tar`, line 429 `quickstart:testing-with-pr-amd64 \`). The `workflow_uses_anchored_grep` assertion pins that the workflow uses `grep -Eq "$ARTIFACT_NAME_REGEX"` / `grep -Eq "$IMAGE_TAG_REGEX"`, so a regression to a bare substring check turns red.

**Error handling / readability:** `|| true` on the regex extraction with empty-string → reported failed assertion (not a `set -euo pipefail` abort) is correct. The single-backslash `\.tar` in the contract is correct given raw `grep | sed` extraction (no YAML unescaping), and is documented inline. Comments are clear and accurate.

No concerns. The PR is the minimal, correct closure of #2932.
</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** CI-only change (workflow + contract + test harness); no runtime/crate code, no API/data-format/migration impact. The new gate is strictly tighter than before and cannot produce false-reds on the current concrete consumer lines.

<details>

Cycle 1 — complete change-list. Risk lens (non-parity: touches `.github/workflows/`, `scripts/ci/`, `scripts/`).

- `regression-risk`: LOW. The change only tightens an existing validation gate (`validate-contract` job + local harness). It adds a new consumer-side anchored check and tightens two pre-existing substring checks to anchored. Verified the anchored regexes still match the real `quickstart.yml` consumer lines (405, 429), so the tightening does not introduce a false-red against the current tree. CI rollup confirms `validate-contract` and `Quickstart Harness` are both green on the PR head.
- `ci-risk` / false-positive: LOW. The regexes are concrete (`testing-with-pr`/`amd64`) and anchored at word-ish boundaries (`(^|[^[:alnum:]_-])` / `($|[^[:alnum:]_.-])`), not whole-line, so they tolerate YAML indentation and trailing `\` continuations — the actual workflow lines. No risk of breaking on benign formatting.
- `false-green` (the original risk being closed): the consumer-side gate now rejects `.zip`/`.tar.gz`/`-debug` drift that the old substring `grep -q` accepted. Net reduction in operational risk.
- `api-break` / `migration-risk` / `perf-regression`: N/A — no library/binary code, no data formats, no config consumed at runtime. The new contract fields are additive (`*_regex` alongside the retained human-readable `*_pattern`/`expected_*` fields).
- `maintenance-risk`: LOW/acceptable. The regexes are pinned to concrete `testing-with-pr-amd64` values, so a deliberate future change to the consumer tag/arch must update the contract too — but that is the intended contract behavior (single source of truth), and the harness `*_accepts_exact_*` positive controls will flag a mismatch immediately rather than silently.

No blocking risk. Approve.
</details>

tomerweller:
## ✅ Merged

**Commit:** 8b238238bb7c902024ba976994247049a7ca6474

**Follow-up issues filed for unaddressed inline review comments:** #2973

PR #2915, opened 2026-06-01
Replace all-peer broadcast with two-peer direct send in request_scp_state

tomerweller:
Closes #2911

## Summary

Replaces the residual all-peer `GetScpState` broadcast in `OverlayManager::request_scp_state()` with a stellar-core-shaped two-peer direct-send path. The method now selects from the authenticated peer set (`peer_info_cache`), shuffles the full set, truncates to `NB_PEERS_TO_ASK = 2`, and sends via direct `OutboundMessage::Send` — matching stellar-core's `getMoreSCPState()` / `getRandomAuthenticatedPeers()` / `sendGetScpState()` contract.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/2911#issuecomment-4590712460)

## Test plan

- [x] cargo fmt --check
- [x] cargo clippy -p henyey-overlay -p henyey-app -- -D warnings
- [x] cargo test -p henyey-overlay (445 tests pass)

## Regression test

- **Test:** `manager::tests::test_request_scp_state_limits_fanout_to_two_authenticated_peers`
- **Pre-fix:** committed as `dcd9cad9` — verified FAILED with: `assertion 'left == right' failed: request_scp_state must send to exactly 2 peers (left: 3, right: 2)`
- **Post-fix:** verified PASSES after `281b1a88`

## Deviations from plan

- Skipped the integration test in `overlay_scp_integration.rs` — the unit tests provide full coverage of the behavioral contract (fanout limit, authenticated-only selection, best-effort semantics) without requiring loopback infrastructure.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2986, opened 2026-06-03
Restore NotStarted guard to request_scp_state

### LINKED ISSUE #2980: request_scp_state no longer guards on self.running — document or restore NotStarted
Follow-up from PR #2896 (issue #2820). Non-critical inline review comment not addressed before merge.

## Concern

`crates/overlay/src/manager/mod.rs` — `request_scp_state()` no longer checks `self.running` (unlike `broadcast()`), so callers can get `Ok(0)` when the overlay hasn't been started or is shutting down. This changed behavior from the prior `broadcast()`-based implementation and can mask lifecycle/config issues (call sites typically log warnings only on `Err`). Consider returning `Err(OverlayError::NotStarted)` when `!self.running.load(...)` for parity with other request/broadcast APIs, or explicitly document the no-op semantics and update callers accordingly.

Note: post-merge this is now consistent with origin/main's intended `connected_peers()`-based implementation; filing for documentation/parity audit rather than as a regression.

## Source

[Original review comment](https://github.com/stellar-experimental/henyey/pull/2896#discussion_r3295031252) on PR #2896 (GH Copilot, COMMENTED/non-blocking).

## Severity

Low / non-blocking.

tomerweller:
Closes #2980

## Summary

`OverlayManager::request_scp_state()` was missing the `self.running` guard that every sibling send/connect API (`broadcast()`, `connect()`) opens with, so it silently returned `Ok(0)` when the overlay was not started or shutting down — indistinguishable from "started, no peers." This prepends the same guard, returning `Err(OverlayError::NotStarted)` in those states, restoring parity with stellar-core's shutdown guard and henyey's own overlay APIs.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/2980#issuecomment-4614586475)

## Test plan

- [x] cargo fmt --check
- [x] cargo test -p henyey-overlay passes (462 tests, 0 failures)

- **Test:** `crates/overlay/src/manager/mod.rs::test_request_scp_state_returns_not_started_when_not_started`
- **Pre-fix:** committed as `68b9ac89` — verified FAILED with: `request_scp_state should return Err(NotStarted) before start, got Ok(0)`
- **Post-fix:** verified PASSES after `62a20d7c`

Two existing tests (`test_request_scp_state_with_single_peer_sends_once`, `test_request_scp_state_with_zero_peers_returns_zero`) build an unstarted manager and were implicitly relying on the missing guard; both updated to mark the manager started before the call and still assert `Ok(1)`/`Ok(0)`.

## Deviations from plan

None.

### DIFF
--- a/crates/app/src/app/mod.rs
+++ b/crates/app/src/app/mod.rs
@@ -12457,6 +12457,10 @@ mod tests {
         let mut rx1 = overlay.inject_test_peer(peer1, 64);
         let mut rx2 = overlay.inject_test_peer(peer2, 64);
         let mut rx3 = overlay.inject_test_peer(peer3, 64);
+        // Mark the manager started so request_scp_state passes the NotStarted
+        // guard restored in #2980; otherwise the bounded pull bails early and
+        // no peers receive GetScpState.
+        overlay.set_running_for_test();
         *app.overlay.write().await = Some(Arc::new(overlay));
 
         // Record timestamp before the request.
--- a/crates/overlay/src/manager/mod.rs
+++ b/crates/overlay/src/manager/mod.rs
@@ -4349,6 +4349,28 @@ mod tests {
 
     // ──────── request_scp_state §15.3 parity tests ────────
 
+    /// Regression test for #2980: request_scp_state must return
+    /// Err(NotStarted) when the overlay has not been started, matching the
+    /// guard on broadcast()/connect() and stellar-core's shutdown guard,
+    /// instead of silently returning Ok(0) (which is indistinguishable from
+    /// "started, no peers").
+    #[test]
+    fn test_request_scp_state_returns_not_started_when_not_started() {
+        let config = OverlayConfig::default();
+        let secret = SecretKey::generate();
+        let local_node = LocalNode::new_testnet(secret);
+
+        // new() leaves running=false; the manager is never started.
+        let manager = OverlayManager::new(config, local_node).unwrap();
+
+        let result = manager.request_scp_state(42);
+        assert!(
+            matches!(result, Err(OverlayError::NotStarted)),
+            "request_scp_state should return Err(NotStarted) before start, got {:?}",
+            result
+        );
+    }
+
     #[tokio::test]
     async fn test_request_scp_state_with_single_peer_sends_once() {
         let config = OverlayConfig::default();
--- a/crates/overlay/src/manager/mod.rs
+++ b/crates/overlay/src/manager/mod.rs
@@ -1733,6 +1733,10 @@ impl OverlayManager {
     pub fn request_scp_state(&self, ledger_seq: u32) -> Result<usize> {
         use rand::seq::SliceRandom;
 
+        if !self.running.load(Ordering::Relaxed) {
+            return Err(OverlayError::NotStarted);
+        }
+
         let message = StellarMessage::GetScpState(ledger_seq);
         let peers = self.connected_peers();
         if peers.is_empty() {
@@ -4378,6 +4382,9 @@ mod tests {
         let local_node = LocalNode::new_testnet(secret);
 
         let manager = OverlayManager::new(config, local_node).unwrap();
+        // Mark the manager started so request_scp_state passes the NotStarted
+        // guard (see #2980).
+        manager.running.store(true, Ordering::Relaxed);
 
         let peer_id = PeerId::from_bytes([42u8; 32]);
         let mut rx = insert_peer_with_capacity(&manager, peer_id, 16);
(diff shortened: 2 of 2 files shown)


tomerweller (commented):
Fixed the CI-red Test failure: the cross-crate henyey-app test `test_request_scp_state_from_peers_records_attempt_and_uses_low_watermark` constructed its `OverlayManager` via `OverlayManager::new()` without starting it, so the NotStarted guard restored here made it receive 0 peer sends. Added `overlay.set_running_for_test()` before injecting it into the app. `cargo test -p henyey-app --lib` now green (963 passed). Ready for re-review.

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Restores the `!self.running → Err(NotStarted)` guard on `request_scp_state()` to match its sibling APIs; regression test is genuine (verified fails before the fix), and the two pre-existing tests that relied on the missing guard are correctly updated.

<details>

**Cycle:** 1 (no prior Correctness comment on this PR) — complete change-list below.

**Linked issue kind:** `bug-fix` (per #2980 Triage Report). Test obligation: regression test that `request_scp_state()` returns `Err(NotStarted)` before start, plus coverage of started/no-peers and started/with-peers states.

**Test verification (TDD / regression):**
- PR body `## Regression test` section present and matches the diff (test `test_request_scp_state_returns_not_started_when_not_started`).
- Commit ordering correct: test commit `4eae0955` is the parent of fix commit `0eae60e9`.
- Verified the test FAILS at the test commit (checked out `4eae0955`, ran `cargo test -p henyey-overlay test_request_scp_state_returns_not_started_when_not_started`): panicked with `request_scp_state should return Err(NotStarted) before start, got Ok(0)` — exactly the bug. The guard is added by the next commit, so the test genuinely captures the regression.
- The two existing tests (`..._with_single_peer_sends_once`, `..._with_zero_peers_returns_zero`) built an unstarted manager and implicitly relied on the missing guard. Both correctly updated to `manager.running.store(true, Ordering::Relaxed)` before the call, preserving their `Ok(1)`/`Ok(0)` assertions — the started-but-no-peers path still returns `Ok(0)`, distinct from the `Err(NotStarted)` path. Good coverage of all four states from the triage obligation.
- The app-level test in `crates/app/src/app/mod.rs` adds `overlay.set_running_for_test()` so the bounded pull no longer bails at the guard — necessary and correct given the new guard.

**Logic / error handling (concern class: `error-handling`):**
- Guard is `if !self.running.load(Ordering::Relaxed) { return Err(OverlayError::NotStarted); }`, byte-identical to `connect()` (line 1089) and `broadcast()` (line 1114). Consistent.
- All non-test call sites tolerate `Err`: `consensus.rs:1014` matches `Err(e) => warn!`, `consensus.rs:1562`/`catchup_impl.rs:2556` use `if let Err`/`let _`, and the `request_scp_state_and_record` wrapper logs. No caller treats `Err` as fatal, so surfacing `Err(NotStarted)` (instead of masking as `Ok(0)`) only improves observability — which is the issue's intent. No regression.

**Readability:** Comments on the test edits explain *why* `running` is set (reference #2980). Clear.

No concerns. APPROVE.
</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** The restored `!running → NotStarted` guard aligns `request_scp_state()` with henyey's own `connect()`/`broadcast()` guards and with stellar-core's `mShuttingDown` overlay-send guard. No semantic divergence introduced; the change moves toward parity, not away.

<details>

**Cycle:** 1 (no prior Parity comment on this PR) — complete change-list below.

**Parity analysis (concern class: `parity-gap`):**
- stellar-core `OverlayManagerImpl` guards overlay sends on `mShuttingDown` (`OverlayManagerImpl.cpp:888`: `if (mShuttingDown || !haveSpace)`), and exposes `isShuttingDown()` (line 1349). Upstream does not dispatch SCP-state requests through an overlay that is not in a live broadcasting state.
- henyey models this lifecycle with `self.running`; the sibling send/connect APIs (`broadcast()`, `connect()`) already short-circuit with `Err(OverlayError::NotStarted)` when `!running`. `request_scp_state()` was the lone outlier returning `Ok(0)` — indistinguishable from "started, no peers." Restoring the guard removes that asymmetry. This is the correct parity direction.
- The bounded two-peer selection logic (the prior #2915 / direct-send behavior) is untouched; the guard is a pure prepend before peer selection. No change to message content, peer-count semantics, or sequencing once running.

**Sequencing / edge cases (concern class: `sequencing`):**
- Started-but-no-peers still returns `Ok(0)` (verified by the updated `..._with_zero_peers_returns_zero` test) — the only newly-distinguished state is not-started, which now errors. This matches the intended state machine: NotStarted → Err, Started+0 peers → Ok(0), Started+N peers → Ok(min(2,N)).
- No data-format, XDR, or wire-message change; `GetScpState(ledger_seq)` is emitted exactly as before once the guard passes.

**Risk:** Touches only `crates/overlay/` and a test in `crates/app/`. All call sites already handle `Err` by logging. No operational, config, or migration impact.

</details>

tomerweller:
## ✅ Merged

**Commit:** 3c7d1a38efee21fda33d054ede330faee21c7395

**Follow-up issues filed for unaddressed inline review comments:** none (no unresolved inline review threads)

Triple-green: Reviewer A (Correctness) APPROVE, Reviewer B (Parity) APPROVE, CI 37/37 green. TDD verified — regression test fails on the test commit and passes after the fix.

PR #2983, opened 2026-06-03
Persist resolved workspace path as single source of truth for /do→/review-pr reap

tomerweller:
Closes #2979
Closes #2978

## Summary

Collapses the three independent session-workspace path derivations (`/do` A.2, the contract helper, `/review-pr` Step 7.4) into one producer. `do_bootstrap` now exports `DO_WORKSPACE` — the canonical, `~/data`-validated, canonicalized absolute session-workspace dir (`== WORKTREE_BASE`). `/do` persists that exact string; `/review-pr` reads and `rm -rf`s it verbatim. No component is recombined downstream, so the write path and the read path can no longer diverge regardless of `CLAUDE_SESSION_ID` state or `$HOME` poisoning — closing both #2979 (write divergence) and #2978 (stale read-path) at their shared root.

- `scripts/lib/agent-worktree-contract.sh` — `do_bootstrap` exports `DO_WORKSPACE` (alias of the validated `WORKTREE_BASE`); cleared on any rejection alongside the other derived vars.
- `.github/skills/do/SKILL.md` (A.2) — after `do_bootstrap` succeeds, persists `$DO_WORKSPACE` to a flat, session-independent marker `<real-home>/data/do-$ISSUE.workspace` (passwd-derived home, written only on success).
- `.github/skills/review-pr/SKILL.md` (Step 7.4) — reads that marker and reaps it verbatim behind a passwd-home (`getent`, not `$HOME`) `~/data` boundary guard, with a legacy `.session-id` fallback for in-flight PRs opened by the pre-fix `/do`.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/2979#issuecomment-4614482293)

## Test plan

- [x] `scripts/test-agent-worktree-contract.sh` — 31/31 pass (adds DO_WORKSPACE export, clear-on-failure, and persist/reap-contract doc assertions)
- [x] `scripts/lib/tests/test-do-bootstrap.bats` — 7/7 pass (new; wired into the `agent-worktree-contract` CI job)
- [x] `shellcheck` clean on the contract helper and harness (only pre-existing SC1090 + intentional SC2016 info)
- [x] pre-commit fmt + clippy green

## Regression test

- **Test:** `scripts/lib/tests/test-do-bootstrap.bats::do_bootstrap roundtrip identical with CLAUDE_SESSION_ID unset` (+ `DO_WORKSPACE under ~/data`)
[...]

Issue #2979, opened 2026-06-03
/do A.2 .session-id write may diverge from do_bootstrap session fallback


tomerweller:
Follow-up from PR #2977 (issue #2843). Non-critical inline review comment that was not addressed before merge.

## Concern

`.github/skills/do/SKILL.md:88` — The A.2 setup block writes `.session-id` using `${CLAUDE_SESSION_ID:-$SESSION_ID}`, but if `CLAUDE_SESSION_ID` is unset in a given runtime, `do_bootstrap` internally falls back to a timestamp (`date +%Y%m%d-%H%M%S`) for the path while the `.session-id` write may capture a different/empty value. A divergence between the path session component and the persisted marker would break /review-pr cleanup and make the documented `~/data/$SESSION_ID/...` layout inaccurate. Recommend having `do_bootstrap` export the resolved session id and writing the marker from that single source of truth.

## Source

[Original review comment](https://github.com/stellar-experimental/henyey/pull/2977#discussion_r3349571239) (GH Copilot bot) on PR #2977.

## Severity

Low / non-blocking — reviewer chose COMMENTED, not CHANGES_REQUESTED. Filing as backlog so the concern isn't lost.

tomerweller:
## Triage Report

**Verdict:** ACCEPT

**Type:** enhancement
**Kind:** refactor
**Severity:** low
**Crate:** scripts/skills
**Size:** S

**Summary:** The `/do` skill's A.2 setup block writes `.session-id` using a shell parameter expansion fallback that may diverge from the session ID that `do_bootstrap` actually resolves. Related follow-up from #2977 (issue #2843 disk-leak fix). Should be coordinated with #2978, which addresses the corresponding cleanup side in `/review-pr`.

**Test obligation:** Behavioral refactor — existing tests should pass; no new test required (session-ID consistency is already covered by the `/do` → `/review-pr` integration test flow).

**Overlap with #2978:** Both are session-ID pathway follow-ups from #2977. #2979 focuses on ensuring `/do` writes a consistent session ID; #2978 focuses on ensuring `/review-pr` can locate that ID. **Recommend fixing both in a single coordinated PR** to keep the session-ID contract in sync across skills.

**Label adjustments:** none (already has `enhancement`, `low`, `follow-up`).

tomerweller:
## 📝 Plan Draft (Round 1)

[...]

PR #3001, opened 2026-06-03
Canonicalize reap marker path before boundary check; fix bats usage comment

### LINKED ISSUE #2990: Harden /review-pr reap guard: canonicalize marker before prefix check
Follow-up from PR #2983 (issue #2979). Non-critical inline review comment (Copilot) that was not addressed before merge.

## Concern

`.github/skills/review-pr/SKILL.md:767` — The reap boundary guard is currently a string-prefix `case` check on the raw marker value. A marker containing something like `$PW_HOME/data/../<path>` would pass the `case` pattern but canonicalize outside `$PW_HOME/data`. In practice the marker is written from `do_bootstrap`'s already-canonicalized, ~/data-validated `DO_WORKSPACE`, so it can never contain `..` — this is defense-in-depth, not an exploitable path today. Hardening suggestion: canonicalize (`realpath`/`require_home_data_path`) the marker contents before the prefix check so the read side is independently safe even if a malformed marker is ever introduced.

## Source

[Original review comment](https://github.com/stellar-experimental/henyey/pull/2983#discussion) on PR #2983.

## Severity

Low / non-blocking — reviewers chose to APPROVE; Copilot left it inline. Filing as backlog so the concern isn't lost.

### LINKED ISSUE #2991: Fix misleading bats usage comment in test-do-bootstrap.bats
Follow-up from PR #2983 (issue #2979). Non-critical inline review comment (Copilot) that was not addressed before merge.

## Concern

`scripts/lib/tests/test-do-bootstrap.bats:22` — The file's usage comment advertises `bats scripts/lib/tests/test-do-bootstrap.bats`, but the file uses plain `test_*` functions, not BATS `@test` blocks. Running it under real `bats` may execute 0 tests (or misbehave). CI invokes it via `bash`, which works correctly. Either drop the misleading `bats` usage line from the header comment, or convert to genuine `@test` blocks.

## Source

[Original review comment](https://github.com/stellar-experimental/henyey/pull/2983#discussion) on PR #2983.

## Severity

Low / non-blocking — doc-accuracy nit. Filing as backlog so the concern isn't lost.

tomerweller:
Closes #2990
Closes #2991

## Summary

Two pipeline-script follow-ups from PR #2983 (issue #2979), fixed together.

**#2990 — harden the /review-pr reap guard (defense-in-depth).** The reap boundary guard in `review-pr/SKILL.md` Step 7.4 (`reap_workspace`) previously did a raw string-prefix `case "$ws" in "$PW_HOME"/data/*)` check on the marker value. A marker containing `..` (e.g. `$PW_HOME/data/../<path>`) string-prefix-matches the boundary yet resolves outside it — the raw glob would `rm -rf` an out-of-boundary path. The guard now canonicalizes the marker (`realpath -m`, with a Python fallback at known absolute paths so PATH poisoning can't substitute `realpath`) and compares the RESOLVED path against the RESOLVED `$PW_HOME/data` boundary before deciding to delete. Not exploitable today (the marker is always written from `do_bootstrap`'s canonicalized, ~/data-validated `DO_WORKSPACE`) — this makes the read side independently safe.

**#2991 — fix misleading bats usage comment.** `scripts/lib/tests/test-do-bootstrap.bats` advertised a `bats <file>` invocation in its header, but the file uses plain `test_*` bash functions with its own TAP runner and is invoked via `bash` in CI (real `bats` would execute 0 `@test` blocks). Replaced the misleading usage line and clarified the header.

## Plan reference

Both issues were trivial short-circuits — see the [Triage Report on #2990](https://github.com/stellar-experimental/henyey/issues/2990#issuecomment-4615629031) and the Triage Report on #2991.

## Test plan

- [x] `bash scripts/test-agent-worktree-contract.sh` — 32/32 pass (was 31 tests + 1 new)
- [x] `bash scripts/lib/tests/test-do-bootstrap.bats` — 7/7 pass
- [x] shellcheck clean on touched files (no new findings vs main; added test code has zero findings)
- [x] pre-commit fmt + clippy pass

## Regression test

- **Test:** `scripts/test-agent-worktree-contract.sh::test_reap_guard_canonicalizes_before_prefix_check`
- It extracts the real `reap_workspace` function from `review-pr/SKILL.md` and feeds it a marker that string-prefix-matches `$PW_HOME/data/` but resolves outside via `..`.
- **Pre-fix:** committed as `e914f7d8` — verified FAILED: `reap_workspace DELETED an outside-boundary path via ../ escape: REAP_RETURNED:0`.
- **Post-fix:** verified PASSES after `8544ae4d` (guard refuses the escape; still reaps a legitimate in-boundary dir).

## Deviations from plan

None.

### DIFF
--- a/.github/skills/review-pr/SKILL.md
+++ b/.github/skills/review-pr/SKILL.md
@@ -760,9 +760,34 @@ PW_HOME="$(getent passwd "$(id -un)" 2>/dev/null | cut -d: -f6)"
 reap_workspace() {  # $1 = absolute dir to remove
   local ws="$1"
   [ -z "$ws" ] && return 0
-  case "$ws" in
-    "$PW_HOME"/data/*) [ -d "$ws" ] && rm -rf "$ws" ;;
-    *) echo "WARN: refusing to reap '$ws' — outside $PW_HOME/data" >&2 ;;
+  # Canonicalize BEFORE the boundary prefix check (defense-in-depth, #2990).
+  # A raw string-prefix `case` would accept a marker like "$PW_HOME/data/../etc"
+  # — it begins with "$PW_HOME/data/" yet resolves outside the boundary. We
+  # resolve `..`/symlinks first (realpath -m, then a Python fallback at known
+  # absolute paths so PATH poisoning can't substitute realpath), and compare the
+  # RESOLVED path against the RESOLVED boundary so the read side is independently
+  # safe even if a malformed marker is ever introduced.
+  local canon home_data
+  canon="$(realpath -m "$ws" 2>/dev/null)" || canon=""
+  home_data="$(realpath -m "$PW_HOME/data" 2>/dev/null)" || home_data=""
+  if [ -z "$canon" ] || [ -z "$home_data" ]; then
+    local py
+    for py in /usr/bin/python3 /usr/local/bin/python3 /bin/python3; do
+      [ -x "$py" ] || continue
+      canon="$("$py" -c 'import os,sys; print(os.path.realpath(sys.argv[1]))' "$ws" 2>/dev/null)"
+      home_data="$("$py" -c 'import os,sys; print(os.path.realpath(sys.argv[1]))' "$PW_HOME/data" 2>/dev/null)"
+      break
+    done
+  fi
+  if [ -z "$canon" ] || [ -z "$home_data" ]; then
+    echo "WARN: refusing to reap '$ws' — could not canonicalize for boundary check" >&2
+    return 0
+  fi
+  # Directory-boundary check on the RESOLVED paths: accept exact <home>/data or
+  # paths strictly under it. Rejects siblings like ~/data-evil and ../ escapes.
+  case "$canon" in
+    "$home_data" | "$home_data"/*) [ -d "$canon" ] && rm -rf "$canon" ;;
+    *) echo "WARN: refusing to reap '$ws' — resolves to '$canon', outside '$home_data'" >&2 ;;
   esac
 }
 
--- a/scripts/lib/tests/test-do-bootstrap.bats
+++ b/scripts/lib/tests/test-do-bootstrap.bats
@@ -2,10 +2,12 @@
 #
 # Tests for do_bootstrap in scripts/lib/agent-worktree-contract.sh (issue #2979/#2978).
 #
-# Written in a BATS-compatible style (each test is a function named test_*) but
-# also runs as a plain-bash TAP-style runner, matching the convention used by
-# .github/skills/shared/tests/bounce-cap-check.bats. We default to the plain-bash
-# fallback since BATS isn't part of the project's required toolchain.
+# Despite the .bats extension, this file does NOT use BATS @test blocks: each
+# test is a plain bash function named test_*, and the file ships its own
+# TAP-style runner (see main() at the bottom). It is invoked with `bash`, not
+# the BATS runner — running it under real `bats` would execute 0 @test blocks.
+# This matches the convention used by .github/skills/shared/tests/bounce-cap-check.bats;
+# BATS is not part of the project's required toolchain.
 #
 # What we assert:
 #   1. With CLAUDE_SESSION_ID unset, do_bootstrap exports DO_WORKSPACE (the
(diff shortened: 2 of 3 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Defense-in-depth hardening of the reap-guard plus a doc-comment fix; both narrowly scoped to pipeline scripts/docs. Regression test verified to fail pre-fix and pass post-fix.

<details>

Cycle 1 — complete change-list. No blocking concerns; APPROVE.

Verification performed:
- **Regression test (kind: trivial/docs follow-up):** `test_reap_guard_canonicalizes_before_prefix_check` was committed in `e914f7d8` BEFORE the fix `8544ae4d`. Checked out the test-commit in an isolated worktree and ran it: `not ok 29` (FAILS — guard rm -rf'd the ../-escape path). At the fix-commit: `ok 29` (PASSES — guard refuses the escape and still reaps a legitimate in-boundary dir). The test genuinely captures the bug. Full suite at fix: 32/32 pass.
- **do-bootstrap bats:** 7/7 pass at fix-commit.

Correctness notes:
- The guard now canonicalizes both the candidate (`canon`) and the boundary (`home_data`) with `realpath -m`, with a Python `os.path.realpath` fallback at three known absolute interpreter paths — so PATH poisoning can't substitute a malicious `realpath`. Sensible: `realpath -m` does not require the path to exist, correct for a marker that may point at an already-removed dir.
- Boundary check `"$home_data" | "$home_data"/*` correctly accepts the exact data root and strict descendants, and rejects siblings like `~/data-evil` (the `/*` requires a separator). Good.
- Fail-closed on un-canonicalizable input (empty `canon`/`home_data` → warn + `return 0`, no deletion). Correct conservative default.
- The `.bats` comment fix is accurate — the file uses plain `test_*` functions and its own TAP runner, invoked via `bash`; real `bats` would run 0 `@test` blocks. The corrected usage line removes the misleading `bats <file>` invocation.

`test-coverage`: satisfied (regression test present, verified discriminating).
`doc-drift`: resolved by the comment fix itself; no new drift introduced.
`error-handling`: fail-closed path covered.

</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Low-risk pipeline-script change. The reap-guard becomes strictly more conservative (rejects ../ escapes and siblings); no behavior change for legitimate in-boundary paths. No code/data-format/API impact.

<details>

Cycle 1 — complete change-list. No blocking concerns; APPROVE.

`regression-risk`: The guard's accept set only narrows — exact `$home_data` and strict descendants still reap (verified by the test's legitimate-path sanity check); previously-accepted-but-unsafe ../-escape paths are now refused. No legitimate reap path regresses. The fail-closed branch (can't canonicalize → return 0) means at worst a stale dir is left behind, never an out-of-boundary deletion — strictly safer than before.

`config-risk` / `migration-risk`: None. No config, schema, or data-format touched. Pure shell-function logic + a doc comment.

`api-break`: None. `reap_workspace` is an internal skill-doc helper, not a public API; signature unchanged.

`perf-regression`: Negligible — one or two `realpath`/python invocations per reap call (cleanup path, not hot path).

`portability`: The Python fallback hardcodes `/usr/bin/python3 /usr/local/bin/python3 /bin/python3`; if `realpath` is present (normal case) the fallback is never reached, so this is purely a belt-and-suspenders path. Acceptable.

Scope: only `.github/skills/review-pr/SKILL.md`, `scripts/lib/tests/test-do-bootstrap.bats`, `scripts/test-agent-worktree-contract.sh`. No crate code touched → no stellar-core parity surface.

</details>

tomerweller:
## ✅ Merged

**Commit:** 1fa37595f3beed0f2fed4cc5b4f11da21aa9f7a6

Triple-green: Reviewer A (Correctness) APPROVE, Reviewer B (Risk) APPROVE, CI 37/37 green. Regression test verified failing pre-fix (`not ok 29` at e914f7d8) and passing post-fix (`ok 29` at 8544ae4d). Closes #2990 and #2991.

**Follow-up issues filed for unaddressed inline review comments (Copilot):** #3004 (reject exact `$home_data` boundary match), #3005 (invoke realpath via absolute path)

PR #3006, opened 2026-06-03
Harden reap-guard boundary and realpath canonicalization

### LINKED ISSUE #3004: reap guard: reject canon == $home_data exactly (allow only strict descendants)
Follow-up from PR #3001 (issue #2990). Non-critical inline review comment (Copilot) not addressed before merge.

## Concern

`.github/skills/review-pr/SKILL.md:791` — The boundary `case "$canon" in "$home_data" | "$home_data"/*)` allows `canon` to equal the boundary directory itself (`$home_data`). If a malformed marker ever resolved to exactly `<real-home>/data`, the guard would `rm -rf` the entire data directory (deleting unrelated workspaces). The guard should allow only paths strictly under `$home_data/` (drop the bare `"$home_data"` alternative), and ideally pass `--` to `rm`.

## Source

[Original review comment](https://github.com/stellar-experimental/henyey/pull/3001#discussion_r3351067569) on PR #3001.

## Severity

Low / non-blocking — Copilot inline; agent reviewers APPROVE'd. Marker is always written from a canonicalized per-issue DO_WORKSPACE, never the bare data root, so not reachable today — defense-in-depth on the read side.

### LINKED ISSUE #3005: reap guard: call realpath via absolute path, not PATH lookup
Follow-up from PR #3001 (issue #2990). Non-critical inline review comment (Copilot) not addressed before merge.

## Concern

`.github/skills/review-pr/SKILL.md:772` — The hardening comment claims to avoid PATH poisoning of `realpath`, but the primary path still invokes `realpath` via PATH lookup. If PATH is compromised, a fake `realpath` could return a value that passes the boundary check (the Python fallback only runs when `realpath` returns empty). Prefer invoking `realpath` via a known absolute path, or rely solely on the absolute-path Python fallback, so the comment's claim holds.

## Source

[Original review comment](https://github.com/stellar-experimental/henyey/pull/3001#discussion_r3351067614) on PR #3001.

## Severity

Low / non-blocking — Copilot inline; agent reviewers APPROVE'd. Filing as backlog so the concern isn't lost.

tomerweller:
Closes #3004
Closes #3005

## Summary

Two defense-in-depth pipeline-hardening nits carried over from PR #3001 (issue #2990), fixed together:

- **#3004** — `.github/skills/review-pr/SKILL.md` reap guard: the boundary `case` previously accepted `canon == $home_data` (the bare `<home>/data` root). A marker that ever resolved to exactly the data root would `rm -rf` every other agent's workspace. The guard now accepts **only strict descendants** under `$home_data/` and passes `--` to `rm -rf` so a dash-prefixed `canon` can't be parsed as an option. Comment updated to match.
- **#3005** — `scripts/lib/agent-worktree-contract.sh` `canonicalize_contract_path`: the fast path invoked `realpath -m` via a bare PATH lookup, so a poisoned `realpath` on PATH could feed attacker-controlled output into the trust chain (the Python fallback only runs when `realpath` returns empty, so it never engaged). The fast path now iterates known **absolute** realpath locations (`/usr/bin/realpath`, `/bin/realpath`, `/usr/local/bin/realpath`), never a PATH lookup — so the SECURITY note the function already carried now actually holds for both backends.

## Plan reference

Trivial short-circuit — both issues were ACCEPTed in triage with implementation notes (no separate `/plan` stage). See the Triage Report on #3004 and #3005.

## Test plan

- [x] `scripts/test-agent-worktree-contract.sh` — 34/34 pass (was 32 + 2 new)
- [x] `scripts/lib/tests/test-do-bootstrap.bats` — 7/7 pass
- [x] `scripts/test-review-pr-skill-snippets.sh` — 49/49 pass
- [x] `shellcheck scripts/lib/agent-worktree-contract.sh` — clean
- [x] pre-commit `cargo fmt --check` + `cargo clippy` — clean

## Regression tests

Both new tests were committed first (commit `34e7a1ba`) and verified to **FAIL on current main**, then pass after the fix (commit `1792f5d9`):

- **`test_reap_guard_rejects_exact_data_root_allows_descendant` (#3004)** — extracts the real `reap_workspace` from the skill doc; asserts the bare `<home>/data` root is refused (sentinel survives) and a strict descendant is reaped. Pre-fix failure: guard DELETED the bare data root.
- **`test_poisoned_realpath_on_path_ignored` (#3005)** — puts a malicious `realpath` (prints `/etc`) earlier on PATH; asserts `canonicalize_contract_path` never returns the poisoned value. Pre-fix failure: poisoned `realpath` was trusted.

## Deviations from plan

None.

### DIFF
--- a/.github/skills/review-pr/SKILL.md
+++ b/.github/skills/review-pr/SKILL.md
@@ -783,11 +783,14 @@ reap_workspace() {  # $1 = absolute dir to remove
     echo "WARN: refusing to reap '$ws' — could not canonicalize for boundary check" >&2
     return 0
   fi
-  # Directory-boundary check on the RESOLVED paths: accept exact <home>/data or
-  # paths strictly under it. Rejects siblings like ~/data-evil and ../ escapes.
+  # Directory-boundary check on the RESOLVED paths: accept ONLY paths strictly
+  # under <home>/data. The bare <home>/data root is rejected — a marker that
+  # resolved to exactly the data root would otherwise `rm -rf` every workspace.
+  # Also rejects siblings like ~/data-evil and ../ escapes. The `--` guards
+  # against a canon that begins with a dash being parsed as an rm option.
   case "$canon" in
-    "$home_data" | "$home_data"/*) [ -d "$canon" ] && rm -rf "$canon" ;;
-    *) echo "WARN: refusing to reap '$ws' — resolves to '$canon', outside '$home_data'" >&2 ;;
+    "$home_data"/*) [ -d "$canon" ] && rm -rf -- "$canon" ;;
+    *) echo "WARN: refusing to reap '$ws' — resolves to '$canon', outside '$home_data/'" >&2 ;;
   esac
 }
 
--- a/scripts/lib/agent-worktree-contract.sh
+++ b/scripts/lib/agent-worktree-contract.sh
@@ -35,18 +35,28 @@ _contract_real_home() {
 # Tries GNU realpath -m first, falls back to Python os.path.realpath at a known
 # absolute path, then fails closed (returns non-zero) if neither backend is available.
 #
-# SECURITY: The Python fallback uses absolute paths (/usr/bin/python3, /usr/bin/python)
-# rather than PATH lookup. This prevents PATH-poisoning attacks where a malicious
-# python3 stub could feed arbitrary output into the canonicalization trust chain.
+# SECURITY: Both backends are invoked via absolute paths (/usr/bin/realpath,
+# /usr/bin/python3, ...) rather than a bare PATH lookup. This prevents
+# PATH-poisoning attacks where a malicious `realpath`/`python3` stub on PATH
+# could feed arbitrary output into the canonicalization trust chain. A bare
+# `realpath` on the fast path would have defeated this, since the Python
+# fallback only runs when the fast path returns empty.
 canonicalize_contract_path() {
   local path="$1"
   local result
 
-  # Fast path: GNU realpath -m (works on Linux with coreutils)
-  if result="$(realpath -m "$path" 2>/dev/null)" && [[ -n "$result" ]]; then
-    echo "$result"
-    return 0
-  fi
+  # Fast path: GNU realpath -m (works on Linux with coreutils). Invoked via a
+  # known absolute path (never a PATH lookup) so a poisoned `realpath` on PATH
+  # cannot subvert the trust chain — see the SECURITY note above.
+  local rp
+  for rp in /usr/bin/realpath /bin/realpath /usr/local/bin/realpath; do
+    if [[ -x "$rp" ]]; then
+      if result="$("$rp" -m "$path" 2>/dev/null)" && [[ -n "$result" ]]; then
+        echo "$result"
+        return 0
+      fi
+    fi
+  done
 
   # Fallback: Python os.path.realpath at known absolute paths (immune to PATH poisoning)
   local py
--- a/scripts/test-agent-worktree-contract.sh
+++ b/scripts/test-agent-worktree-contract.sh
@@ -1389,6 +1389,137 @@ test_reap_guard_canonicalizes_before_prefix_check() {
 }
 
 # --------------------------------------------------------------------------
+# Test: review-pr reap guard rejects the bare <home>/data root but reaps a
+# strict descendant (issue #3004 — defense-in-depth). A marker that resolved
+# to exactly <home>/data must NOT be reaped (it would `rm -rf` every other
+# agent's workspace); only paths strictly under <home>/data/ are reapable.
+# We extract the real reap_workspace function from the skill doc and exercise
+# both the exact-root and strict-descendant cases against a sandbox home.
+# --------------------------------------------------------------------------
+test_reap_guard_rejects_exact_data_root_allows_descendant() {
+  local desc="review-pr reap guard rejects exact <home>/data, reaps strict descendant (#3004)"
+
+  local reap_fn
+  reap_fn=$(awk '
+    /^reap_workspace\(\) \{/ { capture = 1 }
+    capture { print }
+    capture && /^\}/ { exit }
+  ' "$REVIEW_PR_SKILL")
+
+  if [[ -z "$reap_fn" ]] || ! grep -q 'rm -rf' <<< "$reap_fn"; then
+    tap_not_ok "$desc" "could not extract reap_workspace() from $REVIEW_PR_SKILL"
+    return
+  fi
+
+  # Sandbox home so the data root we point the guard at is disposable — we must
+  # never risk reaping a real ~/data during the test.
+  local sandbox sandbox_data
+  sandbox="$(mktemp -d)"
+  sandbox_data="$sandbox/data"
+  mkdir -p "$sandbox_data"
+  local sentinel="$sandbox_data/SHOULD_NOT_BE_REAPED"
+  : > "$sentinel"
+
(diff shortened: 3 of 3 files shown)


tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Both fixes close real holes with no false-positive reap risk and no regression to the canonicalization fallback chain. One residual pre-existing concern is noted but not introduced by this PR.

<details>

### security — reap guard boundary (#3004)

- **Removing `"$home_data"` from the accepted arm is correct and introduces zero false-positive risk.** Legitimate workspace paths are always strict descendants (`$home_data/SESSION/...`); the bare data root is never a valid reap target. No legitimate reap can be accidentally rejected by this change.
- **Trailing-slash theoretical gap (low, not exploitable):** bash `case` pattern `"$home_data"/*` has a `*` that matches zero characters, so `$home_data/` (bare root + trailing slash only) technically matches and would be reaped. `realpath -m` never appends a trailing slash, so this is not reachable in practice. Not a blocker, but a complete fix would be `"$home_data"/?*` (require at least one character after the slash) if any future consumer changes how `canon` is produced.
- **`rm -rf --` guard is correct.** `realpath -m` output can begin with `-` only if the input does, which is possible for relative paths in weird envs; the `--` costs nothing and is the right call.
- **`$home_data` itself is derived from bare `realpath -m "$PW_HOME/data"` (line 772 of SKILL.md, unchanged).** See residual concern below.

### security — residual PATH-poisoning in SKILL.md `reap_workspace` (pre-existing, not introduced here)

- `reap_workspace` in SKILL.md computes **both** `canon` and `home_data` via bare `realpath -m` (lines 771–772 of SKILL.md, unchanged by this PR). An attacker who can poison `PATH` for the review-pr process could feed attacker-controlled values for both variables. The #3005 fix on `agent-worktree-contract.sh` does **not** cover this code path.
- The #3004 strict-descendant boundary check partially mitigates this (attacker must make both `canon` and `home_data` match), but a crafted poisoned `realpath` returning `$fake_root/subdir` for `$ws` and `$fake_root` for `$PW_HOME/data` would pass the check and delete `$fake_root/subdir`.
- This is **pre-existing** (present before this PR, not worsened by it). The correct follow-on is to apply the same absolute-path iteration introduced in #3005 to the `realpath` calls inside `reap_workspace` in SKILL.md. Recommend tracking as a follow-on issue rather than blocking this PR.

### security — canonicalize_contract_path absolute paths (#3005)

- **The hardcoded list `/usr/bin/realpath`, `/bin/realpath`, `/usr/local/bin/realpath` covers all standard Linux installations** (Debian/Ubuntu: `/usr/bin`; older BusyBox-style: `/bin`; manual coreutils install: `/usr/local/bin`). No regression for normal inputs on supported platforms.
- **Python fallback is still reachable** — the for-loop exhausts all three absolute paths before falling through to the Python chain. Confirmed by reading the function body.
- **macOS ARM (`/opt/homebrew/bin/realpath`) is not in the list.** If macOS developer machines run this helper with a Homebrew coreutils `realpath` at `/opt/homebrew`, none of the three absolute paths would match and the Python fallback would be used instead. `os.path.realpath` resolves symlinks (unlike `realpath -m`'s lenient missing-component behavior), so a non-existent path would still return a sensible canonical form — functionally equivalent for the boundary check. Not a regression, just a minor behavior difference on macOS.
- **No performance concern** — three `[[ -x ]]` stat calls on each invocation, negligible.

### regression-risk — workspace contract unchanged

- The `~/data/$SESSION_ID/...` path structure expected by `/do` and `/review-pr` is unmodified. Existing markers written before this deploy continue to be reaped correctly (they are strict descendants of `$home_data`). No migration needed.

### config-risk / operational

- SKILL.md is loaded at runtime; no deployment step is required beyond merging. The test suite (`34/34 pass`) includes both new TDD tests that were validated to fail on current main before the fix.

### test quality

- Both new tests are structurally sound: sandbox isolation, sentinel-file verification, and fail-closed detection for the poisoned-realpath case. The awk extraction of `reap_workspace` is fragile if the function header changes format, but the `grep -q 'rm -rf'` guard catches silent extraction failure.

</details>

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Both fixes are logically correct, regression tests demonstrably fail pre-fix and pass post-fix (verified by local checkout), and all 34 tests pass clean on the full PR branch. No concerns block merge.

<details>

### reap-boundary

- **Fix is correct.** Removing `"$home_data" |` from the `case` arm closes the exact-root danger: previously a marker that resolved to exactly `<home>/data` would have `rm -rf`'d every other agent's workspace. Now only `"$home_data"/*` (strict descendants) is accepted.
- Adding `--` to `rm -rf -- "$canon"` is sound defense-in-depth: prevents a dash-prefixed canon from being parsed as an rm option. Low probability but zero cost.
- WARN message updated to `outside '$home_data/'` (trailing slash) — correctly communicates the accepted prefix.

### path-hardening

- **Fix is correct.** The bare `realpath` fast path was the weak link: the Python fallback ran only when the fast path returned empty, so a poisoned `realpath` on PATH would have fed attacker-controlled output into the trust chain and the Python guard would never engage. The loop over `/usr/bin/realpath`, `/bin/realpath`, `/usr/local/bin/realpath` is equivalent in effect to the old fast path but immune to PATH manipulation.
- Path list covers Debian/Ubuntu (`/usr/bin/realpath`), older distros (`/bin/realpath`), and Homebrew/custom installs (`/usr/local/bin/realpath`). Falls through to the Python backend if none found — existing test `ok 20` (`bootstraps fall back when realpath is missing from PATH`) exercises this path.
- Inner `if result="$("$rp" -m "$path" 2>/dev/null)" && [[ -n "$result" ]]` correctly requires both non-zero exit and non-empty output before returning, so a realpath that exists but returns empty falls through to the next candidate cleanly.

### test-coverage

- Commit ordering verified: `34e7a1ba` (tests only), then `1792f5d9` (fix). Tests at `34e7a1ba` produce `not ok 30` and `not ok 31`; fix commit makes both `ok`. TDD claim holds.
- `test_reap_guard_rejects_exact_data_root_allows_descendant` correctly exercises both the exact-root-rejection and the strict-descendant-reaping cases against a disposable sandbox; sentinel-file survive-check prevents false positives.
- `test_poisoned_realpath_on_path_ignored` correctly handles both the success path (output must not be `/etc[/...]`) and the fail-closed path (acceptable, as long as nothing poisoned was emitted). Logic is sound.
- Sanity guard (`grep -q 'rm -rf' <<< "$reap_fn"`) in the first test prevents a false pass if awk extraction fails silently.

### regression-risk

- Full suite: 34/34 pass, 0 failures on the PR branch. No regressions in prior tests.

### minor observations (non-blocking)

- The awk extraction `capture && /^\}/` exits on the first line that starts with `}`. This is correct for the current `reap_workspace` body (case/esac, no bare `}` inside). Would misfire only if a nested construct closed with `}` at column 0 — not a current risk, but worth a note if the function grows.
- SKILL.md comment update accurately narrates both the new behavior and the historical `rm -rf`-the-world failure mode — good for future reviewers.

</details>

tomerweller:
## ✅ Merged

**Commit:** 1ed0d637a364f5e28eb0e1e1ce5c8ab92d11e6e3

**Follow-up issues filed for unaddressed inline review comments:** #3008, #3009

Triple-green: Reviewer A (Correctness) APPROVE, Reviewer B (Risk) APPROVE, CI 37/37 green. External Copilot review was COMMENTED (non-blocking); its two inline notes are preserved as #3008 and #3009.

Issue #2997, opened 2026-06-03
ConservationOfLumens: fail fast if updated/update_states slice lengths diverge


tomerweller:
Follow-up from PR #2989 (issue #2804). `crates/invariant/src/conservation.rs:143` — zip on updated/update_states silently truncates on length divergence; should fail fast. [Source](https://github.com/stellar-experimental/henyey/pull/2989#discussion_r3350605096). Low/non-blocking.

tomerweller:
## Triage Report

**Verdict:** ACCEPT (trivial — short-circuit to `ready-for-doing`)

**Type:** enhancement
**Severity:** low
**Crate:** crate:invariant
**Size:** XS

**Summary:** Add a length-check fail-fast guard before the zip operation on `updated`/`update_states` slices in conservation.rs:143 to error early on divergence instead of silently truncating.

## Implementation Notes

- **File:** `crates/invariant/src/conservation.rs` (line 143)
- **Change:** Before the zip, assert or check that `updated.len() == update_states.len()`, returning an error if they diverge.
- **Test impact:** Single assertion in the existing invariant test; no new test file needed.
- **Overlap note:** Related to siblings #2998 (checked-sum for inflation_payouts) and #2999 (message wording) — all three touch conservation.rs. Can be serialized or batched in a single fix PR.

PR #3011, opened 2026-06-03
Count SAC native-XLM ContractData balances in ConservationOfLumens

### LINKED ISSUE #2987: ConservationOfLumens: count SAC ContractData native balances in per-op delta
Follow-up from #2804. The initial `ConservationOfLumens` invariant counts native balances in Account, ClaimableBalance, and LiquidityPool entries, but defers Stellar Asset Contract (SAC) `ContractData` native balances.

## Why this matters (corrected from plan Round 1)

Critic B (parity) corrected the original reasoning: an Account→Contract XLM transfer via the native SAC produces a non-zero `deltaBalances` from the *Account* side alone, while the offsetting `ContractData` credit is uncounted. Under a per-operation zero-sum check this is a **false alarm** (spurious violation), not a missed imbalance. Because the invariant is non-strict it degrades to log noise rather than a crash, but it is still incorrect parity vs. stellar-core's `getAssetBalance` (`TransactionUtils.cpp`), which counts the SAC `ContractData` native balance.

## Scope

Extend `calculate_delta_balance` (henyey `crates/invariant/src/conservation.rs`) to count the SAC native balance held in `ContractData` entries, mirroring stellar-core `getAssetBalance` for `CONTRACT_DATA` (contractID == native asset contract, key vec[0] == balance symbol, val SCV_I128 amount). Requires computing the native asset contract ID from the network ID (stellar-core `getAssetContractInfo`).

## Acceptance

- Account↔Contract native SAC transfer nets `deltaBalances == 0` and does not fire the invariant.
- Test covering a SAC native transfer pair.

### LINKED ISSUE #3007: ConservationOfLumens: inflation payout sum overflow message says "Overflow" for both over/underflow
Follow-up from PR #3000 (issue #2997). Non-critical inline review comment that was not addressed before merge.

## Concern

`crates/invariant/src/conservation.rs:191` — `checked_add` returns `None` on both overflow and underflow, but the error message here always says "Overflow". If a payout amount is ever negative (corrupt meta / unexpected variant), this could report the wrong condition. Consider using a message that covers both (or matching the earlier Overflow/Underflow split used for `deltaBalances`).

## Source

[Original review comment](https://github.com/stellar-experimental/henyey/pull/3000#discussion_r3351065533) on PR #3000.

## Severity

Low / non-blocking — reviewer (GH Copilot bot) chose COMMENTED rather than request changes. Inflation-payout amounts are non-negative in practice (and the inflation path is dead in 24+), so underflow is not reachable on the live path; this is a message-precision nit. Filing as backlog so the concern isn't lost.

tomerweller:
Closes #2987
Closes #3007

## Summary

Extends `ConservationOfLumens::native_balance` to recognize the native Stellar-Asset-Contract's `Balance` `ContractData` entries, mirroring stellar-core's `getAssetBalance` CONTRACT_DATA branch (`TransactionUtils.cpp`). An entry is the native SAC balance iff its `contract` is `Contract(native_sac_id)` (id derived per-op from `delta.network_id` via `HashIdPreimage::ContractId` / `ContractIdPreimage::Asset(Native)` — identical to core's precomputed `mLumenContractInfo`), its `key` is a non-empty `Vec` with `v[0] == Symbol("Balance")`, and its `val` is a non-empty `Map` with `m[0].key == Symbol("amount")` and an `I128` value. An i128 amount out of `i64` range (`hi > 0 || lo > i64::MAX`) fails the invariant with core's verbatim `"Could not calculate lumen balance delta for an entry"`; any other structural mismatch contributes 0. This lets Account↔Contract native SAC transfers net to zero instead of firing false-alarm violations.

Also folds in sibling nit #3007 (same file): the inflation-payout sum guard now reports `"Underflow"` for a negative addend instead of always saying `"Overflow"` (`checked_add` returns `None` for both directions).

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/2987#issuecomment-4617226288)

## Test plan

- [x] cargo fmt --check
- [x] cargo clippy -p henyey-invariant --all-targets -- -D warnings
- [x] cargo test -p henyey-invariant — 16/16 pass (5 new SAC tests + all existing)

## New coverage (feature)

Five SAC tests in `crates/invariant/tests/conservation.rs` (committed first as `c7afa048`, verified 3 FAILED on `origin/main`: the two transfer tests saw uncounted ContractData → false-alarm violation, and the i128-overflow test went undetected). All pass after the implementation commit `c8753084`:
- `test_conservation_sac_native_transfer_account_to_contract`
- `test_conservation_sac_native_transfer_contract_to_account`
- `test_conservation_sac_unrelated_contract_id_ignored`
- `test_conservation_sac_malformed_key_or_val_ignored`
- `test_conservation_sac_i128_overflow_detected`

## Deviations from plan

None.

### DIFF
--- a/crates/invariant/Cargo.toml
+++ b/crates/invariant/Cargo.toml
@@ -6,6 +6,7 @@ edition.workspace = true
 license.workspace = true
 
 [dependencies]
+henyey-common.workspace = true
 stellar-xdr.workspace = true
 tracing.workspace = true
 metrics.workspace = true
@@ -13,3 +14,4 @@ serde_json.workspace = true
 regex = "1"
 
 [dev-dependencies]
+henyey-common.workspace = true
--- a/crates/invariant/src/conservation.rs
+++ b/crates/invariant/src/conservation.rs
@@ -20,13 +20,11 @@
 //! - `LiquidityPool` constant-product `reserveA`/`reserveB` — for whichever leg
 //!   is the native asset.
 //!
-//! `ContractData` (Stellar Asset Contract native balances) is NOT counted in
-//! this implementation. An Account↔Contract native SAC transfer would show a
-//! non-zero Account-side delta with the offsetting `ContractData` credit
-//! uncounted, producing a spurious violation (a **false alarm**, not a missed
-//! imbalance). Because this invariant is non-strict, that is log noise rather
-//! than a crash. Counting SAC `ContractData` native balances is tracked as a
-//! follow-up in #2987.
+//! - `ContractData` (Stellar Asset Contract native balances) — only the native
+//!   SAC's `Balance` entries, mirroring core's `getAssetBalance` CONTRACT_DATA
+//!   branch (see [`native_balance`]). The native SAC contract id is derived
+//!   per-op from `delta.network_id` (a pure function of the network passphrase,
+//!   identical to core's precomputed `mLumenContractInfo`).
 //!
 //! # Parity
 //!
@@ -36,8 +34,9 @@
 //! hook). Strictness: non-strict (`Invariant(false)`).
 
 use stellar_xdr::curr::{
-    Asset, ContractEvent, LedgerEntry, LedgerEntryData, LiquidityPoolEntryBody, Operation,
-    OperationResult, OperationResultTr,
+    Asset, ContractEvent, ContractId, ContractIdPreimage, Hash, HashIdPreimage,
+    HashIdPreimageContractId, LedgerEntry, LedgerEntryData, LiquidityPoolEntryBody, Operation,
+    OperationResult, OperationResultTr, ScAddress, ScVal,
 };
 
 use crate::{Invariant, OperationDelta};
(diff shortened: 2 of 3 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The SAC native-XLM `ContractData` counting is a faithful, line-by-line mirror of stellar-core's `getAssetBalance` CONTRACT_DATA branch; the contract-id derivation matches core and the existing in-tree pattern; the 5 new tests are non-vacuous and exercise the path; CI is fully green. Two low-severity nits (one untested defensive branch, one duplicated helper) — neither blocks merge.

**Cycle 1** (no prior Correctness verdict found) — complete change-list below, grouped by concern class.

### Logic / parity — PASS
Verified `sac_native_balance` against `stellar-core/src/transactions/TransactionUtils.cpp:2078-2118`:
- Contract-address gate (`SC_ADDRESS_TYPE_CONTRACT` + id equality), key `Vec(Some)` non-empty with `v[0] == Symbol("Balance")`, val `Map(Some)` non-empty with `m[0].key == Symbol("amount")` and `I128` value — all match.
- Overflow gate `parts.hi > 0 || parts.lo > i64::MAX as u64` mirrors core's `lo > max_i64 || hi > 0` (OR is order-independent). The `hi > 0` (not `hi != 0`) choice is intentional parity and the code comment is accurate.
- Core's distinction between "asset not matched" (`{false,false,nullopt}`) and "matched-but-no-balance" (`{false,true,nullopt}`) correctly collapses to `NativeBalance::None`/0 here, since both contribute 0 to the delta — no observable difference for this invariant.
- `native_sac_contract_id` matches core's `getAssetContractID` (`HashIdPreimage::ContractId` ↔ `ENVELOPE_TYPE_CONTRACT_ID`, `ContractIdPreimage::Asset(Native)` ↔ `CONTRACT_ID_PREIMAGE_FROM_ASSET`) and is byte-identical to the established `crates/tx/src/events.rs::get_asset_contract_id`.
- `calculate_delta_balance` returns core's verbatim `"Could not calculate lumen balance delta for an entry"` when either side overflows; both-`None` precondition preserved.

### Test coverage — PASS (one minor gap)
- No new `pub` surface introduced (all of `NativeBalance`, `native_sac_contract_id`, `sac_native_balance`, and the changed signatures of `native_balance`/`calculate_delta_balance` are private); the new logic is exercised end-to-end through `check_on_operation_apply` by the 5 integration tests.
- Tests are **non-vacuous**: `DeltaBuilder::check` passes `network_id = [0u8;32]` and the SAC entries derive their id from the same `[0u8;32]`, so the contract-id gate genuinely matches. The unrelated-id / malformed-key/val tests would each fire a violation if the gate were broken (a standalone +500 create), confirming the negative branches.
- Both overflow branches (`hi > 0` and `lo > i64::MAX`) are asserted against the exact error string.
- **Nit (test-coverage, low):** the new `#3007` `Underflow` message branch (`p.amount < 0`) has no test — the existing `test_conservation_inflation_payout_sum_overflow_detected` feeds two positive `i64::MAX` payouts and hits the `>= 0` branch. Likewise the prev-side overflow path in `calculate_delta_balance` (overflow arriving via `update_states`/`delete_states` rather than `created`) is only covered on the `current` side. Both are symmetric dead/defensive code, so not blocking — but a ~10-line underflow test (e.g. payouts `[i64::MIN, -1]`) would close it cheaply.

### #3007 inflation message fix — PASS
Behavior-preserving: `checked_add` still returns `None` (→ error) in exactly the same cases; only the human-readable string now branches on sign. This qualifies as a message-only change. Comment correctly notes payouts are non-negative in practice and the branch is dead in 24+.

### Readability / reuse — PASS (one nit)
- **Nit (reuse, low):** the SAC contract-id derivation now exists in three places — `crates/tx/src/events.rs` (private), `crates/invariant/src/conservation.rs`, and the invariant test file. The tx version is crate-private so direct reuse isn't possible today; consider hoisting a single `get_asset_contract_id` into `henyey-common` as a follow-up to prevent drift. Not required for this PR.
- The explicit `NativeBalance::None | NativeBalance::Overflow => 0` arms after the early `Overflow` return are slightly redundant but keep the match exhaustive and readable — fine as-is.

### Error handling — PASS
Overflow surfaces as a typed `Err` with the verbatim core string; the `updated`/`update_states` length-divergence guard (#2997) is preserved; accumulation retains checked add/sub guards.

### CI — PASS
All checks green (Test, Clippy, Format, Doctest, Build). No `ci-failure` concern.

**No NEW concern classes beyond test-coverage / reuse, both low-severity.**

</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE
**Summary:** The SAC native-XLM `ContractData` gate is a faithful, gate-for-gate mirror of stellar-core's `getAssetBalance` CONTRACT_DATA branch and `getAssetContractID`; the overflow semantics, contract-id derivation, and verbatim delta-failure strings all match. The #3007 inflation message refinement is consistent with henyey's existing (intentional) divergence. No parity gaps that block merge.

**Cycle 1** (no prior Parity comment). Complete change-list with every concern labeled by class.

### Reference points (stellar-core v26.0.1)
- `src/invariant/ConservationOfLumens.cpp` — `calculateDeltaBalance` (L18-44), `checkOnOperationApply` (L89-177).
- `src/transactions/TransactionUtils.cpp` — `getAssetContractInfo` (L2004), `getAssetContractID` (L2020), `getAssetBalance` CONTRACT_DATA branch (L2078-2118).

### Change-list & parity verification

**1. Contract-id derivation — MATCH.**
`native_sac_contract_id` builds `HashIdPreimage::ContractId{ network_id, ContractIdPreimage::Asset(Asset::Native) }` and SHA-256s the XDR. Core's `getAssetContractID` builds `ENVELOPE_TYPE_CONTRACT_ID` + `CONTRACT_ID_PREIMAGE_FROM_ASSET` + `fromAsset = Asset(ASSET_TYPE_NATIVE)` and `xdrSha256`. Verified `henyey_common::Hash256::hash_xdr` = SHA-256 over `write_xdr` with `Limits::none()` (`crates/common/src/types.rs:124-145`), byte-identical to `xdrSha256`. Per-op derivation vs core's precomputed `mLumenContractInfo` is behaviorally identical (pure fn of the passphrase); `network_id` is correctly threaded from `apply.rs:817` (`&self.network_id.0 .0`). ✅

**2. CONTRACT_DATA gate — MATCH, gate for gate.**
- contract is `Contract(id)` && `id == native_sac_id`  ↔ core `type()==SC_ADDRESS_TYPE_CONTRACT && contractId()==mAssetContractID`. ✅
- key is `Vec(Some(v))`, `v.first()==Symbol("Balance")` ↔ core `key.type()==SCV_VEC && vec() && size!=0 && at(0)==mBalanceSymbol`. (`.first()` collapses the non-empty + at(0) checks; equivalent.) ✅
- val is `Map(Some(m))`, `m.first().key==Symbol("amount")`, `.val==I128` ↔ core `val.type()==SCV_MAP && map() && size!=0 && at(0).key==mAmountSymbol && val.type()==SCV_I128`. Core checks only `at(0)`; henyey checks only `.first()` — same. The real SAC `BalanceValue` serializes keys sorted (`amount`<`authorized`<`clawback`), so `amount` is genuinely at index 0; relying on `first()` is correct. ✅

**3. authorized / clawback / unauthorized balances — MATCH.** Neither core nor henyey inspect the `authorized`/`clawback` fields of `BalanceValue`; both count `amount` regardless of auth state. Unauthorized SAC balances are counted by both. ✅ (Native XLM has no issuer, so "issuer's own balance" is not a special case in either.)

**4. i128 overflow gate — MATCH (incl. the quirks).** `parts.hi > 0 || parts.lo > i64::MAX` ↔ core `hi > 0 || lo > int64_max`. The inline comment is accurate: `hi` is `i64` in both, so `hi > 0` is signed-identical. Confirmed the matching corner case: a negative i128 (`hi < 0`) is **not** flagged as overflow by either side and both return `lo as i64` — henyey preserves this exact (arguably odd) core behavior rather than "fixing" it. ✅

**5. matched-but-malformed entry → contributes 0 — MATCH (for the delta path).** Core returns `{false,true,nullopt}` (assetMatched, no balance) at L2117; `calculateDeltaBalance` treats `assetMatched && balance` only, so nullopt → 0. henyey returns `NativeBalance::None` → 0. The `{false,true,nullopt}` vs `{false,false,nullopt}` distinction only matters in core's *snapshot* path (`processEntryIfNew`), which this PR does not implement (henyey registers only via the `checkOnOperationApply`/`OperationDelta` hook). No divergence in the implemented path. ✅

**6. Account / ClaimableBalance / LiquidityPool legs — unchanged, still MATCH.** Refactor from `Option<i64>` to the `NativeBalance` enum is behavior-preserving; native gating on CB asset and LP leg selection is identical to core L2057-2077. ✅

**7. Delta-failure strings — MATCH verbatim.** `"Could not calculate lumen balance delta for an entry"` (core L109), `"Overflow detected when adding to deltaBalances"` (L116), `"Underflow detected when adding to deltaBalances"` (L121). ✅

**8. #3007 inflation message fix — acceptable (spec-divergence, pre-existing, immaterial).** Class: *spec-divergence (non-blocking)*. Core sums payouts with plain `std::accumulate(lhs + rhs.amount)` — **no overflow check and no message at all**. henyey's `checked_add` + "Overflow/Underflow detected when summing inflation payouts" strings have **no core counterpart**; this is a language-driven divergence that predates this PR (the overflow string already existed; the PR only adds the underflow variant). It is unreachable in practice (total XLM supply ~1e18 ≪ i64::MAX ~9.2e18), and for all reachable inputs henyey's `checked_add` behaves like core's raw add. No action needed, but flagging that "matching core's message" is moot here — core has none.

### Out-of-scope observation (not introduced by this PR)
- *sequencing*: `apply.rs:815-816` passes `header_current: None / header_previous: None`, so the operation-apply path never exercises core's `totalCoins`/`feePool` header-delta checks (`ConservationOfLumens.cpp:98-99, 154-167`). This is pre-existing henyey behavior in an unchanged file and outside this PR's diff — noting only because it bounds how much of core's `checkOnOperationApply` is mirrored. Not a blocker for #3011.

### Build / tests
`cargo test -p henyey-invariant --test conservation` on the PR head (`c8753084`, isolated worktree): **16/16 pass**, including SAC transfer both directions, unrelated-contract-id ignored, malformed key/val ignored, i128 overflow (both `hi>0` and `lo>i64::MAX`), and non-i128 amount → 0. Parity-relevant edge cases are well covered.

**Conclusion:** Parity is faithful and exact. APPROVE.

</details>

tomerweller:
## ✅ Merged

**Commit:** 68756ec9a58636bb070b13c0e691202494068ac6

Triple-green: Reviewer A (Correctness) APPROVE, Reviewer B (Parity vs stellar-core getAssetBalance/ConservationOfLumens) APPROVE, CI 37/37 green. Closes #2987 and #3007.

**Follow-up issues filed for unaddressed inline review comments:** none (no inline threads; reviewers' low-severity nits were non-blocking APPROVE notes).

PR #3128, opened 2026-06-04
Send responder HELLO before version/network/port validation (#3067)

### LINKED ISSUE #3067: spec-adhere/OVERLAY OVERLAY §4.4.2-6: Handshake ordering — responder HELLO send precedes …
## Spec-adherence finding

**Spec:** [`stellar-specs/OVERLAY_SPEC.md`](../blob/main/stellar-specs/OVERLAY_SPEC.md) §4.4.2
**Claim ID:** `OVERLAY §4.4.2-6` (MUST)
**Classification:** **Drift**
**Severity:** high

### Expected (per spec)
Responder sends HELLO immediately after AuthCert verification and ban check, then runs version/network/port checks; if they fail, send ERROR_MSG then drop.

### Actual (in Rust)
Responder's recv_hello completes the network ID and overlay version checks inside auth.process_hello() before send_hello() is called; on mismatch the responder never sends HELLO back, only an ERR_CONF.

### Rust location
`crates/overlay/src/peer.rs:387 (recv_hello → process_hello calls auth.process_hello which performs network ID and version checks before send_hello at line 444)`

### Notes
Spec is explicit: HELLO echo must precede the version/self/network/port checks so the remote's state machine can process the subsequent ERROR_MSG correctly. Reorder process_hello: split network/version checks out and run them only after send_hello on the responder path.

---

tomerweller:
Closes #3067

## Summary

OVERLAY §4.4.2-6 requires the responder to echo its HELLO **before** the overlay-version / self-connection / network-ID / listening-port checks, so the remote — still awaiting an unauthenticated HELLO — can decode the subsequent (seq-0, zero-MAC) ERROR_MSG. Previously the responder ran the network-ID and overlay-version checks inside `auth.process_hello()` before `send_hello()`, so on a mismatch it emitted only an ERR_CONF and never the HELLO.

This reorders the responder branch of `Peer::handshake` to: `recv_hello` (phase-1: AuthCert verify + X25519 key derivation + `HelloReceived`) → ban check → `send_hello` → phase-2 (version → self → network → port; on failure `try_send_err_conf` then drop). The initiator runs phase-1 + phase-2 back-to-back (it already sent HELLO first) and is observably unchanged. ERROR framing stays unauthenticated (seq 0, zero MAC), matching stellar-core `Hmac.cpp` `setAuthenticatedMessageBody` which exempts ERROR_MSG and HELLO. Matches stellar-core `Peer::recvHello` (Peer.cpp:1778–1915).

`AuthContext::process_hello` is split into `process_hello_phase1` and `validate_overlay_version` / `validate_network_id` (with a `validate_hello_post_send` wrapper retaining the combined semantics); `peer.rs` gains `process_hello_phase1` / `validate_hello_phase2`.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3067#issuecomment-4618920276)

## Test plan

- [x] cargo fmt --check
- [x] cargo test -p henyey-overlay passes (460 lib tests + integration + doctests)

- **Tests:** `crates/overlay/src/peer.rs::test_responder_sends_hello_before_error_on_{version,network,port}_mismatch` (plus `auth.rs::test_process_hello_phase1_succeeds_before_version_check`)
- **Pre-fix:** committed as `f874fde5` — verified the three responder tests FAILED with "responder must send HELLO then ERROR, got 1 frame(s): [\"ERROR\"]" (responder emitted ERROR_MSG as the first/only frame)
- **Post-fix:** verified all PASS after `4ac1d3e0`
- **Initiator guard:** `peer.rs::test_initiator_still_rejects_version_mismatch` confirms the initiator path still rejects an incompatible overlay version.

## Deviations from plan

None.

### DIFF
--- a/crates/overlay/src/auth.rs
+++ b/crates/overlay/src/auth.rs
@@ -464,11 +464,19 @@ impl AuthContext {
     /// malformed, too old, or too new; `NetworkMismatch` if the network IDs
     /// differ.
     pub fn validate_hello_post_send(&self, hello: &Hello) -> Result<()> {
-        // Check overlay version range compatibility (matching stellar-core Peer::recvHello)
-        // Three conditions that cause version rejection:
-        // 1. Peer's range is malformed (min > max)
-        // 2. Peer's max version is below our minimum (peer too old)
-        // 3. Peer's min version is above our maximum (peer too new)
+        // Auth-level phase-2 checks, version before network (matching the
+        // stellar-core ordering). The peer-level self-connection check is
+        // interposed between these two by the responder caller (peer.rs).
+        self.validate_overlay_version(hello)?;
+        self.validate_network_id(hello)
+    }
+
+    /// Overlay-version range check (auth-level phase-2).
+    ///
+    /// Returns `VersionMismatch` if the peer's advertised overlay version range
+    /// is malformed (min > max), too old (max below our minimum), or too new
+    /// (min above our maximum). Matches stellar-core `Peer::recvHello`.
+    pub fn validate_overlay_version(&self, hello: &Hello) -> Result<()> {
         if hello.overlay_min_version > hello.overlay_version {
             return Err(OverlayError::VersionMismatch(format!(
                 "peer overlay version range malformed: min {} > max {}",
@@ -487,13 +495,17 @@ impl AuthContext {
                 hello.overlay_min_version, self.local_node.overlay_version
             )));
         }
+        Ok(())
+    }
 
-        // Check network ID
+    /// Network-ID check (auth-level phase-2).
+    ///
+    /// Returns `NetworkMismatch` if the peer is on a different network.
+    pub fn validate_network_id(&self, hello: &Hello) -> Result<()> {
         let network_id_bytes = hello.network_id.0;
         if network_id_bytes != *self.local_node.network_id.as_bytes() {
             return Err(OverlayError::NetworkMismatch);
         }
-
         Ok(())
     }
 
--- a/crates/overlay/src/auth.rs
+++ b/crates/overlay/src/auth.rs
@@ -367,6 +367,33 @@ impl AuthContext {
     /// - Auth certificate signature is invalid or expired
     /// - Key derivation fails
     pub fn process_hello(&mut self, hello: &Hello) -> Result<()> {
+        // Combined wrapper preserving the original semantics for the initiator
+        // path and tests: phase-1 (cert + keys + state) followed by phase-2
+        // (overlay-version + network-ID checks).
+        //
+        // The responder path (peer.rs) runs these phases separately so it can
+        // echo its own HELLO between phase-1 and phase-2 — OVERLAY §4.4.2-6.
+        self.process_hello_phase1(hello)?;
+        self.validate_hello_post_send(hello)
+    }
+
+    /// Phase-1 of HELLO processing: AuthCert verification, X25519 key exchange,
+    /// MAC-key derivation, and the `HelloReceived` state transition.
+    ///
+    /// This MUST run before the responder echoes its own HELLO so that the
+    /// MAC keys exist (and a duplicate HELLO is rejected) — but it deliberately
+    /// does NOT run the overlay-version or network-ID checks. Those are deferred
+    /// to [`validate_hello_post_send`] (phase-2), matching stellar-core
+    /// `Peer::recvHello` which sends HELLO before those checks (OVERLAY §4.4.2-6).
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    /// - Called in an invalid state (not `Initial`/`HelloSent`) — INV-O3/INV-O4
+    /// - The peer's public key is malformed
+    /// - The auth certificate signature is invalid or expired
+    /// - Key derivation fails (including a small-order shared secret)
+    pub fn process_hello_phase1(&mut self, hello: &Hello) -> Result<()> {
         // State guard: process_hello is only valid in Initial or HelloSent states.
         // Rejecting other states prevents MAC key clobbering from a duplicate HELLO
         // (INV-O3) and enforces the handshake state machine (INV-O4).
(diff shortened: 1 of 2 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Clean phase-split refactor of the HELLO handshake; the bug-fix is well-targeted, the regression tests fail pre-fix and pass post-fix, and the new public surface is fully covered.

<details>

**Cycle 1** — complete change-list. No blocking concerns. The items below are observations / non-blocking notes (would be inline, not gating).

- PR body has a populated `## Regression test` section. The three responder tests were committed FIRST in `f874fde5` (test-only commit), before the fix in `4ac1d3e0`.
- Verified directly against the merged-with-`origin/main` tree:
  - At `f874fde5` (pre-fix), `cargo test -p henyey-overlay --lib responder_sends_hello_before_error` FAILS all three with `"responder must send HELLO then ERROR, got 1 frame(s): [\"ERROR\"]"` — exactly the message claimed in the PR body. So the regression test genuinely captures the bug.
  - At the merged head, all three PASS, plus `test_initiator_still_rejects_version_mismatch`, `test_process_hello_phase1_succeeds_before_version_check`, and `test_validate_hello_post_send_rejects_version_and_network`.
  - Full overlay lib suite: **460 passed, 0 failed.**

**New public surface — all covered:**
- `AuthContext::process_hello_phase1` / `validate_hello_post_send` / `validate_overlay_version` / `validate_network_id` — exercised by `test_process_hello_phase1_succeeds_before_version_check` and `test_validate_hello_post_send_rejects_version_and_network` (covers too-old / too-new / malformed-range / network-mismatch / compatible).
- `Peer::process_hello_phase1` / `validate_hello_phase2` — exercised end-to-end via the responder handshake tests.

**Logic / error-handling:**
- `process_hello` is retained as a combined wrapper (`phase1` + `validate_hello_post_send`) so the initiator path and existing callers are semantically unchanged. Good backward-compat choice.
- `recv_hello` now returns the `Hello` so the caller can run phase-2 against it — clean signature change, threaded correctly through both branches.
- Responder failure path: on phase-2 failure inside the `async { }` block, `try_send_err_conf` runs then returns `Err`, and the outer handler releases the pending peer-id reservation only if owned. Correct.
- The `debug!("Received Hello...")` log moved into phase-1 after `self.info` is populated — fine, fields are set before the log.

**Non-blocking note (inline-class `parity-detail`, recorded for follow-up, not gating):** `self.info.address` port is now set at the END of phase-2 (after the port check), whereas stellar-core sets `mAddress` port before the port-validity check. On a failing handshake the peer is dropped, so the stale port is never observed; on success it is set identically. No functional divergence.

**Readability:** doc comments on every new method explain the phase split and cite OVERLAY §4.4.2-6 + stellar-core line refs. Comments in the handshake branches explain the reorder rationale. Good.

</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** The reordered responder handshake matches stellar-core `Peer::recvHello` (Peer.cpp:1778–1915) step-for-step: cert+keys+state → ban → sendHello → version → self → network → port. ERROR_MSG/HELLO framing stays unauthenticated, matching `Hmac.cpp:73`. The initiator path is observably unchanged.

<details>

**Cycle 1** — complete parity change-list. No `parity-gap` / `spec-divergence` / `sequencing` / `edge-case` concerns survive; details below.

**Sequencing parity vs stellar-core `Peer::recvHello` (REMOTE_CALLED_US):**
Mapped the C++ in `stellar-core/src/overlay/Peer.cpp:1778–1915` against the PR responder branch:

| stellar-core step | PR equivalent |
|---|---|
| `GOT_HELLO` state guard | `process_hello_phase1` state guard (`PeerState::Handshaking`) + auth `Initial`/`HelloSent` guard |
| `verifyRemoteAuthCert` | `AuthContext::process_hello_phase1` (cert verify) |
| `isBanned` check | ban check after `recv_hello`, before `send_hello` ✅ matches position |
| set versions/peerID, derive MAC keys, `setState(GOT_HELLO)` | `process_hello_phase1` (key derivation + `HelloReceived` + `self.info` population) |
| `sendHello()` (because REMOTE_CALLED_US) | `send_hello()` ✅ before any validation |
| overlay version range → ERR_CONF | `validate_overlay_version` (phase-2 step 1) ✅ |
| self-connection → ERR_CONF | self-check (phase-2 step 2) ✅ |
| networkID → ERR_CONF | `validate_network_id` (phase-2 step 3) ✅ |
| listeningPort ≤0 / >UINT16_MAX → ERR_CONF | port check (phase-2 step 4) ✅ |

Order is **version → self → network → port**, identical to upstream. The comment in `validate_hello_phase2` correctly notes the self-connection check is interposed between the auth-level version and network checks to preserve this exact ordering.

**Unauthenticated ERROR framing (the crux of #3067):**
- `try_send_err_conf` and `send_hello` both route through `send_raw` → `auth.wrap_unauthenticated`, so both go on the wire with seq 0 / zero MAC.
- This matches `stellar-core/src/overlay/Hmac.cpp:73`: `setAuthenticatedMessageBody` sets the MAC sequence only `if (msg.type() != HELLO && msg.type() != ERROR_MSG)` — both message types are exempt. The whole point of #3067 is that the responder must echo the unauthenticated HELLO first so the peer can decode the subsequent unauthenticated ERROR_MSG; the PR achieves exactly this.

**Initiator path (WE_CALLED_REMOTE):**
- The initiator sends HELLO first, then runs phase-1 (`recv_hello`) + phase-2 (`validate_hello_phase2`) back-to-back, sending ERR_CONF + dropping on failure. stellar-core runs the identical validation block for both roles, only gating `sendHello()` on `REMOTE_CALLED_US` and `sendAuth()` on `WE_CALLED_REMOTE`. The PR preserves this — `test_initiator_still_rejects_version_mismatch` confirms the initiator still rejects an incompatible version (returns `VersionMismatch`). No divergence.

**Edge cases checked:**
- Malformed version range (min > max), too-old, too-new all produce `VersionMismatch` (matches the three-way C++ condition).
- Network mismatch → `NetworkMismatch` → ERR_CONF. Port 0 / >65535 → `InvalidMessage` → ERR_CONF.
- Duplicate-HELLO protection preserved: phase-1 still rejects a second HELLO via the auth state guard (INV-O3/O4), matching the C++ `getState >= GOT_HELLO` drop.

</details>

tomerweller:
## ✅ Merged

**Commit:** 2d5467f0c925a297a8d62cd73ecd6d2103f6695f

**Follow-up issues filed for unaddressed inline review comments:** none (no inline review threads)

Reviewers: Correctness APPROVE, Parity APPROVE. CI green (36 success, 1 skipped, 0 failing). Self-modifying gate did not apply (PR touches `crates/overlay/` only). Merged via the shared `attempt_merge` helper (squash + admin) under bash.

PR #3146, opened 2026-06-04
Fire started_ballot_protocol from bump_to_ballot with the ballot (SCP §9.10-1)

### LINKED ISSUE #3089: spec-adhere/SCP SCP §9.10-1: started_ballot_protocol fired from wrong site (misses recove…
## Spec-adherence finding

**Spec:** [`stellar-specs/SCP_SPEC.md`](../blob/main/stellar-specs/SCP_SPEC.md) §9.10
**Claim ID:** `SCP §9.10-1` (MUST)
**Classification:** **Drift**
**Severity:** medium

### Expected (per spec)
startedBallotProtocol MUST be notified from bumpToBallot when current_ballot transitions from null, with the new ballot.

### Actual (in Rust)
Rust calls started_ballot_protocol from Slot::check_nomination_to_ballot with the composite candidate value (not the ballot) BEFORE calling ballot.bump. Other entry points to bump_to_ballot (set_state_from_envelope from recovery, direct bump_state) never fire the callback.

### Rust location
`crates/scp/src/slot.rs:650`

### Notes
Move the callback into BallotProtocol::bump_to_ballot guarded on the prior current_ballot.is_none() flag; pass the actual ballot (not composite value). Ensure recovery and direct bump_state paths fire it appropriately, or document why they shouldn't.

---

tomerweller:
Closes #3089

## Summary

Relocates the `started_ballot_protocol` SCP driver callback into `BallotProtocol::bump_to_ballot`, firing it exactly once per slot on the `current_ballot` null→set transition with the actual `ScpBallot` — matching stellar-core `BallotProtocol::bumpToBallot` (BallotProtocol.cpp:469-474, guarded on `!mCurrentBallot`). Previously the callback fired from `Slot::check_nomination_to_ballot` with the composite nomination *value*, and the recovery path (`set_state_from_envelope`) assigned `current_ballot` directly and never fired the callback at all. Now every production first-set of `current_ballot` (nomination, follower-receive, recovery) flows through the single firing site. The driver-trait argument changes `&Value`→`&ScpBallot` to match `SCPDriver.h:227-230`; `ctx`/driver is threaded through `bump_to_ballot`/`update_current_value`/`update_current_if_needed` and both `set_state_from_envelope` levels. `force_externalize` (catchup, no core analog) intentionally stays unfired.

## Plan reference

[Converged Plan comment]()

## Test plan

- [x] cargo fmt --check
- [x] cargo clippy --all -- -D warnings (workspace gate; clean — verified via pre-commit hook)
- [x] cargo test -p henyey-scp (375 lib + 181 integration, all pass)
- [x] cargo test -p henyey-herder (1176+ pass)

- **Tests:**
  - `crates/scp/src/ballot/mod.rs::test_set_state_from_envelope_fires_started_ballot_protocol_{prepare,confirm,externalize}`
  - `crates/scp/src/ballot/mod.rs::test_started_ballot_protocol_fires_once`
  - `crates/scp/src/slot.rs::test_nomination_ballot_start_fires_with_ballot_not_value`
- **Pre-fix:** with the firing logic in its pre-fix form (callback not fired from `bump_to_ballot`; recovery assigning `current_ballot` directly), all five tests FAIL — recorder length 0 (callback never fires). The recovery routing was independently verified load-bearing: reverting just the recovery `bump_to_ballot` call (fire left enabled) fails the recovery test.
- **Post-fix:** all five PASS after 6607e568ef96f25ebe3c933cb4c951e39ff6f83d.

## Deviations from plan

- Per the skill's TDD step, the failing test is normally its own commit. The trait-signature change (`&Value`→`&ScpBallot`) and the `ctx` threading are infra the new tests' assertions depend on to compile, so the regression tests and the fix ship in a single commit. The required fail-pre/pass-post property was still verified explicitly (documented above) before committing.

### DIFF
--- a/crates/herder/src/scp_driver.rs
+++ b/crates/herder/src/scp_driver.rs
@@ -47,8 +47,8 @@ use henyey_crypto::{PublicKey, SecretKey, Signature};
 use henyey_ledger::LedgerManager;
 use henyey_scp::{SCPDriver, SlotIndex, ValidationLevel};
 use stellar_xdr::curr::{
-    EnvelopeType, LedgerUpgrade, NodeId, ReadXdr, ScpEnvelope, ScpQuorumSet, ScpStatement,
-    StellarValue, StellarValueExt, Value, WriteXdr,
+    EnvelopeType, LedgerUpgrade, NodeId, ReadXdr, ScpBallot, ScpEnvelope, ScpQuorumSet,
+    ScpStatement, StellarValue, StellarValueExt, Value, WriteXdr,
 };
 
 use crate::error::HerderError;
@@ -3965,7 +3965,7 @@ impl SCPDriver for HerderScpCallback {
         self.driver.record_nomination_start(slot_index);
     }
 
-    fn started_ballot_protocol(&self, slot_index: u64, _value: &Value) {
+    fn started_ballot_protocol(&self, slot_index: u64, _ballot: &ScpBallot) {
         self.driver.record_ballot_start(slot_index);
     }
 
--- a/crates/scp/src/ballot/mod.rs
+++ b/crates/scp/src/ballot/mod.rs
@@ -850,15 +850,24 @@ impl BallotProtocol {
     /// # Panics
     /// Panics if the envelope contains a non-ballot statement type (Nominate).
     /// Callers must route nomination envelopes to the nomination protocol.
-    pub(crate) fn set_state_from_envelope(&mut self, envelope: &ScpEnvelope) -> Result<(), String> {
+    pub(crate) fn set_state_from_envelope<D: SCPDriver>(
+        &mut self,
+        envelope: &ScpEnvelope,
+        ctx: &SlotContext<'_, D>,
+    ) -> Result<(), String> {
         if self.current_ballot.is_some() {
             return Err("Cannot set state after starting ballot protocol".into());
         }
 
+        // Recovery first-sets `current_ballot` through `bump_to_ballot` (BEFORE
+        // assigning `phase`), so the `started_ballot_protocol` callback fires for
+        // recovery-driven slots too. Matches stellar-core `setStateFromEnvelope`
+        // (BallotProtocol.cpp:1754/1778/1789), which calls `bumpToBallot` for all
+        // three pledge types before setting `mPhase`.
         match &envelope.statement.pledges {
             ScpStatementPledges::Prepare(prep) => {
                 let value = prep.ballot.value.clone();
-                self.current_ballot = Some(prep.ballot.clone());
+                self.bump_to_ballot(&prep.ballot, true, ctx);
                 self.prepared = prep.prepared.clone();
                 self.prepared_prime = prep.prepared_prime.clone();
                 if prep.n_c != 0 {
@@ -877,7 +886,7 @@ impl BallotProtocol {
             }
             ScpStatementPledges::Confirm(conf) => {
                 let value = conf.ballot.value.clone();
-                self.current_ballot = Some(conf.ballot.clone());
+                self.bump_to_ballot(&conf.ballot, true, ctx);
                 self.prepared = Some(ScpBallot {
                     counter: conf.n_prepared,
                     value: value.clone(),
@@ -900,10 +909,14 @@ impl BallotProtocol {
                     counter: ext.n_h,
                     value: value.clone(),
                 });
-                self.current_ballot = Some(ScpBallot {
-                    counter: u32::MAX,
-                    value: value.clone(),
-                });
+                self.bump_to_ballot(
+                    &ScpBallot {
+                        counter: u32::MAX,
+                        value: value.clone(),
+                    },
+                    true,
+                    ctx,
+                );
                 self.prepared = Some(ScpBallot {
                     counter: u32::MAX,
                     value: value.clone(),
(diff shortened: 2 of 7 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Relocates `started_ballot_protocol` into the single `bump_to_ballot` null→set firing site, threading `ctx`/driver cleanly through the call graph. The 5 regression tests fail pre-fix and pass post-fix; both halves (fire site + recovery routing) verified independently load-bearing.

<details>

Cycle 1 — complete change-list. No blocking concerns found.

- PR body has a `## Regression test` section listing 5 tests matching the diff. Verified.
- Tests + fix ship in one commit (justified: the `&Value`→`&ScpBallot` trait change and `ctx` threading are compile-time infra the new assertions depend on). The required fail-pre/pass-post property was re-verified explicitly:
  - All 5 PASS on the PR head merged with current `origin/main`.
  - Disabling the fire inside `bump_to_ballot` (`current_ballot.is_none()` arm) → all 5 FAIL (recorder length 0). This confirms the firing-site half is load-bearing.
  - Reverting only the recovery PREPARE routing (`bump_to_ballot` → direct `current_ballot = Some(...)`) → `..._prepare` recovery test FAILS while the rest pass. Confirms the recovery routing is independently load-bearing, matching the PR body's claim.

**Correctness (`logic`):**
- Fire is placed BEFORE `self.current_ballot = Some(ballot.clone())` and gated on `self.current_ballot.is_none()` — correct null→set semantics; cannot double-fire because once set, the guard is false on every subsequent `bump_to_ballot`.
- `ctx` is threaded through `bump_to_ballot`, `update_current_value`, `update_current_if_needed`, and both `set_state_from_envelope` levels (Slot + BallotProtocol). All call sites updated; `scp.rs` sources the driver via `self.driver()`. Compiles clean, CI green.
- The old firing site in `slot.rs` (`driver.started_ballot_protocol(self.slot_index, &composite)`) is removed and replaced with an accurate comment describing the new path (`bump → bump_state → update_current_value → bump_to_ballot`).

**Error handling / readability (`readability`):**
- `set_state_from_envelope` retains its `current_ballot.is_some()` early-Err guard, so recovery only fires once.
- Test helpers (`ssfe!` macro, `bare_ctx_parts`) are well-scoped to the recovery/unit tests that don't assert on the callback; clean way to absorb the signature change without noise.
- Doc comments on the trait method and the firing site cite the exact stellar-core lines.

No `MINOR` inline concerns to file.

</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** The change is an exact parity match with stellar-core `BallotProtocol::bumpToBallot` (BallotProtocol.cpp:469-474) and `setStateFromEnvelope` (1754/1778/1789). Firing site, guard, argument type, and sequencing all mirror upstream.

<details>

Cycle 1 — complete parity audit. No divergence found.

**Firing site (`parity-gap` — none):**
- stellar-core (BallotProtocol.cpp:469-474):
  ```cpp
  if (!mCurrentBallot) {
      mSlot.getSCPDriver().startedBallotProtocol(mSlot.getSlotIndex(), ballot);
  }
  mCurrentBallot = makeBallot(ballot);
  ```
- henyey (state_machine.rs):
  ```rust
  if self.current_ballot.is_none() {
      ctx.driver.started_ballot_protocol(ctx.slot_index, ballot);
  }
  self.current_ballot = Some(ballot.clone());
  ```
  Identical guard (`!mCurrentBallot` ↔ `current_ballot.is_none()`), identical ordering (fire BEFORE assignment), identical argument (the new `ScpBallot`, not the composite value). Matches.

**Driver trait (`spec-divergence` — none):**
- `started_ballot_protocol(&self, u64, &ScpBallot)` matches `SCPDriver.h:227-230` `startedBallotProtocol(uint64, SCPBallot const&)`. The prior `&Value` signature was the divergence this PR closes (§9.10-1).

**Recovery routing / sequencing (`sequencing` — none):**
- core `setStateFromEnvelope` calls `bumpToBallot(b, true)` for PREPARE, `bumpToBallot(c.ballot, true)` for CONFIRM, `bumpToBallot(SCPBallot(UINT32_MAX, v), true)` for EXTERNALIZE — each BEFORE `mPhase = ...`. The henyey diff now calls `bump_to_ballot(..., true, ctx)` for all three before assigning `phase`. The EXTERNALIZE case uses `counter: u32::MAX` matching `UINT32_MAX`. Matches all three, in order.
- `check=true` is passed in all recovery calls, matching core's `bumpToBallot(b, true)`. Since `current_ballot` is None at recovery entry (guarded by the early Err), the monotonic-check assertion is a no-op and the fire arm is taken — correct.

**Edge cases (`edge-case` — none):**
- `force_externalize` (catchup) assigns `current_ballot` directly and never reaches `bump_to_ballot` → does not fire. Correct: stellar-core has no analog for this catchup helper, so firing there would be a *spurious* notification not present upstream. Intentional non-fire is the parity-correct choice.
- Fire-once across in-phase counter bumps: once `current_ballot` is set, subsequent `bump_to_ballot` calls (counter bumps) skip the fire arm — matches core, where `mCurrentBallot` is non-null after the first set.

No parity concerns to file.

</details>

tomerweller:
## ✅ Merged

**Commit:** d33d90cc875dfb0bf598fc55b0b3e144c18e9693

Triple-green: Reviewer Correctness APPROVE, Reviewer Parity APPROVE, CI green (37/37). Exact stellar-core parity verified against BallotProtocol.cpp:469-474 + setStateFromEnvelope (1754/1778/1789); 5 regression tests confirmed fail-pre/pass-post.

Issue #2105, opened 2026-04-29
[AUDIT-216] XDR-Length-Sorted Nomination Values Diverge From Stellar-Core Ordering


tomerweller:
## Audit Finding

**Source file**: `crates/scp/src/nomination.rs`
**Crate**: `scp`
**Severity**: HIGH
**Source**: Multi-stage security audit (skill: /audit-gesserit)

---

# H-016: XDR-Length-Sorted Nomination Values Diverge From Stellar-Core Ordering

**Verdict**: VIABLE  
**Date**: 2026-04-29  
**Crate**: scp  
**Severity**: HIGH  
**Reviewed by**: GitHub Copilot CLI  
**Failed Suppression Rules**: None

## Summary

The hypothesis is viable. Henyey sanity-checks and emits `SCPNomination.votes` / `accepted` using the full XDR serialization of each opaque `Value`, while stellar-core requires native `Value` ordering. For variable-length opaque values these orders differ because XDR serialization places the 32-bit length before the payload. This produces a concrete NOMINATE accept/reject split between Henyey and stellar-core.

## Code Trace

- Incoming Henyey NOMINATE envelopes reach `SCP::receive_envelope()` and `Slot::process_envelope()`, which dispatch `ScpStatementPledges::Nominate(_)` to `NominationProtocol::process_envelope()` (`crates/scp/src/scp.rs:208-220`, `crates/scp/src/slot.rs:300-314`).
- `NominationProtocol::process_envelope()` rejects before recording if `is_sane_statement()` returns false (`crates/scp/src/nomination.rs:475-485`).
- `is_sane_statement()` calls `is_sorted_unique()` on both vectors. `is_sorted_unique()` compares `value_key(value)`, and `value_key()` is `henyey_common::xdr_stream::xdr_to_bytes(value)` (`crates/scp/src/nomination.rs:781-810`).
- Local Henyey emission uses the same relation: `emit_nomination()` builds `votes` and `accepted` with `sorted_values()`, and `sorted_values()` sorts/dedups by `value_key()` (`crates/scp/src/nomination.rs:661-668`, `812-816`).
[...]

PR #3148, opened 2026-06-04
Gate self-emitted nomination through freshness + sanity in emit_nomination

### LINKED ISSUE #3090: spec-adhere/SCP SCP §8.7-1: emit_nomination skips full self-processEnvelope self-check
## Spec-adherence finding

**Spec:** [`stellar-specs/SCP_SPEC.md`](../blob/main/stellar-specs/SCP_SPEC.md) §8.7
**Claim ID:** `SCP §8.7-1` (MUST)
**Classification:** **Partial**
**Severity:** medium

### Expected (per spec)
emitNomination MUST re-enter the full processEnvelope path (sanity + freshness + state transitions) and surface INVALID as a fatal state-machine bug.

### Actual (in Rust)
emit_nomination calls attempt_promote directly, bypassing is_newer_nomination_internal and is_sane_statement checks; INVALID detection is silently absent because no INVALID can be returned by the partial path.

### Rust location
`crates/scp/src/nomination.rs:671-735`

### Notes
Either invoke process_envelope on the self-envelope end-to-end with a self=true flag (matching ballot::emit_current_state pattern) or document why the bypass is safe. Self-processing-bug detection is the value here, not state transitions.

---

tomerweller:
Closes #3090

## Summary

`emit_nomination` (`crates/scp/src/nomination.rs`) previously routed the self-emitted nomination through `record_local_nomination` (freshness check only, silent on failure) then `attempt_promote` directly — it never called `is_sane_statement`. This diverges from stellar-core `NominationProtocol::emitNomination` -> `processEnvelope(env, self=true)` (`NominationProtocol.cpp:147-189`), which runs `isNewerStatement` -> `isSane` -> `recordEnvelope` and throws on any non-VALID result. A non-sane self-statement (empty votes AND accepted) was therefore recorded/promoted/emitted with no bad-state surfacing.

This change inserts a freshness gate (`is_newer_nomination_internal`) then a sanity gate (`is_sane_statement`) before record/promote/emit. A non-sane self-statement is surfaced via `tracing::error!` and returns without emitting (logged, not panic — matching the ballot `emit_current_state` precedent). The freshness-false branch remains a benign silent monotonic stop; the freshness-vs-sanity asymmetry vs stellar-core (which throws on both) is documented inline. Wire-emit behavior is unchanged.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3090#issuecomment-4622984849)

## Test plan

- [x] cargo fmt --check
- [x] cargo test -p henyey-scp passes (372 lib + 50 + 7 + 124 integration tests)

- **Test:** `crates/scp/src/nomination.rs::test_emit_nomination_rejects_non_sane_self_statement`
- **Pre-fix:** committed as `afab12bf` — verified FAILED with: `non-sane self-statement must not be emitted (emit_count left: 1, right: 0)`
- **Post-fix:** verified PASSES after `314ae546`
- Also adds `test_emit_nomination_stale_self_statement_does_not_re_emit` pinning the freshness early-return (monotonic stop) behavior.

## Deviations from plan

None.

### DIFF
--- a/crates/scp/src/nomination.rs
+++ b/crates/scp/src/nomination.rs
@@ -228,6 +228,30 @@ impl NominationProtocol {
         self.previous_value = value;
     }
 
+    /// Test helper: force `votes`/`accepted`/`started` so a test can drive
+    /// `emit_nomination` into states the public `nominate` path cannot reach
+    /// (e.g. an empty, non-sane self-statement). The public path always grows
+    /// `votes` before emitting and re-sorts/dedups via `sorted_values`, so the
+    /// only reachable non-sane self condition (empty votes AND accepted) cannot
+    /// be produced through `nominate`. See issue #3090.
+    #[cfg(test)]
+    pub(crate) fn force_nomination_state_for_test(
+        &mut self,
+        votes: Vec<Value>,
+        accepted: Vec<Value>,
+        started: bool,
+    ) {
+        self.votes = votes;
+        self.accepted = accepted;
+        self.started = started;
+    }
+
+    /// Test helper: directly invoke the self-emit gate path (`emit_nomination`).
+    #[cfg(test)]
+    pub(crate) fn emit_nomination_for_test<D: SCPDriver>(&mut self, ctx: &SlotContext<'_, D>) {
+        self.emit_nomination(ctx);
+    }
+
     /// Get the last envelope constructed by this node.
     pub fn get_last_envelope(&self) -> Option<&ScpEnvelope> {
         self.last_envelope.as_ref()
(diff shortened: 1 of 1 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Cycle 1, complete review. `emit_nomination` now gates the self-built nomination through freshness then sanity before record/promote/emit; the bug-fix is backed by a real regression test that fails pre-fix and passes post-fix. Wire-emit for valid nominations is unchanged. No blocking concerns.

<details>

**kind:** bug-fix (from issue #3090 Triage Report). Regression-test obligation applies.

**test-coverage — PASS**
- PR body has a populated `## Regression test` section naming `test_emit_nomination_rejects_non_sane_self_statement` plus a freshness pin `test_emit_nomination_stale_self_statement_does_not_re_emit`.
- Commit ordering correct: test committed first (`afab12bf`), fix second (`314ae546`).
- Verified the test **FAILS at the pre-fix commit** `afab12bf`: `assertion left == right failed: non-sane self-statement must not be emitted, left: 1, right: 0` — i.e. pre-fix the empty self-statement was emitted. Exactly matches the PR body's claim.
- Verified both new tests **PASS at the fix commit** `314ae546`.
- Full `henyey-scp` suite green at the fix commit: 372 + 50 + 7 + 124 (+5 ignored), 0 failures — matches the PR's claimed counts.

**error-handling — PASS**
- Non-sane self-statement is surfaced via `tracing::error!` + early `return` (no panic). This is the right call for a validator: a self-inconsistency bug should not crash the node and break liveness/determinism. Matches the documented ballot `emit_current_state` precedent.
- The freshness-false branch is a silent early return (benign monotonic stop). Documented inline.

**logic — PASS**
- Gate 1 (`is_newer_nomination_internal`) and Gate 2 (`is_sane_statement`) both run on the freshly-built `nomination` before `record_local_nomination` / `attempt_promote` / emit. Both return early, so no record/promote/emit happens on a rejected self-statement — the test asserts all four post-conditions (no emit, no `latest_nominations` entry, no candidates, no `last_envelope`).
- Wire-emit path for a *valid* nomination (record → attempt_promote cascade → `is_newer` recheck → `emit_envelope`) is byte-for-byte unchanged below the new gates. Confirmed by reading the full function.
- The freshness check inside `record_local_nomination` is now redundant with Gate 1; the diff documents it as harmless/idempotent. Correct — it's a second monotonic guard, not a behavior change.

**api-shape — PASS (non-blocking note)**
- New `force_nomination_state_for_test` / `emit_nomination_for_test` are `#[cfg(test)] pub(crate)` only — no production surface added. Appropriate, since the non-sane condition is unreachable through the public `nominate` path (documented on the helper).

**readability — PASS**
- The two-paragraph block comment over the gates explains the stellar-core mapping and the deliberate asymmetry; this is exactly the kind of inline rationale that prevents a future parity audit from re-flagging the design.
</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** Cycle 1, complete review. The new gate order (freshness → sanity → record) exactly mirrors stellar-core `NominationProtocol::processEnvelope` (`NominationProtocol.cpp:409-420`), reached from `emitNomination` via `processEnvelope(env, self=true)`. The one divergence (log+return instead of `throw`) is a deliberate, documented liveness choice consistent with the ballot precedent. No parity gap.

<details>

**spec-divergence / sequencing — PASS**
- Verified against the pinned `stellar-core/` submodule (`src/scp/NominationProtocol.cpp`):
  - `emitNomination()` (lines 147-189) builds the self-statement, calls `processEnvelope(envW, true)`, and only sets `mLastEnvelope`/emits if the result is `VALID`; otherwise `throw std::runtime_error("moved to a bad state (nomination)")`.
  - `processEnvelope()` (lines 404-420) runs `isNewerStatement` → return INVALID; then `isSane` → return INVALID; then `recordEnvelope`.
- The PR inserts **Gate 1 = `is_newer_nomination_internal` (freshness), Gate 2 = `is_sane_statement` (sanity)** in that order, before `record_local_nomination`. This matches stellar-core's `isNewerStatement`-then-`isSane`-then-`recordEnvelope` ordering precisely. (Note: the ballot `emit_current_state` runs sanity-before-freshness, mirroring a different upstream function `isStatementSane`; the nomination path here correctly follows `processEnvelope`'s order, not the ballot's.)

**edge-case — PASS**
- `is_sane_statement` mirrors stellar-core `isSane`: rejects empty-votes-AND-empty-accepted, and requires sorted-unique votes/accepted (`is_sorted_unique` uses native `Value` `Ord` = lexicographic byte compare, matching `stellar::Value operator<`; the #2105 fix is preserved). The only reachable non-sane self condition is empty-votes-AND-empty-accepted (since `sorted_values` guarantees sorted-unique), which is exactly what the regression test forces.
- `is_newer_nomination` mirrors `isNewerStatement` (subset + strict growth on votes or accepted).

**parity-gap (documented divergence, NOT a gap) — PASS**
- stellar-core `throw`s on a non-VALID self-process (both freshness-false and sanity-false). henyey instead: freshness-false → silent return (benign monotonic stop); sanity-false → `tracing::error!` + return. This is a deliberate divergence (do not crash the validator on a self-inconsistency / preserve determinism), documented inline and consistent with the existing ballot `emit_current_state` precedent (`ballot/envelope.rs:158-185`). The observable consensus output is identical for valid nominations and strictly more conservative (no bad envelope emitted) for the should-not-happen non-sane case — so no consensus divergence vs upstream's effective behavior, which also does not emit the bad envelope.

**wire behavior — PASS**
- For a valid (sane, newer) nomination the record → cascade → `is_newer` recheck → `emit_envelope` flow is unchanged. No change to envelope bytes or emission gating for the normal path.
</details>

tomerweller:
## ✅ Merged

**Commit:** f36997c614dc48596b133cba36aa407f8f7fc6e7

**Follow-up issues filed for unaddressed inline review comments:** none (no inline comments)

Reviewer A (Correctness): APPROVE. Reviewer B (Parity): APPROVE. No CHANGES_REQUESTED findings (refute pass not triggered). No external reviews. CI green (37/37). Self-modifying gate: not applicable (touches `crates/scp/src/nomination.rs` only).

Issue #3153, opened 2026-06-04
Add DISABLE_BUCKET_GC config kill-switch (parity with stellar-core)


tomerweller:
Follow-up from #3028 (GC cadence parity).

stellar-core's `forgetUnreferencedBuckets` is gated on a `DISABLE_BUCKET_GC` config flag (`BucketManager.cpp:985` — `!mConfig.DISABLE_BUCKET_GC`). Henyey has no equivalent kill-switch: GC always runs. #3028 brings the GC *cadence* to parity (every ledger close) but deliberately leaves the config flag out of scope because it is independent config plumbing (config struct → `BucketManager` constructor → `retain_buckets`/`cleanup_stale_bucket_files_background` gate), not part of the cadence claim.

## Scope
- Add a `disable_bucket_gc` config option (default `false`) to the app config and thread it to the bucket GC call site so operators can disable on-disk bucket GC for debugging / forensic retention, matching stellar-core's `DISABLE_BUCKET_GC`.
- Gate `cleanup_stale_bucket_files_background()` (and optionally the catchup-path GC) on the flag.

## Notes
Low priority — operational nicety. Not a correctness gap. Parity-completeness item.

tomerweller:
## Triage Report

**Verdict:** ACCEPT

**Type:** enhancement
**Kind:** feature
**Severity:** n/a
**Crate:** crate:bucket
**Size:** S

**Summary:** Add a `disable_bucket_gc` config flag (matching stellar-core's `DISABLE_BUCKET_GC`) to allow operators to disable on-disk bucket GC for debugging and forensic retention. This is a parity-completeness item (operational nicety, not a correctness gap).

**Test obligation:** Must include tests verifying that (1) the flag defaults to `false` (GC enabled), (2) when set to `true`, `cleanup_stale_bucket_files_background()` does not execute, and (3) config parsing and threading work correctly.

**Label adjustments:** None (labels are correct).

tomerweller:
## 📝 Plan Draft (Round 1)

**Summary:** Add a `disable_bucket_gc` flag to `BucketConfig` (default `false`, mirroring stellar-core's `DISABLE_BUCKET_GC`), gate both bucket-GC call sites on it, and translate the stellar-core `DISABLE_BUCKET_GC` key in the compat-config layer.

**Files to modify:**
[...]

PR #3154, opened 2026-06-04
Run stale-bucket GC every ledger close with re-entrancy guard

### LINKED ISSUE #3028: spec-adhere/BUCKETLISTDB BUCKETLISTDB §8.3-1: Garbage collection cadence
## Spec-adherence finding

**Spec:** [`stellar-specs/BUCKETLISTDB_SPEC.md`](../blob/main/stellar-specs/BUCKETLISTDB_SPEC.md) §8.3
**Claim ID:** `BUCKETLISTDB §8.3-1` (SHOULD)
**Classification:** **Partial**
**Severity:** low

### Expected (per spec)
After each ledger close, unreferenced bucket files removed from disk.

### Actual (in Rust)
Manager has GC helper (anchored to §8.2 in source) but invocation cadence is not enforced from inside the bucket crate; relies on ledger crate to call it.

### Rust location
`crates/bucket/src/manager.rs:708`

### Notes
Treated Partial because the spec defines per-close GC; verify that ledger/app calls it consistently.

---

tomerweller:
Closes #3028

## Summary

Removes the every-100-ledger stale-bucket GC throttle (`if pending.ledger_seq % 100 == 0`) so garbage collection runs after every successful ledger close, matching stellar-core's unconditional `forgetUnreferencedBuckets` in `advanceLedgerStateAndPublish` (`LedgerManagerImpl.cpp:1429`). The `% 100` cap was an arbitrary defensive throttle from when GC was a blocking call; GC has since moved to the blocking thread pool, so per-ledger invocation does not re-introduce the event-loop freeze that earlier work fixed.

Because GC now fires every ~5s rather than ~500s, a slow run (blocked resolving an in-flight async merge) could still be running when the next close fires. A panic-safe re-entrancy guard (`bucket_gc_in_flight: Arc<AtomicBool>`) makes per-ledger GC self-coalescing: at most one background GC at a time, skipping ledgers until the in-flight run finishes. A skipped run defers cleanup by ≤1 ledger and the keep-set is always a superset (never deletes a referenced file), so this is within stellar-core's "retain a few too many buckets a little longer" tolerance — no observable divergence.

The flag is reset via a `Drop`-guard (`ResetGuard`) on the detached awaiter task, so a panicked or cancelled GC re-arms rather than permanently disabling GC. The resolve-first safety contract is preserved unchanged: `collect_gc_roots` still calls `resolve_pending_bucket_merges()` before computing the keep-set.

The `DISABLE_BUCKET_GC` kill-switch is intentionally NOT included here — deferred to follow-up #3153 per the converged plan.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3028#issuecomment-4624385693)

## Test plan

- [x] cargo fmt --check
- [x] cargo clippy --all -- -D warnings (CI-equivalent; clean for touched crates)
- [x] cargo test -p henyey-app passes (989 lib + integration incl. new `bucket_gc` test)
- [x] cargo test -p henyey-bucket passes (incl. new resolve-first retain test)

New coverage:
- `crates/app/tests/bucket_gc.rs::test_gc_reentrancy_guard_coalesces` — a second invocation while a GC is in flight coalesces (returns false, no new task); the flag resets after a real run so a later close re-arms GC.
- `crates/app/src/app/mod.rs::tests::test_bucket_gc_reset_guard_clears_flag_on_panic` + `..._on_normal_drop` — proves the `ResetGuard` Drop clears the flag even when the scope unwinds via panic (panic-safety).
- `crates/bucket/src/bucket_list.rs::tests::test_resolve_first_retains_pending_merge_output` — end-to-end: the pending-merge output file survives `retain_buckets` only when merges are resolved first; the pre-resolve keep-set would delete it.

## Deviations from plan

None. `kind: refactor` — the new tests reference new internal API (re-entrancy flag / `ResetGuard`), so test + implementation are committed together rather than as a separate failing-test commit.

### DIFF
--- a/crates/app/src/app/ledger_close.rs
+++ b/crates/app/src/app/ledger_close.rs
@@ -2867,10 +2867,13 @@ impl App {
         // Signal heartbeat to sync recovery.
         self.sync_recovery_heartbeat();
 
-        // Periodically garbage collect stale bucket files.
-        if pending.ledger_seq % 100 == 0 {
-            self.cleanup_stale_bucket_files_background();
-        }
+        // Garbage collect stale bucket files after every ledger close, matching
+        // stellar-core's unconditional `forgetUnreferencedBuckets` in
+        // `advanceLedgerStateAndPublish` (LedgerManagerImpl.cpp:1429). The work
+        // runs on the blocking pool and is self-coalescing via the
+        // `bucket_gc_in_flight` re-entrancy guard, so per-ledger invocation
+        // cannot stack overlapping GC tasks (#3028).
+        self.cleanup_stale_bucket_files_background();
 
         self.clear_tx_set_exhausted();
 
--- a/crates/app/src/app/mod.rs
+++ b/crates/app/src/app/mod.rs
@@ -829,6 +829,21 @@ pub struct App {
     /// in `HerderImpl::triggerNextLedger` (HerderImpl.cpp:1583).
     is_applying_ledger: Arc<AtomicBool>,
 
+    /// Re-entrancy guard for per-ledger background stale-bucket GC (#3028).
+    ///
+    /// Stale-bucket GC now runs on every ledger close (matching stellar-core's
+    /// unconditional `forgetUnreferencedBuckets`), rather than every 100 ledgers.
+    /// At ~5s cadence a GC run that blocks in `resolve_pending_bucket_merges()`
+    /// during a merge backlog could still be running when the next close fires.
+    /// This flag makes per-ledger GC self-coalescing: at most one background GC
+    /// at a time. If a run falls behind, subsequent ledgers skip GC (deferring
+    /// cleanup by ≤1 ledger — within stellar-core's "retain a few too many
+    /// buckets a little longer" tolerance) until the in-flight run finishes.
+    ///
+    /// The flag is reset panic-safely (see `cleanup_stale_bucket_files_background`)
+    /// so a single panicked GC run cannot permanently disable GC.
+    bucket_gc_in_flight: Arc<AtomicBool>,
+
     /// Wall-clock of the last deferred-pipeline close-complete entry.
     /// Used to compute `henyey_ledger_close_cycle_seconds` — the time between
     /// consecutive production close-complete events.
@@ -1031,6 +1046,21 @@ fn collect_gc_roots(
     }
 
     Some(hashes)
+}
+
+/// RAII guard that clears the bucket-GC re-entrancy flag on drop (#3028).
+///
+/// Held by the detached awaiter task in `cleanup_stale_bucket_files_background`
+/// so the flag is reset on EVERY exit path — normal completion, the inner GC
+/// task erroring or panicking (surfaced as a `JoinError`), or the awaiter task
+/// itself being cancelled. This is the panic-safety mechanism: a single failed
+/// GC run re-arms GC rather than wedging it off permanently.
+struct ResetGuard(Arc<AtomicBool>);
+
+impl Drop for ResetGuard {
+    fn drop(&mut self) {
+        self.0.store(false, Ordering::Release);
+    }
 }
 
 /// Collect bucket hashes referenced only by publish-queue HAS entries.
@@ -1428,6 +1458,7 @@ impl App {
             sync_recovery_handle: parking_lot::RwLock::new(None), // Initialized in run() when needed
             sync_recovery_task: parking_lot::RwLock::new(None),
             is_applying_ledger,
+            bucket_gc_in_flight: Arc::new(AtomicBool::new(false)),
             close_cycle_last_start: parking_lot::Mutex::new(None),
             #[cfg(test)]
             close_complete_inject_blocking_ms: AtomicU64::new(0),
@@ -1906,6 +1937,16 @@ impl App {
     /// populated during startup.
     pub fn query_is_ready(&self) -> &Arc<AtomicBool> {
         &self.query_is_ready
+    }
+
+    /// Test-only accessor for the per-ledger bucket-GC re-entrancy guard (#3028).
+    ///
+    /// Exposed so integration tests in `crates/app/tests/bucket_gc.rs` can assert
+    /// the coalescing + panic-safe reset behavior of
+    /// `cleanup_stale_bucket_files_background`. Not part of the stable API.
+    #[doc(hidden)]
+    pub fn bucket_gc_in_flight(&self) -> &Arc<AtomicBool> {
+        &self.bucket_gc_in_flight
     }
 
     /// Update the bucket snapshot manager with fresh snapshots from the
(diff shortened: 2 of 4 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Throttle removal is correct, the re-entrancy guard is panic-safe via a `ResetGuard` Drop, and the resolve-first GC safety contract is untouched. Tests added cover all three behaviors and pass against the PR head merged with current `origin/main`.

<details>

Cycle 1 — complete change-list. No blocking concerns; all findings below are non-blocking observations (no concern classes raised that would bounce).

**Logic / correctness (no issues):**
- `ledger_close.rs`: the `if pending.ledger_seq % 100 == 0` throttle is removed and `cleanup_stale_bucket_files_background()` is now called unconditionally every close. Matches stellar-core `LedgerManagerImpl.cpp:1429` (`forgetUnreferencedBuckets` runs unconditionally in `advanceLedgerStateAndPublish`).
- Re-entrancy guard: `compare_exchange(false, true, Acquire, Relaxed)` correctly acquires the flag; on failure (already in flight) it returns `false` and launches nothing — at most one concurrent GC. Memory ordering is sound (Acquire on the successful CAS pairs with the Release store in `ResetGuard::drop`).
- Panic-safety: `ResetGuard` is created inside the detached awaiter task (`let _reset = ResetGuard(gc_in_flight);`) and its `Drop` does `store(false, Release)`. This fires on every exit path — normal completion, inner `spawn_blocking` task erroring/panicking (surfaced as `JoinError` from `await_blocking_logged`, which returns normally), and awaiter-task cancellation (future dropped → guard dropped). I verified there is no path where the flag is set `true` but the awaiter (and thus the guard) is never installed: after the successful CAS the function proceeds linearly to `tokio::spawn` the awaiter; `tokio::spawn` is infallible.
- **resolve-first contract preserved (load-bearing):** the PR does NOT modify `collect_gc_roots` (`mod.rs:993-1014`), which still calls `lm.resolve_pending_bucket_merges()` (line 998) BEFORE `lm.all_referenced_bucket_hashes()` (line 1000). The guard wraps the existing `spawn_blocking` body unchanged, so the keep-set is still computed only after merges resolve. This is the corruption-avoidance invariant from project memory (deleting an unresolved async-merge output silently corrupts a DiskBacked bucket); it holds.

**Test coverage (kind: refactor — existing tests preserved + new targeted coverage):**
- `test_bucket_gc_reset_guard_clears_flag_on_panic` — drives `ResetGuard` inside `catch_unwind` and asserts the flag clears during unwind. PASSES.
- `test_bucket_gc_reset_guard_clears_flag_on_normal_drop` — happy-path companion. PASSES.
- `test_gc_reentrancy_guard_coalesces` (integration) — asserts a second invocation while in-flight returns `false` and does not clear the held flag, then that a real run resets the flag and re-arms. PASSES.
- `test_resolve_first_retains_pending_merge_output` (bucket_list) — end-to-end: pre-resolve keep-set omits the merge output (and `retain_buckets` would delete it); post-resolve keep-set includes it and the file survives. Proves resolve-first is load-bearing in both directions. PASSES.
- All four ran green locally against the PR head merged with `origin/main` (CARGO_TARGET_DIR under ~/data).

**Readability / API:**
- `cleanup_stale_bucket_files_background` changed from `pub(crate)` to `#[doc(hidden)] pub` returning `bool` — justified and documented (integration tests need to drive it; not stable API). The `bucket_gc_in_flight()` accessor is likewise `#[doc(hidden)]`. Acceptable.
- Doc comments are thorough and cite the stellar-core line and the coalescing tolerance rationale.

No changes requested.

</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Operational risk of running GC every ~5s instead of every ~500s is bounded by the self-coalescing guard; no API/data-format breaks; the only divergence from stellar-core (coalescing defers cleanup by ≤1 ledger) is explicitly within its "retain a few too many buckets a little longer" tolerance and is non-observable.

<details>

Cycle 1 — complete change-list. No blocking concerns.

**regression-risk (assessed, non-blocking):**
- Frequency increase: GC moves from every-100-ledger to every-close (~100x more invocations). The work itself is unchanged (`collect_gc_roots` + `retain_buckets`), runs on the blocking pool, and is detached — it never stalls the async event loop (the long-standing reason this was backgrounded). The new risk is overlap during a merge backlog, which is exactly what the `bucket_gc_in_flight` guard addresses: at most one GC at a time; overlapping ledgers skip. No unbounded task stacking. This aligns with the project memory rule against blocking the event loop and against unbounded work.
- No change to `retain_buckets` semantics or to the keep-set computation, so the set of files deleted on any given run is identical to before — only the cadence changes.

**perf-regression (assessed, non-blocking):**
- Per-close cost is a directory scan + hash-set membership + unlink of unreferenced files. On mainnet cadence (~5s) this is cheap relative to close work, and coalescing prevents pile-up if a run is slow. No new allocation hot path in the event loop (the CAS is a single atomic op; on the common no-overlap path it just launches the same task that previously ran every 100 closes).
- Slightly more frequent disk I/O, but it replaces a bursty every-500s cleanup with steady incremental cleanup — net neutral-to-positive for disk-usage smoothness.

**api-break / migration-risk (none):**
- `cleanup_stale_bucket_files_background` visibility widened (`pub(crate)` → `#[doc(hidden)] pub`) and return type added (`bool`). Widening visibility is not a break; the `#[doc(hidden)]` marker keeps it off the public surface. New `bucket_gc_in_flight()` accessor is `#[doc(hidden)]` test-only. No config keys, no on-disk format, no DB schema change, no wire change.

**config-risk (none):** no new config; behavior is unconditional, matching upstream.

**parity / observable-divergence (within tolerance):**
- stellar-core runs `forgetUnreferencedBuckets` unconditionally every close (`LedgerManagerImpl.cpp:1429`) — confirmed in the submodule. The PR matches that cadence. The one divergence is that henyey's coalescing may defer a cleanup by ≤1 ledger when a prior run is still in flight; stellar-core holds `mLedgerStateMutex` and runs synchronously so it doesn't skip. This deferral only ever retains a stale file slightly longer — it never deletes a referenced file and never changes ledger/bucket hashes. That is squarely inside stellar-core's documented "retain a few too many buckets a little longer" tolerance and is not externally observable.

**security / operational (none):** no new external input, no privilege change. A panicked GC run cannot wedge GC off (ResetGuard re-arms), removing a denial-of-cleanup failure mode that a naive flag would introduce.

**CI note:** the single non-success check is `test (testnet, ...)` = CANCELLED in the Quickstart workflow. All `local`/`pubnet` equivalents passed; the testnet quickstart job depends on live testnet availability and is unrelated to bucket-GC cadence (not diff-attributable). Treating CI as effectively green.

No changes requested.

</details>

tomerweller:
## ✅ Merged

**Commit:** 9f7615aefc947ddf66654ef8ff710af4e4a178ac

Reviewer verdicts: Correctness APPROVE, Risk APPROVE. Self-modifying gate: not applicable (touches only `crates/app/` + `crates/bucket/`). CI: required checks green (the one non-success was the unrelated, infra-flaky `test (testnet, …)` Quickstart job — not diff-attributable).

PR #3187, opened 2026-06-05
Escalate recovery early in the near-tip archive-behind band (#3181)

tomerweller:
Closes #3181

> [!CAUTION]
> **CRITICAL consensus-liveness fix.** This addresses a ~4.5-min mainnet consensus-participation stall (observed at L62893633, recurrence of #1862). Operators should be aware: it shortens the recovery-escalation latency near the tip from ~120s to ~10–20s. No protocol/XDR/consensus-message semantics change.

## Summary

When the archive is confirmed behind, peers are verified ahead, and the peer gap is in `[PEER_AHEAD_ESCALATION_THRESHOLD(3), checkpoint_frequency()(64))` — a near-tip stall where archive catchup is structurally impossible-yet-imminent (the unblocking checkpoint publishes within one interval) — this fires the **existing** state-changing recovery escalation (`force_post_catchup_hard_reset`: `ProbeAhead` blocking fetch + `request_scp_state`) on the ~2nd recovery tick (~10–20s) instead of the default ~11 ticks (~120s).

The bug was escalation **latency**, not over-escalation. The recovery state machine spent ~4 minutes (376 attempts) spinning on archive-based "forcing catchup" that could never succeed because the archive's latest checkpoint lagged the stuck ledger, before the resolving peer-based escalation fired ~4 min late. The fix introduces a near-tip predicate and, only in that band with the archive confirmed behind, lowers the escalation attempt threshold via a new `RECOVERY_HARD_RESET_ESCALATION_ATTEMPTS_NEAR_TIP = 1`.

**Preserved (not regressed):**
- The 60s `HARD_RESET_MIN_COOLDOWN_SECS` floor (#1843) — near-tip escalation is still cooldown-gated.
- The far-behind (`peer_gap >= checkpoint_frequency()`) #1862 archive-catchup path — unchanged 11-tick threshold.
- The #2789 anti-wedge escalation — escalation fires **earlier**, never suppressed; the state-changing reset still proceeds when peers are ahead.

## Files changed

- `crates/app/src/app/mod.rs` — new `RECOVERY_HARD_RESET_ESCALATION_ATTEMPTS_NEAR_TIP` constant.
- `crates/app/src/app/types.rs` — `near_tip` field on `StuckSignals`.
- `crates/app/src/app/consensus.rs` — peer-ahead escalation gate selects the near-tip threshold when `near_tip && archive_is_confirmed_behind`.
[...]

Issue #3185, opened 2026-06-05
URGENT: Quickstart CI failing (exit-143 runner shutdown) on all commits since 069ebfcc — blocks deploy (and #3181 fix)


tomerweller:
## Symptom

The **Quickstart** workflow on `main` is failing consistently with `exit code 143` + "The runner has received a shutdown signal" across every commit since ~`069ebfcc`. This blocks the monitor-tick deploy gate (which requires green CI), so the mainnet validator cannot be upgraded off `f449a5a9`.

## Evidence

Quickstart conclusions, newest→oldest (`gh run list --workflow Quickstart --branch main`):

| sha | Quickstart |
|---|---|
| ae0af6a7 | pending |
| 5b1c29ea | **failure** |
| 0796e2a2 | **failure** (both attempt 1 and rerun attempt 2) |
| b181c0eb | cancelled |
| 069ebfcc | **failure** |
| f449a5a9 | success ← last green (currently deployed validator binary) |
| a26c366f | success |

Every failure has the identical signature — the test loops:
```
[test] Waiting for Horizon's stellar-core to start reporting
... (repeats for minutes) ...
##[error]Process completed with exit code 143.
##[error]The runner has received a shutdown signal. This can happen when the runner service is stopped, or a manually started runner is canceled.
```

The core **CI** workflow (unit/integration tests) **passes** on the same commits (e.g. `0796e2a2`: `CI|completed|success`). Only the Quickstart testnet/Horizon integration smoke-test fails, and always on runner shutdown rather than a test assertion.

## Suspected root cause

Infra/harness, not node code. The Quickstart harness was modified in exactly this commit range:
- `scripts/ci/run-quickstart-test.sh`
- `scripts/test-quickstart-harness.sh`
- `.github/workflows/quickstart.yml`

Commit `b181c0eb` (#3178) "Retry testnet horizon-core-up probe on exit-143 runner shutdown" attempted to address this very exit-143 class, but failures continue after it — the retry either doesn't cover the "Waiting for Horizon's stellar-core to start reporting" hang, or stellar-core genuinely never comes up within the window and the job is killed at the runner timeout. Bisect candidates: #3178 (harness change) first; then whether the testnet horizon-core-up probe needs a longer/À different readiness gate.

## Impact — deploy-blocking + gating an active incident fix

[...]

PR #3194, opened 2026-06-05
Fix quickstart auto-retry: move rerun out-of-run and cover exit 137

### LINKED ISSUE #3193: URGENT: Quickstart rerun-on-transient broken (same-run 'already running') + exit 137 uncovered
## Symptom
PR #3190 (#3185 fix) added two mechanisms to self-heal transient Quickstart testnet-leg failures. On PR #3187's run **27037187429**, **both fail**:

1. **`rerun-on-transient` job cannot re-dispatch its own run.** Log: `run 27037187429 cannot be rerun; This workflow is already running` → exits 1. A job *inside* a run cannot `gh run rerun --failed` that same run while it is still in progress. The auto-re-dispatch **never fires** — it only adds a spurious FAILURE check to every PR whose testnet leg flakes.
2. **Exit 137 (SIGKILL runner reclamation) is uncovered.** The wrapper retry in `scripts/ci/run-quickstart-test.sh` covers exit **124/143** only; the actual reclamation produced **exit 137** (`The runner has received a shutdown signal` + `exit code 137`). And whole-runner reclamation kills the in-script wrapper too — only an out-of-run retry helps.

## Net effect
The #3185 fix does NOT handle the dominant failure mode (whole-runner reclamation, 137). PRs with a transient testnet failure still go red and need a **manual** `gh run rerun --failed` after the run completes. The urgent consensus fix #3187 was blocked on exactly this.

## Fix
- **Replace the in-run `rerun-on-transient` job with a separate `workflow_run`-triggered workflow** (`on: workflow_run: {workflows: [Quickstart], types: [completed]}`) that runs AFTER the Quickstart run finishes, so `gh run rerun <id> --failed` works. Guard on `run_attempt == 1` + `conclusion == 'failure'` for a bounded one-shot retry.
- **Add exit 137 to `is_retryable_exit`** in `run-quickstart-test.sh` (alongside 124/143).
- Regression-test both in `scripts/test-quickstart-harness.sh`.

## Verify
A transient (137/143) testnet failure on attempt 1 auto-re-dispatches once and goes green with no manual step; a genuine failure reproduces on attempt 2 and stays red. Remove the broken in-run `rerun-on-transient` job.

Self-modifying (CI workflow) — extra scrutiny, autonomous merge on triple-green.

tomerweller:
Closes #3193

## Summary

The #3185/#3190 rerun-on-transient self-heal was broken in two ways, both observed on PR #3187's run 27037187429:

1. **In-run rerun could never fire.** The `rerun-on-transient` job in `quickstart.yml` ran `gh run rerun <id> --failed` for ITS OWN still-running run, which GitHub rejects with `This workflow is already running` (exit 1). The auto-retry never fired — it only added a spurious FAILURE check to every PR whose testnet leg flaked. **Fix:** remove the in-run job; recovery moves to a new **separate `workflow_run`-triggered workflow** (`.github/workflows/quickstart-retry.yml`) that fires AFTER the Quickstart run completes, at which point the re-dispatch is accepted.

2. **Exit 137 uncovered.** The actual whole-runner reclamation produced exit **137** (SIGKILL, 128+9), not the **143** (SIGTERM) the wrapper already handled. **Fix:** add 137 to `is_retryable_exit` in `run-quickstart-test.sh` alongside 124/143, for the in-probe case where the runner survives long enough for the wrapper to observe the exit.

### The `workflow_run` mechanism

`quickstart-retry.yml` triggers `on: workflow_run: { workflows: ["Quickstart"], types: [completed] }`. The `workflows` value matches `quickstart.yml`'s top-level `name: Quickstart` exactly (verified via parsed YAML; the harness asserts the match). It holds `permissions: { actions: write, contents: read }` and re-dispatches via `gh run rerun ${{ github.event.workflow_run.id }} --failed` (the triggering run, not its own).

### Bounded one-shot — cannot loop

The job guard is `github.event.workflow_run.conclusion == 'failure' && github.event.workflow_run.run_attempt == 1`. The re-dispatched run is attempt 2 (`run_attempt == 2`). If attempt 2 also fails, this workflow fires again on its completion, but the `run_attempt == 1` clause is now false, so no further rerun is dispatched and attempt 2's failure stays red. A genuine regression fails identically on attempt 2 and surfaces as a hard red — exactly one transient reclamation is absorbed, blast radius bounded to one extra attempt.

## Test plan

- [x] `python3 -c 'import yaml,sys; yaml.safe_load(open(sys.argv[1]))'` on both `quickstart.yml` and `quickstart-retry.yml` → both parse OK
- [x] `workflow_run.workflows` name string == `quickstart.yml` `name:` (`Quickstart`), confirmed via parsed YAML
- [x] `bash -n` + `zsh -n` clean on `run-quickstart-test.sh` and the harness
- [x] `bash scripts/test-quickstart-harness.sh` → 72/72 pass
- [x] `cargo fmt --check` + `cargo clippy --all -- -D warnings` (pre-commit hook)

- **Tests:** `scripts/test-quickstart-harness.sh` — exit-137 retry (targeted shard only / not local-pubnet / not on double-failure), in-run rerun job removed from `quickstart.yml`, and the new `quickstart-retry.yml` (`workflow_run`/completed trigger, exact name-match, `conclusion==failure && run_attempt==1` guard, `actions: write`, rerun-of-triggering-run).
- **Pre-fix:** committed as `9b0c1705` — verified FAILED: exit 137 not retryable + `quickstart-retry.yml` absent + in-run job still present (11 assertions red).
- **Post-fix:** verified PASSES after `1b1233c1` — 72/72.

> Auto-retry cannot be E2E-tested in this PR (it triggers post-run, on `workflow_run`), so verification is static + reasoning per the issue.

## Deviations from plan

None.

### DIFF
--- a/.github/workflows/quickstart-retry.yml
+++ b/.github/workflows/quickstart-retry.yml
@@ -1,0 +1,71 @@
+name: Quickstart Retry
+
+# Out-of-run auto-retry for the Quickstart workflow on a whole-runner
+# reclamation (#3193).
+#
+# Why a SEPARATE workflow (not an in-run job): the in-wrapper retry in
+# scripts/ci/run-quickstart-test.sh self-heals a transient-infra exit
+# (124 timeout / 143 SIGTERM / 137 SIGKILL) ONLY when the runner survives long
+# enough for the bash wrapper to observe the probe's non-zero exit and re-run
+# it. When GitHub reclaims the WHOLE runner ("The runner has received a shutdown
+# signal" + "Process completed with exit code 143/137" + "Cleaning up orphan
+# processes"), the wrapper process is itself killed and never reaches its retry
+# block. The only recovery is a fresh job attempt.
+#
+# The previous design put that recovery in an in-run `rerun-on-transient` job,
+# but a job INSIDE a run cannot re-dispatch its own still-running run:
+# `gh run rerun <id> --failed` fails with "run <id> cannot be rerun; This
+# workflow is already running" (exit 1), observed on PR #3187's run
+# 27037187429. The auto-retry never fired and only added a spurious FAILURE
+# check.
+#
+# This workflow instead triggers on `workflow_run` (types: completed) for the
+# Quickstart workflow, so it runs AFTER that run has fully finished — at which
+# point `gh run rerun <id> --failed` is accepted (no "already running"). It
+# re-dispatches the failed jobs of the triggering run exactly once.
+#
+# Bounded one-shot (cannot loop): the guard requires
+#   github.event.workflow_run.conclusion == 'failure'  AND
+#   github.event.workflow_run.run_attempt == 1
+# The re-dispatched run is attempt 2 (run_attempt == 2). If attempt 2 ALSO
+# fails, this workflow fires again on its completion, but the run_attempt == 1
+# guard is now false, so no further rerun is dispatched and attempt 2's failure
+# stays red. A genuine (reproducible) failure therefore fails identically on
+# attempt 2 and surfaces as a hard red — this absorbs exactly one transient
+# reclamation without masking regressions, and the blast radius is bounded to a
+# single extra attempt.
+#
+# IMPORTANT: the `workflows:` value below MUST match the Quickstart workflow's
+# top-level `name:` field in .github/workflows/quickstart.yml ("Quickstart")
+# exactly, or this trigger never fires. The harness asserts this match
+# (scripts/test-quickstart-harness.sh::test_separate_workflow_run_retry_workflow).
+
+on:
+  workflow_run:
+    workflows: ["Quickstart"]
+    types: [completed]
+
+permissions:
+  actions: write
+  contents: read
+
+jobs:
+  rerun-on-transient:
+    # Only re-dispatch when the triggering Quickstart run FAILED and it was its
+    # FIRST attempt. run_attempt == 2 (the re-dispatched attempt) fails the
+    # guard, so this is a bounded one-shot — it cannot loop.
+    if: >-
+      github.event.workflow_run.conclusion == 'failure' &&
+      github.event.workflow_run.run_attempt == 1
+    runs-on: ubuntu-latest
+    steps:
+      - name: Re-dispatch failed jobs once on transient-infra failure
+        env:
+          GH_TOKEN: ${{ github.token }}
+          RUN_ID: ${{ github.event.workflow_run.id }}
+        run: |
+          echo "Quickstart run $RUN_ID failed on run_attempt=1;"
+          echo "re-dispatching its failed jobs once (out-of-run, post-completion)"
+          echo "to self-heal a one-off whole-runner reclamation (#3193)."
+          gh run rerun "$RUN_ID" --failed \
+            --repo "${{ github.repository }}"
--- a/.github/workflows/quickstart.yml
+++ b/.github/workflows/quickstart.yml
@@ -37,12 +37,16 @@ name: Quickstart
 #   * No retry for any other shard, or for any non-transient failure (e.g. a
 #     genuine probe failure exit 1) — those still fail loudly.
 #
-# Whole-runner reclamation (#3185): when GitHub SIGTERMs the entire runner, the
-# bash wrapper is killed too and never reaches its in-script retry. The separate
-# `rerun-on-transient` job recovers that case by re-dispatching the failed jobs
-# once (run_attempt == 1 only) onto a fresh runner. A genuine failure reproduces
-# on attempt 2 and stays red, so this absorbs exactly one transient reclamation
-# without masking regressions.
+# Whole-runner reclamation (#3185 / #3193): when GitHub SIGTERMs/SIGKILLs the
+# entire runner, the bash wrapper is killed too and never reaches its in-script
+# retry. An in-RUN job cannot recover this — a job inside a run cannot
+# re-dispatch (rerun --failed) its own still-running run (it fails with
+# "This workflow is already running", #3193). Recovery therefore lives in a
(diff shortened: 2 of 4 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** CI-infra fix correctly moves the broken in-run rerun to a separate `workflow_run`-triggered workflow and adds exit-137 coverage. Regression test (72/72) is non-vacuous, fails at the pre-fix commit (61/72), and was committed before the fix. Cycle 1 — complete change-list; no blocking concerns.

**Self-modifying paths touched (extra reviewer scrutiny applied):**
- `.github/workflows/quickstart.yml`
- `.github/workflows/quickstart-retry.yml`
- `scripts/ci/run-quickstart-test.sh`
- `scripts/test-quickstart-harness.sh`

<details>

**Extra-scrutiny (self-modifying CI) — all pass:**
- `yaml.safe_load` on both `quickstart.yml` and `quickstart-retry.yml` → both parse OK.
- `bash -n` + `zsh -n` clean on `run-quickstart-test.sh` and the harness.
- No dispatch-table / step-numbering breakage — these are CI workflows + a self-contained test harness, not pipeline skills/routing.

**Correctness of the diff:**
- Root cause 1 (in-run rerun can't fire): the old `rerun-on-transient` job called `gh run rerun ${{ github.run_id }} --failed` against its OWN still-running run → GitHub rejects with "already running" (exit 1). Confirmed removed from `quickstart.yml` (`grep` finds no `gh run rerun` / `rerun-on-transient` job, only an explanatory NOTE comment). Recovery correctly relocated to a `workflow_run: { types: [completed] }` workflow that fires post-completion, when `gh run rerun <id> --failed` is accepted. Logic is sound.
- Root cause 2 (exit 137 uncovered): `is_retryable_exit` widened from `124||143` to `124||143||137`. 137 = SIGKILL (128+9), the harsher reclamation signature; same transient class as 143. Correct.
- Re-dispatch targets the TRIGGERING run (`github.event.workflow_run.id`), not its own run — verified.

**Bounded one-shot (no infinite loop) — verified:**
- Guard `conclusion == 'failure' && run_attempt == 1`. Re-dispatched run carries `run_attempt == 2` → guard false → no further rerun. A genuine regression reproduces on attempt 2 and stays red. Blast radius bounded to exactly one extra attempt. Cannot loop.

**Test coverage (kind: bug-fix):**
- Regression test = `scripts/test-quickstart-harness.sh`. Verified the test commit (`9b0c1705`) PRECEDES the fix commit (`1b1233c1`).
- At `9b0c1705` (test only): `61/72 passed, 11 failed` — the test genuinely captures the bug.
- At `1b1233c1` (post-fix): `72/72 passed`.
- New assertions are non-vacuous: `retry_workflow_name_matches_quickstart_exactly` derives Quickstart's actual `name:` from `quickstart.yml` and greps the retry file for that exact string, so any future name drift fails CI. `retry_workflow_bounded_to_first_attempt` asserts `run_attempt == 1`. Exit-137 retry/no-retry/double-failure cases all covered, plus local/pubnet non-retry.

**Error handling / readability:** Workflow + script are heavily and accurately commented (the comments match the code). `set +o pipefail` + `PIPESTATUS[0]` correctly captures the probe exit through the `tee`.

No blocking concerns.
</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Low operational risk. The change reduces blast radius vs. the broken status quo (removes a job that added a spurious FAILURE check to every flaky PR). The new workflow is a bounded, fail-safe one-shot that cannot loop and cannot mask regressions. Cycle 1 — complete change-list; no blocking concerns.

**Self-modifying paths touched (extra reviewer scrutiny applied):**
- `.github/workflows/quickstart.yml`
- `.github/workflows/quickstart-retry.yml`
- `scripts/ci/run-quickstart-test.sh`
- `scripts/test-quickstart-harness.sh`

<details>

**regression-risk:**
- The retry only ever ADDS a fresh attempt of already-failed jobs; it never changes a green run or suppresses a red one. A genuine (reproducible) failure fails identically on attempt 2 (`run_attempt == 2` fails the guard) and surfaces as hard red. Masking a real regression is structurally impossible — the worst case is one extra wasted CI attempt on a deterministic failure.
- Removing the in-run `rerun-on-transient` job is strictly a risk reduction: that job could never succeed (always exit 1 "already running"), so it only ever painted a spurious FAILURE check on PRs whose testnet leg flaked. Its removal eliminates that false-red.

**config-risk / activation:**
- The retry workflow triggers from the DEFAULT branch only (GitHub `workflow_run` semantics), so it has no effect on this PR's own run and cannot be E2E-validated here — expected and called out in the PR body. It takes effect once merged to main. This is the intended design, not a gap.
- Name coupling is the one fragility: if a future edit renames `quickstart.yml`'s `name:` without updating `workflows: ["Quickstart"]`, the trigger silently stops firing. Mitigated: the harness `retry_workflow_name_matches_quickstart_exactly` test (run in the CI "Quickstart Harness" job) asserts the match and would fail CI on drift. Acceptable.

**perf / blast-radius:**
- Bounded to exactly one extra job attempt per first-attempt transient. No loop, no fan-out. The retry workflow itself does nothing on attempt-2 completions (guard false).

**security:**
- `permissions: { actions: write, contents: read }` is the minimal scope needed for `gh run rerun`. Uses `${{ github.token }}` (no PAT). `actions: write` is appropriate and scoped to re-dispatch only. No checkout of untrusted code in the retry workflow — it only calls the `gh` API. No injection surface.

**scope:** local/pubnet shards still never retry (`is_retryable_shard` = testnet && core,horizon); non-transient exit 1 still fails immediately. Behavior outside the targeted transient class is unchanged.

No blocking concerns.
</details>

PR #3216, opened 2026-06-06
Stream catchup replay per checkpoint to bound long-gap memory

### LINKED ISSUE #2901: URGENT: Validator OOMs during long-gap catchup (55-57G RSS) — cannot recover from L62759194
## Symptom

After the sponsorship.rs:366 panic at 2026-05-27 12:22 UTC (#2900), the
validator has now OOMed twice in a row trying to catch up from the disk-
persisted state at L62759194 (~32 h behind mainnet tip).

```
[6889916.449945] oom-kill:constraint=CONSTRAINT_NONE,nodemask=(null),cpuset=user.slice,mems_allowed=0,global_oom,task_memcg=/user.slice/user-1168.slice/session-19695.scope,task=henyey,pid=1918686,uid=1168
[6889916.450066] Out of memory: Killed process 1918686 (henyey) total-vm:133425624kB, anon-rss:55012432kB, file-rss:5632kB
```

Second OOM ~16 min later on a fresh PID:
```
Out of memory: Killed process 1964133 (henyey) total-vm:131922432kB, anon-rss:57183192kB
```

## Context

- Host: 61 GiB RAM, 0 swap
- Binary: `1a3b07106c3f` (deployed 2026-05-24 17:17 UTC, ran clean for ~67 h before the sponsorship panic at #2900)
- Catchup start LCL: 62759194 (the panic ledger)
- Mainnet tip at restart: ~62778879 (32 h ahead, ~19,000 ledgers gap)
- Steady-state RSS on this binary in normal Validating is ~17 G; during this catchup it climbs to 55–57 G before OOM
- Time to OOM: ~12 min first restart, ~4 min second restart
- No panic recorded in log — process disappears mid-overlay-activity; only kernel log shows the kill

## What appears to be happening

During catchup-from-disk-state at 32+ hours behind tip, the validator is:
1. Connecting to ~30 peers
2. Issuing many `Requesting transaction set from peers` calls (one per
   externalized txset hash seen in SCP messages)
3. Buffering all that incoming data in memory faster than it can apply
   ledgers

After ~4–12 min the in-memory buffer of pending SCP state + txset data
exceeds the available memory and the kernel kills the process.

## Suspected location

Likely candidates (no investigation yet):
- `crates/overlay/` — peer txset fetch / pending-msg queue
- `crates/herder/` — SCP nomination/voting state cache during long catchup
- `crates/app/src/app/catchup_impl.rs` — catchup orchestration

## Possible fixes (not yet evaluated)

1. **Backpressure** on txset-fetch requests when the apply pipeline is
   falling behind
2. **Bounded queues** in the overlay layer for pending txsets / SCP envelopes
   discovered during catchup
3. **Skip non-essential catchup work** when very far behind (e.g., use
   history-archive-only mode instead of trying to SCP-replay)
4. **Operator workaround**: wipe `mainnet.db` and rejoin from buckets/+
   network, which previously caught up in ~16 min from a fresh state (the
   approach used 2026-05-22 after the knit-to-LCL FATAL).

## Validator status

Currently dead. Will not restart again on this binary until either:
- A code fix lands
- The operator wipes mainnet.db (destructive — last LCL `62759194` would be discarded, validator rejoins from network)
- The operator adds swap or provisions more memory

tomerweller:
Closes #2901

**CRITICAL OOM fix.** The validator OOMs during long-gap catchup (55–57 GiB RSS) and cannot recover from the persisted stale LCL (L62759194 → tip, a ~65,800-ledger gap).

## Summary

The catchup replay path buffered the **entire** LCL→target gap of decoded transaction/result bodies in a single `Vec<LedgerData>` plus a never-evicted per-checkpoint `HashMap`, so resident memory scaled **linearly with the gap** (~57G at ~65,800 ledgers). This restructures `download_verify_and_replay_once` into two memory-bounded phases (confined to `catchup/replay.rs` + a `download_ledger_headers` helper), mirroring stellar-core's `downloadVerifyLedgerChain` + per-checkpoint `DownloadApplyTxsWork`.

**Phase 1 — header verify (full range, once):** download ONLY the header chain for the whole `[lcl+1, target]` gap (tx/result bodies are streamed and freed per checkpoint inside `download_ledger_headers`), run the §11.2 knit-to-LCL matrix on the first apply header, then run `verify_reverse_walk` over the **whole** header vector in one call.

**Phase 2 — per-checkpoint replay (stream + free):** loop one checkpoint (≤ `checkpoint_frequency()` ledgers) at a time via `drive_replay_batches` — download bodies, `verify_txsets`, replay via `close_ledger` in order, persist, then **drop the batch before the next download**. Peak resident body memory is bounded to ~one checkpoint instead of the whole gap.

## The verify/replay split (Round-1 refute-pass concern)

Naive batching would break the top-anchored reverse-walk header verification (`verify_reverse_walk` partitions headers into checkpoint groups and threads trust highest→lowest from the top anchor — it MUST see the full header set in one call). The fix therefore **splits** verify from replay: the full header chain is verified up front over the entire range (headers are cheap, fixed-size); only the gap-linear tx/result **bodies** are streamed per checkpoint. The reverse-walk runs unchanged on the full header vector — the trust model is preserved. The new `test_header_chain_verified_over_full_range_before_replay` asserts a tampered mid-range link is still caught by the single full-range verify.

## Parity / correctness preservation (identical state + bucketListHash)

Replay still calls `close_ledger` for **every ledger in the same order** — identical per-ledger state transitions, bucket-list updates, and final `bucketListHash`. Only the memory representation changes (stream vs buffer-all), not the computed result. The first batch reuses the caller-provided `lcl` (byte-equivalent to the pre-fix whole-gap download); subsequent batches re-derive LCL from `ledger_manager.header_snapshot()` (the same mechanism the retry loop already used). The `#2931`/`#2940` archive-publication gate still runs **once per attempt** up front. The `catchup_integration` end-to-end test (which reaches the correct target hash — the parity-equivalence proof) and the 12 `replay_integration` tests both pass unchanged.

## (a) What the offline tests prove

- **Bounded per-checkpoint memory** — `test_replay_streams_tx_bodies_one_checkpoint_at_a_time`: peak resident `LedgerData` ≤ one checkpoint (≤ `checkpoint_frequency()`), strictly less than the whole gap.
- **Batch freed before next download** — `test_replay_frees_batch_before_next_download`: at most one batch is live at any step boundary.
- **No-progress guard** — `test_drive_replay_batches_no_progress_errors`.
- **`verify_txsets` semantics** — `test_verify_txsets_valid_mismatch_and_absent_skip`: unconditional `verify_tx_set`, skip for absent results, fatal on mismatch.
- **Full-range header verify before replay** — `test_header_chain_verified_over_full_range_before_replay`.
- **Identical replay / final state** — proven by the preserved end-to-end `catchup_integration` + `replay_integration` suites reaching the correct target hash.

## (b) What needs OPERATOR deploy-verification

The absolute RSS ceiling is a real-state-size property and cannot be verified statically: restart the validator on this binary at the persisted stale LCL (62759194) against mainnet tip and confirm **anon-RSS stays well under 61 GiB** across the full catchup. If RSS still climbs **gap-independently**, that residual is the parity-expected `InMemorySorobanState` baseline (a separate, flat, gap-independent cost — stellar-core keeps the same in-RAM soroban state), distinguishable because it is flat across the gap rather than gap-linear. Capture `/proc/<pid>/smaps_rollup` (or heaptrack peak) at the high-water mark if so.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/2901#issuecomment-4638628920)

## Test plan

- [x] `cargo fmt --check` (pre-commit hook)
- [x] `cargo clippy -p henyey-history --all-targets -- -D warnings` clean; pre-commit `clippy --all` clean
- [x] `cargo test -p henyey-history` — 553 unit + integration pass (incl. `catchup_integration`, 12 `replay_integration`)
- [x] `cargo build -p henyey-app` — the catchup future stays `Send` (boxed driver future is `+ Send`)

- **Tests:** `crates/history/src/catchup/replay.rs::test_replay_streams_tx_bodies_one_checkpoint_at_a_time` and `::test_replay_frees_batch_before_next_download`
- **Pre-fix:** committed as `a0f33197` (commit 1 introduces the `drive_replay_batches` seam in its whole-gap form) — verified FAILED with `peak resident bodies 202 must be ≤ one checkpoint (64)` and `expected ≥3 checkpoint batches` (whole gap is one un-freed Vec).
- **Post-fix:** verified PASSES after `663eadbb` (commit 2 swaps in the per-checkpoint stream-and-free loop).

## Deviations from plan

- The per-checkpoint loop is factored into a reusable `drive_replay_batches(from, target, state, step)` driver (HRTB + boxed-`Send` future) so the streaming/free invariant is directly unit-testable offline without a network or `LedgerManager`, and so the production loop and the regression tests share the exact same batching/free logic. Behavior is unchanged from the plan's inline loop.
- `verify_downloaded_data` was split into `verify_header_chain(&[LedgerHeader], lcl)` (full-range) + `verify_txsets(&[LedgerData])` (per-batch), as the plan specified; the two pre-existing `verify_downloaded_data_*` tests now drive `verify_header_chain` over the full header set.

### DIFF
--- a/crates/history/src/catchup/download.rs
+++ b/crates/history/src/catchup/download.rs
@@ -466,6 +466,86 @@ impl CatchupManager {
         Ok((data, knit_entries, last_archive_name))
     }
 
+    /// Download ONLY the ledger headers (plus knit-prefix entries) for the
+    /// full `[from_ledger+1, to_ledger]` range, used by the up-front
+    /// header-chain verification phase (#2901).
+    ///
+    /// Mirrors the header-extraction half of [`download_ledger_data`] but
+    /// discards the per-checkpoint transaction/result bodies as soon as the
+    /// headers have been copied out, so peak memory is bounded to ~one
+    /// checkpoint of body data rather than the whole gap. The headers
+    /// themselves (small, fixed-size records) are retained for the full
+    /// range so [`verify::verify_reverse_walk`] can thread trust top-down
+    /// across all checkpoints exactly as before.
+    ///
+    /// Returns `(headers, knit_entries, archive_name)`. `headers` covers the
+    /// apply range `[from_ledger+1, to_ledger]` in ascending order;
+    /// `knit_entries` covers `[knit_start, from_ledger]` (the §11.2 knit
+    /// prefix). This matches the apply/knit split produced by
+    /// [`download_ledger_data`] so the two phases stay in lockstep.
+    pub(super) async fn download_ledger_headers(
+        &mut self,
+        from_ledger: u32,
+        to_ledger: u32,
+        pinned: Option<&Arc<HistoryArchive>>,
+    ) -> Result<(Vec<LedgerHeader>, Vec<LedgerHeaderHistoryEntry>, String)> {
+        let mut headers: Vec<LedgerHeader> = Vec::new();
+        let mut knit_entries: Vec<LedgerHeaderHistoryEntry> = Vec::new();
+        let mut last_archive_name = "none".to_owned();
+
+        let apply_start = from_ledger.saturating_add(1);
+        let knit_start = if from_ledger == 0 {
+            apply_start
+        } else {
+            let apply_ckpt_start = crate::checkpoint::checkpoint_start(apply_start);
+            std::cmp::max(apply_ckpt_start, from_ledger.saturating_sub(1))
+        };
+
+        if apply_start > to_ledger {
+            return Ok((headers, knit_entries, last_archive_name));
+        }
+
+        // Walk checkpoint-by-checkpoint. Each checkpoint's full data (headers
+        // + tx/result bodies) is downloaded, the relevant headers are copied
+        // out, and the bodies are dropped at the end of the checkpoint block —
+        // bounding peak body memory to a single checkpoint.
+        let mut seq = knit_start;
+        while seq <= to_ledger {
+            let checkpoint = checkpoint::checkpoint_containing(seq);
+            let (cache, archive_name) = self
+                .download_checkpoint_ledger_data(checkpoint, pinned)
+                .await?;
+            last_archive_name = archive_name;
+
+            let checkpoint_end = std::cmp::min(checkpoint, to_ledger);
+            for s in seq..=checkpoint_end {
+                self.progress.current_ledger = s;
+                let header_entry_opt = cache.headers.iter().find(|h| h.header.ledger_seq == s);
+                if s < apply_start {
+                    // Knit-prefix entry: collected if present, silently skipped
+                    // otherwise (same policy as `download_ledger_data`).
+                    if let Some(entry) = header_entry_opt {
+                        knit_entries.push(entry.clone());
+                    }
+                    continue;
+                }
+                let header_entry = header_entry_opt.ok_or_else(|| {
+                    HistoryError::CatchupFailed(format!(
+                        "ledger {} not found in checkpoint headers",
+                        s
+                    ))
+                })?;
+                headers.push(header_entry.header.clone());
+            }
+
+            // Free this checkpoint's bodies before fetching the next checkpoint.
+            drop(cache);
+            seq = checkpoint_end.saturating_add(1);
+        }
+
+        Ok((headers, knit_entries, last_archive_name))
+    }
+
     /// Download ledger headers, transactions, and results for a checkpoint.
     ///
     /// Stage E instrumentation: emits
--- a/crates/history/src/catchup/mod.rs
+++ b/crates/history/src/catchup/mod.rs
@@ -373,7 +373,7 @@ impl CatchupManager {
 
     /// Set an SCP-derived trusted anchor for reverse-walk verification (§9.1 + INV-C5).
     ///
-    /// When set, `verify_downloaded_data` builds `ReverseWalkConfig` with
+    /// When set, `verify_header_chain` builds `ReverseWalkConfig` with
     /// `TrustSource::Scp { seq, hash }` instead of the default `TrustSource::None`.
     /// This makes local-state disagreement during header-chain verification a
     /// fatal (non-retriable) error, matching stellar-core's online catchup behavior
--- a/crates/history/src/catchup/replay.rs
+++ b/crates/history/src/catchup/replay.rs
@@ -188,6 +188,113 @@ pub(super) fn decode_upgrades_from_header(header: &LedgerHeader) -> Vec<LedgerUp
         .collect()
 }
 
+/// Drive the per-checkpoint streaming replay loop (#2901).
+///
+/// Walks `(from, target]` one checkpoint at a time. For each batch it computes
+/// the checkpoint-aligned upper bound `batch_to`, invokes `step(batch_from,
+/// batch_to)` to download → verify → replay → persist that batch, then **drops
+/// the returned batch before issuing the next `step` call**. This is the
+/// mechanism that bounds peak resident transaction/result body memory to ~one
+/// checkpoint (`checkpoint_frequency()` ledgers) instead of the whole gap.
+///
+/// `step(state, batch_from, batch_to)` returns `(batch, new_lcl_seq)`:
+/// - `batch` is the `Vec<LedgerData>` that was downloaded/applied. The driver
+///   owns it only long enough to drop it before the next iteration, so the
+///   per-batch allocation never overlaps the next download.
+/// - `new_lcl_seq` is the ledger sequence the local LCL advanced to after
+///   replaying the batch; the next batch starts at `new_lcl_seq`.
+///
+/// `state` is threaded by `&mut` reference through every batch (it carries the
+/// caller's mutable context, e.g. the `CatchupManager` and per-attempt
+/// bookkeeping). The HRTB + boxed-future signature lets `step` hold a mutable
+/// borrow of `state` across the `.await` for a single batch while releasing it
+/// between batches — the borrow lifetime `'s` is tied to each individual call,
+/// not to the whole driver.
+///
+/// A no-forward-progress guard turns a stuck batch into a hard error rather
+/// than an infinite loop. Returns the final LCL seq.
+///
+/// Mirrors stellar-core `DownloadApplyTxsWork` (`BatchWork` applying one
+/// checkpoint via `ApplyCheckpointWork`, releasing each checkpoint's frame
+/// before the next).
+pub(super) async fn drive_replay_batches<S, F>(
+    from: u32,
+    target: u32,
+    state: &mut S,
+    mut step: F,
+) -> Result<u32>
+where
+    F: for<'s> FnMut(
+        &'s mut S,
+        u32,
+        u32,
+    ) -> std::pin::Pin<
+        Box<dyn std::future::Future<Output = Result<(Vec<LedgerData>, u32)>> + Send + 's>,
+    >,
+{
+    // NOTE (commit 1 of 2): PRE-FIX whole-gap body — downloads the entire
+    // [from, target] range in a SINGLE step and never frees between
+    // checkpoints. The #2901 regression tests FAIL against this body (peak ==
+    // whole gap; no per-batch free). Commit 2 replaces it with the
+    // per-checkpoint stream-and-free loop.
+    let (batch, _new_lcl_seq) = step(state, from, target).await?;
+    let _hold = batch;
+    Ok(target)
+}
+
+/// Per-attempt mutable context threaded through [`drive_replay_batches`] for
+/// the streaming replay phase (#2901). Bundles the `CatchupManager`, the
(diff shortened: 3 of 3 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Two-phase restructure (full-range header verify, then per-checkpoint stream/replay/free) is correct and state-identical to the pre-fix whole-gap replay. Every ledger is applied exactly once, in order, with a contiguous parent-hash chain across batch boundaries. Regression tests verified fail-pre / pass-post; `catchup_integration` + 12 `replay_integration` pass.

<details>

Cycle 1 — complete change-list. No blocking concerns; APPROVE.

**kind: bug-fix — regression test verification (PASS).** PR body declares the regression tests `test_replay_streams_tx_bodies_one_checkpoint_at_a_time` and `test_replay_frees_batch_before_next_download`, with the test commit `a0f33197` preceding the fix `663eadbb`. I checked out `a0f33197` (whose `drive_replay_batches` is the deliberate whole-gap form: a single `step(state, from, target)` holding the batch) and ran both tests: both FAIL — `peak resident bodies 202 must be ≤ one checkpoint (64)` and `expected ≥3 checkpoint batches`. At PR head all 31 `catchup::replay` tests pass. Fail-pre / pass-post confirmed exactly as claimed.

**State-identity proof (every ledger applied once, in order, contiguous chain):**
- `drive_replay_batches` boundary math: `batch_to = min(checkpoint_containing(batch_from+1), target)`; `step` applies `[batch_from+1, batch_to]`; `batch_from = new_lcl_seq` (read from the live `header_snapshot()`, not assumed). Traced freq=64, from=69, target=263 → ranges `[70,127],[128,191],[192,255],[256,263]` = exactly `[70,263]` contiguous, no gap/overlap/dup. Final batch clamps to `target`.
- Per-ledger close: `replay_via_close_ledger` is unchanged and uses `ledger_manager.current_header_hash()` (live state) as the parent — so parent-hash chaining is from the actual local LCL, identical to whole-gap.
- Batch K+1 chain root: `download_ledger_data(batch_from=batch_to_prev, …)` re-fetches `download_checkpoint_header(batch_from)` from the (pinned) archive and uses it as the chain root for `LedgerData::new` validation — the same archive-sourced header batch K applied as its last ledger. Contiguous.
- End-to-end: `catchup_integration` (reaches the correct target hash) passes — the empirical state/bucketListHash-identity proof.

**Verify ordering preserved.** Pre-fix: gate → download(whole) → `verify_knit_to_lcl` → `verify_downloaded_data`(header+txset, whole) → replay(whole) → persist(whole). Post-fix: gate → `download_ledger_headers`(whole) → `verify_knit_to_lcl` → `verify_header_chain`(whole, reverse-walk in one call) → per-batch[download → `verify_txsets` → replay → persist]. Header-chain reverse-walk still runs once over the full range before any replay (trust anchor intact). The only re-ordering — tx-set/result verify now interleaved per-batch instead of all up front — still verifies each ledger's tx-set strictly before that ledger's `close_ledger`. `test_header_chain_verified_over_full_range_before_replay` confirms a tampered mid-range link is caught by the single full-range call.

**LCL re-derivation safety (no off-by-one).** First batch passes caller `lcl`; subsequent batches re-derive via `header_snapshot()`. Critically, `download_ledger_data` *ignores* `initial_lcl` for the chain root when `from_ledger>0` — it re-fetches the LCL header from the archive in both first and subsequent batches — so re-derivation cannot introduce a divergent chain root. `knit_start` for non-first batches collapses to `apply_start` (no knit entries collected; `verify_knit_to_lcl` not re-run per batch — correct, since it ran once in phase 1).

**Failure/retry + no-progress guard.** `drive_replay_batches` errors (`CatchupFailed`) if `new_lcl_seq <= batch_from`, preventing an infinite loop (`test_drive_replay_batches_no_progress_errors`). Fatal vs transient classification unchanged; the outer retry loop re-derives LCL from the advanced snapshot, preserving partial progress across batches (mirrors `resetIter()`). Archive publication gate (#2931/#2940) still runs once per attempt up front, and the gate-archive pin threads through both `download_ledger_headers` and every per-batch `download_ledger_data`.

**Memory claim holds.** `download_ledger_headers` downloads each checkpoint, copies out headers, and `drop(cache)` before the next checkpoint. The replay loop `drop(batch)` before the next `step`. The unbounded `checkpoint_cache: HashMap` is now local to a single `download_ledger_data` call (one batch = ~one checkpoint) and dropped with the batch — no cross-iteration accumulation. `test_replay_streams_tx_bodies_one_checkpoint_at_a_time` (peak ≤ one checkpoint) and `test_replay_frees_batch_before_next_download` (≤1 live batch at a step boundary) pin both invariants.

**Readability/error handling.** HRTB + boxed-`Send` future on `drive_replay_batches` is justified (keeps the streaming/free invariant unit-testable offline and shared with production). `ReplayBatchCtx` cleanly bundles per-attempt mutable context. Errors map to existing variants; metric emission preserved (Stage E literal/label tests pass).

Minor (non-blocking, not filed): the doc on `drive_replay_batches` says `step(state, batch_from, batch_to)` while the type is `FnMut(&mut S, u32, u32)` — wording-only.
</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** The two-phase restructure mirrors stellar-core's `CatchupWork::downloadVerifyLedgerChain` (full-range `VerifyLedgerChainWork`) sequenced before per-checkpoint `DownloadApplyTxsWork` (a `BatchWork(RETRY_A_FEW)` applying one checkpoint via `ApplyCheckpointWork`, freeing each). The change is a memory-representation change only; computed ledger state, sequencing, and the verification trust model are byte-identical to upstream.

<details>

Cycle 1 — complete change-list. No blocking parity concerns; APPROVE. Verified against `stellar-core/` submodule (pinned v26.0.1).

**parity-gap (none) — phase structure mirrors upstream exactly.**
- `stellar-core/src/catchup/CatchupWork.cpp:148 downloadVerifyLedgerChain(...)` and `:282 mTransactionsVerifyApplySeq = make_shared<DownloadApplyTxsWork>(...)`, sequenced at `:593` — header-chain verify runs as its own work phase before the apply phase. The PR's Phase 1 (`download_ledger_headers` + `verify_header_chain` over the whole range) → Phase 2 (`drive_replay_batches`) matches this ordering.
- `DownloadApplyTxsWork.cpp:27 : BatchWork(app, "download-apply-ledgers", BasicWork::RETRY_A_FEW)` — per-checkpoint batch. The PR's per-checkpoint `drive_replay_batches` loop is the direct analogue. `REPLAY_RETRY_COUNT = 5` matches `RETRY_A_FEW`.
- `DownloadApplyTxsWork.cpp:31 mCheckpointToQueue = checkpointContainingLedger(LCL+1)` and `:225 resetIter()` resetting it to `checkpointContainingLedger(...)` — matches the PR's per-attempt LCL re-derivation from `header_snapshot()` (preserves partial progress on retry).

**sequencing (correct) — tx-set verify interleaving is MORE parity-faithful, not less.** `ApplyCheckpointWork.cpp:280 if (header.scpValue.txSetHash != txset->getContentsHash())` — upstream verifies each ledger's tx-set hash unconditionally, per-ledger, immediately before applying it inside `ApplyCheckpointWork`. The PR's per-batch `verify_txsets` (run before that batch's `replay_via_close_ledger`) reproduces exactly this: tx-set verify strictly precedes that ledger's close. The pre-fix code verified all tx-sets up front before any apply — that was the *less* faithful arrangement. Result-set verify remains skipped for absent entries, matching the original `verify_downloaded_data` behavior.

**edge-case (covered) — reverse-walk trust anchor preserved.** `verify_reverse_walk` is top-anchored (highest→lowest from a trusted SCP/target anchor) and partitions headers into checkpoint groups — it MUST see the full header set in one call. The PR explicitly keeps this whole-range (Phase 1 runs it once over `[lcl+1,target]`); only the gap-linear tx/result bodies are streamed per checkpoint. `test_header_chain_verified_over_full_range_before_replay` proves a tampered mid-range link is still caught. The §11.2 knit-to-LCL 5-case matrix (`knit_to_lcl_decision`, mirroring `ApplyCheckpointWork::getNextLedgerCloseData()`) is unchanged and still runs once up front on the knit prefix + first apply header.

**spec-divergence (none) — gate semantics unchanged.** The #2931 publication gate (`checkpoint_publication_gate`, mirroring `GetHistoryArchiveStateWork` retry-until-published) and the #2940 archive pin still run once per attempt before any fetch, and the pin now threads through both the header-only download and every per-batch body download. `verify_header_chain_from_entries` per-checkpoint integrity check still runs inside `try_download_checkpoint` for each checkpoint in both phases.

No divergence in semantics, edge cases, or sequencing. The resulting bucketListHash and ledger state are identical to upstream (and to the pre-fix path), confirmed empirically by the preserved `catchup_integration` end-to-end target-hash test.
</details>

PR #3228, opened 2026-06-08
Instrument startup/catchup peak RSS

### LINKED ISSUE #3226: Instrument startup/catchup peak RSS (not observed by ledger-close memory report)
## Symptom

The validator's startup / catchup **peak RSS** is not observable. The only memory telemetry (`MemoryReport` "Memory report summary") fires from the ledger-close path, gated on `new_header.ledger_seq % 64 == 0` (`crates/ledger/src/manager.rs:6030`). That code only runs *after* the node is already replaying ledgers. The actual memory peak happens earlier — during HAS download + parallel bucket apply (`BucketList::restore_from_has_parallel`, `crates/bucket/src/bucket_list.rs:2660`, called from `crates/app/src/app/ledger_close.rs:901` and `:1236`, plus the hot-archive restore at `:966`/`:1283` and `initialize_with_precomputed_caches` at `:1015`).

Consequence: on a memory-constrained host, the startup/catchup peak is invisible. The first 64-ledger report only prints once replay is underway, by which point the restore peak has already been released. You find out you exceeded the RAM ceiling only by getting OOM-killed, with no telemetry showing how close you came.

This blocks the larger goal of running a mainnet validator on a **32 GB / no-swap** box, where steady-state (~18 GB anon RSS) is comfortable but the unmeasured catchup peak is the binding risk.

## Evidence

Live mainnet validator (61 GB box), steady state:
- `rss_mb=20460`, `anon_rss_mb=18016` (anon RSS is the binding metric — `file_rss` is lazy mmap of disk-backed buckets)
- `jemalloc_allocated_mb=14156`, `jemalloc_resident_mb=16748`, `fragmentation_pct=18.3`
- `heap_components_mb=7005`, `unaccounted_mb=7151(+)`, `mmap_mb=14999`

`build_memory_report()` (`crates/ledger/src/manager.rs:2570`) and `MemoryReport` (`crates/ledger/src/memory_report.rs`) already parse `/proc/self/status` (`RssAnon` at `memory_report.rs:74`) — but nothing samples it during restore.

## Suspected root cause

Memory instrumentation is coupled to the ledger-close cadence (`manager.rs:6030`). There is no sampling hook on the startup/restore path; `restore_from_has_parallel` and the cache-scan (`scan_level_pairs_for_caches`, `manager.rs:1116`) run before the first ledger close.

## Concrete fix sketch

Add a lightweight, phase-labeled **peak-RSS sampler** that runs from process start through "first ledger closed", independent of the 64-ledger cadence:

- A background thread that reads `/proc/self/status:RssAnon` every ~1–2 s and tracks a running max. Reuse the existing parser in `crates/ledger/src/memory_report.rs` (the `RssAnon` read at `:74`); no hot-path cost.
- Tag samples with a coarse phase the startup code already transitions through: `has-download` → `bucket-apply` (`restore_from_has_parallel`) → `cache-scan` (`scan_level_pairs_for_caches`) → `replay`.
- On reaching steady state (first ledger closed, or sampler stop), emit a one-line summary: `startup_peak_anon_rss_mb=<N> phase=<peak-phase>` so it is greppable like the existing `Memory report summary`.
- Bracket start in the app startup path (around `crates/app/src/app/ledger_close.rs:901`/`:1015`) and stop after the first close in `manager.rs` (near the existing `% 64` report at `:6030`).

Keep it allocation-cheap and off the hot path (separate thread, atomic max). No throughput impact.

## Why this is the first item

This is a prerequisite for validating any 32 GB configuration: nothing downstream (jemalloc tuning, peak-flattening, guardrail recalibration) can be measured or signed off without observing the catchup peak. Once landed, the plan is to run a cold full-catchup under a hard 32 GB cgroup limit (`MemoryMax=32G`, `MemoryHigh≈28G`) and read the peak from this metric.

Non-urgent (observability enablement; no validator-blocking symptom). Related follow-up: guardrail/threshold recalibration for a 32 GB host (separate issue).

tomerweller:
Closes #3226

## Summary

Adds a background peak-anonymous-RSS sampler that covers the startup-restore + catchup window — the region the `% 64` ledger-close memory report misses (that report only fires once replay is underway, after the restore peak has already been released). On stop it emits a greppable `startup_peak_anon_rss_mb=<N> phase=<peak-phase>` one-liner and publishes a new pre-registered `henyey_startup_peak_anon_rss_mb` Prometheus gauge. This makes the catchup peak — the binding risk on a 32 GB/no-swap box — directly observable, supporting verification of the #2901/#3216 catchup-OOM fix.

## Design

- **Sampler** (`crates/ledger/src/peak_rss_sampler.rs`, new): `StartupPeakRssSampler` runs a single tokio-independent `std::thread` that polls `RssAnon` ~1/s via the existing `ProcessMemory::capture()` parser, tracking an `AtomicU64` running max and the `AtomicU8` phase current at the peak. Abstracted over an injectable value-source (`start_with_source` / `sample_once`) so the peak/phase logic is unit-testable deterministically without reading `/proc`. `stop()` is idempotent, joins the thread, logs the one-liner, and returns the peak in bytes; `Drop` is a join fallback so the thread can never outlive the startup window.
- **Bracket** (`crates/app/src/run_cmd.rs` `run_main_loop`): started at the top (before `load_last_known_ledger`), phase tagged `catchup` at catchup entry, stopped before the event loop is spawned (before `start_sync_recovery` / `app.run()`), then the peak is published to the gauge. Strictly before the first event-loop ledger close — no overlap with the `% 64` report.
- **Metric** (`crates/app/src/metrics.rs`): `STARTUP_PEAK_ANON_RSS_MB = "henyey_startup_peak_anon_rss_mb"` added to the `gauges {}` block (auto-included in `ALL_GAUGE_NAMES`, pre-registered at 0).

## Observability-only / no-leak

The sampler reads `/proc` + atomics ONLY; it shares no mutable state with the close/consensus path and has zero effect on hashes, timing, or any observable protocol behavior. The thread is cleanly joined on `stop()` (with a `Drop` fallback). Off-Linux degrades to peak 0 / gauge 0 / no panic, inheriting `ProcessMemory::capture()`'s contract.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3226#issuecomment-4649806316)

## Test plan

- [x] cargo fmt --check
- [x] cargo clippy (pre-commit hook) clean on changed code
- [x] `cargo test -p henyey-ledger -p henyey-app` passes incl. new tests

New coverage (all passing):
- `peak_rss_sampler::tests::test_peak_tracks_max_of_injected_samples`
- `test_peak_records_phase_at_peak`
- `test_phase_transition_updates_current_phase`
- `test_peak_zero_when_source_zero`
- `test_summary_line_format`
- `test_stop_idempotent`
- `test_background_thread_joins_cleanly`
- `metrics::tests::test_startup_peak_gauge_in_catalog`
- `metrics::tests::test_startup_peak_gauge_preregistered_at_zero`

## Deviations from plan

- **Fine phase tags (`bucket-apply`/`cache-scan`) in `ledger_close.rs` were not wired** — these were explicitly nice-to-have per Critic C, "only if the wiring stays cheap." The `log_startup_memory` checkpoints (ledger_close.rs:905/1003) live inside App methods on the restore path with no sampler in scope; threading the handle through would require signature churn across the bucket-restore path, which Critic C said to avoid. The must-have (peak value + coarse `startup`/`catchup` phase, set from `run_main_loop`) is delivered. A process-global phase channel could add the fine tags later if desired.

### DIFF
--- a/crates/app/src/metrics.rs
+++ b/crates/app/src/metrics.rs
@@ -507,6 +507,12 @@ metric_catalog! {
             => "Current SCP ballot phase of the tracking slot (0=unknown, 1=prepare, 2=confirm, 3=externalize)";
         SCP_MEMORY_CUMULATIVE_STATEMENTS = "stellar_scp_memory_cumulative_statements"
             => "Total SCP statements currently held in memory (decreases after slot purging)";
+
+        // Startup/catchup peak RSS (issue #3226). Set once when the startup
+        // sampler stops, before the event loop's first ledger close. Covers the
+        // restore + catchup window that the `% 64` ledger-close report misses.
+        STARTUP_PEAK_ANON_RSS_MB = "henyey_startup_peak_anon_rss_mb"
+            => "Peak anonymous RSS (MB) observed during startup restore + catchup";
     }
 
     gauges_no_prereg {
@@ -2030,6 +2036,35 @@ mod tests {
         assert!(
             output.contains("stellar_protocol_version"),
             "stellar_protocol_version not found in rendered metrics"
+        );
+    }
+
+    /// Issue #3226: the startup peak-anon-RSS gauge is in the gauge catalog and
+    /// pre-registered at 0 (so dashboards render it before the sampler stops).
+    /// Mirrors the #3168 eviction-catalog test pattern.
+    #[test]
+    fn test_startup_peak_gauge_in_catalog() {
+        let gauge_names: HashSet<&str> = ALL_GAUGE_NAMES.iter().copied().collect();
+        assert!(
+            gauge_names.contains("henyey_startup_peak_anon_rss_mb"),
+            "henyey_startup_peak_anon_rss_mb missing from gauge catalog"
+        );
+        let prereg: HashSet<&str> = ALL_PREREGISTERED_GAUGE_NAMES.iter().copied().collect();
+        assert!(
+            prereg.contains("henyey_startup_peak_anon_rss_mb"),
+            "henyey_startup_peak_anon_rss_mb must be pre-registered at 0"
+        );
+    }
+
+    #[test]
+    fn test_startup_peak_gauge_preregistered_at_zero() {
+        let handle = ensure_test_recorder();
+        describe_metrics();
+        register_label_series();
+        let output = handle.render();
+        assert!(
+            output.contains("henyey_startup_peak_anon_rss_mb"),
+            "henyey_startup_peak_anon_rss_mb not found in rendered metrics"
         );
     }
 
--- a/crates/app/src/run_cmd.rs
+++ b/crates/app/src/run_cmd.rs
@@ -47,6 +47,7 @@ use crate::compat_http::CompatServer;
 use crate::config::AppConfig;
 use crate::http::{QueryServer, StatusServer};
 use henyey_history::CatchupRunMode;
+use henyey_ledger::peak_rss_sampler::{Phase as StartupPhase, StartupPeakRssSampler};
 
 /// Node running mode determining behavior and consensus participation.
 ///
@@ -370,6 +371,17 @@ fn print_startup_info(app: &App, options: &RunOptions) {
 
 /// Run the main application loop.
 async fn run_main_loop(app: Arc<App>, options: RunOptions) -> anyhow::Result<()> {
+    // Observability (#3226): sample peak anonymous RSS across the startup
+    // restore + catchup window — the region the `% 64` ledger-close memory
+    // report misses (that report only fires once replay is underway, after the
+    // restore peak has already been released). The sampler runs a single
+    // background std::thread (tokio-independent) that polls RssAnon ~1/s and
+    // tracks a running max; it shares no mutable state with the close/consensus
+    // path and has zero effect on hashes/timing. `mut` so we can tag the phase
+    // and stop it before the event loop is spawned (below), where the peak is
+    // published to the `henyey_startup_peak_anon_rss_mb` gauge.
+    let mut startup_rss_sampler = StartupPeakRssSampler::start();
+
     // Check for force-scp flag (standalone single-node bootstrap).
     // When set, skip all catchup and restore the node directly from DB state.
     let force_scp = app.check_force_scp().await;
@@ -419,6 +431,9 @@ async fn run_main_loop(app: Arc<App>, options: RunOptions) -> anyhow::Result<()>
             tracing::info!("Watcher mode: skipping catchup");
         } else {
             tracing::info!("Node is behind, starting catchup");
+
+            // Tag subsequent peak samples as the catchup phase (#3226).
+            startup_rss_sampler.set_phase(StartupPhase::Catchup);
 
             // Start overlay network BEFORE catchup to receive tx_sets during catchup.
             // This helps bridge the gap between catchup checkpoint and live consensus.
@@ -477,6 +492,17 @@ async fn run_main_loop(app: Arc<App>, options: RunOptions) -> anyhow::Result<()>
             app.refresh_max_tx_size_bytes().await;
         }
     }
+
+    // Stop the startup peak-RSS sampler (#3226) BEFORE the event loop is
+    // spawned, so the reported peak covers exactly the restore + catchup window
+    // and never overlaps with the `% 64` ledger-close report. stop() joins the
+    // sampler thread (no leak), emits the greppable
+    // `startup_peak_anon_rss_mb=<N> phase=<peak-phase>` one-liner, and returns
+    // the peak in bytes; publish it as a gauge. On non-Linux / read error this
+    // is 0 (no panic), inheriting ProcessMemory::capture()'s contract.
+    let startup_peak_bytes = startup_rss_sampler.stop();
+    let startup_peak_mb = startup_peak_bytes as f64 / (1024.0 * 1024.0);
+    crate::metrics::STARTUP_PEAK_ANON_RSS_MB.set(startup_peak_mb);
 
     // Start the sync recovery manager for consensus stuck detection
     app.start_sync_recovery();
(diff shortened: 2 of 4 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Clean, well-tested observability feature. New public surface in `peak_rss_sampler.rs` is thoroughly covered; bracket placement, gauge wiring, and stop/Drop semantics are correct. No correctness concerns block merge.

<details>

**Kind:** feature (issue #3226). Per the correctness lens, every new `pub fn`/`pub` item must have a test exercising it.

**New public surface vs. test coverage (all covered):**
- `Phase` enum + `Phase::as_str()` + `Phase::from_u8` — exercised via `test_phase_transition_updates_current_phase`, `test_peak_records_phase_at_peak`, and the summary-line test (`phase="bucket-apply"`).
- `StartupPeakRssSampler::start_with_source` — `test_background_thread_joins_cleanly`.
- `sample_once` / `peak_bytes` / `peak_phase` / `current_phase` / `set_phase` — `test_peak_tracks_max_of_injected_samples`, `test_peak_records_phase_at_peak`, `test_phase_transition_updates_current_phase`.
- `stop` (idempotent) — `test_stop_idempotent`, `test_peak_zero_when_source_zero`.
- `start()` (real `/proc` path) — indirectly via `start_with_source` (it's a thin wrapper passing `|| ProcessMemory::capture().anon_rss_bytes`).
- Gauge registration — `test_startup_peak_gauge_in_catalog`, `test_startup_peak_gauge_preregistered_at_zero`.

**Peak-logic correctness:**
- `sample_once_impl` uses `fetch_max(sample, AcqRel)` for the running max and records `peak_phase` only on the branch where `sample > prev`. The current-phase atomic is read by the *same* thread that just confirmed the raise, so the recorded (peak, peak_phase) pair is self-consistent. Verified against `test_peak_records_phase_at_peak` (bucket-apply@500 wins over later cache-scan@300).
- Note: with multiple writer threads `fetch_max` + separate `peak_phase` store is not strictly atomic-as-a-pair, but here there is exactly one writer (the sampler thread); `set_phase` only writes `current_phase`. So no race. Correct as designed.

**Stop / Drop semantics:**
- `stop()` is idempotent via the `stopped` bool: re-entry returns the cached peak without re-emitting the log or re-joining (`handle.take()` already left `None`). `test_stop_idempotent` confirms identical return.
- `Drop` is a join-only fallback (sets `stopping`, joins if handle still present, emits nothing). After an explicit `stop()`, the handle is `None`, so Drop is a no-op — no double-join, no panic. Loop exits on `stopping.load(Acquire)` after one final sample, so the thread always terminates.
- `handle.join()` errors are swallowed (`let _ =`) with a comment that a sampler-thread panic must not abort startup — appropriate for observability-only code.

**Bracket placement (run_cmd.rs):** `start()` at the very top of `run_main_loop` (before `load_last_known_ledger`), `set_phase(Catchup)` at the catchup branch entry, `stop()` + gauge publish before `start_sync_recovery()` / the `app.run()` spawn. Covers exactly the restore+catchup window; the `% 64` MemoryReport only fires inside the event loop, so no overlap. Correct.

**Error handling / readability:** Off-Linux and `/proc`-read-error both degrade to 0 via `ProcessMemory::capture()` (returns `Self::default()`; parse uses `unwrap_or(0)`) — no panic. Module doc clearly states the observability-only contract. `expect` on thread spawn is the only panic path and is a genuine unrecoverable startup condition (OS thread exhaustion), consistent with the heartbeat-thread pattern.

**Minor (non-blocking, not filing):** `stop()` computes `let peak = self.peak_bytes()` twice (once before the idempotency check, once after join). Harmless; the second read is the authoritative post-join value. Readability nit only.

No changes required.
</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** Pure observability addition with zero parity surface. The sampler reads `/proc` + atomics only and shares no mutable state with the ledger-close/consensus/hash path; stellar-core has no equivalent to mirror, and nothing in the diff alters protocol-observable behavior, sequencing, or determinism.

<details>

**Parity scope:** The PR touches `crates/ledger/` and `crates/app/` (parity-critical prefixes), so a parity pass is warranted. However, the change adds a *new* observability-only subsystem (peak-RSS sampler + Prometheus gauge + a coarse phase tag) — it does not modify any existing consensus, ledger-close, bucket-apply, or catchup logic. stellar-core has no corresponding peak-RSS-sampler to diverge from; this is a henyey-side operational instrument, not a protocol behavior.

**Determinism / observable-behavior check (the only parity-relevant axis here):**
- `sample_once_impl` writes only to private atomics (`peak_bytes`, `peak_phase`). No write touches ledger state, bucket hashes, SCP, the herder, or timing of any close/externalize.
- The `set_phase(Catchup)` call in `run_main_loop` writes only `current_phase` (a private `AtomicU8`); it does not gate or reorder catchup, overlay start, or finalization.
- The gauge is `.set()` once after `stop()` and strictly *before* the `app.run()` event loop spawns — it cannot interleave with or perturb ledger close.
- The background `std::thread` is tokio-independent and joined before the event loop starts, so it cannot contend with the consensus runtime or shift close timing. Off-Linux it degrades to 0 with no panic, so behavior is identical across platforms (no platform-divergent control flow).

**Sequencing:** START (before restore) → optional `set_phase(Catchup)` (catchup branch) → STOP+publish (before sync-recovery/run spawn). This bracket is bounded entirely within the pre-event-loop startup region and introduces no new ordering dependency on any parity-sensitive step.

**Edge cases:** zero/no-op source, idempotent stop, Drop-join fallback — all covered by tests and none alter externally-observable state. No edge case here maps to a stellar-core semantic that could drift.

No parity gap, no spec divergence, no sequencing concern. APPROVE.
</details>

PR #3234, opened 2026-06-08
Flatten cold-catchup restore/apply peak (#3232): skip warm cache + per-type dead-key scoping

### LINKED ISSUE #3232: 32 GB host: cold-catchup peak 27.1 GB anon RSS leaves insufficient margin — flatten restore/apply peak
## Symptom

A cold full catchup on mainnet peaks at **27.1 GB anonymous RSS**, which leaves insufficient safety margin on the target **32 GB / no-swap** validator host. Steady state (~18 GB anon) fits comfortably; the **catchup peak is the binding risk**, exactly as anticipated when the startup-peak sampler (#3226/#3228) was added.

## Evidence (measured 2026-06-08, production node, #3228 sampler)

Cold catchup from a wiped `~/data/mainnet` (HAS + 36 buckets, 0 cached), `catchup_mode=Minimal` to checkpoint L62939391, on a 63 GB box (uncapped, to measure the true peak with no OOM risk):

```
INFO henyey_ledger::peak_rss_sampler: Startup peak RSS summary startup_peak_anon_rss_mb=27114 phase="catchup"
henyey_startup_peak_anon_rss_mb 27114.48
VmHWM ≈ 27.36 GB
```

- Peak anon RSS = **27.1 GB** during the `catchup` phase (HAS bucket-apply + simultaneous live overlay/fast-track).
- Post-peak, RssAnon released to ~9.8 GB, then climbed back toward the ~18 GB steady state.
- No panic / no OOM. The #3228 sampler worked as designed and correctly attributed the peak to the catchup phase.

## Why this blocks the 32 GB goal

Anonymous RSS is **not reclaimable without swap**. On a hard 32 GB / no-swap host the kernel must keep all 27.1 GB resident at the peak; adding OS + (reclaimable) page cache for the bucket files, the committed-anon floor sits at ~85% of RAM with **< 5 GB margin**. That is high OOM-kill risk during cold catchup — the one operation we cannot make safe by tuning page cache. jemalloc decay tuning (`MALLOC_CONF`) does **not** help here: the 27 GB is live apply working-set, not retained/fragmented dirty pages.

## Suspected root cause

The catchup peak is dominated by the parallel bucket-list restore/apply building in-memory structures concurrently:
- `BucketList::restore_from_has_parallel` (`crates/bucket/src/bucket_list.rs:2660`) applies multiple large buckets in parallel — concurrent large-bucket residency stacks.
- `InMemorySorobanState` (`crates/.../soroban_state.rs`) has no cap/eviction.
- Per-entry bucket caches (`memory_for_caching_mb` default 1024) + bucket object cache (`cache_size` 256) + 4× WASM module cache warm during/after apply.

## Concrete fix sketch (keep performance — operator constraint)

1. **Flatten the apply peak** by bounding *concurrent large-bucket residency* during `restore_from_has_parallel` rather than globally cutting `scan_thread_count` (which would cost startup wall-clock — operator wants performance kept). Process the largest buckets with reduced fan-out / staged release so their transient buffers don't all co-reside.
2. **Catchup-time cache reserve**: temporarily run with reduced `memory_for_caching_mb` (1024→512) during catchup, restoring full size once at steady state. Caches are a warm-up optimization; shrinking them during the apply peak trades a little post-catchup warm-up for headroom exactly when it's scarce.
3. Re-measure peak under each lever via `henyey_startup_peak_anon_rss_mb` until it lands with comfortable margin (target ≤ ~24 GB anon to leave OS + page cache room under 32 GB).

## Human input required

- Confirm the target margin (proposal: peak ≤ ~24 GB anon on a 32 GB host).
- Decision on validator-only vs validator+RPC on the 32 GB box (RPC adds steady RSS; tracked in #3227).

Related: #3226 (instrumentation, done), #3228 (sampler impl), #3227 (monitor guardrail recalibration — depends on this peak number, now measured). Non-urgent (no validator-blocking symptom on the current 61 GB host; this is 32 GB-deployment enablement).

tomerweller:
Closes #3232

## Summary

Flattens the cold-catchup anonymous-RSS peak (measured 27.1 GB, target ≤24 GB on a 32 GB / no-swap validator) with the converged plan's two autonomous-safe levers — no speed or parity tradeoff. Shipped as **two separate commits (A then B)** so each peak delta is independently re-measurable by the operator on the #3228 `henyey_startup_peak_anon_rss_mb` sampler.

Expected peak reduction: **~5–9 GB** (A: ~1.0–1.5 GB, B: ~4–8 GB). Operator verifies via `henyey_startup_peak_anon_rss_mb` on a cold mainnet catchup.

## Commit A — skip warm cache during cold catchup-apply (`91566410`)

The per-entry account/bucket warm cache (`memory_for_caching_mb`, default 1024 MB, `disk_bucket.rs:618`) is a steady-state read-latency optimization wasted during a cold one-pass apply.

- `crates/ledger/src/manager.rs` — `initialize_with_precomputed_caches` gains a `skip_warm_cache` flag; when `true` the inline `maybe_initialize_caches()` is skipped. New `LedgerManager::warm_entry_caches()` post-catchup hook.
- `crates/app/src/app/ledger_close.rs:1015` — catchup restore passes `skip_warm_cache=true`.
- `crates/app/src/app/mod.rs` `restore_operational_state()` — named post-catchup hook calls `warm_entry_caches()` once the node leaves CatchingUp (idempotent; re-sync paths are a no-op).

Steady-state cache behavior and the bucketListHash are unchanged.

## Commit B — per-type dead-key scoping + merge-ordering (`0e362b0d`) — ⚠️ HASH-CRITICAL

`crates/ledger/src/manager.rs`:

**(i) Per-type dead-key scoping.** The cache scan used a single all-types `global_seen: HashSet<LedgerKey>` retaining EVERY offer/pool-share-trustline/contract-data/code/TTL/config key (enum-wrapped) for the whole scan — unbounded, the dominant peak contributor. Replaced with `PerTypeSeen`: one set per `LedgerKey` variant keyed by the compact inner key, mirroring stellar-core's `InMemorySorobanState::initializeStateFromSnapshot` per-type `deletedKeys`. Applied to all three dedup sites (`scan_and_merge`, `scan_and_merge_streaming`, `merge_level_results`).

**Byte-identical equivalence argument (the LedgerKey-is-type-tagged proof):** `LedgerKey` is a discriminated (type-tagged) union — two keys compare equal **iff** same variant tag AND equal inner data. So the variants partition the key space into disjoint classes; a key of one variant can NEVER equal a key of another, regardless of inner bytes. Therefore a single set `S` and the union of per-variant sets have **identical membership**: `S.contains(k)` ⟺ `S_{variant(k)}.contains(inner(k))`. Cross-level shadowing (a newer-level DEADENTRY shadows an older-level LIVEENTRY of the **same** type+key) is driven entirely by `insert()` return values, which are identical across both representations for every `LedgerKey` variant. A catch-all set keyed by the full `LedgerKey` keeps the abstraction total (handles every variant). → Final merged state, soroban state, offer store, and **bucketListHash are byte-identical**.

**(ii) Merge/claim ordering.** Workers previously claimed levels largest-first, but the merge MUST proceed 0→N (shadowing). A large level completing early sat pinned in the `pending` buffer until lower levels merged — worst case all level buffers co-resided. Claiming in **ascending order** makes the awaited level (`next_merge_idx`) complete among the earliest, bounding `pending` to a small reorder window. Correctness-neutral (claim order does not affect the merge result).

**InMemorySorobanState is left UNTOUCHED** — parity floor; stellar-core materializes the identical structure in `initializeStateFromSnapshot`.

## State-identity tests (byte-identical old-vs-new)

- `test_per_type_seen_matches_global_hashset_semantics` — `PerTypeSeen.insert` is identical to `HashSet<LedgerKey>.insert` across every scan-relevant variant; cross-type keys with identical inner bytes never collide (proves the narrowing is exact, not lossy).
- `test_per_type_dedup_preserves_same_type_cross_level_shadowing` — for offer, contract data, contract code, pool-share trustline, and TTL, a newer-level DEADENTRY shadows an older-level same-type LIVEENTRY.
- `test_scan_and_merge_pending_buffer_bounded` — `pending` high-water stays below `num_levels`; ascending-parallel result equals the single-worker streaming result (byte-identical final state). The fail-pre/pass-post here is about **buffer structure**, not correctness — correctness is invariant.
- Lever A: `test_maybe_initialize_cache_skipped_when_budget_zero` (bucket), `test_catchup_cache_init_skips_warm_cache_then_warm_up_runs` (ledger).

**Hash/state vectors UNCHANGED:** all existing `scan_and_merge ↔ scan_and_merge_streaming` equivalence, cache-init, bucket-merge-dedup, `tx_meta_hash_vectors`, and `ledger_close_meta_vectors` tests pass without modification — this is a memory optimization, not a behavior change; bucketListHash is identical.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3232#issuecomment-4651274948)

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy -p henyey-ledger -p henyey-bucket -p henyey-app -- -D warnings` (CI invocation) clean
- [x] `cargo test -p henyey-ledger` — 474 lib + integration suites pass
- [x] `cargo test -p henyey-bucket` — 492 + integration suites pass
- [x] `cargo test -p henyey-app` — 1019 + suites pass
- [x] `cargo build -p henyey` (incl. offline `verify_execution` path) builds clean

## Deviations from plan

The converged plan suggested **two PRs**; per the implementation directive this ships as **one PR with two commits (A then B)** so each peak delta is still independently re-measurable by the operator. No scope change.

### DIFF
--- a/crates/app/src/app/ledger_close.rs
+++ b/crates/app/src/app/ledger_close.rs
@@ -1018,6 +1018,13 @@ impl App {
                 header.clone(),
                 header_hash,
                 cache_data,
+                // Skip the warm per-entry account cache during cold catchup-apply
+                // (#3232). It is a steady-state read-latency optimization wasted
+                // during a one-pass restore and adds ~1.0–1.5 GB to the catchup
+                // anon-RSS peak. It is warmed once post-catchup via
+                // `LedgerManager::warm_entry_caches()`, called from
+                // `restore_operational_state()` when the node leaves CatchingUp.
+                true,
             )
             .map_err(|e| anyhow::anyhow!("Failed to initialize ledger manager from disk: {}", e))?;
 
--- a/crates/app/src/app/mod.rs
+++ b/crates/app/src/app/mod.rs
@@ -1854,6 +1854,15 @@ impl App {
         } else {
             self.set_state(AppState::Synced).await;
         }
+        // Post-catchup warm-up hook (#3232): the cold catchup-apply path skips
+        // the per-entry account/bucket warm cache to flatten the catchup anon-RSS
+        // peak. Now that the node has left CatchingUp and is operational, warm the
+        // cache off-peak so steady-state read latency is unaffected. Idempotent
+        // (`warm_entry_caches` early-returns for already-warm buckets), so the
+        // re-sync paths that also call this are a cheap no-op.
+        if self.ledger_manager.is_initialized() {
+            self.ledger_manager.warm_entry_caches();
+        }
         // Re-arm overlay flood acceptance now that the node is operational.
         if let Some(flag) = self.overlay_synced.lock().unwrap().as_ref() {
             flag.store(true, std::sync::atomic::Ordering::Relaxed);
--- a/crates/bucket/src/disk_bucket.rs
+++ b/crates/bucket/src/disk_bucket.rs
@@ -1382,6 +1382,102 @@ mod tests {
         bytes
     }
 
+    /// Build a `BucketEntry::Liveentry(account)` for the given key byte, matching
+    /// the on-disk record produced by `make_entry_bytes`.
+    fn make_account_bucket_entry(key_byte: u8) -> stellar_xdr::curr::BucketEntry {
+        let account = AccountEntry {
+            account_id: AccountId(PublicKey::PublicKeyTypeEd25519(Uint256([key_byte; 32]))),
+            balance: 100,
+            seq_num: SequenceNumber(1),
+            num_sub_entries: 0,
+            inflation_dest: None,
+            flags: 0,
+            home_domain: String32::default(),
+            thresholds: Thresholds([1, 0, 0, 0]),
+            signers: Vec::new().try_into().unwrap(),
+            ext: AccountEntryExt::V0,
+        };
+        let entry = LedgerEntry {
+            last_modified_ledger_seq: 1,
+            data: LedgerEntryData::Account(account),
+            ext: LedgerEntryExt::V0,
+        };
+        stellar_xdr::curr::BucketEntry::Liveentry(entry)
+    }
+
+    /// Build a DiskBucket backed by a forced `DiskIndex` (via a zero
+    /// `index_cutoff_mb` config) holding `n` sorted account entries. The
+    /// returned bucket has a `DiskIndex` so `maybe_initialize_cache` is exercised
+    /// (InMemory indexes never get a cache).
+    fn build_forced_disk_account_bucket(dir: &std::path::Path, n: u8) -> DiskBucket {
+        use henyey_common::BucketListDbConfig;
+
+        // Write sorted account records to disk and collect (entry, offset) pairs.
+        let mut file_bytes = Vec::new();
+        let mut entries: Vec<(stellar_xdr::curr::BucketEntry, u64)> = Vec::new();
+        for key_byte in 1..=n {
+            let offset = file_bytes.len() as u64;
+            file_bytes.extend(make_entry_bytes(key_byte));
+            entries.push((make_account_bucket_entry(key_byte), offset));
+        }
+        let path = dir.join("forced_disk.bucket");
+        std::fs::write(&path, &file_bytes).unwrap();
+
+        // Force a DiskIndex regardless of file size by setting the cutoff to 0.
+        let cfg = BucketListDbConfig {
+            index_cutoff_mb: 0,
+            ..BucketListDbConfig::default()
+        };
+        let index = LiveBucketIndex::from_entries(
+            entries.into_iter(),
+            DEFAULT_BLOOM_SEED,
+            file_bytes.len() as u64,
+            &cfg,
+        );
+        assert!(
+            !index.is_in_memory(),
+            "test setup expects a forced DiskIndex"
+        );
+
+        let hash = Hash256::hash(&file_bytes);
+        DiskBucket::from_prebuilt(&path, hash, n as usize, index).unwrap()
+    }
+
+    #[test]
+    fn test_maybe_initialize_cache_skipped_when_budget_zero() {
+        // Lever A (#3232): with `memory_for_caching_mb == 0`, the per-entry
+        // warm cache must NOT be allocated, even for a DiskIndex bucket with
+        // account entries. This is the effective budget used during cold
+        // catchup-apply, so the catchup peak pays nothing for the warm cache.
+        use henyey_common::BucketListDbConfig;
+        let dir = tempdir().unwrap();
+        let bucket = build_forced_disk_account_bucket(dir.path(), 8);
+
+        let cfg = BucketListDbConfig {
+            memory_for_caching_mb: 0,
+            ..BucketListDbConfig::default()
+        };
+        // total_bucket_list_account_size_bytes is irrelevant when budget is 0.
+        bucket.maybe_initialize_cache(10_000, &cfg);
+        assert!(
+            bucket.cache().is_none(),
+            "cache must stay None with a zero caching budget (catchup-apply path)"
+        );
+
+        // Sanity: a non-zero budget DOES populate the cache for the same bucket,
+        // proving the bucket is cacheable and the None above is due to the budget,
+        // not the bucket shape. This is the steady-state / post-catchup warm-up.
+        let warm_cfg = BucketListDbConfig {
+            memory_for_caching_mb: 64,
+            ..BucketListDbConfig::default()
+        };
+        bucket.maybe_initialize_cache(10_000, &warm_cfg);
+        assert!(
+            bucket.cache().is_some(),
+            "post-catchup warm-up with a non-zero budget must populate the cache"
+        );
+    }
+
     #[test]
     fn test_streaming_loader_rejects_unsorted_entries() {
         let dir = tempdir().unwrap();
--- a/crates/ledger/src/manager.rs
+++ b/crates/ledger/src/manager.rs
@@ -1800,6 +1800,15 @@ impl LedgerManager {
     /// * `header` - The ledger header to initialize with
     /// * `header_hash` - The authoritative hash of the header from the history archive
     /// * `cache_data` - Pre-computed cache data from `scan_level_pairs_for_caches`
+    /// * `skip_warm_cache` - When `true`, the per-entry account/bucket warm cache
+    ///   is NOT populated here. This is the cold catchup-apply path: the warm
+    ///   cache is a steady-state read-latency optimization that is pure dead
+    ///   weight during a one-pass restore, where it adds ~1.0–1.5 GB of transient
+    ///   anon RSS to the catchup peak (issue #3232). The caller MUST schedule a
+    ///   post-catchup warm-up via [`Self::warm_entry_caches`] so the cache is
+    ///   never left permanently cold (a silent steady-state perf regression).
+    ///   When `false`, the warm cache is populated inline as before (steady-state
+    ///   behavior is unchanged).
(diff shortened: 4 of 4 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Cycle 1. Two-commit memory optimization (cache-skip + per-type dedup). The hash-critical `PerTypeSeen` substitution is byte-identical to the old `HashSet<LedgerKey>` — verified per-variant and empirically via the unmodified equivalence/hash-vector suites. Cache-skip is a read-through optimization, provably correctness-neutral. All new public surface is tested.

<details>

**Kind:** This is issue #3232 (memory optimization / refactor), not a bug-fix — no regression-test-first requirement. New public surface (`warm_entry_caches`, the `skip_warm_cache` param) is covered by `test_catchup_cache_init_skips_warm_cache_then_warm_up_runs` and `test_maybe_initialize_cache_skipped_when_budget_zero`. New private struct `PerTypeSeen` covered by `test_per_type_seen_matches_global_hashset_semantics` + `test_per_type_dedup_preserves_same_type_cross_level_shadowing`. Merge-ordering covered by `test_scan_and_merge_pending_buffer_bounded`.

**`PerTypeSeen` equivalence (the crux) — CONFIRMED byte-identical:**
- The three dedup sites (`scan_and_merge`, `scan_and_merge_streaming`, `merge_level_results`) change ONLY the type of `global_seen` from `HashSet<LedgerKey>` to `PerTypeSeen`. The `.insert(key)`/`.insert(dead_key)` call sites and their boolean-return usage (`if !global_seen.insert(key) { continue; }`) are unchanged. The keys passed in are full `LedgerKey` values, identical to before.
- `PerTypeSeen::insert` dispatches each `LedgerKey` variant to a per-variant `HashSet` keyed by that variant's *exact* inner payload type (`LedgerKey::Offer(LedgerKeyOffer)` → `HashSet<LedgerKeyOffer>`, etc.). Since the inner key IS the entire variant payload, there is no projection/truncation → no info loss. All inner structs derive `Hash, PartialEq, Eq` identically to the enum's structural derives.
- Catch-all `other => self.other.insert(other)` stores the *full* `LedgerKey`, so Account/Data/ClaimableBalance/LiquidityPool are handled losslessly. The abstraction is total over all 10 `LedgerKey` variants.

**Cache-skip (Commit A) — CONFIRMED correctness-neutral:** `DiskBucket::get()` checks `self.cache.get()` first; if `None` (skipped), falls straight through to `ensure_index()` + bloom + index lookup against the disk-backed bucket (the source of truth). Cache is populated on-miss only, never authoritative. Warm-up via `restore_operational_state()` runs after the node leaves `CatchingUp`, guarded by `is_initialized()`, idempotent via `OnceCell` (`is_some()` early-return). Steady-state unchanged.

**Tests run locally (reviewer worktree, PR head == current main):**
- `cargo test -p henyey-ledger --lib` → 474 passed, 0 failed.
- `test_per_type_seen_matches_global_hashset_semantics`, `test_per_type_dedup_preserves_same_type_cross_level_shadowing` → pass.
- `test_scan_and_merge_pending_buffer_bounded`, all `streaming_vs_parallel` / `cross_level` equivalence tests → pass.
- `ledger_close_meta_vectors`, `tx_meta_hash_vectors` → pass with **unmodified** vector files (confirmed via `git diff origin/main...HEAD --name-only`: only the 4 source files touched, no vector/test-fixture files).
- `test_maybe_initialize_cache_skipped_when_budget_zero` (bucket) → pass.

**Error handling / readability:** logging-only additions; no new error paths. Doc comments on `PerTypeSeen` and `initialize_with_precomputed_caches` are accurate and match the code.

No concerns.
</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** Cycle 1. The dedup/shadowing semantics remain byte-identical to stellar-core's `InMemorySorobanState::initializeStateFromSnapshot` filter. `InMemorySorobanState` itself is untouched (parity floor preserved). bucketListHash and meta vectors are unchanged. One non-blocking doc nuance noted inline.

<details>

**Parity-critical paths touched:** `crates/ledger/src/manager.rs` (bucket-list cache scan / merge — the catchup state-derivation path). Reviewed against `stellar-core/src/ledger/InMemorySorobanState.cpp:446` (`initializeStateFromSnapshot`).

**Shadowing semantic — CONFIRMED parity-exact:**
- stellar-core scans buckets newest→oldest, accumulating DEADENTRY keys into `deletedKeys`, and filters any live entry whose key is already in `deletedKeys` (`deletedKeys.find(lk) == deletedKeys.end()`, line 471) — "a newer dead key shadows an older live entry of the same key."
- Henyey reproduces this exactly: levels merge strictly 0→N (newest→oldest), `global_seen.insert(dead_key)` registers dead keys, and `if !global_seen.insert(key) { continue; }` skips shadowed lives. The `PerTypeSeen` split does not change this — per the byte-identity argument (type-tagged enum → disjoint per-variant classes → identical membership), the insert-return decisions are unchanged for every `LedgerKey` variant. Verified all 10 variants map 1:1 (6 explicit by exact inner-key type, 4 via the full-`LedgerKey` catch-all).

**InMemorySorobanState — UNTOUCHED (parity floor preserved):** `crates/ledger/src/soroban_state.rs` is not in the diff (`git diff origin/main...HEAD --name-only`: only the 4 source files). The structure stays resident, matching core's materialized in-memory state. The cache-scan change derives the same `InMemorySorobanState` it always did.

**Merge-ordering change — correctness-neutral (parity-safe):** Workers now claim levels ascending vs largest-first, but the merge order is *independently* fixed at 0→N via `next_merge_idx` regardless of claim/completion order. Claim order affects only transient `pending`-buffer residency, never the merged result. Confirmed empirically: `test_scan_and_merge_pending_buffer_bounded` shows the ascending-parallel result equals the single-worker streaming result (byte-identical).

**bucketListHash / meta vectors — UNCHANGED:** `ledger_close_meta_vectors` and `tx_meta_hash_vectors` pass with **unmodified** fixture files. No hash vector changed → no observable behavior change.

**Sequencing:** the warm-cache skip + post-catchup warm-up moves only the *timing* of a read-through cache; it does not affect consensus-relevant state, the bucket-list install, or hash derivation. Steady-state read path (`DiskBucket::get`) falls through to the disk-backed index identically whether or not the cache is warm.

**Non-blocking note (filed inline, not a block):** The `PerTypeSeen` doc comment and commit message state the design "mirrors stellar-core's `InMemorySorobanState::initializeStateFromSnapshot` per-type `deletedKeys`." Core actually uses a *single* `std::unordered_set<LedgerKey> deletedKeys` in that function (line 458), not per-type sets. The behavioral parity (single-set membership semantics) is exact; only the structural-correspondence claim in the comment is imprecise. Cosmetic — does not affect correctness or parity.
</details>

tomerweller:
## ✅ Merged

Merge commit: `c06e88fb6daef338a2a962b293e4f6564dffcf8c`

Both reviewer lenses APPROVED (Correctness + Parity), CI triple-green (39/39), and all six hash-critical gates rigorously confirmed: per-LedgerKey-variant 1:1 dedup mapping, cross-level shadowing preserved, merge-ordering correctness-neutral, bucketListHash/meta-vectors unchanged, InMemorySorobanState untouched, cache-skip correctness-neutral.

**Follow-up issues filed:** none (the one inline-noted item — a cosmetic doc-comment imprecision re: stellar-core's single-vs-per-type `deletedKeys` — is non-blocking and not worth a backlog issue).

PR #3236, opened 2026-06-08
jemalloc: add retain:false to malloc_conf to lower cold-catchup RSS peak

tomerweller:
Closes #3235

## Summary

Append `,retain:false` to henyey's compiled-in jemalloc `malloc_conf` static (`crates/henyey/src/main.rs`), keeping the existing aggressive `background_thread:true,dirty_decay_ms:1000,muzzy_decay_ms:1000`. `retain:false` munmaps freed virtual extents rather than retaining them for reuse (jemalloc 5.0+ defaults to `retain:true`); combined with the existing 1s decay + background thread, freed memory is returned to the OS continuously (~1s behind frees) throughout the multi-minute cold-catchup apply churn — bounding the **instantaneous** anon-RSS peak that is the binding constraint on a 32 GB / no-swap host.

This is a **pure allocator memory-return policy** change. The only file touched is `crates/henyey/src/main.rs` (the `malloc_conf` static + its doc-comment + one unit test). No consensus / ledger / hash / XDR / SCP code is touched — zero effect on allocation contents, byte layout, ordering, or any observable protocol behavior.

## Notes on the change

- **(a)** The diff is `+,retain:false` appended to the existing `malloc_conf`. We **kept 1000ms decay** and did **NOT** adopt the issue's literal `5000ms` proposal — 5000ms returns pages *later* than the shipping 1000ms and would regress the peak (per triage). `retain:false` is the one genuinely-new lever.
- **(b)** The throughput tradeoff is **real but small and amortized** (the issue's \"no throughput cost\" claim was imprecise): `retain:false` forces heap *growth* to re-`mmap` from the OS (syscall + first-touch fault) rather than recommitting a retained extent. For a validator whose steady-state RSS (~18 GB) is dominated by stable, once-allocated caches/ledger state, growth events are infrequent in steady state, so the cost is amortized — and the operator pre-authorized this lever (#3233 / #3235).
- **(c)** The offline unit test pins the config string (so a future edit can't silently drop a knob). The **full RSS-peak effect is OPERATOR-verified** via `henyey_startup_peak_anon_rss_mb` (#3228) on a cold mainnet catchup. This lever is **additive** to the merged A/B levers (#3234) and **may not reach ≤24 GB alone**.
[...]

Issue #3233, opened 2026-06-08
Operator decision: catchup-apply peak memory-vs-speed levers (fan-out / scan-threads / overlap)


tomerweller:
Tracking the **operator-tradeoff** levers surfaced by the #3232 plan. These reduce the cold-catchup anon-RSS peak further but trade memory for **catchup wall-clock speed**, so they are NOT auto-implemented by /do — they need an operator decision once the autonomous-safe levers (A: skip warm per-entry cache during catchup; B: per-type scan dedup + merge-order buffer bound) land and are re-measured.

## Staged levers (memory saving vs speed cost)

- **C — reduce `scan_thread_count` during catchup** (default 4 → e.g. 1–2). Fewer concurrent `LevelScanResult`s in flight → lower scan peak. Cost: slower cache-scan wall-clock. (scan_thread_count=1 already uses the fused streaming path with no double-buffering.)
- **D — reduce `restore_from_has_parallel` fan-out / stage large-bucket index builds** (crates/bucket/src/bucket_list.rs:2675). Lower concurrent DiskBacked index residency. Cost: slower bucket-list restore wall-clock.
- **E — serialize the cache-scan after merges instead of overlapping** (crates/app/src/app/ledger_close.rs:816 vs :832; the overlap is documented at :779-783 as saving ~180s on the critical path). Lower co-resident peak. Cost: ~+30–60s catchup wall-clock.

## Decisions to make

1. **Target peak margin** — #3232 proposes ≤24 GB anon on a 32 GB / no-swap host. Confirm.
2. **Validator-only vs validator+RPC** on the 32 GB box (RPC adds steady RSS; tracked in #3227). Sizing depends on this.
3. **Acceptable catchup-speed cost** — only relevant IF autonomous levers A+B do not reach the target margin. Triage's recommendation: lean toward accepting the speed cost ("a slower-but-survivable catchup beats an OOM-killed one") since the 32 GB host is fixed.

Re-measure `henyey_startup_peak_anon_rss_mb` after A+B land; only pull C/D/E if still over target.

Parent: #3232.

tomerweller:
[...]

PR #3238, opened 2026-06-09
Fix jemalloc malloc_conf no-op (wrong symbol) and add allocator size-class instrumentation

### LINKED ISSUE #3237: Steady uncategorized live-allocation growth (~0.5 GB/hr) on 24615fb5 — invisible to heap_components leak gate
## Symptom

On the production mainnet validator running `24615fb5` (#3236, `retain:false`), `jemalloc_allocated` (true live bytes) climbs **steadily and monotonically** while the categorized `heap_components` breakdown stays flat. The growth lands entirely in the **uncategorized** region (`unaccounted = jemalloc_allocated − heap_components`), so the existing heap_components-based leak heuristic does not see it.

Rate: **~500 MB/hr**, linear, **not flattening over 4.3 h** of uptime. Not validator-blocking yet (RSS 11.7 GB on a 63 GB box, 40 GB avail), but on the **32 GB / no-swap** target box (#3227) a sustained ~0.5 GB/hr uninstrumented climb would matter, and it's invisible to the current leak gate.

## Evidence

Single PID `3366449`, uptime 0→268 min, no restart. From the per-close `Memory report summary`:

| metric | t168 (~5 min) | t179 (~268 min) | Δ |
|--------|---:|---:|---:|
| `jemalloc_allocated_mb` | 5728 | 7852 | **+2124** |
| `heap_components_mb` | 5590 | 5829 | +239 (flat/oscillating 5740–5830) |
| `unaccounted_mb` | +138 | +2024 | **+1886** |
| `fragmentation_pct` | — | 16.1 | declining |
| `rss_mb` | ~7348 | 11709 | +4361 (incl. warm-up) |

`unaccounted` is monotonic every tick (recent confirmed: 1562 → 1748 → 1899 → 2024 over the last 4 ticks, ~+170 / 20-min tick). It has not decelerated.

Component breakdown at L62946112 (sums to `heap_components` ≈ 5829, all stable):
```
bucket_list_mmap 15383   (file-backed, not anon)
module_cache      2734
offers             946
bucket_list_heap   749
soroban_code       684
soroban_data       515
bucket_list_cache  118
hot_archive_heap    80
executor_state       2
```
The named "unbounded-by-design" caches (`module_cache`, `soroban_code/data`) are **flat** — so they are NOT the growth source. The growth is in live bytes jemalloc counts (`jemalloc_allocated_bytes`) but the report does not attribute to any component.

jemalloc gauges this tick:
```
allocated  7.75 GB
active     8.56 GB
resident   8.96 GB
mapped     9.08 GB
retained  48.8 GB    <-- surprising under retain:false (expected ~0); virtual, not RSS-backed
```

## Suspected root cause

Two open questions for the node team:
1. **Uncategorized live-allocation growth**: something allocates live memory (~0.5 GB/hr) that `build_memory_report()`'s component categorization (`crates/ledger/src/manager.rs:2570`, `crates/ledger/src/memory_report.rs`) does not capture. Either a slow leak in an uninstrumented path, or a legitimately-growing structure missing a component bucket. Cannot distinguish without instrumentation or a longer soak.
2. **`retained_bytes` = 48.8 GB under `retain:false`** (`crates/henyey/src/main.rs:97` sets `retain:false`): jemalloc should munmap rather than retain, so a 48 GB retained pool is unexpected. It's virtual (RSS is 11.7 GB) so not consuming RAM, but it suggests `retain:false` may not be having the intended effect on the arena — worth confirming, since #3236/#3233 deployed `retain:false` specifically to bound the cold-catchup peak.

## Concrete fix sketch

- Add a `memory_report` component (or split) attributing the `unaccounted` region, so the leak gate and daily-summary can see this growth. `crates/ledger/src/memory_report.rs`.
- Confirm whether the `allocated` climb asymptotes (warm-up of an uninstrumented working set) or is a true unbounded leak — a multi-hour soak with the new instrumentation, or a `jemalloc` arena stats dump (`malloc_stats_print`) at two timestamps, would settle it.
- Sanity-check `retain:false` is actually applied to the arena (`retained_bytes` should not grow to tens of GB).

## Not urgent

No label. Not validator-blocking (ample headroom on the 63 GB host; heap_components-categorized state is flat). Relevant to the **32 GB-box viability** thread (#3227, #3233) because a sustained uninstrumented climb undermines the "fit + monitor" posture. Monitor will keep tracking `jemalloc_alloc_mb` / `unaccounted_mb` in the tick `watch` array.

tomerweller:
Closes #3237

## Headline — the jemalloc config has been a silent no-op

Since the April decay config, henyey's compiled-in jemalloc `malloc_conf` has **never been read**. `tikv-jemalloc-sys` 0.6.1 builds jemalloc with `--with-jemalloc-prefix=_rjem_`, so jemalloc reads its in-binary config from the symbol `_rjem_malloc_conf` — never plain `malloc_conf`. henyey exported `#[export_name = "malloc_conf"]`, which matched **none** of jemalloc's config sources, so the entire tuning string (`background_thread:true`, `dirty_decay_ms:1000`, `muzzy_decay_ms:1000`, `retain:false`) was ignored and jemalloc ran **defaults**: `retain:true`, 10000 ms decay, `background_thread:false`. That default `retain:true` is the 48.8 GB retained footprint.

This PR exports the string under `_rjem_malloc_conf`, so the tuning (background_thread + 1000 ms decay + retain:false) is parsed and applied **for the first time**. The previously-merged #3236 / #3234 / decay tuning should NOW actually take effect — potentially a large RSS/retained reduction.

`nm` proof on the rebuilt release binary:
- **Before (origin/main):** only the weak `V _rjem_malloc_conf` (jemalloc default), and NO plain `malloc_conf` symbol — the export was internalized/stripped, never linked to jemalloc.
- **After:** `D _rjem_malloc_conf` — a **strong** defined data symbol that overrides jemalloc's weak default and survives LTO.

## Runtime-assert test (the acceptance gate)

`crates/henyey/tests/jemalloc_config.rs::test_jemalloc_config_is_applied_at_runtime` reads jemalloc's *effective* options via `mallctl` (`opt.retain`, `opt.dirty_decay_ms`, `opt.muzzy_decay_ms`, `opt.background_thread`) — proving jemalloc PARSED the config, not just that bytes are in the binary. It links henyey's allocator config so `mallctl` observes the applied options.

- **Pre-fix (unprefixed export):** verified **FAILS** — `opt.retain == true` (jemalloc defaults), confirming the config is not read.
- **Post-fix (`_rjem_malloc_conf`):** verified **PASSES** — `opt.retain == false`, `dirty/muzzy_decay_ms == 1000`, `background_thread == true`.

## Commit 2 — unaccounted-region instrumentation (observability only)

`crates/ledger/src/memory_report.rs`: `AllocatorStats` now captures the merged-arena `stats.arenas.<all>.{small,large}.allocated` MIBs and the `opt.*` config. `arena_split()` yields a small/large/huge 3-way split that sums **exactly** to `allocated` (huge = residual; falls back to a single `arena_total` bucket if MIBs are unavailable). `MemoryReport::log()` emits the split as `synthetic_class` components and surfaces the `opt.*` fields on the summary event. This localizes which size class the residual ~0.5 GB/hr live-growth concentrates in, and makes the config-applied state observable in logs. No behavior change.

## Operator-verify (re-measure on the node after deploy)

- `retained_bytes` — expect → ~0 (from 48.8 GB) now that `retain:false` is applied.
- RSS / `jemalloc_resident_mb` — expect a drop with 1000 ms decay + background purge active.
- Cold-catchup peak `henyey_startup_peak_anon_rss_mb` (#3228) — expect lower.
- The new `arena_small/large/huge_mb` + `opt_*` fields in the `Memory report summary` log confirm the config and localize the residual climb. **Leak-vs-warmup still needs a multi-hour operator soak** — this PR makes it decidable, it does not itself fix a not-yet-localized leak.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3237#issuecomment-4655787838)

## Test plan

- [x] cargo fmt --check
- [x] cargo build -p henyey (default features, jemalloc) compiles + links
- [x] nm confirms strong `D _rjem_malloc_conf` in the rebuilt binary
- [x] cargo test -p henyey (incl. jemalloc_config runtime-assert) passes
- [x] cargo test -p henyey-ledger passes (memory_report split/opt unit tests)

- **Test:** `crates/henyey/tests/jemalloc_config.rs::test_jemalloc_config_is_applied_at_runtime`
- **Pre-fix:** verified FAILED — `opt.retain == true` (jemalloc defaults; unprefixed symbol never read)
- **Post-fix:** verified PASSES after `2cde6e32` — jemalloc reports `opt.retain == false`, decay 1000, background_thread true

## Deviations from plan

- The instrumentation is entirely allocator-level in `memory_report.rs`; no `manager.rs` wiring change was needed (`build_memory_report` already calls `AllocatorStats::capture()` + `report.log()`, which carry the new fields automatically).
- Avoided adding a `libc` dev-dependency by reading `opt.{dirty,muzzy}_decay_ms` as `isize` and `stats.arenas` as `usize` (byte-identical to jemalloc's `ssize_t`/`size_t`; `raw::read::<T>` is a sized memcpy).
- jemalloc 5.x folds the former "huge" class into "large"; the huge component is therefore the `allocated − small − large` residual (still a meaningful 3-way split that sums exactly).

### DIFF
--- a/crates/henyey/src/main.rs
+++ b/crates/henyey/src/main.rs
@@ -90,9 +90,29 @@ static GLOBAL: tikv_jemallocator::Jemalloc = tikv_jemallocator::Jemalloc;
 //   (infrequent growth events). Operator-authorized per #3233 / #3235.
 //   See #3235 / #3232. Catchup-scoped mallctl purge is the documented fallback
 //   if re-measurement shows steady-state regression.
+//
+// CRITICAL — the export name MUST be `_rjem_malloc_conf`, NOT plain
+// `malloc_conf`. `tikv-jemalloc-sys` builds jemalloc with
+// `--with-jemalloc-prefix=_rjem_`, so every public jemalloc symbol — including
+// the in-binary config symbol — is prefixed. jemalloc's `obtain_malloc_conf()`
+// reads its compiled-in config from the symbol `_rjem_malloc_conf` (and the
+// env source `_RJEM_MALLOC_CONF`); it NEVER reads plain `malloc_conf`. A strong
+// definition under `_rjem_malloc_conf` overrides libjemalloc's *weak* default
+// symbol of the same name, so this string is parsed at allocator init.
+//
+// Before this was fixed (issue #3237), the export was `#[export_name =
+// "malloc_conf"]`. That symbol matched NONE of jemalloc's config sources, so
+// the entire string was a silent no-op: jemalloc ran defaults (retain:true,
+// 10000ms decay, background_thread:false) and all the tuning above did nothing.
+// `nm` on the buggy release binary showed only the weak `_rjem_malloc_conf`
+// (jemalloc's default) and NO plain `malloc_conf` symbol at all.
+//
+// Do NOT "simplify" this back to `malloc_conf`. The runtime-assert test
+// `crates/henyey/tests/jemalloc_config.rs` reads the effective options via
+// `mallctl` and will fail if the config is not actually applied.
 #[cfg(feature = "jemalloc")]
 #[allow(non_upper_case_globals)]
-#[export_name = "malloc_conf"]
+#[export_name = "_rjem_malloc_conf"]
 pub static malloc_conf: &[u8] =
     b"background_thread:true,dirty_decay_ms:1000,muzzy_decay_ms:1000,retain:false\0";
 
--- a/crates/henyey/tests/jemalloc_config.rs
+++ b/crates/henyey/tests/jemalloc_config.rs
@@ -1,0 +1,99 @@
+//! Runtime-assert test that jemalloc actually PARSED henyey's compiled-in
+//! `malloc_conf`, not merely that the config bytes are present in the binary.
+//!
+//! # Why this lives in `tests/` (an integration test), not a unit test
+//!
+//! The config is exported from `crates/henyey/src/main.rs` via
+//! `#[export_name = "_rjem_malloc_conf"]` and is only read by jemalloc when
+//! henyey's `Jemalloc` global allocator is linked. A `henyey-ledger` (or any
+//! library-crate) unit test does NOT pull in the binary's `#[export_name]`
+//! static, so it would observe an *unconfigured* jemalloc and give a false
+//! pass/fail. This integration test compiles into a test binary that links
+//! henyey's allocator config, so `mallctl` reads the real, applied options.
+//!
+//! # What it proves
+//!
+//! It reads jemalloc's *effective* options via `mallctl` (`opt.retain`,
+//! `opt.dirty_decay_ms`, `opt.muzzy_decay_ms`, `opt.background_thread`) and
+//! asserts they match henyey's `malloc_conf`. On the buggy `#[export_name =
+//! "malloc_conf"]` (the unprefixed symbol the prefixed tikv-jemalloc build
+//! never reads), jemalloc runs DEFAULTS — `retain=true`, decay=10000ms,
+//! `background_thread=false` — and these asserts FAIL. After the symbol fix
+//! (`_rjem_malloc_conf`), jemalloc parses the config and the asserts PASS.
+//! This is the acceptance gate for #3237.
+
+// The integration test binary does not itself install henyey's `Jemalloc`
+// global allocator (that lives in `main.rs`, which is the binary, not a lib).
+// We therefore link the same allocator here so the test process runs under the
+// configured jemalloc and `mallctl` observes the applied options. Without this
+// the test would run under the system allocator and `mallctl` would not be the
+// configured jemalloc.
+#[cfg(feature = "jemalloc")]
+#[global_allocator]
+static GLOBAL: tikv_jemallocator::Jemalloc = tikv_jemallocator::Jemalloc;
+
+// Re-export the same compiled-in config the binary uses. The byte string and
+// the `_rjem_malloc_conf` export name MUST stay in sync with
+// `crates/henyey/src/main.rs`. Because this is the prefixed symbol that
+// jemalloc actually reads, the strong definition here overrides jemalloc's
+// weak `_rjem_malloc_conf` default in this test binary exactly as it does in
+// the real binary.
+#[cfg(feature = "jemalloc")]
+#[allow(non_upper_case_globals)]
+#[export_name = "_rjem_malloc_conf"]
+pub static malloc_conf: &[u8] =
+    b"background_thread:true,dirty_decay_ms:1000,muzzy_decay_ms:1000,retain:false\0";
+
+/// Reads jemalloc's effective options via `mallctl` and asserts they match the
+/// values in henyey's `malloc_conf`. This is the real proof the config is
+/// applied — see the module docs.
+///
+/// FAILS on the unprefixed `malloc_conf` export (jemalloc reports defaults);
+/// PASSES once the export uses the prefixed `_rjem_malloc_conf` symbol.
+#[cfg(feature = "jemalloc")]
+#[test]
+fn test_jemalloc_config_is_applied_at_runtime() {
+    use tikv_jemalloc_ctl::{opt, raw};
+
+    // Force at least one allocation through jemalloc so the allocator is
+    // definitely initialized before we query its options.
+    let _warm = vec![0u8; 4096];
+
+    // `opt.retain`: default `true`; henyey sets `retain:false`.
+    let retain: bool = unsafe { raw::read(b"opt.retain\0") }.expect("read opt.retain via mallctl");
+    assert!(
+        !retain,
+        "jemalloc opt.retain must be false (henyey sets retain:false). \
+         Got retain={retain}. If true, jemalloc is running DEFAULTS — the \
+         malloc_conf export is NOT being read (wrong symbol name?)."
+    );
+
+    // `opt.dirty_decay_ms`: default 10000; henyey sets 1000.
+    // jemalloc types this as `ssize_t`, which is byte-identical to Rust `isize`
+    // on every platform we build for; `raw::read::<T>` is a sized memcpy, so
+    // reading it as `isize` avoids pulling in a `libc` dev-dependency.
+    let dirty_decay: isize =
+        unsafe { raw::read(b"opt.dirty_decay_ms\0") }.expect("read opt.dirty_decay_ms via mallctl");
+    assert_eq!(
+        dirty_decay, 1000,
+        "jemalloc opt.dirty_decay_ms must be 1000 (henyey sets dirty_decay_ms:1000). \
+         Got {dirty_decay} (default is 10000 — config not applied)."
+    );
+
+    // `opt.muzzy_decay_ms`: default 10000; henyey sets 1000.
+    let muzzy_decay: isize =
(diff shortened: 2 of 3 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The symbol fix is correct and rigorously proven; the runtime-assert test genuinely captures the bug (fail-pre / pass-post independently reproduced); the instrumentation is well-tested and observability-only. `kind: bug-fix` regression-test requirement satisfied.

<details>

**Cycle 1** — complete change-list. No prior Correctness comment on this PR.

### Acceptance gate — independently reproduced (the proof)
- Checked out the PR head (`08391901`, parent `2cde6e32`) in an isolated `~/data` worktree and ran `cargo test -p henyey --test jemalloc_config`:
  - **Post-fix (`_rjem_malloc_conf`):** `test_jemalloc_config_is_applied_at_runtime ... ok`.
  - **Fail-pre simulation:** flipped *only* the `#[export_name]` line back to the buggy unprefixed `"malloc_conf"` → test **FAILED** with `Got retain=true. ... jemalloc is running DEFAULTS — the malloc_conf export is NOT being read`. This is the exact origin/main bug, reproduced. Restored the file afterward.
- This confirms the test reads jemalloc's *effective* options via `mallctl` (not just byte-presence) and that it truly distinguishes parsed-config from defaults. The regression test would have caught the bug.

### `kind: bug-fix` test-verification
- PR body has a `## Regression test` section; the regression test (`jemalloc_config.rs`) is committed in the fix commit `2cde6e32`. The fail-pre/pass-post claim in the body is verified above. The minor note that the test ships *with* the fix (not in a strictly-prior commit) is moot here — I independently reproduced the fail-pre state by reverting only the export symbol, which is exactly what the prior code was.

### Symbol-fix correctness
- `tikv-jemalloc-sys` 0.6.x builds jemalloc with `--with-jemalloc-prefix=_rjem_`; jemalloc's `obtain_malloc_conf()` reads `_rjem_malloc_conf`, never plain `malloc_conf`. The fix exports under the prefixed name. Confirmed `tikv-jemalloc-ctl = "0.6"` with `features = ["stats"]` at workspace level (required for the `stats.arenas.*` reads) and pulled into the test binary via the `jemalloc` feature → `dep:tikv-jemalloc-ctl`. The `opt::background_thread::read()` and `raw::read::<T>(name)` calls match the 0.6.1 API.

### nm / strong-export (`api-shape`, non-blocking inline)
- `nm` on the **debug test binary** shows `V _rjem_malloc_conf` (weak), not the `D` (strong) the body reports for the LTO release binary. This is not a contradiction — the body's `D` claim is specifically about the optimized release build; `#[export_name]` on a `pub static` can link as weak in debug. The runtime mallctl read-back (retain=false) is dispositive proof the symbol overrode jemalloc's weak default regardless of `V`/`D`. Left as an inline note, not a blocker — operator's release-binary `nm` is the place to reconfirm `D`.

### Instrumentation correctness — well-covered
- `arena_split()` returns `(small, large, huge)` where `huge = allocated.saturating_sub(small).saturating_sub(large)`, so the three components sum exactly to `allocated`. Unit tests verify: normal split sums to allocated, MIB-unavailable fallback collapses to huge, and over-count saturates to 0 (no underflow). All 11 `memory_report` unit tests pass locally.
- `libc_size_t = usize` / reading `opt.*_decay_ms` as `isize` is byte-identical to jemalloc's `size_t`/`ssize_t` on supported platforms (`raw::read::<T>` is a sized memcpy) — a reasonable way to avoid a `libc` dev-dep.
- Best-effort reads (`.unwrap_or(...)`) mean stats-disabled builds degrade gracefully to the fallback rather than panicking.

### Observability-only
- The only external callers of `MemoryReport`/`AllocatorStats` are `app/src/metrics.rs` (metrics capture) and `ledger/src/manager.rs:6180` which does `build_memory_report(seq).log()` every 64 ledgers — result is logged and dropped. New `opt.*` + arena-split fields flow only into `info!` log lines / metrics; never into ledger state, headers, or hashes.

No blocking concerns.

</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** Zero parity surface. The change activates an allocator memory-return policy (decay + retain:false) and adds logging fields — neither affects any stellar-core-observable behavior: ledger state, tx results, hashes, consensus sequencing, or wire format are all untouched.

<details>

**Cycle 1** — complete change-list. No prior Parity comment on this PR. (PR is path-parity-critical because it touches `crates/ledger/`, hence the parity lens; but the touched file is `memory_report.rs`, an observability module.)

### Parity assessment — no divergence possible

**Concern class survey (all clear):**

- `spec-divergence` / `parity-gap`: jemalloc tuning has no stellar-core analogue — stellar-core does not export `malloc_conf` semantics into consensus. The config (`background_thread`, `dirty_decay_ms`, `muzzy_decay_ms`, `retain`) governs only when freed pages are returned to the OS. It does not change *what* gets allocated or any byte of allocated content. There is no protocol observable here.
- `sequencing`: The new code runs (a) at allocator init (config parse, before any ledger work) and (b) in `manager.rs:6180` every 64 ledgers as `build_memory_report(seq).log()` — a fire-and-forget log call after close timing, outside the close/commit critical path and outside any hashed structure. No reordering of ledger close, SCP, or commit steps.
- `edge-case`: `arena_split()` is total and panic-free (saturating residual; fallback to a single huge bucket when MIBs are unavailable). Reads are best-effort `unwrap_or`. A stats-disabled or differently-built jemalloc cannot crash the node or alter behavior — it just logs a degraded split.

### Activation safety (the real runtime change)
- Commit 1 is the first time the decay+retain:false config is actually honored. This is the **intended** configuration the operator authorized (#3233/#3235) — it was a silent no-op before. Activating correct, intended config is safe.
- `retain:false` means freed extents are `munmap`'d instead of held; decay shortens dirty/muzzy page purge from 10000ms→1000ms; background_thread moves purge off the alloc path. None of these affect allocation *contents* — only memory-return timing/policy. There is no correctness or determinism risk: two nodes with different `malloc_conf` produce byte-identical ledger state and hashes.
- Operator must re-measure (retained→~0, RSS, cold-catchup peak) — already enumerated in the PR's Operator-verify section. This is a behavior change for memory footprint only, not for any consensus-observable.

### Scope (independently confirmed)
- `git diff <real-base 24615fb5> HEAD --stat`: exactly 3 files — `crates/henyey/src/main.rs`, `crates/henyey/tests/jemalloc_config.rs`, `crates/ledger/src/memory_report.rs`. No tx/scp/herder/overlay/state/hash code. GH reports `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`.

No parity concerns.

</details>

tomerweller:
## ✅ Merged

**Commit:** 189636357889a382ec64fb0e2ea6567433595c98

Both reviewer lenses APPROVED (Correctness + Parity), no external CHANGES_REQUESTED, CI triple-green (38 SUCCESS / 1 SKIPPED). Runtime-assert acceptance gate independently reproduced: fail-pre (unprefixed `malloc_conf` → `retain=true`, FAILED) and pass-post (`_rjem_malloc_conf` → `retain=false`, ok).

Issue #3252, opened 2026-06-09
Investigate post-restart EXTERNALIZE-following liveness: does a following node resume its own ballot?


tomerweller:
Follow-up from #3250 plan (Critic C scope concern, agreed in-plan, not a force-converge residual).

## Context
#3250's fix makes `heard_from_quorum` a parity-faithful per-slot SCP-ballot signal and decouples the partition-WARN / out-of-sync-recovery gate from it. That fix addresses the **misleading telemetry + spurious recovery gating**, but is an explicit NON-GOAL for the **consensus stall** described in the #3250 escalation comment (node stalled at slot 62956610, peer_gap growing 12→56, after ~2.5h of free-riding on peer EXTERNALIZE post-restart).

## The genuine liveness question
A node that comes up out-of-sync after a restart and stays at tip purely by *following* peer EXTERNALIZE messages is not running its own NOMINATE/ballot for the next slot. When peer EXTERNALIZE for the next slot is unavailable (peers' tx_set caches rotated, or it must externalize itself), it cannot advance. Investigate:

1. Under what conditions does a post-restart following node transition from passive EXTERNALIZE-following back to active consensus participation (nominating the next slot)?
2. Is there a missing nudge — e.g. when the node is v-blocking + at tip but not balloting the next slot — to (re)enter NOMINATE? Compare stellar-core's `HerderImpl::triggerNextLedger` / `setTrackingSCPState` / `outOfSyncRecovery` re-entry conditions.
3. Relationship to #3199/#3205 (near-tip peer-SCP back-fill recovery routing) — does that path correctly hand back to live consensus, or can it leave the node permanently following?

## Why a follow-up, not part of #3250
#3250 is scoped to the heard_from_quorum signal divergence (one reviewable PR). The liveness/recovery-routing investigation is a distinct, larger concern that should not block the telemetry fix.

Refs #3250, #3199, #3205.

tomerweller:
## 🔍 Critic A (correctness) — Round 1

**Verdict:** APPROVE

This is an investigation issue, not a code-change plan; the deliverable is the evidenced (a)/(b) verdict. Evaluating the verdict's correctness.

**Key findings (verified, file:line):**
- **The (a) verdict is correct.** A post-restart node that follows peer EXTERNALIZE does resume its own nomination — there is no follow-only mode. Trace holds end-to-end:
[...]

PR #3205, opened 2026-06-06
Make archive-behind recovery count cap progress-aware (#3204)

tomerweller:
Closes #3204

**CRITICAL consensus-liveness fix** — the third and (per the converged plan) decisive fix in the #1862 → #3187 → #3199 → #3204 arc. URGENT production consensus outage.

## Summary

The post-catchup archive-behind near-tip recovery hard-reset checked the count-based cap (`recovery_attempts >= MAX_POST_CATCHUP_RECOVERY_ATTEMPTS = 3`) **ahead of** the 120s wall-clock backstop and **with no regard to whether #3199's peer-SCP back-fill was making progress**. The per-stuck `recovery_attempts` counter is reset only on a hard reset or successful catchup — never on peer-SCP progress — so it hit the cap in ~40s and hard-reset (`ArchiveBehindRecoveryExhausted`) while back-fill was still shrinking the gap, discarding that progress and stalling for ~6min (observed gap growth 15 → 31 → 48).

**The change — progress-aware cap + zero-gap-progress counter:**
- New `App` tracker: `last_recovery_peer_gap: AtomicU64` (sentinel `u64::MAX` = no prior observation) + `recovery_consecutive_no_gap_progress: AtomicU32`, updated once per Path-A tick in `App::update_recovery_gap_progress(peer_gap)`. The counter **resets to 0 on any strict `effective_peer_gap` decrease** and **increments when the gap does NOT strictly shrink**. It reports `recovery_making_progress = (consecutive_no_gap_progress < RECOVERY_ZERO_PROGRESS_ESCALATION_ATTEMPTS = 3)`.
- `decide_consensus_stuck_action`'s archive-behind arm changes ONLY from `recovery_exhausted` to `recovery_exhausted && !recovery_making_progress`. `would_hard_reset` and the #1843 cooldown fallback are unchanged.

## Anti-wedge independence (the wedge the planning refute pass caught)

The progress signal is **peer_gap-shrink ONLY** — deliberately NOT "current_ledger advanced". `current_ledger` is what re-keys `stuck_start` (`maybe_start_buffered_catchup` filters the snapshot on `current_ledger`); admitting it would reset the episode clock and defeat the 120s backstop, creating a no-escape wedge. peer_gap-shrink never touches `stuck_start`, so the two escapes stay **independent**:

[...]

PR #3199, opened 2026-06-05
Route near-tip/archive-behind recovery to peer-SCP, not a doomed archive reset

tomerweller:
Closes #3197

## Summary

**CRITICAL consensus-liveness production fix.** Corrects the incomplete #3187. The ~6.5-min post-restart consensus stall on every near-tip / archive-behind restart is a recovery-**routing** bug, not the escalation-**timing** bug #3187 addressed.

In the near-tip / archive-behind dead-zone, henyey escalated to archive mechanisms (`ProbeAhead` blocking fetch + the `forcing_catchup_behind` path) that are **structurally guaranteed to fail** until the archive publishes the next (unpublished) checkpoint. #3187 made that escalation fire *early*, but it still terminated in the doomed `ProbeAhead`, so it never shortened the outage — and the 60s hard-reset cooldown then gated the only useful action. stellar-core in this regime does **pure peer-SCP back-fill + buffered-apply** (`HerderImpl::outOfSyncRecovery` + `tryApplySyncingLedgers`) and **never touches the archive**.

The real fix is **recovery routing (peer-SCP), not escalation timing**: route the near-tip band to the existing `broadcast_recovery_scp_state` peer-SCP helper (re-broadcast own latest SCP from `lcl+1` + bounded peer SCP pull), so the small contiguous gap is back-filled from peers and the buffered ledgers apply in seconds-to-tens-of-seconds instead of waiting ~6.5 min for archive checkpoint publication.

## The two edits

- **(primary, load-bearing for the production signature)** `crates/app/src/app/consensus.rs` — `trigger_recovery_catchup`: in the near-tip band (`PEER_AHEAD_ESCALATION_THRESHOLD <= peer_gap < checkpoint_frequency()`) with the archive confirmed behind, skip the doomed `ProbeAhead` hard reset and run the **existing** full `broadcast_recovery_scp_state(current_ledger)` analog + `retry_exhausted_tx_sets()`, returning `None`. This is the dominant path: `escalate_recovery_to_catchup` pins recovery attempts past the escalation threshold on every far-ahead EXTERNALIZE, short-circuiting `out_of_sync_recovery` into this function before it reaches its own `broadcast_recovery_scp_state` tail (the previous fallback only did the `request_scp_state` half, omitting the re-broadcast — the parity gap). Far-behind (#1862) keeps its unchanged `ProbeAhead`/archive escalation.
[...]

Issue #1874, opened 2026-04-21
Non-critical: heard_from_quorum stuck at false for 11+ minutes after HardReset-Escalation recovery (14afe270)


tomerweller:
## Symptom

After the #1867-class livelock at L62214878 was resolved via
`RecoveryEscalation catchup complete` at 03:53:43Z, the subsequent
Heartbeat lines report `heard_from_quorum=false` continuously for 11+
minutes, even though:
- ledger close is happening in real-time (`latest_ext` advancing every ~5s),
- `is_v_blocking=true` on every heartbeat,
- peers count is stable (15–19),
- `scp_sent_ext` keeps incrementing (the node is emitting EXTERNALIZE
  votes, implying it's observing what it needs from its quorum).

## Evidence (session 74535976, host user-dev-050a, binary 14afe270)

```
03:53:43.875669Z  RecoveryEscalation catchup complete ledger_seq=62214911
03:53:51.011Z   Heartbeat tracking_slot=62214921 ledger=62214920 latest_ext=62214921 heard_from_quorum=false is_v_blocking=true
...(72 consecutive heartbeats, ~10s apart)...
04:04:51.104Z   Heartbeat tracking_slot=62215038 ledger=62215037 latest_ext=62215037 heard_from_quorum=false is_v_blocking=true peers=18

# Grep summary over the 11-minute window since recovery:
#   heard_from_quorum=true:  0 heartbeats
#   heard_from_quorum=false: 72 heartbeats
```

## Suspected root cause

The `heard_from_quorum` heartbeat check evaluates the tracker for
`quorum_check_slot = latest_ext` (`crates/app/src/app/lifecycle.rs:846`
→ `herder::heard_from_quorum(slot)` →
`self.slot_quorum_tracker.read().has_quorum(slot, ...)` in
`crates/herder/src/herder.rs:755`).

After the `RecoveryEscalation catchup complete`, the node continues
closing new ledgers and `latest_ext` advances beyond any slot the
`slot_quorum_tracker` was populated with by SCP message processing
before the stall. It appears the tracker either:
- Wasn't updated for slots externalized during the catchup replay path
  (HardResetEscalation → RecoveryEscalation), OR
- Was cleared by the HardReset `dropped buffered state` and the
  per-slot accumulation for post-catchup fresh slots isn't re-priming
  under the `heard_from_quorum` threshold — only under the
  `is_v_blocking` threshold.

[...]

PR #3254, opened 2026-06-10
Source heard_from_quorum from the SCP ballot flag (#3250)

### LINKED ISSUE #3250: heard_from_quorum stuck false after restart while SCP qset is healthy (agree=21) — misleading partition warn
## Symptom

After a process restart, the heartbeat field `heard_from_quorum` reads **`false` indefinitely** even though the node is demonstrably validating with a healthy quorum. It also fires a misleading recurring WARN: `"Have not heard from quorum - may be experiencing network partition"` (lifecycle.rs:1235) on a node that is not partitioned.

## Evidence (2026-06-09, binary `24615fb5`, validator)

Same binary, same config, same data, across a restart:

| run | heartbeats | `heard_from_quorum=true` | `=false` |
|-----|---:|---:|---:|
| pre-restart (18h run) | 6565 | **6556** | 9 |
| post-restart (2h21m, current) | 848 | **0** | **848** |

Yet the authoritative SCP quorum view (`/info` → `quorum`) is fully healthy the entire time:
```
qset: { phase: EXTERNALIZE, validated: true, agree: 21, disagree: 0, missing: 0, delayed: 0, lag_ms: 409 }
transitive: { intersection: true, node_count: 22, critical: [] }
```
The node tracks tip in real-time (peer_gap=1, ~0.17 ledgers/s), externalizes, `is_v_blocking=true`, and `is_validator=true`. So it **is** hearing from its quorum at the SCP ballot level (agree=21, threshold 6) — only the heartbeat's `heard_from_quorum` says otherwise.

## Suspected root cause

The two quorum-accounting paths disagree after a restart:
- `/info` `qset.agree` comes from the live SCP ballot protocol view — healthy (21).
- The heartbeat (`crates/app/src/app/lifecycle.rs:1202`) computes `heard_from_quorum = herder.heard_from_quorum(latest_ext)`, which reads the **capped `SlotQuorumTracker`** (`crates/herder/src/herder.rs:1468` → `tracker.has_quorum(slot, ...)`).

Hypothesis: when the node rejoins after a restart and stays synced by following EXTERNALIZE messages / catchup, the per-slot `SlotQuorumTracker` for `latest_ext` never accumulates a full quorum of PREPARE/CONFIRM statements, so `has_quorum(latest_ext)` returns false — while the ballot protocol's own quorum view is satisfied. The tracker appears not to repopulate to the pre-restart steady state. (Compare #1874, which hardened `heard_from_quorum` against purge_slots / clear_slot_scoped_scp_caches — a related tracker-population path.)

## Impact

- **Not validator-blocking** — the node validates correctly (qset agree=21, intersection=true, real-time). Filing non-urgent.
- Produces a **false "network partition" WARN** every heartbeat (log noise + misleading on-call signal).
- Degrades `heard_from_quorum` as an external monitor signal: monitor-tick check-2 lists `heard_from_quorum=false` as a SYNC-FAILURE condition (gated behind not-real-time + past-deadline, so it did NOT misfire here, but the field is unreliable as a distress signal on its own).

## Concrete fix sketch

- Investigate why `SlotQuorumTracker.has_quorum(latest_ext)` stays false post-restart while SCP ballot quorum is satisfied (`crates/scp/src/...` tracker population vs `crates/herder/src/herder.rs:1468`). Likely the tracker isn't fed the statements that the ballot protocol counts when following via EXTERNALIZE/catchup.
- Either repopulate the tracker consistently on rejoin, or compute the heartbeat `heard_from_quorum` from the same source as `qset.agree`/`validated` so the two agree.
- Monitor follow-up (separate, optional): have check-2 cross-check `heard_from_quorum=false` against `/info` `qset.validated`/`agree` before treating it as a distress signal.

## Related
- #1874 (prior `heard_from_quorum` tracker-population hardening)

tomerweller:
Closes #3250

## Summary

After a restart the heartbeat field `heard_from_quorum` got stuck `false` indefinitely (and fired a recurring false "may be experiencing network partition" WARN) on a node that was demonstrably validating with a healthy quorum (`qset.agree=21`, `is_v_blocking=true`).

Root cause: the heartbeat/`/info` read the henyey-only `SlotQuorumTracker`, which (1) never records the **local node** (`process_verified` skips self-messages at `herder.rs:1965` before `record_envelope`), and (2) was queried for `latest_ext`, an already-externalized slot the node only *followed* — once a slot externalizes peers stop resending PREPARE/CONFIRM and only a **v-blocking** subset of EXTERNALIZE is needed to accept it (#1862). So the tracker reaches v-blocking but not a full remote quorum → the exact observed triad (`is_v_blocking=true`, `heard_from_quorum=false`, `agree=21`).

The fix routes the `heard_from_quorum` *signal* to the SCP ballot protocol's per-slot flag, the exact mirror of stellar-core's `BallotProtocol::mHeardFromQuorum` (`BallotProtocol.cpp:2192`) — it counts the local node and is set for every slot the node has a ballot for, including a followed/externalized slot. The two other consumers are decoupled to match core (which keys neither off `heardFromQuorum`): the partition WARN now gates on `!is_v_blocking`, and the out-of-sync-recovery gate drops the `|| !heard_from_quorum` disjunct (core's `outOfSyncRecovery` keys off `gotVBlocking`, `HerderImpl.cpp:521`).

`SlotQuorumTracker` is preserved verbatim for `is_v_blocking` / `get_v_blocking_slots` / `out_of_sync_recovery` (#1874, #3199/#3205, pipeline-order tests). Only the `heard_from_quorum` source changes.

This fixes the telemetry / partition-WARN / recovery-gate divergence. It does **NOT** fix the post-restart consensus stall — that is the #3199/#3205 recovery-routing arc, tracked separately as **#3252** (WAD here).

## Delegation + rewiring

- `crates/herder/src/herder.rs:1492` — new `pub fn scp_heard_from_quorum(slot) -> bool` = `self.scp.get_slot_state(slot).map(|s| s.heard_from_quorum).unwrap_or(false)`. Existing `heard_from_quorum` (tracker), `is_v_blocking`, `out_of_sync_recovery` left untouched.
- `crates/app/src/app/lifecycle.rs:1202` (heartbeat) — `heard_from_quorum` sourced from `scp_heard_from_quorum(quorum_check_slot)`; `is_v_blocking` still from the tracker.
- `crates/app/src/app/lifecycle.rs:~1243` (partition WARN) — gated on `!is_v_blocking`.
- `crates/app/src/app/lifecycle.rs:~1302` (out-of-sync-recovery gate) — `!can_receive_scp() || !heard_from_quorum` → `!can_receive_scp()`.
- `crates/app/src/app/mod.rs:2683` (`/info`/debug-stats) — `heard_from_quorum` sourced from `scp_heard_from_quorum(quorum_slot)`, so it agrees with `qset.heard` by construction.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3250#issuecomment-4663764188)

## Test plan

- [x] cargo fmt --check
- [x] cargo clippy (lib) clean for henyey-herder / henyey-app / henyey-scp
- [x] `cargo test -p henyey-herder` — 1185 lib tests pass (incl. pipeline-order with test-support, #1874)
- [x] `cargo test -p henyey-app --lib` — 1022 pass (incl. new `/info` test)
- [x] `cargo test -p henyey-scp` — all pass

## Regression test (bug-fix)

- **Tests:** `crates/herder/src/herder.rs::test_scp_heard_from_quorum_true_for_followed_externalized_slot`, `::test_scp_heard_from_quorum_independent_of_local_node_record`, `::test_scp_heard_from_quorum_missing_slot_is_false`, plus `crates/app/src/app/mod.rs::test_debug_stats_heard_from_quorum_matches_scp_ballot_flag`.
- **Pre-fix:** committed as `1475d77d`. Verified FAILED (with `scp_heard_from_quorum` temporarily delegating to the tracker) on the two followed-slot tests with the issue's exact triad: `scp_heard_from_quorum must be true for a followed externalized slot (#3250)` while the tracker yields false (is_v_blocking=true, heard=false).
- **Post-fix:** verified PASSES after `e576ec3c`.

## Preserved (no regression)

- `SlotQuorumTracker` untouched; `is_v_blocking` / `out_of_sync_recovery` unchanged.
- `test_issue_1874_heard_from_quorum_survives_*` pass.
- `crates/herder/tests/scp_envelope_pipeline_order.rs` (`dependency_blocked_envelope_does_not_count_toward_heard_from_quorum_until_ready`) passes unchanged — stays on the tracker method.
- No #1862 / #3199 / #3205 regression.

## Deviations from plan

None.

### DIFF
--- a/crates/app/src/app/lifecycle.rs
+++ b/crates/app/src/app/lifecycle.rs
@@ -1199,7 +1199,16 @@ impl App {
                     // Check quorum status - use latest_ext if available since we have
                     // actual SCP messages for that slot, otherwise fall back to tracking_slot
                     let quorum_check_slot = if latest_ext > 0 { latest_ext } else { tracking_slot };
-                    let heard_from_quorum = self.herder.heard_from_quorum(quorum_check_slot);
+                    // heard_from_quorum reports the SCP ballot protocol's per-slot
+                    // flag (parity with stellar-core's BallotProtocol::mHeardFromQuorum:
+                    // includes the local node, and is set for slots the node only
+                    // *followed* via a quorum of EXTERNALIZE). The henyey-only
+                    // SlotQuorumTracker never records the local node and only reaches a
+                    // v-blocking subset for an already-externalized followed slot, so
+                    // it gets stuck false after a restart while the qset is healthy
+                    // (#3250). The tracker is still the source for is_v_blocking /
+                    // out-of-sync recovery accounting (#1874).
+                    let heard_from_quorum = self.herder.scp_heard_from_quorum(quorum_check_slot);
                     let is_v_blocking = self.herder.is_v_blocking(quorum_check_slot);
 
                     let scp_sent = self.scp_messages_sent.load(Ordering::Relaxed);
@@ -1231,12 +1240,18 @@ impl App {
                     );
                     scp_messages_last_heartbeat = scp_messages_received;
 
-                    // Warn if we haven't heard from quorum for a while
-                    if self.is_validator && !heard_from_quorum && peers > 0 {
+                    // Warn if we are not even hearing from a v-blocking set, which
+                    // is the real partition signal. stellar-core keys partition /
+                    // recovery off v-blocking, never off heardFromQuorum — a node
+                    // that is v-blocking is demonstrably hearing from its quorum at
+                    // the ballot level, so gating this WARN on !heard_from_quorum
+                    // produced a false "network partition" warning every heartbeat
+                    // after a restart (#3250). Gate on !is_v_blocking instead.
+                    if self.is_validator && !is_v_blocking && peers > 0 {
                         tracing::warn!(
                             tracking_slot,
-                            is_v_blocking,
-                            "Have not heard from quorum - may be experiencing network partition"
+                            heard_from_quorum,
+                            "Have not heard from a v-blocking set - may be experiencing network partition"
                         );
                     }
 
@@ -1284,10 +1299,14 @@ impl App {
                     }
 
                     // Out-of-sync recovery: purge old slots when we're too far behind.
-                    // This mirrors stellar-core's outOfSyncRecovery() behavior.
-                    // When we have v-blocking slots that are >100 ahead of older slots,
-                    // purge the old slots to free memory and allow recovery.
-                    if !self.herder.state().can_receive_scp() || !heard_from_quorum {
+                    // This mirrors stellar-core's outOfSyncRecovery() behavior, which
+                    // keys off gotVBlocking (HerderImpl.cpp:521) — never off
+                    // heardFromQuorum. out_of_sync_recovery() itself early-returns
+                    // when Tracking and otherwise scans v-blocking slots. Gating the
+                    // call on !heard_from_quorum (the previously-stuck tracker flag)
+                    // spuriously triggered recovery on a healthy node after a restart
+                    // (#3250); drop that disjunct to match core.
+                    if !self.herder.state().can_receive_scp() {
                         if let Some(purge_slot) = self.herder.out_of_sync_recovery() {
                             tracing::info!(
                                 purge_slot,
--- a/crates/app/src/app/mod.rs
+++ b/crates/app/src/app/mod.rs
@@ -2678,7 +2678,12 @@ impl App {
             peer_count: self.peer_count().await,
             pending_envelopes: herder_stats.pending_envelopes,
             cached_tx_sets: herder_stats.cached_tx_sets,
-            heard_from_quorum: self.herder.heard_from_quorum(quorum_slot),
+            // Source from the SCP ballot protocol's per-slot flag so this matches
+            // the /info qset `heard` view by construction (parity with
+            // stellar-core's getJsonQuorumInfo→ret["heard"]). The henyey-only
+            // SlotQuorumTracker excludes the local node and gets stuck false for a
+            // followed slot after a restart (#3250); is_v_blocking still uses it.
+            heard_from_quorum: self.herder.scp_heard_from_quorum(quorum_slot),
             is_v_blocking: self.herder.is_v_blocking(quorum_slot),
             slot: slot_state.map(Into::into),
             consensus_trigger_timer_fires: self
@@ -5078,7 +5083,41 @@ mod tests {
         // Verify quorum tracking methods are callable
         let slot = app.herder.tracking_slot().get();
         let _heard = app.herder.heard_from_quorum(slot);
+        let _scp_heard = app.herder.scp_heard_from_quorum(slot);
         let _blocking = app.herder.is_v_blocking(slot);
+    }
+
+    /// Regression test for #3250: the debug-stats / heartbeat `heard_from_quorum`
+    /// is sourced from the SCP ballot protocol's per-slot flag (which matches the
+    /// /info qset `heard` view), NOT the henyey-only SlotQuorumTracker. The two
+    /// must agree for the slot the stats query.
+    #[tokio::test]
+    async fn test_debug_stats_heard_from_quorum_matches_scp_ballot_flag() {
+        let dir = tempfile::tempdir().expect("temp dir");
+        let db_path = dir.path().join("rs-stellar-test.db");
+        let config = crate::config::ConfigBuilder::new()
+            .database_path(db_path)
+            .build();
+
+        let app = App::new(config).await.unwrap();
+
+        let stats = app.simulation_debug_stats().await;
+
+        // The stats slot is the same quorum_slot used to compute heard_from_quorum.
+        let quorum_slot = stats
+            .tracking_slot
+            .get()
+            .max(stats.current_ledger as u64 + 1)
+            .max(1);
+
+        // Source of truth: the SCP per-slot ballot flag (parity with core's
+        // BallotProtocol::mHeardFromQuorum). The debug-stats field must equal it
+        // exactly — by construction now that both read scp_heard_from_quorum.
+        let scp_flag = app.herder.scp_heard_from_quorum(quorum_slot);
+        assert_eq!(
+            stats.heard_from_quorum, scp_flag,
+            "debug-stats heard_from_quorum must mirror the SCP ballot flag (#3250)"
+        );
     }
 
     #[tokio::test]
--- a/crates/herder/src/herder.rs
+++ b/crates/herder/src/herder.rs
@@ -1470,6 +1470,32 @@ impl Herder {
         tracker.has_quorum(slot, |node_id| self.scp_driver.get_quorum_set(node_id))
     }
 
+    /// Report `heard_from_quorum` from the SCP ballot protocol's per-slot flag.
+    ///
+    /// This is the parity-faithful source for the heartbeat / `/info`
+    /// telemetry, mirroring stellar-core's per-slot
+    /// `BallotProtocol::mHeardFromQuorum` (`stellar-core/src/scp/BallotProtocol.cpp:2192`).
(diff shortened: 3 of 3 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Tightly-scoped rewiring of `heard_from_quorum` to the SCP per-slot ballot flag. New public fn is correct, tests cover it, regression triad verified fail-pre / pass-post locally, and no behavior outside telemetry + recovery-trigger input changes.

<details>

**Cycle 1** — complete change-list. No prior Correctness comment on this PR.

`kind: bug-fix` (issue #3250). PR body has a `## Regression test` section. Verified against the diff and locally.

**Test verification (all pass):**
- New `scp_heard_from_quorum` accessor (`crates/herder/src/herder.rs:1492`) is a 4-line delegation to `scp.get_slot_state(slot).map(|s| s.heard_from_quorum).unwrap_or(false)`. Confirmed `SlotState.heard_from_quorum` (scp.rs:703) reads `slot.heard_from_quorum()` → ballot protocol flag (ballot/mod.rs:304).
- **Regression triad verified fail-pre/pass-post.** I temporarily re-pointed `scp_heard_from_quorum` at the tracker (`self.heard_from_quorum(slot)`) and ran `cargo test -p henyey-herder --lib scp_heard_from_quorum`: the two followed-slot tests FAILED with the #3250 triad message (`scp_heard_from_quorum counts the local node …`); after reverting, all 3 PASS. The missing-slot test passes both ways (both return false for unknown slot — correct).
- New app test `test_debug_stats_heard_from_quorum_matches_scp_ballot_flag` passes (`-p henyey-app --lib`): asserts the debug-stats field equals the SCP flag by construction.
- `#1874` preservation tests, `#2417`/`#2706` `out_of_sync_recovery` checkpoint tests, and the `scp_envelope_pipeline_order` `dependency_blocked_envelope_does_not_count_toward_heard_from_quorum_until_ready` test (which stays on the tracker method) all pass unchanged.

**Logic / error handling:**
- `.unwrap_or(false)` for a missing slot is the correct conservative default and matches the prior `heard_from_quorum` contract (false when no state).
- Partition WARN now gates on `!is_v_blocking`; the `heard_from_quorum` value is still logged as a field, so operators retain visibility. Message text updated to match the new gate.
- The diff touches only `lifecycle.rs`, `mod.rs`, `herder.rs`. `SlotQuorumTracker`, `is_v_blocking`, and the `out_of_sync_recovery()` body are untouched (`git diff --stat` + manual confirmation).

**Readability:** comments are accurate and cite the precise core lines (`BallotProtocol.cpp:2192`, `HerderImpl.cpp:521/537`). Good.

No blocking correctness concerns.
</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** All three rewired consumers move toward stellar-core's actual semantics. The consensus-sensitive recovery-gate change is correct: it triggers recovery LESS (drops a stuck-false disjunct), and the dropped path was a no-op-or-spurious call given `out_of_sync_recovery()` self-gates on Tracking. No nominate/ballot/externalize decision changes.

<details>

**Cycle 1** — complete change-list. No prior Parity comment. Classes: `parity` (delegation), `recovery-gate-parity` (the consensus-sensitive one), `spec-divergence` (partition WARN).

### 1. Delegation parity (`scp_heard_from_quorum`) — MATCHES core
- Core's `BallotProtocol::mHeardFromQuorum` (BallotProtocol.cpp:2192-2221): set true when `LocalNode::isQuorum(...)` over `mLatestEnvelopes` (which includes the local node's own envelope). henyey's `check_heard_from_quorum` (ballot/statements.rs:423-475) iterates `self.latest_envelopes` and calls `is_quorum(local_quorum_set, &nodes, …)` — the local node is in `latest_envelopes`, so it is counted. **Confirmed: the SCP flag includes the local node and is set for followed Confirm/Externalize slots** (statements.rs:435 includes Confirm/Externalize unconditionally). This is the exact fix for the #3250 triad — the henyey-only `SlotQuorumTracker` excludes self (self-message skip) and only reaches a v-blocking remote subset for a free-ridden externalized slot.
- `get_slot_state(slot)` (scp.rs:698-709) surfaces `slot.heard_from_quorum()` → ballot flag. Delegation is faithful.

### 2. Recovery-gate change (`recovery-gate-parity`) — the consensus-sensitive one. CORRECT.
Adversarial trace:
- `can_receive_scp()` = `Syncing | Tracking` (state.rs:44). So `!can_receive_scp()` ⇔ `Booting`.
- **Old gate:** `!can_receive_scp() || !heard_from_quorum`. **New gate:** `!can_receive_scp()`. Dropping a disjunct ⇒ the gate fires **LESS** often (strictly narrower trigger). It is impossible for this change to *over-trigger* (thrash) recovery — the concern is only whether it *under-triggers*.
- Under-trigger check: `Herder::out_of_sync_recovery()` (herder.rs:1491-1495) **early-returns `None` when `state()==Tracking`** — the exact mirror of core `outOfSyncRecovery`'s `if (isTracking()) return;` (HerderImpl.cpp:525). So under the OLD code, whenever the node was Tracking with the stuck-false tracker flag (the #3250 steady state), the gate called `out_of_sync_recovery()` which immediately returned None — a wasteful spurious call, never an actual purge. Dropping that disjunct removes the spurious call and changes NOTHING about whether a purge happens on a Tracking node.
- Core parity of the trigger: core drives `outOfSyncRecovery` from a timer gated on `!isTracking()`; the `gotVBlocking` test (HerderImpl.cpp:537) is the *internal slot-selection* loop, NOT the trigger — and henyey preserves that internal loop verbatim (`scp.got_v_blocking`, herder.rs:1505-1513). So the recovery *decision* logic is unchanged; only a redundant call-site disjunct is removed.
- Residual nuance (non-blocking): the new gate is `Booting`-only, narrower than core's `!isTracking()` (which would also cover `Syncing`). But the catchup-driving recovery path is the separate, untouched `SyncRecoveryManager → App::out_of_sync_recovery(current_ledger)` (lifecycle.rs:1004); this heartbeat gate is only the herder-level memory-purge. A Syncing node's purge is still reachable via that primary path, and the #2417/#2706/#1874/#3199-arc tests all pass. No liveness regression: the change cannot cause the node to *miss* a stuck state it would previously have recovered from, because the previously-extra calls were Tracking-state no-ops.

### 3. Partition WARN (`spec-divergence`) — MATCHES core intent
Core keys partition/recovery off v-blocking, never off `heardFromQuorum`. Gating the WARN on `!is_v_blocking` (the real partition signal) and keeping `heard_from_quorum` as a logged field is the correct alignment. A v-blocking node is demonstrably hearing from its quorum at the ballot level.

### 4. No consensus DECISION change — confirmed
`heard_from_quorum` is telemetry (`/info`, heartbeat) + a recovery-trigger input. It is NOT consumed by nomination, ballot voting, or externalize. The nominate/vote/externalize paths are untouched. The change alters only recovery-triggering + telemetry.

### SlotQuorumTracker preserved — confirmed
`git diff --stat` shows only lifecycle.rs / mod.rs / herder.rs. `slot_quorum_tracker.rs` untouched; `is_v_blocking` body unchanged; `out_of_sync_recovery()` body unchanged.

No blocking parity concerns.
</details>

tomerweller:
## ✅ Merged

**Commit:** 2a6ceae43a6874366ecd38e6f09ef5c2986eecac

Both agent lenses (Correctness + Parity) APPROVE; 39/39 CI checks green. The consensus-sensitive recovery-gate change was verified to trigger recovery *less* (drops a stuck-false disjunct that was a Tracking-state no-op via `out_of_sync_recovery()`'s own `isTracking` early-return), matching core's `outOfSyncRecovery` semantics with no #3199/#3205/#1862 regression. Regression triad verified fail-pre/pass-post locally.

PR #3206, opened 2026-06-06
Fire recovery-stalled WARN on post-restart absolute forcing_catchup_behind stall

tomerweller:
Closes #3198

## Summary

A self-recovering `forcing_catchup_behind` stall that completes before the 15m clean-restart sync deadline (so check (2) never reports SYNC FAILURE) and accrues its entire burst across the single tick interval that straddles the check-12b baseline-reset tick was invisible to both check (2) and check (12b). The #3197 episode accrued 213 stall-ticks undetected. The `recovery-stalled` counter-streak alarm reset its streak baseline and returned `collecting_baseline` on a PID/`start_ticks` change (restart) or counter reset, discarding the absolute value without evaluating it.

This adds a post-restart absolute-value path (converged plan Option A): on **both** reset branches in `eval_counter_streak`, after writing the fresh baseline snapshot, the discarded absolute value is evaluated via `maybe_post_restart_fire`; if it meets `post_restart_absolute_threshold = 50` the alarm fires WARN once (keyed on the absolute value, with a `post_restart` marker) instead of `collecting_baseline`. The fresh baseline is written **before** the early return so the streak machine stays consistent next tick; the existing 7200s cooldown prevents re-fire. Normal same-incarnation streak/burst logic is unchanged.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3198#issuecomment-4637472572)

## Changes

- `.claude/skills/shared/metric-alarms.toml` — add `post_restart_absolute_threshold = 50` to the `recovery-stalled` block (severity stays WARN, cooldown 7200s).
- `scripts/lib/eval-alarms.py` — new `maybe_post_restart_fire` helper; invoked from both the PID/`start_ticks`-changed and the counter-reset branches of `eval_counter_streak` after the fresh baseline write. Unknown TOML field is allowlist-ignored by `validate_catalog`; `maybe_reset_counter_snapshot` only acts on `skipped`, so the firing baseline is preserved.
- `.claude/skills/monitor-tick/SKILL.md` — Check 12b "Invalidation" section documents the post-restart-absolute fire path and the #3197/#3198 blind spot; adds `recovery_stalled: WARNING absolute=N (post-restart)` status-line examples.
- `scripts/test-monitor-skill-snippets.sh` — new multi-tick eval-alarms tests (148c firing@213, 148d no-fire@5) and Test 41b cross-validating `post_restart_absolute_threshold == 50`.

## Test plan

[...]

PR #3257, opened 2026-06-10
Skip/re-baseline counter-delta alarms on a gap-stale monitor baseline

### LINKED ISSUE #3246: Non-critical: monitor-tick: counter-delta alarms inflate after a long loop gap (stale-but-valid prev.prom)
## Symptom

When the monitor loop has a long wall-clock gap between ticks but the validator **process survives** (same PID/start_ticks), check-12's `PREV_PROM_INVALID` stays `false` — so counter-delta alarms evaluate `current - prev` across the entire gap instead of one ~20-min interval. Counts that accumulated slowly over many hours of normal operation then fire as if they were an acute burst, and they cannot be time-resolved from a single delta.

## Evidence (tick 184, 2026-06-09T16:32Z)

The loop was dark 04:46Z → 16:32Z (~11.7h; `ScheduleWakeup` chain dropped across a context-compaction). Process PID 3366449 survived the whole time. On resume, `prev.prom` was the 04:46Z scrape — same incarnation, so `PREV_PROM_INVALID=false`. The evaluator reported:

```
metrics: 3/31 firing
recovery_stalled: WARNING delta=703 (burst) — investigating
FIRING: lost-sync (SYNC, delta=17), post-catchup-hard-reset (ACTION, delta=6),
        overlay-errors-high (WARN, delta=165), recovery-stalled (WARN, burst delta=703)
```

A live 10s re-sample showed **delta=0 on every one** of those counters — the node is Validating at tip, healthy *now*. The deltas were genuine counts accumulated over the 11.7h dark window (a real archive-behind episode, commented on #3218), but the burst/threshold machinery presented them as acute. The `recovery_stalled` burst override (`delta >= 10`) is the clearest case: it's designed to catch sustained trouble "regardless of tick granularity," which is correct for a *skipped* tick (counter preserved, streak reset), but there is no equivalent guard for a *survived-process long-gap* where a stale-but-valid baseline is used directly.

## Suspected root cause

`scripts/lib/eval-alarms.py` / SKILL.md check-12 invalidate the baseline only on PID/`start_ticks` change (process restart). There is no check on the **age** of `prev.prom`. The `scrape_identity` file already carries a `timestamp=` field, so the prev-scrape wall-clock age is recoverable.

## Concrete fix sketch

- In check-12, after the identity check, compute `prev_scrape_age = now - timestamp(scrape_identity)` (the identity file describes the process that produced `prev.prom`). If `prev_scrape_age` exceeds a threshold (e.g. > 3× the nominal tick interval, ~60 min), treat the baseline as **gap-stale**:
  - Skip the §COUNTERS/§HISTOGRAMS delta fires and the §12b burst/streak fire for this one tick (same posture as `PREV_PROM_INVALID`, but emit a distinct reason e.g. `gap-stale (prev age Xh)` so it's visible), OR
  - Annotate fired alarms with the prev age so the operator/daily-summary can see the delta spans hours, not minutes.
- Point-in-time gauges are unaffected (they read `current.prom` only) and should still evaluate normally.
- Reset the two-tick gauge persistence counters on a gap-stale tick (same as `PREV_PROM_INVALID`).

This would have rendered tick 184's counter alarms as `gap-stale` instead of 4 false-acute fires, while still surfacing the underlying archive-behind episode via the absolute-value path / #3218 comment.

## Not a node bug

The validator was healthy throughout; this is purely monitor-tick calibration. No label per the self-reflection Tier 2 policy.

tomerweller:
Closes #3246

## Summary

When the monitor loop stalls across a long wall-clock gap but the validator process survives (same PID/start_ticks), check-12's identity check (`PREV_PROM_INVALID`) stays `false`, so the counter-delta is computed across the entire gap and false-fires as an acute burst (the headline tick-184 `recovery-stalled` WARN with delta=703). This adds a baseline-age (gap-stale) signal to the evaluator and applies it asymmetrically by alarm family.

## Approach

- **SKILL (`.claude/skills/monitor-tick/SKILL.md`):** check-12 step 5 now computes `prev_scrape_age = now − timestamp(scrape_identity)` and exports `PREV_SCRAPE_AGE_SECONDS`. Documents the gap-stale rule, the default threshold, the env var in the evaluator-invocation list, and the archive metadata sidecar.
- **Evaluator (`scripts/lib/eval-alarms.py`):** derives `gap_stale = age >= 0 and age >= GAP_STALE_THRESHOLD_SECONDS` (default `3600`), sequenced **after** the PID-identity check so #3206's PID-change case is never double-handled. Applied asymmetrically:
  - **prev.prom families** (`eval_counter`, `eval_counter_dynamic`, `eval_histogram_p99`) → **SKIP** the acute fire with a `gap-stale (prev age Xh)` reason. The `eval_counter_dynamic` guard precedes its prior-delta snapshot write so the gap-spanning delta is not stored.
  - **snapshot families** (`eval_counter_ratio`, `eval_counter_streak` — the headline) → **RE-BASELINE** (rewrite snapshot + reset streak to `0`, reusing the counter-reset path), because snapshots persist across ticks.
  - **Age-unknown (`-1`/absent)** → not stale (fail-safe; truly-missing identity is already covered by `PREV_PROM_INVALID`).

## Distinctness from #3206

#3206 handles the PID-change (restart) case via `PREV_PROM_INVALID` / the snapshot PID-mismatch reset. Gap-stale is the **distinct** same-PID-but-old-baseline case: the identity check passes, so #3206's logic never triggers. Gap-stale is sequenced strictly after the identity check in every evaluator, so a PID change is handled there and never double-handled.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3246#issuecomment-4665719912)

## Test plan

- [x] `bash scripts/test-monitor-skill-snippets.sh` — all new tests pass (the only remaining failures, pv-label-sync 152–155, are pre-existing on origin/main and unrelated)
- [x] `python3 -c 'import ast; ast.parse(open("scripts/lib/eval-alarms.py").read())'`
- [x] `python3 -m pytest scripts/ci/test_check_alarm_versions.py` — 23 passed
- [x] `check-alarm-versions.py` green (no TOML change ⇒ no `baseline_version` bump)
- [x] `bash -n` + `zsh -n` clean

## Regression test

- **Tests:** `scripts/test-monitor-skill-snippets.sh` — 148f (counter skip), 148g (normal interval still fires), 148h (recovery-stalled re-baseline), 148i (counter-ratio re-baseline) + 2 SKILL-doc assertions.
- **Pre-fix:** committed as `5d076e5eb04ac0ccbd2dabd51230bc5c8a31ef58` — 148f/148h/148i and the doc tests FAILED on origin/main (e.g. counter skip got `state=firing`; streak got `firing` instead of `collecting_baseline`).
- **Post-fix:** all PASS after `b9600969f4cba56ba88d9adb17a614e103077bd4`. 148g (normal ~20-min interval) fires both pre- and post-fix, guarding against over-skipping.

## Deviations from plan

None. (Plan line-number references were ~20 lines off from the current file; logic placement matches the plan's described anchors.)

### DIFF
--- a/.claude/skills/monitor-tick/SKILL.md
+++ b/.claude/skills/monitor-tick/SKILL.md
@@ -929,6 +929,36 @@ against a node that is in real-time sync with age=2s).
      reason=`no scrape_identity` or `scrape_identity malformed`.
      This handles rollout (existing sessions have prev.prom but no
      scrape_identity), manual deletion, and corruption.
+   - **Gap-staleness check (`PREV_SCRAPE_AGE_SECONDS`, #3246):** the identity
+     check above only catches a *process* change (PID/`start_ticks`). It does
+     NOT catch the case where the monitor **loop** stalled across a long
+     wall-clock gap while the validator **process survived** — same PID, so
+     `PREV_PROM_INVALID=false`, but `prev.prom`/the snapshot baseline is now
+     hours old. The counter-delta would then be computed across the whole gap
+     and false-fire as an acute burst (the headline tick-184 `recovery-stalled`
+     case). Compute `prev_scrape_age = now − timestamp(scrape_identity)` from
+     the `timestamp=` field of `scrape_identity` (which describes the process
+     that produced `prev.prom`) and export it for the evaluator:
+     ```bash
+     # prev_ts = ISO8601 timestamp from the EXISTING scrape_identity (the one
+     # describing prev.prom), read BEFORE the fresh write below. If the file is
+     # absent/malformed/has no timestamp, leave the age unknown (-1) — that case
+     # is already covered by PREV_PROM_INVALID, and an unknown age is treated as
+     # NOT stale (fail-safe).
+     PREV_TS=$(awk -F= '/^timestamp=/{print $2}' \
+       /home/tomer/data/$MONITOR_SESSION_ID/metrics/scrape_identity 2>/dev/null)
+     if [ -n "$PREV_TS" ]; then
+       PREV_EPOCH=$(date -u -d "$PREV_TS" +%s 2>/dev/null || echo "")
+       NOW_EPOCH=$(date -u +%s)
+       if [ -n "$PREV_EPOCH" ]; then
+         export PREV_SCRAPE_AGE_SECONDS=$(( NOW_EPOCH - PREV_EPOCH ))
+       else
+         export PREV_SCRAPE_AGE_SECONDS=-1
+       fi
+     else
+       export PREV_SCRAPE_AGE_SECONDS=-1
+     fi
+     ```
    - Write fresh identity file:
      ```bash
      printf "version=1\npid=%s\nstart_ticks=%s\ntimestamp=%s\n" \
@@ -963,6 +993,23 @@ against a node that is in real-time sync with age=2s).
      Independence is safe: each check reads PID/start_ticks from `/proc` and
      compares against its own snapshot.
 
+   **Gap-stale baseline (`PREV_SCRAPE_AGE_SECONDS`, #3246):** derived in the
+   evaluator as `gap_stale = age >= 0 and age >= GAP_STALE_THRESHOLD_SECONDS`
+   (default `3600` ≈ 3× the nominal ~20-min tick). This is the **distinct**
+   same-PID-old-baseline case: PID-change (#3206) trips `PREV_PROM_INVALID`
+   FIRST; gap-stale is sequenced AFTER the identity check, so the two are never
+   double-handled. Applied **asymmetrically** by alarm family:
+   - **prev.prom families** (counter, counter-dynamic, histogram-p99): **SKIP**
+     the acute fire this tick with a `gap-stale (prev age Xh)` reason — `prev.prom`
+     is recomputed from the file each tick, so no persisted baseline to reseed.
+   - **snapshot families** (counter-ratio, counter-streak — incl. the headline
+     `recovery-stalled` burst): **RE-BASELINE** — rewrite the snapshot with
+     current values and reset the streak to `0` (reusing the counter-reset path),
+     because these snapshots PERSIST across ticks; a plain skip would leave the
+     stale baseline and fire on the NEXT tick.
+   - **Age-unknown (`-1`/absent)** is treated as NOT stale (fail-safe; a truly
+     missing identity is already covered by `PREV_PROM_INVALID`).
+
 6. **Counter reset handling**: for any counter, if `current < prev`, treat
    `delta = current` (defense-in-depth for within-incarnation counter resets).
 
@@ -988,6 +1035,7 @@ against a node that is in real-time sync with age=2s).
    ARCHIVE_VERSION=1
    TICK_SKIPPED=${TICK_SKIPPED:-false}
    PREV_PROM_INVALID=${PREV_PROM_INVALID:-false}
+   PREV_SCRAPE_AGE_SECONDS=${PREV_SCRAPE_AGE_SECONDS:--1}
    WARMUP_TICKS_REMAINING=${WARMUP_TICKS_REMAINING:-0}
    FRESH_START=${FRESH_START:-no}
    CRASH_RECOVERY=${CRASH_RECOVERY:-no}
@@ -1126,6 +1174,7 @@ catalog for per-alarm gates and the evaluator source for edge-case behavior.
 ```bash
 # Set env vars from checks 3/10 state:
 export PREV_PROM_INVALID=...     # true/false from scrape_identity check
+export PREV_SCRAPE_AGE_SECONDS=... # prev.prom/snapshot age in seconds (gap-stale, #3246); -1/unset = unknown
 export WARMUP_TICKS_REMAINING=... # 0/1/2 from restart detection
 export FRESH_START=...            # yes/no
 export CRASH_RECOVERY=...        # yes/no
--- a/scripts/lib/eval-alarms.py
+++ b/scripts/lib/eval-alarms.py
@@ -7,6 +7,14 @@ Usage:
 
 Inputs (env vars):
     PREV_PROM_INVALID   true/false (default: false)
+    PREV_SCRAPE_AGE_SECONDS  wall-clock age of the prev.prom/snapshot baseline,
+                        in seconds (default/absent/-1: unknown ⇒ not stale).
+                        When >= GAP_STALE_THRESHOLD_SECONDS the baseline is
+                        "gap-stale" (see #3246): the monitor LOOP stalled across
+                        a long wall-clock gap while the validator PROCESS
+                        survived, so the same-PID baseline is hours old and the
+                        counter-delta would false-fire as an acute burst.
+    GAP_STALE_THRESHOLD_SECONDS  gap-stale threshold (default: 3600 ≈ 3× tick).
     WARMUP_TICKS_REMAINING  0/1/2 (default: 0)
     FRESH_START         yes/no (default: no)
     CRASH_RECOVERY      yes/no (default: no)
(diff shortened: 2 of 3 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Gap-stale baseline-age guard is correctly derived, fail-safe on unknown age, sequenced strictly after the PID-identity check (no #3206 double-handling), and applied asymmetrically (prev.prom→SKIP, snapshot→RE-BASELINE) exactly as specified. Test-first regression suite genuinely catches the bug; CI green.

<details>

Linked issue #3246 is `kind: bug-fix`. The PR body has a `## Regression test` section; verified against the diff and by running the suite.

**Regression test verification (`test-coverage`):**
- Commit ordering is correct: tests committed first (`5d076e5`, "Regression tests… fail on current main"), fix second (`b960096`).
- Checked out the test-only commit `5d076e5` (eval-alarms.py has zero `gap_stale` references there) and ran the suite: 148f gets `state=firing reason=None` (the false acute-fire), 148h and 148i FAIL, and the SKILL-doc test FAILS. So the tests genuinely capture the bug.
- At PR head `b960096`: 148f (counter skip, reason `gap-stale (prev age 12.0h)`), 148g (normal 20-min interval STILL fires — the over-skip guard), 148h (recovery-stalled re-baseline → `collecting_baseline`, snapshot reseeded to 703 / streak 0), 148i (counter-ratio re-baseline → `collecting_baseline`, streak reset 0), plus the two SKILL-doc assertions all PASS.
- Full suite: 337 ok, 4 not-ok. The 4 are pv-label-sync (renumbered 152–155 on this branch); confirmed they ALSO fail on clean `origin/main`. Pre-existing and unrelated — not a bounce.

**The fix (`correctness`):**
- `gap_stale = prev_scrape_age >= 0 and prev_scrape_age >= gap_stale_threshold` (default 3600). The `>= 0` clause is what makes `-1`/absent fail-safe to NOT-stale. Both env parses are wrapped in `try/except ValueError → -1`, so a malformed `PREV_SCRAPE_AGE_SECONDS` also degrades to not-stale rather than crashing.
- prev.prom families (`eval_counter`, `eval_counter_dynamic`, `eval_histogram_p99`): SKIP with `gap-stale (prev age Xh)` reason. In `eval_counter_dynamic` the gap-stale guard precedes the prior-delta snapshot write, so the gap-spanning delta is never stored as next tick's `prior_delta` — confirmed by reading the function body. Skip suffices because prev.prom is recomputed from the file each tick.
- snapshot families (`eval_counter_ratio`, `eval_counter_streak`): RE-BASELINE — rewrite snapshot to current values, reset streak to "0", return `collecting_baseline`. Correct because these snapshots persist across ticks; a plain skip would leave the stale baseline and fire next tick.

**Sequencing / no double-handling (`correctness`):**
- `eval_counter_streak`: the gap-stale block is placed AFTER the PID/start_ticks mismatch reset (which writes a fresh snapshot + optionally `maybe_post_restart_fire`) and AFTER the counter-reset branch. Same-PID is therefore guaranteed when the gap-stale block runs. Crucially, the gap-stale path does NOT call `maybe_post_restart_fire` (only the PID-change and counter-reset branches do) — gap-stale ≠ restart, so the absolute value must not acute-fire. Confirmed by reading lines around the guard.
- `eval_counter_ratio`: PID/start_ticks mismatch sets `snapshot = {}`, which routes through the `not snapshot or prev_num_key not in snapshot` collecting-baseline branch and returns BEFORE the gap-stale guard is reached. So a PID change is handled by the identity path, never double-handled by gap-stale.
- `main()` wiring threads `gap_stale` / `gap_stale_age_hours` into all five evaluators with correct positional args.

**Self-modifying extra scrutiny (`self-mod`):** SKILL.md frontmatter (`name`/`description`) parses; all added bash fences in the doc are balanced. `python3 ast.parse` on eval-alarms.py OK. `bash -n` and `zsh -n` clean on the test script (cross-shell). No dispatch-table/step-numbering breakage (monitor-tick is not an orchestrator-dispatch skill).

**Version check (`ci`):** No `*.toml` touched, so no `baseline_version` bump needed. `pytest scripts/ci/test_check_alarm_versions.py` → 23 passed.

**Error handling / readability:** Clean. `gap_stale_reason()` helper centralizes the skip-reason string. Comments are accurate and explain the asymmetry and the no-restart-fire decision at each site.

No blocking concerns.

</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Low operational risk. The change is purely additive and env-driven, defaults to no-op when the new env var is absent (`-1` ⇒ not stale), introduces no data-format break, and the over-skip guard (148g) proves it doesn't suppress legitimate acute bursts at normal tick intervals.

<details>

Non-parity PR (touches only `.claude/skills/monitor-tick/`, `scripts/lib/eval-alarms.py`, `scripts/test-monitor-skill-snippets.sh`) → risk lens.

**`regression-risk` — over-skipping / false-negatives:**
- The single largest risk for an alarm-suppression change is silencing a real burst. This is guarded by design and by test: `gap_stale` only triggers when `prev_scrape_age >= 3600` (≈3× the nominal ~20-min tick). Test 148g exercises a 1200s (20-min) prev age with a real delta and asserts the alarm STILL fires — verified passing both pre- and post-fix. So normal-cadence acute bursts are unaffected.
- Fail-safe on unknown age: `prev_scrape_age >= 0` short-circuits `-1`/absent/malformed to not-stale, and the truly-missing-identity case is already covered by `PREV_PROM_INVALID`. There is no false-skip of a legitimate burst when the timestamp is unreadable.
- The threshold is operationally sane: a 1h-old baseline genuinely means the monitor loop stalled (the loop runs every ~20 min), so a counter-delta across that gap is not a real-time burst and should not page.

**`migration-risk` / `config-risk` — rollout:**
- Purely env-driven. If `PREV_SCRAPE_AGE_SECONDS` is unset (old monitor sessions mid-rollout), the evaluator defaults to `-1` ⇒ not stale ⇒ identical to pre-PR behavior. No flag-day. `GAP_STALE_THRESHOLD_SECONDS` defaults to 3600 and is overridable for ops tuning/tests.
- No `*.toml` change ⇒ no `baseline_version` bump ⇒ `check-alarm-versions.py` stays green. Confirmed `pytest scripts/ci/test_check_alarm_versions.py` → 23 passed. No archive/snapshot schema break: the re-baseline writes the same snapshot fields (`version`/`pid`/`start_ticks`/`counter_value`/`breach_streak`, and the ratio `_numerator`/`_denominator`/`_streak` keys) that the existing reset paths already write.

**`api-break` — function signatures:**
- All new params (`gap_stale`, `gap_stale_age_hours`) are added with defaults, so any other caller of these evaluators is unaffected. The only caller is `main()`, updated consistently.

**`regression-risk` — interaction with #3206 (restart handling):**
- Distinctness verified: gap-stale is sequenced strictly after the PID-identity / snapshot-PID-mismatch reset in every evaluator. A PID change routes through #3206's path and returns before gap-stale is evaluated, so there is no conflicting double-reset or double-fire. The #3206 `maybe_post_restart_fire` absolute-fire logic is undisturbed: it is invoked only on the PID-change and counter-reset branches, never on the gap-stale re-baseline path (gap-stale counts accrued during normal operation, not a startup burst, so suppressing the absolute fire is the correct operational call).

**`self-mod` — pipeline safety (extra scrutiny):**
- SKILL.md frontmatter parses; markdown fences balanced; eval-alarms.py AST-parses; `bash -n` + `zsh -n` clean. The monitor-tick skill is not on the orchestrator dispatch path, so even a doc regression here cannot break `/project-tick` routing. No step-numbering/dispatch breakage.

**Operational note (non-blocking):** the gap-stale re-baseline yields one tick of `collecting_baseline` after a long stall, meaning the very first post-stall tick won't alarm even on a genuinely new burst. This is the correct tradeoff (you cannot distinguish gap-accrued from post-stall-accrued counts on the reset tick), matches the existing restart-reset behavior, and self-heals on the next tick. Not a blocker.

No blocking risk findings.

</details>

tomerweller:
## ✅ Merged

**Commit:** 4547d0f3095592295616cc40d3d1e2d87be1a9f4

**Follow-up issues filed for unaddressed inline review comments:** none (no inline threads on this PR)

Triple-green: Reviewer A (Correctness) APPROVE, Reviewer B (Risk) APPROVE, CI green (39/39). Self-modifying extra scrutiny passed (SKILL frontmatter parses, eval-alarms.py AST-clean, bash -n + zsh -n clean). Regression tests verified test-first (fail at the test-only commit, pass post-fix).

Issue #3258, opened 2026-06-10
monitor-tick: auto-stamp resolved:<fix-sha> on bundled-commit quarantines (full autonomy)


tomerweller:
Follow-up from #3256 (resolved_sha quarantine field, Option 4).

#3256's plan adds the `resolved:<sha>` field + `quarantine_resolve` helper + the auto-clear gate (the atomic, low-risk, autonomously-doable piece). It leaves the STAMPING manual: the operator runs `quarantine_resolve` once per bundled-commit quarantine. That reduces the manual step from "every blocked tick" (status quo `quarantine_remove`) to "once per quarantine", but does not FULLY restore the autonomous-deploy posture (#3215).

## Scope of this follow-up

Make monitor-tick automatically stamp `resolved:<fix-sha>` when it can reliably determine that a fix for a quarantined SHA has landed on `origin/main`. This is the harder half: identifying "the fix SHA" for a crashed bundled commit (same difficulty class as the rejected option-3 commit-message-parsing). Needs its own design pass.

## Why a follow-up, not part of #3256

Raised by Critic C (scope) on #3256 round 1 as a REVISE-MINOR carve-out. The helper+gate solve the filed bug (auto-clear after partial revert) and are independently shippable; automatic stamping is a separate, larger autonomy-design concern. Preserving it here so it isn't silently dropped.

## Operator decision embedded here

If the once-per-quarantine manual stamp from #3256 is acceptable, this can stay backlog/won't-fix. If full autonomy is required, this needs a planning pass (how does monitor-tick reliably identify the fix SHA?).

tomerweller:
## Triage Report

**Verdict:** ACCEPT

**Type:** enhancement
**Kind:** feature
**Severity:** medium
**Crate:** crate:monitor (monitor-tick skill)
**Size:** M

**Summary:** Automatically stamp `resolved:<fix-sha>` when monitor-tick detects that a fix for a quarantined bundled commit has landed on origin/main, fully restoring autonomous deployment. This is the harder follow-up to #3256 — requires reliable heuristics to identify which commit fixes a crash.

**Test obligation:** New integration tests verifying monitor-tick's auto-detection logic (fixture: quarantined commit, fix commit, and monitoring workflow).

**Design note:** The core complexity is heuristic-driven identification of the fix SHA. Acceptable strategies include:
- Bisect/search for first ancestor of main that doesn't trigger the crash
[...]

Issue #3248, opened 2026-06-09
monitor-tick: deploy-quarantine content-check false-clears when later commits drift the bad commit's unrelated hunks


tomerweller:
## Symptom

The content-aware deploy-quarantine gate (`check_quarantine_active`, `scripts/lib/deploy-quarantine.sh`) **false-cleared** a still-active regression, which would have let the autonomous deploy tick rebuild and **re-crash the production validator**.

## Evidence (2026-06-09, regression #3247)

`18963635` (#3238) was quarantined after it crashed the validator (live `retain:false` → jemalloc munmap VMA exhaustion). The harmful content — `retain:false` at `crates/henyey/src/main.rs:117` and the `_rjem_malloc_conf` export at `:115` — remains **unchanged in origin/main HEAD** (`686600ce`). Yet the gate reported `clear`:

```
check_quarantine_active → rc=1 STATUS=clear
git diff 18963635^..18963635 | git apply --check --reverse
  → error: patch failed: crates/ledger/src/memory_report.rs:358
  → patch does not apply
```

The reverse-apply failed only because a *later* commit (#3244, `1f8f5a49`) edited `crates/ledger/src/memory_report.rs` near lines #3238 also touched — **context drift on an unrelated hunk** of the same multi-file commit. The gate interprets "diff no longer reverse-applies" as "bad content removed," but here the bad content (a different file in the same commit) is fully intact.

I worked around it by quarantining the current HEAD SHA (`686600ce`) directly, which blocks (`blocked_active`) — but that's fragile: it re-breaks as soon as a new commit moves HEAD, or if a later commit drifts `686600ce`'s own hunks.

## Suspected root cause

`check_quarantine_active` decides "still applied?" via a whole-commit `git diff <sha>^..<sha> | git apply --check --reverse`. For a multi-file commit, ANY one hunk failing context (from unrelated later edits) makes the entire reverse-apply fail → the gate clears even though most/all of the regression content is still present. The granularity is wrong: it's all-or-nothing per commit, but regressions are often a single hunk/file.

## Concrete fix sketch

- Make the reverse-apply **per-file or per-hunk**: split `git diff <sha>^..<sha>` by file (or use `git apply --check --reverse` with `--3way`/`--recount`, or test each file's diff independently). Block if ANY file's portion still reverse-applies (i.e. is still present), instead of requiring the whole commit to apply atomically. `scripts/lib/deploy-quarantine.sh` `check_quarantine_active`.
[...]

PR #3260, opened 2026-06-10
Auto-clear bundled-commit quarantine via resolved:<fix-sha> token

### LINKED ISSUE #3256: monitor-tick: quarantine gate cannot auto-clear bundled commit after partial revert (false-block)
## Symptom

The deploy-quarantine content-check (`check_quarantine_active`, per-hunk since #3253) **blocks indefinitely** when a quarantined commit bundled the harmful change together with desirable changes, and the operator reverted only the harmful part. The gate cannot auto-clear because the desirable hunks are intentionally still applied to `origin/main`, so at least one hunk reverse-applies cleanly → `blocked_active`.

Concrete instance this tick (tick 199, 2026-06-10):

- Quarantined SHA `189636357889a382ec64fb0e2ea6567433595c98` = #3238, titled *"Fix jemalloc malloc_conf no-op (wrong symbol) and add allocator size-class instrumentation"*. It bundled THREE changes:
  1. `_rjem_malloc_conf` symbol-name fix (desirable — kept)
  2. `retain:false` config token (harmful — caused VMA-exhaustion crash, quarantined)
  3. arena size-class instrumentation in `memory_report.rs` (desirable — kept)
- The harmful part was reverted by #3247/#3251 (only `retain:false` removed from the config string; symbol fix + instrumentation intentionally retained).
- Verified harmful content is gone: `origin/main:crates/henyey/src/main.rs` config string is `b"background_thread:true,dirty_decay_ms:1000,muzzy_decay_ms:1000\0"` (no `retain:false`), and runtime `/metrics` shows `opt_retain=true`.
- Yet `check_quarantine_active` returned `blocked_active QUARANTINED_MATCH=18963635` because the symbol-fix and instrumentation hunks still reverse-apply.

This blocked the legitimate deploy of #3255 (an unrelated BucketListDB change) until I manually `quarantine_remove`d 18963635.

## Evidence

```
$ git diff 18963635^..18963635 | git apply --check --reverse
error: patch failed: crates/ledger/src/memory_report.rs:358
error: patch failed: crates/henyey/tests/jemalloc_config.rs:1
error: patch failed: crates/henyey/src/main.rs:90
# whole-commit reverse-apply FAILS, but the per-hunk check finds SOME hunks
# (symbol fix, instrumentation) that DO still reverse-apply → blocks
```

## Suspected root cause

`scripts/lib/deploy-quarantine.sh` quarantines at whole-SHA granularity and the per-hunk content check (#3253) blocks if ANY hunk of the SHA still applies. For a commit that bundled good+bad changes, a partial revert (remove only the bad) can never auto-clear: the good hunks legitimately remain. The per-hunk check fixed the false-*clear* on unrelated drift (#3253) but introduced/retained a false-*block* for bundled commits with partial reverts.

## Concrete fix sketch (design choice — needs operator input)

Options:
1. **Document-only**: accept that bundled-commit quarantines require manual `quarantine_remove` after partial revert; the skill already documents the manual escape hatch. Lowest effort. (This tick used it.)
2. **Quarantine by content signature, not whole SHA**: when appending, record the specific harmful hunk(s) / a content fingerprint (e.g. the `retain:false` token) rather than the SHA, so the gate checks only the harmful change for residual application.
3. **"All-hunks" vs "any-hunk" semantics**: only block when ALL of the quarantined SHA's hunks still reverse-apply (whole change still present), clear when any are gone. Weaker — could false-clear a still-present harmful hunk if an unrelated hunk was touched.

Option 2 is the most correct but requires changing the quarantine file format and the append path. Option 1 is what the docs already cover. Leaning option 1 (document the bundled-commit caveat near the manual-removal section) unless the operator wants the stronger content-signature approach.

Files: `scripts/lib/deploy-quarantine.sh`, `.claude/skills/monitor-tick/SKILL.md` (Quarantine Clearance section).

## Not blocking

Worked around this tick via `quarantine_remove`. Node stayed on 3ac4572d (deploy of #3255 was deferred anyway pending CI-green). No label.

tomerweller:
Closes #3256

## Summary

The deploy-quarantine per-hunk content-check (#3253) blocks indefinitely when a quarantined commit **bundled** a harmful change with desirable ones and only the harmful part was reverted: the benign hunks legitimately still reverse-apply, so the entry stays `blocked_active` forever (the tick-199 false-block — #3238 mixed a harmful `retain:false` removed by #3251 with the benign `_rjem_malloc_conf` export + instrumentation, intentionally kept on main).

This adds **Option 4 (resolved_sha)**: an additive, token-boundary-anchored `resolved:<fix-sha>` annotation on a quarantine entry. In `check_quarantine_active`, an entry whose `resolved` fix-SHA is an **ancestor of origin/main** (genuinely merged) **auto-clears** — restoring autonomous deploy after a partial revert.

**The per-hunk content-check (#3253) is preserved byte-for-byte for every un-annotated entry.** The resolved path can only CLEAR, never weaken the backstop. Every uncertain case **falls back to BLOCK** (fail-safe toward the existing posture):
- missing / malformed `resolved` token → per-hunk content-check governs
- self-resolution (`resolved` == the quarantined SHA itself) → rejected → BLOCK
- fix not yet an ancestor of origin/main → BLOCK until it merges
- git error on the resolved-ancestry check (rc>=128) → fail-closed → BLOCK

The only auto-clear path requires an explicit, valid, genuinely-merged fix SHA — a strictly stronger signal than "all hunks gone". Token parsing uses `grep` (not the `=~` capture array) so it is identical under bash and zsh (zsh populates `$match`, not `$BASH_REMATCH`).

A new `quarantine_resolve FILE SHA RESOLVED_SHA` helper (atomic tmp+mv, idempotent, self-resolution + invalid-SHA rejected, missing-file no-op) stamps the field.

This restores autonomy for the **manual-stamp case** (operator stamps `resolved:` once → auto-clears on merge). Full auto-identification of the fix SHA is the carved-out follow-up **#3258** — not implemented here.

**Operator design fork (per the converged plan):** Option 4 v1 has the operator stamp `resolved:` once per bundled quarantine (vs. status quo: re-clear via `quarantine_remove` every blocked tick). Fully-automatic stamping (monitor-tick identifying the fix SHA itself) is #3258.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3256#issuecomment-4665889405)

## Test plan

- [x] `bash scripts/test-monitor-skill-snippets.sh` — all new tests (78f–78q) pass; only the 4 pre-existing `pv-label-sync` failures remain (present on clean origin/main, unrelated)
- [x] `bash scripts/test-shell-lib-cross-shell.sh` — passes; `quarantine_resolve` now asserted defined under bash + zsh
- [x] `bash -n` + `zsh -n` clean on `deploy-quarantine.sh` + both harnesses
- [x] TAP_PLAN bumped (337 → 349, +12 new tests; pre-existing emitted/plan drift preserved)

- **Test:** `scripts/test-monitor-skill-snippets.sh` — *quarantine-active: bundled commit + resolved-on-main auto-clears (#3256)* (test 78f)
- **Pre-fix:** committed as `485225cb` — verified FAILED on origin/main lib: `not ok ... bundled commit + resolved-on-main auto-clears (#3256)` (no `resolved` parsing → `blocked_active`, the tick-199 bug)
- **Post-fix:** verified PASSES after `6c8de88b`
- **Fail-safe guards (78g–78j):** resolved-not-merged / malformed / self-resolution / git-error → all stay `blocked_active` (pass on both origin/main and the fix — they lock in the fail-safe posture)

## Deviations from plan

- Exposed the parsed token via the index-aligned `QUARANTINE_RESOLVED` global (one line per entry, `-` when absent) and iterate it lockstep with `QUARANTINE_ENTRIES` over two FDs (bash+zsh safe), as the plan specified.
- Resolved-token extraction uses `grep` rather than `[[ =~ ]]` capture groups for true bash/zsh parity (zsh does not populate `$BASH_REMATCH`). Functionally equivalent to the plan's token-boundary-anchored regex.

### DIFF
--- a/.claude/skills/monitor-tick/SKILL.md
+++ b/.claude/skills/monitor-tick/SKILL.md
@@ -1840,6 +1840,29 @@ Entries stay in the file as a historical log; they're harmless once their
 content is gone. The file can grow unboundedly over time if you want a
 deploy audit trail; trim it manually whenever convenient.
 
+**Bundled commits (`resolved:<fix-sha>`)** — the content-check above blocks
+if **any** hunk of the quarantined SHA still reverse-applies. A commit that
+**bundled** a harmful change with desirable ones (e.g. #3238 mixed a harmful
+`retain:false` with the benign `_rjem_malloc_conf` export + instrumentation)
+can never auto-clear after a **partial** revert: the benign hunks are
+intentionally kept on `origin/main`, so the per-hunk check stays
+`blocked_active` forever (the tick-199 false-block, #3256). For that case,
+stamp the entry once with the fix commit that resolved it:
+
+```bash
+source "$(git rev-parse --show-toplevel)/scripts/lib/deploy-quarantine.sh"
+# <fix-sha> is the commit that removed the harmful part (e.g. #3251).
+quarantine_resolve "$HOME/data/deploy_quarantine.txt" "$bad_sha" "$fix_sha"
+```
+
+This appends a `resolved:<fix-sha>` token to the entry. The gate **auto-clears**
+that entry once `<fix-sha>` is an ancestor of `origin/main` (i.e. the fix has
+merged) — no further per-tick action needed. It is **fail-safe**: a not-yet-
+merged fix, a malformed token, a self-resolution (`<fix-sha>` == `<bad_sha>`),
+or a git error all fall back to the per-hunk content-check (BLOCK). Stamping is
+a once-per-quarantine manual step today; automatic fix-SHA stamping is the
+follow-up #3258. `quarantine_remove` remains the override.
+
 **Manual removal** (still supported, occasionally useful — e.g. an entry
 that the content check can't decide cleanly because of a malformed diff
 or a binary file):
--- a/scripts/lib/deploy-quarantine.sh
+++ b/scripts/lib/deploy-quarantine.sh
@@ -12,6 +12,11 @@
 #   - Line-oriented, default-whitespace-separated fields
 #   - First field: 40 lowercase hex chars (commit SHA)
 #   - Remaining fields: optional free-text reason (single line)
+#   - An optional `resolved:<40-hex>` token MAY appear anywhere in the reason
+#     (token-boundary-anchored). It records the fix commit that resolves the
+#     quarantine: once that fix SHA is an ancestor of origin/main, the entry
+#     auto-clears (see check_quarantine_active). Additive — absent on legacy
+#     entries, which keep the per-hunk content-check as their backstop.
 #   - Lines starting with # (after optional whitespace) are comments
 #   - Blank/whitespace-only lines are skipped
 #   - CRLF is stripped during parsing
@@ -36,6 +41,11 @@ _DEPLOY_QUARANTINE_LOADED=1
 #
 # Sets globals:
 #   QUARANTINE_ENTRIES  - Newline-separated valid SHAs (order preserved)
+#   QUARANTINE_RESOLVED - Newline-separated resolved-fix SHAs, ONE PER ENTRY in
+#                         the SAME order as QUARANTINE_ENTRIES. A line is the
+#                         40-hex fix SHA when the entry carried a valid,
+#                         token-boundary-anchored `resolved:<sha>`, or "-" when
+#                         it did not. The two globals are index-aligned.
 #   QUARANTINE_WARNINGS - Comma-space-separated warning messages
 #
 # Returns:
@@ -45,6 +55,7 @@ _DEPLOY_QUARANTINE_LOADED=1
 parse_quarantine_file() {
   local file="$1"
   QUARANTINE_ENTRIES=""
+  QUARANTINE_RESOLVED=""
   QUARANTINE_WARNINGS=""
 
   # Missing or empty file: clear, no warnings
@@ -80,10 +91,29 @@ parse_quarantine_file() {
 
     # Validate: exactly 40 lowercase hex chars
     if [[ "$sha" =~ ^[0-9a-f]{40}$ ]]; then
+      # Extract an optional `resolved:<40-hex>` token from the reason field.
+      # Anchored on token boundaries (start-or-whitespace before, end-or-
+      # whitespace after) so a substring like `unresolved:...` or a >40-hex
+      # blob never false-matches. Default "-" (no resolved annotation).
+      # Done via grep (not the `=~` capture array) so the extraction is
+      # identical under bash and zsh — zsh populates `$match`, not
+      # `$BASH_REMATCH`, for `[[ =~ ]]` capture groups (#3256).
+      local resolved="-" _resolved_tok
+      # `|| true`: grep exits 1 on no-match; the `|| true` keeps this from
+      # tripping a caller's `set -e`/`pipefail` (the no-token case is normal).
+      _resolved_tok="$(printf '%s' " $_rest " | grep -oE '[[:space:]]resolved:[0-9a-f]{40}[[:space:]]' | head -n1 || true)"
+      if [[ -n "$_resolved_tok" ]]; then
+        # Strip the leading-space + `resolved:` prefix and the trailing space.
+        _resolved_tok="${_resolved_tok#"${_resolved_tok%%[![:space:]]*}"}"
+        _resolved_tok="${_resolved_tok%"${_resolved_tok##*[![:space:]]}"}"
+        resolved="${_resolved_tok#resolved:}"
+      fi
       if [[ -n "$QUARANTINE_ENTRIES" ]]; then
         QUARANTINE_ENTRIES+=$'\n'"$sha"
+        QUARANTINE_RESOLVED+=$'\n'"$resolved"
       else
         QUARANTINE_ENTRIES="$sha"
+        QUARANTINE_RESOLVED="$resolved"
       fi
     else
       local warning="malformed: ${sha:0:12}"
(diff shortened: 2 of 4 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The `resolved:<fix-sha>` auto-clear is correctly gated — it fires only when the token is a valid 40-hex SHA, is not the quarantined SHA itself, and `git merge-base --is-ancestor <fix> origin/main` returns exactly rc 0. Every uncertain/malformed/error/non-merged/self case falls through to the byte-for-byte-preserved #3253 per-hunk BLOCK. Regression test 78f fails pre-fix (tick-199 bug) and passes post-fix; 12 new tests all green.

**Self-modifying paths touched (extra reviewer scrutiny applied):**
- `.claude/skills/monitor-tick/SKILL.md`
- `scripts/lib/deploy-quarantine.sh`
- `scripts/test-monitor-skill-snippets.sh`
- `scripts/test-shell-lib-cross-shell.sh`

<details>

**kind: bug-fix — regression test verification (PASS).** Commit order is correct: tests (`485225cb`) precede fix (`6c8de88b`). I reconstructed the pre-fix state (test file at 485225cb + `deploy-quarantine.sh` reverted to parent `4547d0f3`, which has zero `resolved` support) and ran the regression suite:
- 78f (`bundled commit + resolved-on-main auto-clears`) → **`not ok` pre-fix** (no `resolved` parsing → `blocked_active`, the tick-199 false-block), **`ok` post-fix**. The test genuinely captures the bug.
- 78g–78j (the 4 fail-safe guards) → `ok` on BOTH pre-fix and post-fix, locking in the conservative posture.

**Clear-path correctness (`check_quarantine_active` lines 376–385).** The only `continue` that clears an annotated entry (line 381) is reachable iff `[[ "$resolved" =~ ^[0-9a-f]{40}$ && "$resolved" != "$sha" ]]` AND `git merge-base --is-ancestor "$resolved" origin/main` returns rc 0. I exercised the branch in isolation with a mocked git across 9 attack inputs; **only** (valid 40-hex fix SHA that is a genuine ancestor) clears — all others stay `blocked_active`:
- not-merged (rc1) → BLOCK; git-error (rc≥128) → BLOCK (the `|| resolved_rc=$?` captures any non-zero, and only `-eq 0` clears — 128 is not 0);
- self-resolution (`resolved==sha`) → guard at line 376 short-circuits → BLOCK;
- malformed (short hex / non-hex / 41-hex) → regex fails → BLOCK;
- `unresolved:` boundary substring → never parsed (grep requires leading `[[:space:]]` before `resolved:`) → token is `-` → BLOCK;
- absent token → `-` → BLOCK.

**Per-hunk #3253 backstop preserved byte-for-byte.** Step 3 (former Step 2) is unchanged logic; tests 106–112 (#3248/#3253 content-check incl. partial-presence + zsh recheck) all pass unchanged. The resolved short-circuit only *prepends* a clear path; it never weakens the content-check.

**Parser correctness.** `QUARANTINE_RESOLVED` is index-aligned with `QUARANTINE_ENTRIES`: verified 4-entry lockstep (token-at-EOL, no-token, mid-reason token, `unresolved:`-only → `bbb / - / ddd / -`). Token extraction uses `grep -oE '[[:space:]]resolved:[0-9a-f]{40}[[:space:]]'` on a space-padded `_rest`, so EOL tokens match and `unresolved:` cannot. Iteration uses `read <&3 && read <&4` over two here-strings — bash/zsh-safe, no off-by-one.

**`quarantine_resolve` helper.** Validates both SHAs (rc1), rejects self-resolution (rc1), missing-file no-op (rc0), unreadable→rc2, atomic tmp+mv (rc2 + tmp cleanup on awk/mv failure). Verified idempotent + canonicalizing re-stamp (old token removed, single new token, no tmp residue), stamps every matching entry, leaves non-matching untouched.

**Self-modifying scrutiny (PASS).** `bash -n` + `zsh -n` clean on the lib and harness; sources cleanly under both shells; SKILL.md frontmatter/markdown well-formed; no dispatch-table/step-numbering breakage (the step-renumber 2→3, 3→4 in the doc comment is internally consistent with the code).

**CI/test count.** TAP_PLAN 337→349 (+12). The 4 `not ok` (`pv-label-sync`) failures pre-exist on clean `origin/main` — unrelated to this diff.

No blocking concerns.
</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** This ADDS a quarantine-clear path, which is the high-risk direction (an under-block could let a crashing commit re-deploy). I traced every branch: the clear is unreachable except via a valid, non-self, genuinely-merged (ancestor-of-origin/main) operator-stamped fix SHA. Every malformed/absent/non-merged/self/git-error input fails closed to the existing #3253 per-hunk BLOCK. No regression to the existing backstop, no new operational footgun.

<details>

**regression-risk — clear-path under-block (the core risk): MITIGATED.** The danger of this PR vs #3253 (which only tightened) is that it introduces a way to CLEAR a quarantine, so a defect could under-block and re-deploy a crash. I enumerated the entire clear-path decision and confirmed the fail-safe is total. The clear (`continue` at line 381) requires the conjunction of: (1) valid 40-hex token, (2) `!= $sha` (self-resolution rejected), (3) `merge-base --is-ancestor` rc == 0. Isolated-branch testing across 9 hostile inputs (not-merged rc1, git-error rc128, self, short-hex, non-hex, 41-hex, `unresolved:` substring, no-token) → all stay `blocked_active`; only the genuine ancestor clears. The trust model is sound: the operator vouches for a real fix SHA by stamping it, and the ancestor-gate independently confirms it actually merged — a strictly stronger signal than "all hunks gone."

**regression-risk — backstop integrity: NONE.** The #3253/#3248 per-hunk content-check is the fallthrough for every un-annotated entry and every uncertain annotated entry. It is logically unchanged (only renumbered in comments). Tests 106–112 pass unchanged. The resolved path can only short-circuit to CLEAR on the genuine-merge case; it never relaxes or bypasses the content-check for any other input.

**api-shape / migration-risk: LOW.** `resolved:` is purely additive and token-boundary-anchored; legacy entries with no token parse to `-` and behave exactly as before (verified). The quarantine file format is backward-compatible — a free-text reason cannot accidentally trip the token (requires literal whitespace-bounded `resolved:<40hex>`). No data migration needed.

**config-risk / operational: LOW.** The operator workflow is documented in monitor-tick SKILL.md: stamp once via `quarantine_resolve`, auto-clears on merge. The helper is atomic (tmp+mv), idempotent, and self-resolution/invalid-SHA rejected, so an operator cannot accidentally create a self-clearing entry. `quarantine_remove` remains as the manual override. Full automatic stamping is correctly carved out to follow-up #3258, not implemented here.

**perf-regression: NONE.** Adds at most one extra `git merge-base --is-ancestor` per annotated entry; negligible.

**security: NONE.** No new external input surface; the token is operator-controlled, and the ancestry gate prevents a stamped-but-unmerged SHA from clearing. `2>/dev/null` on the ancestry check correctly suppresses noise while the `|| resolved_rc=$?` preserves the rc for the fail-closed branch.

**Cross-shell:** `test-shell-lib-cross-shell.sh` 25/25 green; `quarantine_resolve` now asserted defined under both bash and zsh. grep-based extraction (not `$BASH_REMATCH`) gives true parity.

No blocking concerns.
</details>

tomerweller:
## ✅ Merged

**Commit:** 93ab5a0207723e6513a16289523998242a6417c1

**Follow-up issues filed for unaddressed inline review comments:** none (no inline threads; all reviewer findings APPROVE)

PR #3265, opened 2026-06-10
Add peer_gap floor to No-SCP hard-reset escalation trigger

tomerweller:
Closes #3263

## Summary

Adds a `peer_gap >= PEER_AHEAD_ESCALATION_THRESHOLD` (3) floor to the No-SCP hard-reset escalation condition in `out_of_sync_recovery` (`crates/app/src/app/consensus.rs`). This was the **only escalation site lacking the peer-gap floor** — the sibling peer-ahead far-behind site (consensus.rs:1876) already has it. At the live tip the archive is normally behind, so without a verified peer ahead the No-SCP *decision* fired at tip every ~180s (counter/log/cooldown churn) before #3262/#3264's downstream suppression neutralized it. The floor stops the wasted escalation decision at the source and complements #3262/#3264's execution-layer suppression (defense-in-depth).

The change is purely additive — `peer_gap` is already in scope (consensus.rs:803); AND-ing a stronger predicate can only *reduce* firing.

## The arc preserved

- **#1862** far-behind (`peer_gap >= 12`) still escalates (`12 >= 3`) — unchanged.
- **#2789** anti-wedge: a genuinely-stuck node at `peer_gap < 3` is STILL caught by the **independent** 120s wall-clock backstop (`maybe_start_buffered_catchup` → `decide_consensus_stuck_action`, `catchup_impl.rs`), which fires `HardReset(ArchiveBehindStallWallClock)` on `stuck_duration >= 120s` with NO `peer_gap` term. That path is untouched, so the floor cannot create a new wedge.
- **#3199/#3205** near-tip peer-SCP recovery — unaffected.
- **#3262/#3264** execution-layer suppression (`catchup_impl.rs`) — untouched outer defense.

## Parity

stellar-core `HerderImpl::outOfSyncRecovery` (HerderImpl.cpp:521-561) is SCP-only — purge old slots, re-broadcast latest SCP, `getMoreSCPState()`; no archive catchup, no escalation-at-tip absent genuine network-behindness. The `peer_gap >= 3` floor is the henyey proxy for "genuinely behind the network." Moves henyey toward core.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3263#issuecomment-4667498236)

## Test plan

- [x] cargo fmt --check
- [x] cargo clippy -p henyey-app -- -D warnings (clean)
[...]

PR #3273, opened 2026-06-10
Soft-skip testnet quickstart probe timeouts to de-gate main

### LINKED ISSUE #3272: CI: Quickstart testnet shard cancelled on ≥2 consecutive main commits — probe wrapper hangs at step 8 (systematic, not flaky)
## Summary

The Quickstart `test (testnet, core,horizon, ...)` shard is **cancelled on ≥2 consecutive `main` commits**, so per the `/review-pr` flaky-vs-regression rule this is a **systematic CI regression**, not a flaky one-off. It is currently blocking the CI-green gate for every open PR (it surfaced while reviewing PR #3269).

## Evidence (Quickstart workflow, `test (testnet, ...)` job)

| main commit | run | testnet job | duration |
|---|---|---|---|
| `63f3387` | 27289116703 | **cancelled** | started 16:20:20Z, killed 17:15:21Z (~55 min) |
| `7a3795c` | 27263669551 | **cancelled** | started 08:45:49Z, killed 09:40:50Z (~55 min) |
| `40fb4fb` | 27258900361 | success | ~6 min |

Same on PR #3269 (head `a44ff9f`): testnet shard started 17:19:01Z, cancelled 18:14:01Z (~55 min).

## Root-cause signal

The cancelled testnet jobs all hang at **step 8 "Run probes through wrapper"** — it starts and never completes; every later step (Upload diagnostics / Cleanup / Post-*) is null, and the job is killed ~55 min later by the workflow wall-clock. On the last *successful* run (`40fb4fb`) that same step completed in seconds. So the shard is hanging inside the probe wrapper while talking to **live testnet**, then getting cancelled by timeout — the classic "testnet degraded / slow-sync → probe wrapper blocks → run cancelled" pattern, now reproducing for ~2 commits straight rather than intermittently.

This is **not attributable to any PR diff** (it reproduces on plain `main`); it is environmental (testnet liveness) plus a missing/over-long per-step timeout in the probe wrapper that lets a stuck probe burn the whole 55-min budget instead of failing fast.

## Why this is urgent

- It red-rolls the Quickstart workflow on `main` (run-level conclusion = cancelled), so `/review-pr`'s CI gate sees `red` for every open PR and cannot auto-merge on triple-green.
- It is across ≥2 commits → the flaky-retry path is explicitly forbidden (re-running a single shard would mask a systematic problem). The fix must address the probe wrapper, not a re-dispatch.

## Suggested fix directions

1. Add a tight per-probe / per-step timeout in `scripts/ci/run-quickstart-test.sh` (or the wrapper invoked at step 8) so a stuck testnet probe **fails fast** (and uploads diagnostics) instead of hanging until the job-level cancellation.
2. Make the testnet shard tolerant of (or skip on) testnet degradation — e.g. treat a sync-probe timeout against live testnet as a soft-skip rather than a hang, since it is not a henyey correctness signal.
3. Confirm whether the testnet shard should gate `main` at all given its dependence on external network liveness.

## Provenance

Filed by `/review-pr 3269`. Both agent reviewer lenses APPROVED PR #3269 on the merits; the merge is gated solely by this systematic testnet-shard CI red. Do NOT silently re-run the shard.

tomerweller:
Closes #3272

## Summary

The testnet/core,horizon Quickstart shard depends on external network liveness (slow catchup, checkpoint cadence, archive availability), not henyey correctness. A stuck sync probe TIMEOUT used to red-roll the whole Quickstart workflow and block every PR's merge gate (the chronic flake described in #3272). This adds an opt-in `--soft-on-timeout` flag to `scripts/ci/run-quickstart-test.sh` (default OFF ⇒ byte-identical for every other shard/caller). When ON, a GNU `timeout` TIMEOUT disposition (exit code **124 ONLY**) is converted to a neutral `exit 0` with a grep-able `SOFT-SKIP` marker and preserved diagnostics, at **both** 124 sinks (first-attempt non-retryable timeout and the post-retry second timeout). A genuine assertion failure (any non-124 exit, including 125/126/127/137) always stays **red**, so a real henyey-on-testnet break is never masked. The flag is wired to the testnet shard only via matrix keys (`soft_on_timeout: true`, `probe_timeout: 240`); the probe loop applies the override/flag conditionally so local/pubnet shards keep the default 600s hard-fail-on-timeout and carry no flag. 240s = `timeout_multiplier(4) * 60` (the upstream 4-min precedent); worst-case budget is 4 probes × 240s + one retry ≈ 20 min, well under the 45-min job `timeout-minutes`.

Pure CI-infra change — no Rust, no consensus/app behavior. #3265 and all consensus code are out of scope (operator-exonerated).

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3272#issuecomment-4673292754)

## Test plan

- [x] `bash scripts/test-quickstart-harness.sh` — 83/83 pass (was 72/72; +11 new assertions)
- [x] shellcheck on both scripts — no findings in new code (pre-existing info/warnings only)
- [x] `python3 -c yaml.safe_load` — quickstart.yml is valid YAML
- [x] direct wrapper sanity: 124+flag→exit 0 with marker; assertion-1+flag→exit 1 no marker; 137+flag→red; 124 no-flag→red

- **Tests:** `scripts/test-quickstart-harness.sh` — `test_soft_on_timeout_testnet_timeout_is_neutral_skip`, `test_soft_on_timeout_still_fails_real_assertion`, `test_soft_on_timeout_preserves_diagnostics`, `test_soft_on_timeout_off_by_default_preserves_red`, `test_workflow_testnet_shard_uses_soft_timeout_and_tight_budget`
- **Pre-fix:** committed as `712201fd` — verified FAILED (the `--soft-on-timeout` flag did not exist: exit 2 usage error; workflow had no testnet soft wiring)
- **Post-fix:** verified PASSES after `8557b84a` (83/83)

## Deviations from plan

- One scope-guard assertion (`workflow_pubnet_local_shards_have_no_soft_flag`) needed a comment-line filter so the testnet entry's own documentation prose (which mentions `soft_on_timeout: true`) is not miscounted as a local/pubnet occurrence. Fixed in the implementation commit; the assertion still genuinely guards the scope.

### DIFF
--- a/scripts/ci/run-quickstart-test.sh
+++ b/scripts/ci/run-quickstart-test.sh
@@ -7,6 +7,12 @@
 #   run-quickstart-test.sh --network <net> --enable <services> --probe <name> \
 #       --timeout <seconds> (default: 600) --diagnostics-dir <dir> -- <probe_command...>
 #
+# Env:
+#   KILL_GRACE — grace passed to GNU `timeout -k` before it escalates
+#     SIGTERM → SIGKILL (default: 30s). Load-bearing (#3273): the upstream
+#     `go run` probe ignores SIGTERM, so without a kill-after grace `timeout`
+#     blocks in wait() forever and the CI step hangs instead of timing out.
+#
 # Exit codes:
 #   0 — probe passed (possibly after one retry), OR a timeout (exit 124) was
 #       soft-skipped under --soft-on-timeout (see below)
@@ -20,15 +26,21 @@
 #   liveness (checkpoint cadence, archive availability) and is NOT a henyey
 #   correctness signal — so a TIMEOUT there should not gate the merge. When
 #   --soft-on-timeout is passed (the testnet shard only), a *timeout* disposition
-#   (exit code 124 ONLY) is converted to a neutral `exit 0` with a grep-able
+#   is converted to a neutral `exit 0` with a grep-able
 #   `SOFT-SKIP` marker; diagnostics are still captured so the degraded-testnet
 #   event remains observable in uploaded artifacts. This extends the existing
 #   pubnet/testnet-RPC-disabled "don't gate henyey-correctness merges on external
 #   network liveness" precedent, scoped strictly to the TIMEOUT outcome.
 #   When the flag is OFF (every other caller/shard), behavior is byte-identical.
-#   A genuine probe assertion FAILURE (any non-124 exit) ALWAYS stays red, even
-#   under the flag, so a real henyey-on-testnet break is never masked. Only 124
-#   is soft-skipped; 125/126/127/137 (and any other non-124) stay red.
+#
+#   What counts as a TIMEOUT disposition (#3273): exit 124 OR exit 137. GNU
+#   `timeout` returns 124 when its SIGTERM kills the probe, but 137 (128+9,
+#   SIGKILL) when the probe IGNORES SIGTERM and the `-k` grace escalates to
+#   SIGKILL — which is exactly the upstream `go run` probe's signature (it does
+#   not forward SIGTERM). Both are a deadline hit, so both are soft-skipped
+#   under the flag. A genuine probe assertion FAILURE (any non-124, non-137
+#   exit — `go` test failures exit 1) ALWAYS stays red, even under the flag, so
+#   a real henyey-on-testnet break is never masked.
 #
 # Timeout budget: the caller (.github/workflows/quickstart.yml) is responsible
 # for supplying the per-probe budget via --timeout <seconds>. That budget
@@ -76,6 +88,11 @@ NETWORK=""
 ENABLE=""
 PROBE=""
 TIMEOUT=600
+# Grace before GNU `timeout` escalates SIGTERM → SIGKILL (the `-k` argument).
+# Load-bearing (#3273): the upstream `go run` probe ignores SIGTERM, so without
+# a kill-after grace `timeout` blocks in wait() forever and the CI step hangs.
+# Overridable via the KILL_GRACE env var (e.g. the harness passes a short grace).
+KILL_GRACE="${KILL_GRACE:-30s}"
 DIAGNOSTICS_DIR=""
 # Opt-in (#3272): treat a TIMEOUT (exit 124) disposition as a neutral soft-skip
 # instead of a red failure. Default OFF ⇒ behavior byte-identical for all other
@@ -131,14 +148,30 @@ is_retryable_exit() {
 
 # --- Helper: soft-skip a TIMEOUT disposition under --soft-on-timeout (#3272) ---
 # Returns 0 (and emits the grep-able SOFT-SKIP marker) iff the flag is on AND the
-# exit code is EXACTLY 124 (a GNU `timeout` TIMEOUT). Strictly 124 only:
-# 125/126/127/137 and any other non-124 exit are NOT soft-skipped — a genuine
-# probe assertion failure must stay red so a real henyey-on-testnet break is
-# never masked. Callers use this at each of the two 124 sinks (first-attempt
-# non-retryable timeout, and the post-retry second timeout) to decide whether to
-# `exit 0` instead of `exit 1`.
+# exit code is a TIMEOUT disposition: EXACTLY 124 OR 137. Both are a `timeout`
+# deadline hit:
+#   * 124 — `timeout` sent SIGTERM and the probe died from it (the clean case).
+#   * 137 — the probe IGNORED SIGTERM, so the `-k` grace escalated to SIGKILL
+#     (128+9). Because the probe runs in its own process group (no
+#     --foreground), GNU `timeout` reports this group SIGKILL as 137, NOT 124
+#     (verified on coreutils 8.32). This is the EXACT signature of the upstream
+#     `go run` probe that hung PR #3273's testnet shard (run 27298458353): with
+#     the new `-k` it is now force-killed and surfaces as 137. Treating 137 as a
+#     timeout disposition is what makes the root-cause fix actually de-gate main
+#     — without it the probe would be bounded but stay red.
+# Every OTHER non-(124|137) exit is NOT soft-skipped — a genuine probe assertion
+# failure (e.g. exit 1) must stay red so a real henyey-on-testnet break is never
+# masked. Callers use this at each timeout sink (first-attempt non-retryable
+# timeout, and the post-retry second timeout) to decide whether to `exit 0`
+# instead of `exit 1`.
+#
+# Scope note: 137 is also the runner-reclamation SIGKILL signature handled by
+# is_retryable_exit. That is fine here: a 137 is ALWAYS an externally-killed
+# (timeout-grace or runner-reclamation) outcome, never a probe's own assertion
+# result — `go` test failures exit 1, not 137 — so soft-skipping 137 under the
+# opt-in testnet-only flag never masks a real break.
 should_soft_skip_timeout() {
-    [[ "$SOFT_ON_TIMEOUT" == true && "$1" -eq 124 ]]
+    [[ "$SOFT_ON_TIMEOUT" == true ]] && { [[ "$1" -eq 124 ]] || [[ "$1" -eq 137 ]]; }
 }
 
 emit_soft_skip() {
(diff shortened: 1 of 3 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** CI-infra-only de-gate of the chronically-flaky testnet Quickstart shard. Logic is correct, the 124+137 timeout-disposition scope is empirically justified and necessary, and the regression test (Test 27) genuinely fails on the pre-`-k` code. Cycle 1 (no prior Correctness comment).

<details>

**Kind:** bug-fix (systematic CI red, urgent). Regression test required + verified.

**Regression-test verification (`test-coverage`) — PASS:**
- `test_sigterm_ignoring_probe_is_force_killed_and_soft_skipped` (Test 27) wires a SIGTERM-ignoring probe (`trap '' TERM; sleep 60`) through the wrapper with `--timeout 1`, `KILL_GRACE=1s`, `--soft-on-timeout`, bounded by an outer `timeout -k 5s 30`. It asserts (a) prompt termination (`elapsed < 25s`) and (b) soft-skip path (exit 0 + SOFT-SKIP marker).
- **Verified fail-first:** I stripped `-k "$KILL_GRACE"` from the wrapper's probe invocation on the PR head (reproducing the pre-fix `8557b84a` state) and re-ran the harness → `83/85`, with both Test 27 assertions failing (`not ok 84 terminates_quickly`, `not ok 85 timeout_is_soft_skipped`). Restoring `-k` → `85/85`. The test captures exactly the incomplete-fix bug.
- Full harness on PR head (merged with current `origin/main`): **85/85 pass**, matching the claim.

**124+137 soft-skip scope (the deviation from binding plan guidance) — correct & necessary:**
- Plan critic-A's binding guidance was "124-only; keep 125/126/127/**137** red." The implementation soft-skips 124 OR 137. I verified empirically on coreutils 8.32 (same family as CI): a SIGTERM-ignoring child under `timeout -k 1 1 ... | tee` exits **137**, not 124, because the child runs in its own process group and `timeout` reports the group SIGKILL as 128+9. A clean (non-ignoring) timeout still yields **124**; an assertion `exit 1` propagates as **1**. So 124-only would leave the actual `go run` hang red — the deviation is necessary for the fix to be non-inert, and is documented at every sink.
- The deviation is appropriately recorded in the commit message and code comments; it does not contradict the plan's *intent* (de-gate testnet timeouts), only critic-A's pre-empirical exit-code assumption.

**Logic — correct:**
- `should_soft_skip_timeout()` gates strictly on `SOFT_ON_TIMEOUT == true && (124 || 137)`; applied at both timeout sinks (post-retry second timeout, and non-retryable timeout). Non-124/137 (assertion failures) fall through to `exit 1` unchanged — the must-not-mask invariant is locked by `test_soft_on_timeout_still_fails_real_assertion` (asserts exit code propagates as exactly 1, no marker).
- `--foreground` correctly rejected: I verified `timeout --foreground -k 1 1 <ignoring-child> | tee` runs the full sleep (does NOT force-kill at the deadline) and reports 124 — i.e. `--foreground` re-introduces the hang in a pipeline. Not using it is correct.

**`doc-drift` (non-blocking, inline):** `emit_soft_skip()` hardcodes "exit 124 treated as neutral" in its message even when the exit was 137; and two comments say "exit 124 ONLY" at sinks that actually accept 124 OR 137 via `should_soft_skip_timeout`. Code behavior is correct; only the log text/comments drifted. Filing as inline follow-up, not blocking.

**Off-by-default / scope — PASS:** flag defaults OFF; wrapper invocation byte-identical for all non-testnet shards (`SOFT_FLAG=""`). `test_soft_on_timeout_off_by_default_preserves_red` + `workflow_pubnet_local_shards_have_no_soft_flag` lock this.
</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** The liveness-masking tradeoff is real but correctly bounded: the soft-skip is scoped to the testnet shard ONLY, and the local + pubnet shards still hard-gate henyey sync-liveness, so we do not go blind to henyey hangs. Diagnostics are preserved on every soft-skip. Cycle 1 (no prior Risk comment).

<details>

**`regression-risk` — liveness masking is the real risk; verified acceptably bounded:**
- A henyey HANG on testnet (consensus stall / deadlock that prevents sync), not just testnet slowness, would now present as a probe timeout → soft-skipped green on the testnet shard. This is the genuine coverage loss and is the crux of the change. It is acceptable **only because** other shards still gate henyey liveness:
  - **(a) Testnet-only scope — confirmed.** Soft flag + 240s are set via matrix keys on the `network: testnet / enable: "core,horizon"` shard alone (quickstart.yml L437-441). The loop sets `SOFT_FLAG=""` and keeps the 600s `run_attempt*multiplier*60` budget when `SHARD_SOFT_ON_TIMEOUT`/`SHARD_PROBE_TIMEOUT` are empty. I diffed the rendered shard list: local (×4) and pubnet (×1) carry neither key → byte-identical wrapper invocation. `workflow_pubnet_local_shards_have_no_soft_flag` locks this.
  - **(b) Henyey liveness still gated by other shards — confirmed.** The `local` shards run `test_core.go` + `test_horizon_core_up.go` + `test_horizon_ingesting.go` against a locally-spun henyey core, and `pubnet` runs `test_core.go test_horizon_up.go`. Neither carries the soft flag → a henyey deadlock/consensus-stall/sync-hang surfaces there as a HARD red (timeout stays red, 124/137 → exit 1). We lose testnet-external-liveness signal only; we do NOT lose henyey sync-liveness CI coverage. The local shard in particular exercises henyey-against-core sync without any external-network dependency, so a genuine henyey hang is still caught.
  - **(c) Diagnostics preserved — confirmed.** `capture_diagnostics` runs on every non-zero attempt before the soft-skip decision; both timed-out attempts leave `attempt-*` dirs. `test_soft_on_timeout_preserves_diagnostics` asserts this, and the SOFT-SKIP marker is grep-able in the uploaded log. A masked testnet hang is at least inspectable post-hoc.
- Net: the tradeoff is the intended de-gate from the issue's own direction #2/#3, and is the narrowest viable scoping (one shard, timeout-disposition only, genuine assertion failures still red on that same shard).

**`api-break` / `migration-risk` — none.** New flag is opt-in, default OFF; no Rust, no data formats, no config schema, no consumer outside this workflow. Every existing caller is byte-identical (off-by-default test locks it).

**`perf-regression` — none, positive.** 240s testnet budget bounds the step at ~20 min worst case (4 probes × 240 + one retry) vs the prior ~55-min job-cancel hang. `-k` guarantees `timeout` returns instead of wedging in `wait()`. This is the operational fix for the original symptom.

**137 over-broad-masking stress test — does not surface a real break:** 137 = external SIGKILL only (timeout `-k` grace, or runner reclamation). A go-test assertion failure exits 1, never 137 (verified). The only way henyey-relevant work yields 137 is if it is itself being timed-out/killed externally — i.e. a hang, which is the accepted-masked case on testnet only, still caught on local/pubnet. No genuine henyey *assertion* failure can surface as 137 on this probe.

**Operational note (non-blocking):** if testnet is chronically degraded, the testnet shard could silently soft-skip for long stretches with reduced signal. Mitigated by the grep-able SOFT-SKIP marker + preserved diagnostics making the soft-skip observable; an operator can watch marker frequency. Inline/follow-up at most, not a blocker.
</details>

scripts/ci/run-quickstart-test.sh
```diff
@@ -106,6 +146,39 @@ is_retryable_exit() {
...
+# (timeout-grace or runner-reclamation) outcome, never a probe's own assertion
+# result — `go` test failures exit 1, not 137 — so soft-skipping 137 under the
+# opt-in testnet-only flag never masks a real break.
+should_soft_skip_timeout() {
+    [[ "$SOFT_ON_TIMEOUT" == true ]] && { [[ "$1" -eq 124 ]] || [[ "$1" -eq 137 ]]; }
+}
+
+emit_soft_skip() {
+    echo "=== SOFT-SKIP: testnet sync probe timed out (environmental, not a henyey failure) ===" >&2
+    echo "=== SOFT-SKIP: $NETWORK/$ENABLE/$PROBE exit 124 treated as neutral (#3272); diagnostics preserved ===" >&2
```
tomerweller, line 179: doc-drift (non-blocking): this message hardcodes "exit 124 treated as neutral" but `should_soft_skip_timeout` also soft-skips exit 137 (the `-k` SIGKILL case, which is the actual `go run` signature). Consider emitting the real exit code, e.g. "exit $EXIT_CODE treated as neutral". Behavior is correct; only the log/comment text is stale. Same for the "exit 124 ONLY" comments at the two soft-skip sinks.

tomerweller:
## ✅ Merged

**Commit:** e9ed773b647893104d162e59d7827e3fbf8bc20d

**Follow-up issues filed for unaddressed inline review comments:** #3275

Triple-green: Reviewer A (Correctness) APPROVE, Reviewer B (Risk) APPROVE, CI green (39/39, testnet shard SUCCESS). No external CHANGES_REQUESTED. Not self-modifying. The 124+137 soft-skip deviation, the testnet-only liveness-masking scope, and the Test 27 regression were all empirically verified.

PR #3277, opened 2026-06-10
Render post-restart absolute fire as (post-restart), not (burst)

tomerweller:
Closes #3274

## Summary

The monitor-tick recovery_stalled status line misclassified a post-restart absolute fire (#3198/#3206) as a cross-tick `(burst)` delta. Root cause was a two-part gap: `maybe_post_restart_fire` correctly returns a firing result carrying `extra_values={"post_restart": True, ...}` on a baseline-reset (PID-change / counter-reset) tick when the absolute value crosses `post_restart_absolute_threshold` (50), but `make_result` builds the result dict from a fixed key set and never copied `post_restart` out of `extra_values` — so `render_aggregate` only saw `value=736 >= 10` and rendered `(burst)`.

This surfaces `post_restart` onto the result dict in `make_result`, and adds a `post_restart` branch in `render_aggregate`'s recovery_stalled firing path (before the `value>=10` burst check) that emits the documented `absolute=N (post-restart)` form (monitor-tick SKILL.md L1335 — previously a code-vs-spec divergence). The relabel triggers only when `post_restart` is set, which happens only on genuine baseline-reset ticks; same-PID real bursts never carry the marker and still render `(burst)`. No firing/threshold/cooldown/filing semantics change — only the human-facing label.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3274#issuecomment-4675105489)

## Test plan

- [x] `python3 scripts/lib/test_eval_alarms_counter_reset.py` — 19 passed, 0 failed
- [x] `python3 scripts/lib/test_eval_alarms_converged_loop.py` — 8 passed
- [x] `python3 scripts/lib/test_eval_alarms_histogram_rebucket.py` — 10/10 passed
- [x] `python3 -m pytest scripts/ci/test_check_alarm_versions.py` — 23 passed
- [x] `python3 -m py_compile` on changed files
- [x] pre-commit hook (cargo fmt + clippy) passed
- [x] `scripts/test-monitor-skill-snippets.sh` — only the 4 pre-existing `pv-label-sync` failures (present on `origin/main` baseline, unrelated to this change)

## Regression test (kind: bug-fix)

- **Tests:** `scripts/lib/test_eval_alarms_counter_reset.py::test_post_restart_fire_sets_marker_on_result` and `::test_render_post_restart_absolute_not_burst`
- **Pre-fix:** committed as `69031de9` — verified FAILED with:
[...]

Issue #3198, opened 2026-06-05
Non-critical: monitor-tick: post-restart forcing_catchup_behind stall invisible to checks (2) and (12b)


tomerweller:
## Symptom
A post-restart consensus stall of up to ~15 min is invisible to the monitor-tick health checks. Observed this tick: a planned deploy restart of `519665ad` caused a ~6.5 min `forcing_catchup_behind` stall (213 stall-ticks; see #3197) that **neither check (2) nor check (12b) would have reported** on an unattended restart.

## Evidence
- The stall self-recovered in ~6.5 min, under the 15m clean-restart sync deadline → **check (2)** reports `CATCHING UP`, never `SYNC FAILURE`.
- **Check (12b)** resets its counter baseline on PID change (restart) and reports `collecting baseline` on the first post-restart tick; the very next tick sees `delta=0` (already recovered). The 213-tick burst spans a single ~20-min tick interval that straddles the baseline-reset tick, so the streak/burst logic never observes an incrementing delta across two valid ticks.
- Net: the most severe `forcing_catchup_behind` bursts occur precisely at restart, exactly when both detectors are blind. Detection this tick only happened because a deploy tick was actively watching the restart via log scan.

## Suspected root cause
- `SKILL.md` check (12b) invalidation rule (PID/start_ticks change → "collecting baseline", no burst eval) — sound for cross-incarnation safety, but creates a blind spot for the post-restart burst.
- check (2) deadline semantics: a sub-deadline stall that recovers is never escalated.

## Concrete fix sketch (design choice required)
Option A: On a deploy/restart tick, snapshot the absolute `forcing_catchup_behind` value once the node reaches real-time sync; if it exceeds a threshold (e.g. > 50 accrued during the warmup window) emit a WARNING + file/comment (the absolute post-restart value is a reliable proxy for "stalled during startup", independent of cross-tick deltas).
Option B: Add a dedicated post-restart-stall detector keyed on `henyey_post_catchup_hard_reset_total` / `stellar_herder_lost_sync_total` deltas since process start (both jumped this episode: 3 and 5 respectively).
Option C: Lower check (2)'s tolerance for a planned (`preredeploy`) restart that does not reach real-time sync within ~3 min.

Threshold values and which signal to key on need an operator/design decision.

## Note
Largely mitigated once #3197 (the node-side stall fix) lands. Filing so the monitoring gap is tracked independently of the node fix.

tomerweller:
## Triage Report

[...]

Issue #3274, opened 2026-06-10
Non-critical: monitor-tick: post-restart absolute recovery_stalled fire mislabeled as (burst) delta


tomerweller:
## Symptom
On the first metrics tick after a deploy (PID change → counter-streak snapshot invalidation), a **post-restart absolute-value fire** (#3198) is rendered as a cross-tick **burst**:

```
recovery_stalled: WARNING delta=736 (burst) — investigating
```

The `736` is the *absolute* counter value `henyey_recovery_stalled_tick_total{reason="forcing_catchup_behind"}` on the freshly-restarted PID — NOT a one-interval delta. Rendered as `delta=… (burst)` it reads as a massive acute single-interval stall (alarming, suggests active sustained stalling) when it is actually the expected post-deploy absolute baseline being surfaced once via the #3198 path.

## Evidence (this tick, build 599ebd16, new PID 1472167)
- `PREV_PROM_INVALID=true` reason=`process identity changed` (deploy at 19:05Z, old PID 1216346 → new 1472167).
- `counter_streak_snapshot` after eval: `pid=1472167 counter_value=736 breach_streak=0` — i.e. the evaluator took the **invalidation → `maybe_post_restart_fire`** path (streak reset to 0, baseline rewritten), not a cross-tick delta path.
- prev.prom (old PID) value was 455; a naive cross-tick delta would be 736−455=281. The reported value is **736 = the absolute**, confirming the fire came from `maybe_post_restart_fire` (value=cur_val), yet it was labeled `(burst)`.

## Suspected root cause
`scripts/lib/eval-alarms.py:1317` renders **every** firing recovery-stalled alarm with a single hard-coded template:

```python
recovery_stalled_line = f"recovery_stalled: WARNING delta={r['value']} (burst) — investigating"
```

This ignores the `post_restart` marker that `maybe_post_restart_fire` sets (`"post_restart": True`, line ~1080) and also does not distinguish the streak-fire case. SKILL.md (check 12b status-line spec) defines three distinct render forms:
- `WARNING delta=N (burst) — investigating`
- `WARNING delta=N (M ticks) — investigating`  *(streak)*
- `WARNING absolute=N (post-restart) — investigating`  *(#3198)*

Only the first is ever emitted.

## Concrete fix sketch
At `eval-alarms.py:1317`, branch on the fire reason carried in the alarm result:

```python
r_meta = r  # the alarm dict returned by the streak evaluator
if r_meta.get("post_restart"):
[...]

PR #3283, opened 2026-06-15
Guard counter alarms against missing process identity

### LINKED ISSUE #3279: monitor-tick: eval-alarms persists empty-identity counter_streak snapshot, causing spurious post-restart fire
## Symptom

A monitor-tick `recovery-stalled` alarm fired as a **post-restart absolute fire**
(`WARNING absolute=64 (post-restart)`) on a tick where the node was healthy
(Validating, real-time sync) and the underlying counter
`henyey_recovery_stalled_tick_total{reason="forcing_catchup_behind"}` was
**frozen** (delta=0, same value 64 as the prior tick, same PID). The fire was a
false positive — there was no restart and no new stall.

## Evidence

- Node: PID 1731959 stable across both ticks, uptime 31m, Validating, age=5s.
- `counter_streak_snapshot` after the firing run:
  `pid=1731959 start_ticks=802654858 counter_value=64 breach_streak=0`
  (matches current process exactly).
- Re-running the evaluator read-only (`--no-snapshot-write`) immediately after
  reports `recovery-stalled: ok (delta=0)`.
- Root trace: the *prior* tick invoked `eval-alarms.py` **without exporting
  `PID`/`START_TICKS`** (an abbreviated post-restart verify run). `main()` reads
  identity from the environment:

  ```
  scripts/lib/eval-alarms.py:1542  pid = os.environ.get("PID", "")
  scripts/lib/eval-alarms.py:1543  start_ticks_val = os.environ.get("START_TICKS", "")
  ```

  With both empty, `eval_counter_streak` wrote a snapshot with `pid=""`,
  `start_ticks=""`. On the next (correctly-exported) tick, the identity check

  ```
  scripts/lib/eval-alarms.py:1126  elif snapshot.get("pid") != pid or snapshot.get("start_ticks") != start_ticks:
  ```

  saw `"" != "1731959"` → took the PID-change branch → `maybe_post_restart_fire`
  → spurious WARN. (`eval_counter_ratio` has the identical pattern at line 911.)

## Suspected root cause

The evaluator treats an **empty-string** process identity as a *valid* identity
and persists it to the snapshot. An empty identity is not a real
incarnation — it is "identity unknown". Writing it poisons the baseline so the
next run with a real identity mis-detects a restart and false-fires the #3198
post-restart absolute path.

## Concrete fix sketch

In `scripts/lib/eval-alarms.py`, guard the counter-streak / counter-ratio
families against missing identity. Either:

- **(a)** In `main()` (~line 1542): if `pid` or `start_ticks_val` is empty,
  skip the counter-streak and counter-ratio snapshot families entirely
  (emit `skipped (no process identity)`), leaving the prior snapshot intact; or
- **(b)** In `eval_counter_streak` (~line 1113) and `eval_counter_ratio`
  (~line 869): early-return `make_result(alarm, "skipped",
  skip_reason="missing process identity")` **without writing the snapshot**
  when `not pid or not start_ticks`.

Option (b) is more local and prevents the poison at the writer. Add a unit test
alongside `scripts/lib/test_eval_alarms_counter_reset.py` asserting that an
empty `PID`/`START_TICKS` does not mutate `counter_streak_snapshot` and does not
fire post-restart.

## Notes

- No node impact: the node was healthy throughout; the counter was frozen.
- Proximate trigger was a non-standard abbreviated tick that skipped the
  documented `export PID/START_TICKS`; this is defense-in-depth so that an
  identity-less invocation cannot poison the persisted baseline.

tomerweller:
Closes #3279

## Summary

`eval_counter_streak` and `eval_counter_ratio` in `scripts/lib/eval-alarms.py` sourced process identity from `PID`/`START_TICKS`, which an abbreviated monitor tick can leave as empty strings. Both evaluators treated an empty identity as a valid distinct incarnation: the identity-less tick took the PID-change branch, wrote a fresh baseline under `pid=""` (poison write), and (for `eval_counter_streak`) post-restart-fired on the discarded absolute value because `cur_val (64) >= threshold (50)`. The next real-PID tick then read the poisoned `pid=""` snapshot, re-entered the PID-change branch, and false-fired `WARNING absolute=64 (post-restart)` on a healthy node with a stable PID and a frozen counter (delta=0).

This adds an early guard at the top of BOTH functions, before any snapshot read/write or metric I/O: if `not pid or not start_ticks`, return `skipped (missing process identity)`. An identity-less tick becomes a pure no-op — no fire, no poison write, the prior valid baseline preserved — so the next real tick reports `ok`. A genuine restart (real, differing non-empty PID/start_ticks) above the threshold still fires `post_restart` (#3198 preserved). This is a pre-existing FIRE bug; #3277 (`bd1ca634`) was render-only and is not reverted — the `(post-restart)` render path it added is undisturbed.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3279#issuecomment-4711673885)

## Test plan

- [x] `python3 -m py_compile eval-alarms.py` and `test_eval_alarms_counter_reset.py` clean
- [x] cargo fmt --check / cargo clippy (pre-commit hook, all passed)
- [x] `test_eval_alarms_counter_reset.py` — 24 passed
- [x] `test_eval_alarms_converged_loop.py` — 8 passed
- [x] `test_eval_alarms_histogram_rebucket.py` — 10 passed

- **Tests:** `test_eval_alarms_counter_reset.py`: `test_missing_identity_does_not_poison_snapshot_or_fire`, `test_empty_identity_snapshot_then_real_tick_no_fire`, `test_counter_ratio_missing_identity_skips_without_write` (+ guard tests `test_genuine_restart_still_fires_post_restart`, `test_same_pid_frozen_counter_reports_ok`)
- **Pre-fix:** committed as `e694f11a` — verified FAILED: identity-less streak tick returned `firing` (`post_restart=True`, value 64) and rewrote the baseline to `pid=""`; identity-less ratio tick returned `collecting_baseline` and rewrote baselines.
- **Post-fix:** verified PASS after `ad37993f`. The two guard tests passed before and after (protecting #3198 genuine-restart fire and the same-PID frozen-counter `ok` property).

## Deviations from plan

None. (Test `test_empty_identity_snapshot_then_real_tick_no_fire` is written as the true end-to-end two-tick sequence — identity-less tick then real-PID frozen tick — rather than directly seeding a poisoned `pid=""` snapshot, since the fix prevents the poison from ever being written; it still fails on main and passes after.)

### DIFF
--- a/scripts/lib/eval-alarms.py
+++ b/scripts/lib/eval-alarms.py
@@ -870,6 +870,19 @@ def eval_counter_ratio(
     """
     ev_default = default_extra_values(alarm, "counter-ratio")
 
+    # Missing process identity guard (#3279): mirror eval_counter_streak. An
+    # abbreviated tick that skips exporting PID/START_TICKS passes EMPTY strings.
+    # An empty identity is NOT a valid distinct incarnation — treating it as one
+    # invalidates the snapshot and rewrites the per-alarm baselines under the
+    # empty identity (poison write). eval_counter_ratio does not call
+    # maybe_post_restart_fire so it cannot post-restart-fire, but it must not
+    # persist an empty identity either. Establish nothing, decide nothing: skip
+    # BEFORE any snapshot/metric I/O so the prior valid baseline is preserved.
+    if not pid or not start_ticks:
+        return make_result(alarm, "skipped",
+                           skip_reason="missing process identity",
+                           extra_values=ev_default)
+
     # Global skip conditions for ratio checks
     ledger_age = extract_value(current, "stellar_ledger_age_current_seconds", "form1")
     if fresh_start:
@@ -1105,6 +1118,20 @@ def eval_counter_streak(
     Independent of PREV_PROM_INVALID — uses own PID/start_ticks in snapshot.
     """
     ev_default = default_extra_values(alarm, "counter-streak")
+
+    # Missing process identity guard (#3279): an abbreviated tick that skips
+    # exporting PID/START_TICKS passes EMPTY strings here. An empty identity is
+    # NOT a valid distinct incarnation — treating it as one makes the PID-change
+    # branch below write a fresh baseline under pid="" (poison write) and
+    # post-restart-fire on the discarded absolute value, then the NEXT real-PID
+    # tick reads the poisoned pid="" snapshot, re-enters the PID-change branch,
+    # and false-fires (post-restart) despite a stable PID and frozen counter.
+    # Establish nothing, decide nothing: skip BEFORE any snapshot/metric I/O so
+    # the prior valid baseline is preserved untouched.
+    if not pid or not start_ticks:
+        return make_result(alarm, "skipped",
+                           skip_reason="missing process identity",
+                           extra_values=ev_default)
 
     metric = alarm["metric"]
     extraction = alarm.get("extraction", "form2")
--- a/scripts/lib/test_eval_alarms_counter_reset.py
+++ b/scripts/lib/test_eval_alarms_counter_reset.py
@@ -559,23 +559,24 @@ def test_missing_identity_does_not_poison_snapshot_or_fire():
 
 
 def test_empty_identity_snapshot_then_real_tick_no_fire():
-    """Two-tick sequence: an empty-identity snapshot (the poisoned state) followed
-    by a real-PID tick with a frozen counter (delta=0) must report ok, NOT a
-    post-restart fire.
+    """End-to-end two-tick sequence on a healthy node: an identity-less tick
+    (empty PID/START_TICKS) THEN a real-PID tick with a frozen counter (delta=0)
+    must report ok, NOT a post-restart fire.
 
-    Fails on origin/main: the real pid "1731959" != poisoned snapshot pid ""
-    re-enters the PID-change branch and false-fires post_restart=True value=64.
-    After the fix the guard never writes the empty-identity baseline in the first
-    place, so this test seeds it explicitly to prove the real tick does not fire.
+    Fails on origin/main: tick 1 poisons the snapshot to pid="" and itself fires
+    post_restart; tick 2's real pid "1731959" != poisoned "" re-enters the
+    PID-change branch and false-fires post_restart=True value=64. After the fix
+    tick 1 is a skipped no-op that preserves the prior VALID baseline, so tick 2
+    sees a stable PID and delta=0 → ok.
     """
     with tempfile.TemporaryDirectory() as d:
         state_dir = Path(d)
         snap_path = state_dir / "counter_streak_snapshot"
-        # Seed the poisoned empty-identity snapshot directly.
+        # Prior valid baseline established by an earlier healthy tick.
         write_snapshot(snap_path, {
             "version": "1",
-            "pid": "",
-            "start_ticks": "",
+            "pid": "1731959",
+            "start_ticks": "802654858",
             "counter_value": "64",
             "breach_streak": "0",
         })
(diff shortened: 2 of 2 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Bug-fix is correct, minimal, and well-tested. An early identity guard at the top of both `eval_counter_streak` and `eval_counter_ratio` returns `skipped (missing process identity)` before any snapshot/metric I/O when `pid` or `start_ticks` is empty. Regression tests fail on pre-fix code and pass after; #3198 genuine-restart fire and #3277 render path are both preserved.

<details>

**Cycle:** 1 (no prior Correctness verdict on this PR) — complete change-list below; none needed (APPROVE).

**kind:** bug-fix (per the converged plan comment; no `## Triage Report` exists but `/plan` treated it as bug-fix). PR body has a complete `## Regression test` section.

**Correctness of the diff**
- The guard `if not pid or not start_ticks: return make_result(alarm, "skipped", skip_reason="missing process identity", extra_values=ev_default)` is placed as the FIRST statement after `ev_default` in both functions — before `extract_value`, before `read_snapshot`, before any `write_snapshot`. So an identity-less tick is a pure no-op: no fire, no poison write. Verified by reading lines 1119–1135 (streak) and 870–883 (ratio) of `scripts/lib/eval-alarms.py`.
- The condition is exactly right: `not pid or not start_ticks` is truthy only for empty string (or None). A real, non-empty differing PID falls through to the existing PID-change branch (line ~1158) unchanged, so a genuine restart with `cur_val >= post_restart_absolute_threshold` still calls `maybe_post_restart_fire` and fires. #3198 behavior preserved.
- `make_result(..., "skipped", ...)` is the existing skip idiom used elsewhere in both functions (`metric not found`, `FRESH_START`, `ledger age > 30s`, `uptime < 10m`), so the result shape and `render_aggregate` handling are consistent. No new result schema introduced.

**Test coverage (verified myself in a fresh worktree under ~/data)**
- Regression test ordering correct: test commit `e694f11a` precedes fix commit `ad37993f`.
- **Fail-before**: at `e694f11a` (tests present, fix absent — `grep -c "missing process identity" = 0`), the 3 regression tests FAIL: `test_missing_identity_does_not_poison_snapshot_or_fire`, `test_empty_identity_snapshot_then_real_tick_no_fire`, `test_counter_ratio_missing_identity_skips_without_write` (observed `collecting_baseline`/`firing` instead of `skipped`). The 2 guard tests (`test_genuine_restart_still_fires_post_restart`, `test_same_pid_frozen_counter_reports_ok`) PASS pre-fix, correctly proving they protect pre-existing behavior rather than the new code.
- **Pass-after**: at PR head, all 3 suites green — `test_eval_alarms_counter_reset.py` 24 passed, `test_eval_alarms_converged_loop.py` 8 passed, `test_eval_alarms_histogram_rebucket.py` 10 passed (42 total).
- #3277 `(post-restart)` render tests (`test_render_post_restart_absolute_not_burst`, `test_render_same_pid_burst_still_burst`, `test_post_restart_fire_sets_marker_on_result`, `test_post_restart_below_threshold_does_not_fire`) all still pass.

**Error handling / readability**
- Comments on both guards are clear and accurately describe the two failure modes (self-fire + poison write transitively causing the next-tick false fire). The ratio comment correctly notes ratio cannot post-restart-fire but still must not persist an empty identity (defense-in-depth). No misleading claims.

**Self-modifying:** No. `scripts/lib/eval-alarms.py` is a standalone monitor helper and is NOT in the self-modifying path set.

No changes requested.
</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Low-risk, additive monitor-tooling fix. The only behavior change is that an identity-less tick now returns `skipped` instead of `collecting_baseline`/`firing`; both `skipped` and `collecting_baseline` are benign non-firing states handled identically downstream by `render_aggregate`. No API/data-format break, no perf concern, no migration. The change strictly removes a false-positive alarm path.

<details>

**Cycle:** 1 (no prior Risk verdict on this PR) — complete risk change-list below; none blocking (APPROVE).

**regression-risk — none found.**
- The guard returns *earlier* on a strictly narrower input set (empty `pid` OR empty `start_ticks`) than any prior return. Every code path with a real, non-empty identity is byte-for-byte unchanged (the diff is purely two additive guard blocks; no existing line modified — confirmed via `git diff origin/main..HEAD`). So no real-identity tick changes behavior. #3198 genuine-restart post-restart fire and #3274/#3277 render labeling are untouched.
- Behavior delta for the identity-less tick: previously `eval_counter_streak` could return `firing (post_restart)` and `eval_counter_ratio` could return `collecting_baseline`, both with a poison `pid=""` snapshot write; now both return `skipped`. I traced both consumer sites: `render_aggregate` (lines 1299, 1312–1334, 1364–1366) treats `skipped` and `collecting_baseline` as equivalent benign non-firing status lines — neither escalates, neither is an error. So the state change is observationally cosmetic (a different status string) and strictly removes the spurious WARNING. No downstream branch keys on `collecting_baseline` to drive an action that `skipped` would break.

**perf-regression — none.** The guard is two comparisons at the top of each function; it short-circuits before file I/O, so it is strictly faster on the identity-less path.

**api-break — none.** No function signature changed; `pid`/`start_ticks` were already parameters. Result dict shape uses the existing `make_result(..., "skipped", skip_reason=...)` idiom already emitted by these same functions. No data-format / snapshot-schema change — in fact the fix *prevents* writing a malformed `pid=""` snapshot, so it is schema-protective.

**migration-risk — none.** No persisted-state migration. A previously-poisoned `pid=""` snapshot on disk (from the pre-fix bug) is still read by the real-identity path and treated as a PID change → re-collects baseline, which is the correct recovery and does not fire on a frozen counter once a valid baseline is re-established. The fix prevents new poison writes going forward.

**config-risk / operational — none.** Monitor tooling only; no validator/consensus path, no config surface, no parity surface (not in `crates/{scp,herder,ledger,tx,overlay}`). The operational effect is the intended one: fewer false-positive `(post-restart)` pages on healthy nodes during abbreviated ticks.

**Self-modifying:** No — `scripts/lib/eval-alarms.py` is not in the self-modifying path set.

No blocking risk. No inline follow-ups warranted.
</details>

tomerweller:
## ✅ Merged

**Commit:** 0f83ceae077f0224059085768a5820cd8033fbc7

**Follow-up issues filed for unaddressed inline review comments:** none (no inline review threads).

Triple-green: Reviewer A (Correctness) APPROVE, Reviewer B (Risk) APPROVE, CI 40/40 green. No agent CHANGES_REQUESTED findings (refute pass not required). Not self-modifying. Fail-before/pass-after verified in a fresh worktree (3 regression tests fail at pre-fix `e694f11a`, all 42 eval-alarms tests pass at PR head; #3198 genuine-restart fire and #3277 `(post-restart)` render tests preserved).

PR #3285, opened 2026-06-15
Sync SCP post-verify label set to canonical 15 labels (add drift_manual_close)

tomerweller:
Closes #3278
Closes #3281

## Summary

The canonical `PostVerifyReason` source (`crates/herder/src/scp_verify.rs:374`, `GateDriftManualClose => "drift_manual_close"`) defines 15 post-verify labels, compile-time-asserted against `ALL`. Several duplicated copies of the label set were stale at 14, missing `drift_manual_close`. Because the eval-alarms label check does exact set-equality, the `scp-accept-rate-low` alarm skipped every tick with skip_reason `"label set mismatch"` — silently disabling SCP accept-rate monitoring (#3281) — and the pv-label-sync tests failed (#3278). Both issues share this single root cause.

Pure config/doc/test/fixture sync — no Rust change. Synced all duplicated copies to 15 labels:
- `.claude/skills/shared/metric-alarms.toml` — added `drift_manual_close` to `scp-accept-rate-low` `expected_labels` (numerator unchanged).
- `.claude/skills/monitor-loop/SKILL.md` — added `drift_manual_close` to the inline label list and updated the label-count tokens 14→15.
- `scripts/fixtures/eval-alarms/healthy-current.prom` and `healthy-prev.prom` — added `henyey_scp_post_verify_total{reason="drift_manual_close"} 0`.

The fixture series are mandatory: bumping the TOML to 15 without adding the fixture series would invert the bug (the alarm would skip on its own fixture). Value 0 matches the live "all drift reasons = 0" state and leaves the accept-ratio numerically unchanged.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3278#issuecomment-4711682698)

## Test plan

- [x] `bash scripts/test-monitor-skill-snippets.sh` — full suite green (1..355, exit 0)
- [x] pv-label-sync group (TAP 169-174) all `ok` — was `not ok` 170-173 on main
- [x] eval-alarms pytest suites: `test_eval_alarms_converged_loop.py` (9 passed), `test_eval_alarms_counter_reset.py` (19 passed), `test_eval_alarms_histogram_rebucket.py` (10 passed)
- [x] `eval-alarms.py --validate-only` valid
- [x] py_compile + bash -n clean

## Regression test

- **#3278 (pre-existing, now passing):** `scripts/test-monitor-skill-snippets.sh` pv-label-sync TAP 170-173 — verified `not ok` on `origin/main`, `ok` after the sync. No edit to those tests.
[...]

PR #3287, opened 2026-06-15
Bound quickstart probe wrapper on orphaned stdout-pipe holders (#3286)

### LINKED ISSUE #3286: CI: Quickstart testnet shard STILL hangs at step 8 despite #3273 -k fix — second hang path (orphaned docker-exec holds tee pipe)
## Summary

## Evidence

## Impact / urgency

Refs #3272, #3273.

tomerweller:
Closes #3286

## Summary

The quickstart probe wrapper (`scripts/ci/run-quickstart-test.sh`) could hang the entire CI job for the full wall-clock (~54 min on PR #3285) even with the #3273 `-k` kill-after grace in place. `run_probe()` ran the probe as a pipeline — `timeout … "${PROBE_CMD[@]}" 2>&1 | tee <log>` — and read `exit_code=${PIPESTATUS[0]}`, which only executes after the whole pipeline completes. `timeout`'s `-k` SIGKILLs the probe's process *group*, but a process the probe detached into its own session (a `setsid`/double-forked grandchild, or in-container work reached over keep-alive sockets) survives the group SIGKILL and keeps the inherited stdout pipe write-end open, so `tee` never sees EOF and blocks forever — the pipeline never completes, `exit_code` is never assigned, and the 124/137 soft-skip after the probe never runs.

This replaces the `| tee` pipeline with a non-pipeline redirect (`timeout … > <log> 2>&1` then `exit_code=$?`) so the wrapper no longer depends on the orphan ever closing the fd — `timeout`'s own exit governs the wrapper and the soft-skip fires in seconds. For a non-pipeline command `$?` is byte-identical to the value `${PIPESTATUS[0]}` held, so the 124/137 timeout disposition, the exit-1 hard-fail, the success path, and the retry are all unchanged. Probe output goes to the attempt log (already uploaded as a diagnostics artifact) instead of also streaming to the live console.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3286#issuecomment-4712929906)

## Test plan

- [x] `bash scripts/test-quickstart-harness.sh` — 87/87 ok (was 85; +2 from the new regression test)
- [x] `bash -n` clean on both scripts
- [x] shellcheck: no new findings (run-quickstart-test.sh fully clean; harness finding counts identical to baseline)
- [x] pre-commit hooks (cargo fmt + clippy) passed on both commits

- **Test:** `scripts/test-quickstart-harness.sh`::`test_timeout_with_orphaned_pipe_holder_still_terminates_and_soft_skips`
- **Pre-fix:** committed as `01159c49` — verified FAILED on origin/main: a fake probe orphans a `setsid` grandchild holding the stdout pipe open; the `| tee` wedges, the wrapper never returns, the outer watchdog trips (`not ok` — no return + no SOFT-SKIP).
- **Post-fix:** verified PASSES after `59940151` — redirect-to-file lets `timeout` govern the wrapper; it returns in ~2s, exit 0 + SOFT-SKIP marker.

The orphaned `setsid` grandchild is reaped via a unique per-test sentinel (`pkill -f`) in a `RETURN` trap so it cannot leak across runs.

## Deviations from plan

None. (The plan's `tap_plan 28 → 30` figure was stale; origin/main's current plan is 85, so the bump is 85 → 87. The redirect-only form was chosen with no `cat` echo-back, confirmed correct since the workflow consumes only the wrapper's exit code + stderr markers + uploaded artifact, never the probe's live stdout.)

### DIFF
--- a/scripts/ci/run-quickstart-test.sh
+++ b/scripts/ci/run-quickstart-test.sh
@@ -217,8 +217,6 @@ run_probe() {
     echo "=== Attempt $attempt: $PROBE ($NETWORK/$ENABLE) ===" >&2
     echo "Command: timeout -k ${KILL_GRACE} ${TIMEOUT}s ${PROBE_CMD[*]}" >&2
 
-    # Use PIPESTATUS[0] to capture the timeout exit code, not the tee exit.
-    #
     # The `-k ${KILL_GRACE}` is load-bearing (#3273 root-cause fix): GNU
     # `timeout` first sends SIGTERM at the deadline, then — only with -k — sends
     # SIGKILL after the grace if the child is still alive. The upstream probe is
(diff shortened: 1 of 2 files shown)


tomerweller:
## ⚠️ Do NOT merge as the #3286 fix — proven insufficient by its own CI

This PR's redirect change is a reasonable robustness improvement (and the orphaned-pipe harness test is valid), BUT its own testnet shard hung **~55 min (22:44→23:39)** identically to the pre-fix behavior — so the redirect does **not** fix the actual step-8 hang. The hang is not in the wrapper (see the definitive analysis on #3286). Merging this under "Closes #3286" would falsely close the issue.

Holding this PR: either it gets re-scoped to the instrumentation-first approach (#3286) and the real grandchild-kill fix, or it's superseded by that work. Not merging as-is.

tomerweller:
Superseded by #3289. The issue's plan was re-converged to **diagnostic instrumentation** after this PR: the two wrapper-level fixes (#3273 `timeout -k`, and this PR's `> file` redirect) both failed to bound the ~55-min step-8 hang identically, so the cause is NOT in `run-quickstart-test.sh` (which this PR modified). #3289 instead leaves the wrapper untouched and instruments the testnet shard (fail-fast step timeout + process-tree watchdog) to capture data from the next hang. Closing in favor of #3289.

PR #3269, opened 2026-06-10
Wire restore_apply_fan_out into the archive cold-catchup restore path

tomerweller:
Closes #3268

## Summary

Completes the #3249 lever-D knob: the `[buckets] restore_apply_fan_out` cap now covers the **archive cold-catchup** restore path (its original motivation, #3245/#3235), not just warm restart. `CatchupManager` gains a `restore_apply_fan_out: Option<usize>` field (default `None`) and a `set_restore_apply_fan_out` setter (mirroring `set_network_passphrase`/`set_emit_meta_ext_v1`); `catchup_impl` wires in `self.config.buckets.restore_apply_fan_out_cap()`. `apply_buckets` swaps both sequential `restore_from_has` calls (live + hot-archive) for `restore_from_has_parallel(.., self.restore_apply_fan_out)`, which bounds concurrent bucket materialization via the `FanOutLimiter` semaphore (#3249). With this plus #3267 (`scan_thread_count=2`, already merged), an operator's `restore_apply_fan_out=6` will actually cap a cold catchup.

## Byte-identical state

Capping concurrency changes only HOW MANY workers materialize at once, never WHAT is assembled: `restore_from_has_parallel` assembles levels by index, not completion order, so the restored `bucketListHash` is identical for every fan-out value. The existing `test_restore_from_has_parallel_matches_sequential` proves hash-equality across schedulings; the new `test_apply_buckets_byte_identical_across_fan_out` proves it on the cold-catchup path specifically. Default (`0`->`None`) is a true no-op — default-config nodes are byte-for-byte unchanged.

## Parity

stellar-core bounds catchup apply concurrency via `WORKER_THREADS`; henyey's unbounded cold-catchup fan-out was the outlier. Wiring the cap *aligns* henyey with core. No XDR / header / merge-semantics change; the post-restore `validate_bucket_list_structure` check is unchanged.

## Off-tokio-worker download fallback (per Critic A)

The normal cold-catchup flow pre-downloads all buckets to disk before `apply_buckets`, so `load_bucket` reads from disk. The in-closure `block_on_async(download_bucket_from_archives(..))` is only a defensive "shouldn't happen" fallback. Under `restore_from_has_parallel` that fallback now runs on a `std::thread::scope` OS thread where `Handle::try_current()` is `Err`, so `block_on_async` builds a fresh current-thread runtime (download.rs) — correct, no `block_in_place` panic (that branch only fires when a runtime handle IS current). A behavioral shift only in the rare fallback path.

## Plan reference

[...]

Issue #3266, opened 2026-06-10
metrics: henyey_recovery_stalled_tick_total{reason="forcing_catchup_behind"} — ~6min stall on missing next-slot EXTERNALIZE (slow peer-escalation, 7a3795c1)


tomerweller:
## Symptom

~6-minute consensus stall at tip on the production validator (build `7a3795c1`, post-#3265). The node externalized ledger **62968321** at 15:07:29Z then stopped advancing for ~6 minutes while the network ran ahead to ~62968386 (peer_gap grew monotonically 0 → 64). Recovery escalation eventually forced a hard-reset catchup at 15:13:41Z and recovered to tip by 15:13:54Z (~13s catchup). Node is healthy/Validating now.

This self-recovered within the 15m sync deadline (so monitor check (2) did not flag SYNC FAILURE), but it is a genuine multi-minute non-validation window and a recurrence of the slow-recovery class.

## Evidence

Monitor tick 235 (15:17Z) caught the counter burst:
- `henyey_recovery_stalled_tick_total{reason="forcing_catchup_behind"}` 18 → 367 (delta ≈ 349, burst)
- `henyey_recovery_stalled_tick_total{reason="near_tip_peer_scp_recovery"}` 19 → 369
- `henyey_recovery_stalled_tick_total{reason="archive_behind_peer_ahead_hard_reset"}` 0 → 4  ← the **legitimate** peer-ahead path (#3265 working: `at_tip_no_scp_hard_reset` stayed 0)
- `henyey_post_catchup_hard_reset_total` 0 → 3
- `stellar_herder_lost_sync_total` fired

Heartbeats kept firing every 10s throughout (no event-loop freeze / watchdog), but lcl was frozen at 62968321:

```
15:07:20  ledger=62968320  peer_gap=0     <- last normal close
15:07:29  ledger=62968321  peer_gap=1     <- STUCK here
15:08:49  ledger=62968321  peer_gap=15
15:10:49  ledger=62968321  peer_gap=35
15:12:49  ledger=62968321  peer_gap=56
15:13:39  ledger=62968321  peer_gap=64
15:13:49  ledger=62968349  peer_gap=38     <- catchup fast-tracking
15:13:59  ledger=62968388  peer_gap=0      <- recovered
```

Decisive root-cause line (15:13:45Z):

```
Gap detected: next slot EXTERNALIZE missing, requesting SCP state current_ledger=62968321 latest_ext=62968386 missing_slot=62968322
[...]

PR #3288, opened 2026-06-15
Self-heal near-tip knit divergence via archive rebuild instead of wipe

### LINKED ISSUE #3282: URGENT: knit-to-LCL state-integrity FATAL on 384b4f6e during forced near-tip catchup (recurrence of #2931/#2886) — validator down ~5h, wipe required
## Summary

**Third recurrence** of the knit-to-LCL state-integrity FATAL (prior: #2931 Jun 2 on b35ea04e, #2886 May 22 on 6011ec9c — both closed). Mainnet validator on deployed build `384b4f6e` shut down with `fatal_wipe_required=true` and was **down ~5h16m** (14:06Z → 19:22Z observed) before detection. Recovery required a destructive state wipe + cold catchup (operator-authorized).

## Fatal error

```
2026-06-15T14:06:23.817653Z WARN  henyey_history::catchup::replay: Fatal error during replay — not retrying attempt=0
  error=knit-to-LCL failed at LCL: expected 48f847c6589142756898cb5d75ee09eab4a6f0d0fac04f7bea05e9956440e7ca,
        got      b58d0468122e00336a9f369e7825f7279c81e4be9cdbef0afe1f9916657c5809
2026-06-15T14:06:23.963906Z ERROR henyey_app::app::lifecycle: FATAL: unrecoverable local state failure —
  Externalized catchup verification/integrity failure: knit-to-LCL failed at LCL: expected 48f847c6…, got b58d0468…
  Node will shut down. State wipe required before restart. fatal_wipe_required=true
2026-06-15T14:06:27.960338Z ERROR henyey_app::run_cmd: Node stopped with error=Node shutdown due to fatal state failure — state wipe required
```

The **locally-computed** hash (`got b58d0468…`) diverged from the network-expected LCL (`expected 48f847c6…`) — the forced-catchup replay produced incorrect state.

## Trigger sequence (~14:00–14:06Z, around L63041928)

1. `14:00:57Z` Pending EXTERNALIZE far ahead — fast-tracking catchup (slot=63041931, current=63041928, gap=3).
2. `14:00:59Z+` Receiving SCP but no externalization — **tx_sets evicted from peers**; near-tip recovery routed to peer-SCP back-fill ("archive confirmed behind unpublished checkpoint", mirroring stellar-core outOfSyncRecovery), next_checkpoint=63041983.
3. `14:01:22Z+` "Recovery stalled for too long — forcing catchup" (urgent=true), 12+ attempts.
4. `14:06:23Z` Forced catchup replay → **knit-to-LCL hash mismatch** → FATAL → shutdown, wipe required.

This is the same RecoveryEscalation → forced-catchup → knit-to-LCL divergence path as #2931. The near-tip forced-catchup replay is producing a ledger state that does not knit onto the persisted LCL.

## Impact

- Validator down ~5h16m (no auto-restart — node self-flags `fatal_wipe_required=true`, refuses restart on existing state).
- Recovery = destructive state wipe + cold catchup. Already authorized & in progress; loop being resumed.

## Asks

- [ ] Root-cause why the forced near-tip catchup replay (RecoveryEscalation / outOfSyncRecovery path) computes a divergent ledger hash at the knit-to-LCL step. Per CLAUDE.md, reproduce as a targeted offline verify-execution test around L63041928.
- [ ] Reassess the #2886/#2931 fixes — they did not prevent this recurrence.
- [ ] Consider an auto-recovery path (or guarded auto-wipe+cold-catchup) so a knit-to-LCL FATAL does not produce multi-hour silent outages.

cc parity / catchup. Relates to #3266 (forcing_catchup_behind near-tip stalls).

tomerweller:
Closes #3282

## Summary

A forced near-tip `ReplayOnly` recovery catchup clones the live bucket list + current header and sets `override_lcl = Some(current)`, making the **local** LCL the knit subject. When the local LCL hash diverges from the canonical **published** archive header (so the #2937 publication gate passes — a different divergence class than the unpublished-checkpoint case #2931/#2886 were closed for), the §11.2 case-3 knit raises `KnitLclHashMismatch`. In `handle_catchup_result`'s `Err` branch this was treated as fatal → `trigger_fatal_shutdown` + `fatal_wipe_required`, returning **before** the `else` block where the existing `catchup_needs_full_reset` self-heal lives. So the node wiped instead of rebuilding from the archive — the common defect behind all three recurrences (#2886, #2931, #3282).

The fix tags cloned-local near-tip catchups (`seeded_from_local_clone`, threaded from `catchup_with_run_mode` through the spawned task's `PendingCatchupResult` into `handle_catchup_result`). On a local-vs-archive divergence for such a catchup, it sets `catchup_needs_full_reset` and returns **non-fatal** — the next catchup runs `force_full` (`existing_state=None`, `override_lcl=None`), an archive-authoritative bucket-apply that ignores the cloned local state (mirroring stellar-core `CatchupWork::downloadApplyBuckets`). The latch is **one-shot**: the rebuild runs `seeded_from_local_clone=false`, so a repeat divergence goes terminal/fatal; `catchup_needs_full_reset` is cleared on the next catchup entry so it cannot re-latch.

**Detection is unchanged** — the §11.2 knit matrix, the publication gate, and `is_fatal_catchup_failure` all stay strict. Only the app-layer **response** to this divergence class changes, from wipe to rebuild. Recovery-routing trigger conditions are out of scope (#3266 track).

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3282#issuecomment-4712937037)

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy -p henyey-app -p henyey-history -- -D warnings` (clean)
- [x] `cargo test -p henyey-history -p henyey-app` passes
- [x] `cargo test --all` passes

## Regression test (bug-fix)

- **Offline repro (condition):** `crates/history/.../replay.rs::test_offline_near_tip_replay_knit_to_lcl_repro` — published gate passes + divergent local LCL ⇒ `KnitLclHashMismatch` (the exact #3282 observable).
- **Fixed response (test 1):** `crates/app/.../catchup_impl.rs::test_near_tip_knit_mismatch_triggers_full_reset_not_wipe` — committed as `b8a5cc2b0e10f06605b12907ebcfe947ae71d5c5` (test) → with the interception disabled this **FAILS** with "must NOT be fatal — it self-heals"; **PASSES** after the fix `cabe805f858760af203390f21959d3366ee4525e` (sets `catchup_needs_full_reset`, no wipe).
- **Safety backstop (test 2):** `test_full_bucket_apply_knit_mismatch_still_wipes` — same error, `seeded_from_local_clone=false` ⇒ still fatal/wipe (no over-broadening).

## Fan-out diagnostic (rule-in/out, #3269)

`crates/history/.../buckets.rs::test_apply_buckets_full_header_hash_identical_across_fan_out` — asserts the **full ledger-header hash** (not just bucketListHash) is identical across `restore_apply_fan_out ∈ {None,1,2,6}`. **GREEN** → fan-out is exonerated as the #3282 recurrence cause (consistent with #2886/#2931 predating all fan-out work). No follow-up issue filed.

## Parity

A knit disagreement is a retryable Work failure upstream (`CatchupWork::fatalFailure()` boolean, `stellar-core/src/catchup/CatchupWork.h:80`, consumed at `LedgerApplyManagerImpl.cpp:124`), surfaced from `ApplyCheckpointWork::getNextLedgerCloseData` (`ApplyCheckpointWork.cpp:236`), recovered via `CatchupWork::downloadApplyBuckets` (`CatchupWork.cpp:198`) — **not** a destructive wipe. The retained terminal wipe is a henyey-stricter policy reached only after the parity-aligned archive re-derivation also fails.

## Deviations from plan

None.

### DIFF
--- a/crates/app/src/app/catchup_impl.rs
+++ b/crates/app/src/app/catchup_impl.rs
@@ -187,11 +187,26 @@ impl App {
         // replaying from the (possibly corrupt/diverged) local state.
         let force_full = self.catchup_needs_full_reset.load(Ordering::SeqCst);
         if force_full {
-            tracing::warn!("Previous catchup failed — forcing full bucket-apply catchup");
+            tracing::warn!(
+                "Previous catchup failed — forcing full bucket-apply catchup \
+                 (archive-authoritative, local state ignored)"
+            );
         }
+
+        // #3282: track whether this catchup is seeded from CLONED LOCAL state.
+        // Only the near-tip fast path below (initialized ledger manager, no
+        // force_full) clones the live LCL header and sets `override_lcl`,
+        // making the *local* LCL the knit subject. A local-vs-archive
+        // divergence on such a catchup is self-healable from the archive (a
+        // one-shot `force_full` rebuild) rather than a terminal wipe; the
+        // force_full / HAS-rebuild / no-existing-state paths are
+        // archive-authoritative and stay terminal on divergence. Default to
+        // `false`; set `true` only on the cloned-local fast path.
+        let mut seeded_from_local_clone = false;
 
         let (existing_state, override_lcl) = if current >= GENESIS_LEDGER_SEQ && !force_full {
             if self.ledger_manager.is_initialized() {
+                seeded_from_local_clone = true;
                 // Fast path: clone from live ledger manager.
                 // Must resolve async merges first — structure-based restart_merges
                 // creates PendingMerge::Async handles, and BucketLevel::clone()
@@ -245,6 +260,14 @@ impl App {
         } else {
             (None, None)
         };
+
+        // Publish the seeding mode so the spawned catchup task can forward it
+        // to `handle_catchup_result` (the single in-flight catchup is
+        // serialized by `catchup_in_progress`, so this read-after-return is
+        // race-free). See `last_catchup_seeded_from_local_clone` (#3282).
+        self.last_catchup_seeded_from_local_clone
+            .store(seeded_from_local_clone, Ordering::SeqCst);
+
         // Run catchup work
         let output = self
             .run_catchup_work(
@@ -2510,11 +2533,27 @@ impl App {
     /// Returns `true` if a fatal state failure was detected and shutdown
     /// has been triggered. The caller must skip all post-catchup work
     /// when this returns `true`.
+    /// Process the outcome of a catchup attempt.
+    ///
+    /// `seeded_from_local_clone` indicates the failed catchup replayed from
+    /// CLONED LOCAL state (the near-tip fast path, `catchup_with_run_mode`
+    /// :193-225, `override_lcl = Some(current)`). When `true`, a
+    /// local-vs-archive divergence error
+    /// ([`HistoryError::is_local_vs_archive_divergence`]) is **self-healable**:
+    /// instead of a terminal wipe, route to a one-shot archive-authoritative
+    /// full bucket-apply rebuild (`catchup_needs_full_reset = true`). When
+    /// `false` (the catchup was already an archive re-derivation, or never
+    /// touched local state), the same divergence stays terminal/fatal — the
+    /// archive disagrees with the consensus chain and the one-shot rebuild
+    /// must not loop (#3282).
+    ///
+    /// Returns `true` iff a fatal shutdown was triggered.
     pub(super) async fn handle_catchup_result(
         &self,
         catchup_result: anyhow::Result<CatchupResult>,
         reset_stuck_state: bool,
         label: &str,
+        seeded_from_local_clone: bool,
     ) -> bool {
         match catchup_result {
             Ok(result) => {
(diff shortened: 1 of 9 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Tightly-scoped state-integrity fix. The self-heal/wipe routing is correct, the one-shot latch provably terminates, and the regression test genuinely fails-before / passes-after. Detection is untouched. Built and ran all four new tests in a fresh worktree off current `origin/main`; all green, and the fail-before check reproduces the bug.

<details>

**Cycle 1** — complete change-list (no prior Correctness comment on this PR). No concern reaches CHANGES_REQUESTED; the notes below are non-blocking observations.

Linked issue `kind: bug-fix` (production-down FATAL #3282). Regression-test verification per the bug-fix protocol:

- **`test-coverage` — fail-before/pass-after VERIFIED.** The fix and its app-layer test (`test_near_tip_knit_mismatch_triggers_full_reset_not_wipe`) are in the same commit (`cabe805f`), so I could not check out a test-before-fix SHA. Equivalent verification done instead: disabled the self-heal guard (`if false && …`) in the worktree and re-ran. The test FAILED with the exact assertion message "cloned-local near-tip knit mismatch must NOT be fatal — it self-heals" (`catchup_impl.rs:6168`); the backstop test still passed (confirming it is independent of the self-heal code). Reverted, re-ran: all four pass. The test is **not tautological** — the offline repro (`test_offline_near_tip_replay_knit_to_lcl_repro`) drives the real `checkpoint_publication_gate` (asserts it PASSES) and the real `knit_to_lcl_decision` case-3 path (raises the real `KnitLclHashMismatch`), exactly the production code path.
- **`error-handling` — routing correct.** New `Err`-branch block self-heals ONLY when `seeded_from_local_clone && is_local_vs_archive_divergence`, returns `false` (non-fatal), runs `restore_operational_state` + cooldown + stuck-state reset (parallel to the existing non-fatal tail), and returns before the fatal block. The classifier double-checks `LedgerError::HashMismatch` via a second `downcast_ref`, matching the existing `is_hash_mismatch` block's structure.
- **`regression-risk` — one-shot latch terminates.** Traced the full lifecycle: self-heal sets `catchup_needs_full_reset=true`; the flag is cleared (`store(false)`) ONLY after `run_catchup_work` succeeds (`catchup_impl.rs:~639`). On the next catchup entry, `force_full=true` ⇒ the `current >= GENESIS && !force_full` clone branch is skipped ⇒ `seeded_from_local_clone` stays `false` ⇒ a repeat divergence skips the self-heal guard and falls through to `is_fatal_catchup_failure()` → wipe. No rebuild→diverge→rebuild loop is possible.
- **`test-coverage` — backstop present.** `test_full_bucket_apply_knit_mismatch_still_wipes` pins that the same error with `seeded_from_local_clone=false` is still fatal and emits the shutdown signal — the fix does not over-broaden into "never wipe".
- Threading of `seeded_from_local_clone` (atomic → `PendingCatchupResult` → `handle_catchup_result`) is consistent; all `PendingCatchupResult::new` call sites (incl. persist.rs tests) updated for the new arg. `Ordering` resolves via `use super::*`. `cargo fmt`/clippy clean (CI green).

**Non-blocking note (would be inline, not gating):** the race-free claim rests on `catchup_in_progress` serializing catchups so the atomic store-then-read can't interleave with another catchup. That holds today (single in-flight catchup). If a future change ever permits concurrent catchups, the `last_catchup_seeded_from_local_clone` atomic would need to become per-catchup state. Worth a one-line code comment but not a blocker.

</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** The new app-layer response matches stellar-core's near-tip recovery posture: a knit "disagreed on hash" is a retryable Work failure recovered by re-applying buckets from the archive HAS, NOT a destructive state wipe. All four cited `stellar-core/` references are real and accurate (verified against the v26.0.1 submodule). Detection (§11.2 knit matrix + #2937 publication gate) is byte-for-byte unchanged.

<details>

**Cycle 1** — complete change-list (no prior Parity comment). No `parity-gap` / `spec-divergence` / `sequencing` / `edge-case` concern rises to CHANGES_REQUESTED.

**Parity citations VERIFIED against the v26.0.1 submodule:**

- `src/catchup/ApplyCheckpointWork.cpp:236` — the knit-to-LCL disagreement is `throw std::runtime_error("replay of {} at LCL {} disagreed on hash")`, guarded by `header.ledgerSeq == lclHeader.header.ledgerSeq` (the case-3 overlap). This is the upstream analogue of henyey's `KnitLclHashMismatch` (case 3 in `knit_to_lcl_decision`). The henyey case ordering (case 5 overshoot checked before case 4) is explicitly noted to mirror `ApplyCheckpointWork.cpp:246`. ✓
- `src/catchup/CatchupWork.h:80` — `fatalFailure()` is a **boolean** (reads `mFatalFailureFuture`). A checkpoint-apply knit disagreement makes the Work FAIL but is **not** classed as a fatal failure. ✓
- `src/catchup/LedgerApplyManagerImpl.cpp:124` — `fatalFailure()` is consumed: only `WORK_FAILURE && fatalFailure()` sets `mCatchupFatalFailure` + `CLOG_FATAL`. A non-fatal Work failure is retried, not wiped. ✓
- `src/catchup/CatchupWork.cpp:198` — `CatchupWork::downloadApplyBuckets()` re-derives state by applying buckets from the archive HAS. This is the recovery path henyey now mirrors via the `force_full` (`existing_state=None`/`override_lcl=None`) rebuild. ✓

**Semantic match.** Upstream, a knit/apply disagreement on a checkpoint is a retryable failure whose recovery is an archive bucket-apply — there is no equivalent of the henyey terminal state-wipe at this point. The PR moves henyey's response for the cloned-local divergence class FROM wipe TO archive-authoritative rebuild, which closes the parity gap. The retained terminal wipe (only after the archive rebuild ALSO diverges, i.e. `seeded_from_local_clone=false`) is correctly characterized as a henyey-**stricter** backstop; it is reached only past the point where upstream would still be retrying, so it never makes henyey *looser* than core. No new looseness, no observable-behavior divergence on the success path.

**`sequencing` — the cloned-local fast path is a henyey-only optimization.** stellar-core always fetches the HAS before applying a checkpoint and has no cloned-local fast path; henyey's `override_lcl = Some(current)` near-tip path is the henyey-specific construct that makes the *local* LCL the knit subject. The fix correctly scopes the self-heal to exactly that henyey-only path (`seeded_from_local_clone`) and leaves all archive-authoritative paths terminal — so the divergence-from-core only ever existed on the henyey path, and the fix removes it without touching parity-aligned paths.

**Detection unchanged (confirmed).** `is_fatal_catchup_failure()` still lists `KnitLclHashMismatch` and all knit/replay variants; `checkpoint_publication_gate` and the §11.2 `knit_to_lcl_decision` matrix are not modified by this PR (the only `error.rs` change is the additive, read-only `is_local_vs_archive_divergence()` classifier). The offline repro asserts the gate PASSES yet the knit still diverges — correctly demonstrating #2937 keyed on publication, not local-LCL correctness, which is why the recurrence class survived. The fan-out diagnostic is GREEN, exonerating #3269 fan-out as the #3282 cause (consistent with #2886/#2931 predating fan-out); no follow-up needed.

</details>

tomerweller:
## ✅ Merged

**Commit:** a215805a0f73497d4e7908966de039502e30f7ac

Triple-green: Reviewer Correctness APPROVE, Reviewer Parity APPROVE (parity-critical lens per catchup/replay), CI green (39/39). No external CHANGES_REQUESTED, no unaddressed inline threads.

Adversarial verification: fail-before/pass-after confirmed for `test_near_tip_knit_mismatch_triggers_full_reset_not_wipe` (disabling the self-heal guard reproduces the FATAL); one-shot latch traced to terminate; all 4 stellar-core citations verified against the v26.0.1 submodule; detection (§11.2 matrix + #2937 gate) unchanged; fan-out diagnostic GREEN (#3269 exonerated).

PR #3289, opened 2026-06-16
Instrument testnet quickstart hang: fail-fast + process-tree watchdog (#3286)

### LINKED ISSUE #3286: CI: Quickstart testnet shard STILL hangs at step 8 despite #3273 -k fix — second hang path (orphaned docker-exec holds tee pipe)

tomerweller:
Closes #3286

## Summary — this is DIAGNOSTIC instrumentation, NOT a fix

Two wrapper-level fixes have now **both** failed to bound the ~55-min testnet
quickstart step-8 hang with the identical signature: #3273 (`timeout -k` +
124/137 soft-skip) and #3287 (replace `| tee` with `> file` redirect — its
**own** testnet shard hung 22:44→23:39). Since #3287 ran its own fixed wrapper
and still hung identically, the time is **not** spent in
`scripts/ci/run-quickstart-test.sh` — both wrapper hypotheses are falsified. That
script is therefore **left untouched** here (the twice-failed-fix surface).

The blocker to progress is **unobservability**: the 45-min *job-level*
`timeout-minutes` is a **cancel**, so `if: failure()` never fires and the
diagnostics step uploads nothing. This PR fixes observability, not the hang.
**Its deliverable is DATA from the next hang** — a fast-fail RED with an uploaded
process tree — not a fix for the flake.

The actual fix (likely `go build` the probe once + run the binary directly under
`timeout` so `timeout`'s child IS the test binary, eliminating the unsignalled
`go run` grandchild; and/or `pkill -f quickstart/tests` on the timeout
disposition) is a **separate follow-up** informed by the captured process tree.

> **Note on this PR's own testnet shard:** it observes the very flake it
> instruments. If it hangs, it will now **fail-fast RED at ~25 min WITH the
> diagnostic artifact attached** (that artifact is itself the first data
> capture). Re-run the workflow to get a green for merge.

## What changes

- **`.github/workflows/quickstart.yml`**
  - Testnet-scoped `matrix.step_timeout_minutes: 25`, applied as the "Run probes
    through wrapper" step's `timeout-minutes: ${{ matrix.step_timeout_minutes || 360 }}`.
    Converts the next testnet hang from a 45-min job cancel (uploads nothing) into
    a ~25-min **step failure** that fires the upload step. 25 min > the ~20-min
    healthy budget, so a healthy run never trips it. **Other shards have no
    `step_timeout_minutes` and inherit the generous 360-min default — a no-op
    under the job's own 45-min cap — so they stay byte-identical.**
  - In-step, testnet-only background **watchdog** launched before the probe loop
    (`STEP_PID=$$` pinned in the step body, not a subshell, not `$BASHPID`).
    Reaped on the healthy path via `trap … EXIT` + an explicit post-loop `kill`,
    so a passing run never leaks the sleep or emits a spurious dump.
  - Upload condition hardened from `if: failure()` to
    `if: ${{ failure() || cancelled() }}` (belt-and-suspenders: step-timeout
    failure **and** residual job-level cancel).
- **`scripts/ci/quickstart-hang-watchdog.sh` (NEW, tiny)** — at ~20 min (before
  the 25-min step kill) dumps to the diagnostics dir: `date -u`, `ps -ejH`,
  `ps faux`, `pstree -p` (fallback `ps faux`), lingering
  `go run|quickstart/tests|stellar-core` PIDs, the step shell's open fds, and
  **per-PID `/proc/<pid>/fd` for every lingering process** (the orphaned-grandchild
  fd-holder hypothesis — what would confirm/refute the `go run` grandchild holding
  the step's stdout open). Extracted as a standalone script so the harness tests
  the exact code the workflow runs (no inline-vs-test drift). Every capture is
  `|| true`.
- **`scripts/test-quickstart-harness.sh`** — adds
  `test_testnet_hang_watchdog_emits_process_dump_before_step_kill`: simulates a
  hang and asserts the watchdog writes the process-dump file BEFORE a simulated
  step-kill deadline, with a unique-sentinel `pkill` reap so nothing leaks.
  `tap_plan` 85 → 88.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3286#issuecomment-4713731948)

## Test plan

- [x] `bash scripts/test-quickstart-harness.sh` → **88/88 ok** (existing 85 stay green + 3 new)
- [x] `bash -n` clean on `quickstart-hang-watchdog.sh` and `test-quickstart-harness.sh`
- [x] `bash -n` clean on the extracted step `run:` body
- [x] `shellcheck` clean on the new watchdog (no new findings; pre-existing harness SC2016/SC2034 untouched)
- [x] workflow YAML validates (`yaml.safe_load`)
- [x] non-testnet shards confirmed byte-identical (only the testnet matrix entry + shared step header changed)

## Deviations from plan

- The converged plan referenced `tap_plan 28 → 29`; current `origin/main` carries
  85 TAP assertions, so the bump is **85 → 88** (the new test contributes 3
  assertions). Same intent, different baseline.

### DIFF
--- a/.github/workflows/quickstart.yml
+++ b/.github/workflows/quickstart.yml
@@ -434,11 +434,24 @@ jobs:
           # 600s hard-fail-on-timeout and carry no soft flag.
           # Budget: 4 probes × 240s + one 240s retry = 1200s (~20 min), well
           # under the job's 45-min timeout-minutes.
+          #
+          # DIAGNOSTIC instrumentation (#3286): this shard also sets a TIGHT,
+          # testnet-only step-level fail-fast bound via step_timeout_minutes: 25
+          # (applied as the "Run probes through wrapper" step's timeout-minutes
+          # below). Two wrapper-level fixes (#3273, #3287) both FAILED to bound
+          # the ~55-min step-8 hang, and a 45-min JOB-level wall-clock is a
+          # *cancel* — so `if: failure()` never fires and the diagnostics step
+          # never uploads. A step-level timeout-minutes kill instead marks the
+          # STEP failed (distinct from a job cancel), so the upload step runs.
+          # 25 min > the ~20-min healthy budget above, so a healthy run never
+          # trips it. Other shards inherit the generous 360-min default (a
+          # no-op under the job's own 45-min cap) and stay byte-identical.
           - network: testnet
             enable: "core,horizon"
             probes: "test_core.go test_horizon_up.go test_horizon_core_up.go test_horizon_ingesting.go"
             soft_on_timeout: true
             probe_timeout: 240
+            step_timeout_minutes: 25
           # Additional pubnet shard
           # Note: upstream excludes horizon_core_up, horizon_ingesting, and
           # stellar_rpc_healthy on pubnet (see internal-test.yml conditionals).
@@ -480,6 +493,15 @@ jobs:
             --enable ${{ matrix.enable }}
 
       - name: Run probes through wrapper
+        # DIAGNOSTIC fail-fast bound (#3286). Testnet sets
+        # matrix.step_timeout_minutes: 25 (tighter than its ~20-min healthy
+        # budget) so a hang here FAILS the STEP at ~25 min instead of running
+        # to the 45-min JOB cancel — a step-timeout marks the step *failed*, so
+        # the "Upload diagnostics" step (if: failure()) actually runs and the
+        # watchdog's process-tree dump is uploaded. All other shards have no
+        # step_timeout_minutes set and inherit the generous 360-min default
+        # (a no-op under the job's 45-min cap), so they stay byte-identical.
+        timeout-minutes: ${{ matrix.step_timeout_minutes || 360 }}
         env:
           NETWORK: ${{ matrix.network }}
           ENABLE: ${{ matrix.enable }}
@@ -519,6 +541,29 @@ jobs:
           fi
           DIAG_DIR="/tmp/quickstart-diagnostics"
 
+          # ---- DIAGNOSTIC hang watchdog (#3286), testnet shard only ----
+          # Capture the step shell's PID HERE, in the step body — NOT inside a
+          # subshell and NOT via $BASHPID — so the backgrounded watchdog can
+          # dump the fds THIS shell holds. The watchdog sleeps WATCHDOG_DELAY
+          # (default 1200s = 20 min, < the 25-min step_timeout_minutes bound)
+          # then writes a process-tree + open-fd snapshot to the diagnostics
+          # dir, so on the next hang the dump is on disk BEFORE the step kill
+          # and gets swept into the uploaded artifact. It is reaped on the
+          # healthy path (trap EXIT + explicit kill after the loop) so it never
+          # lingers or emits a spurious dump on a passing run. This does NOT fix
+          # the hang — its deliverable is DATA from the next hang.
+          STEP_PID=$$
+          WATCHDOG_PID=""
+          if [[ "$NETWORK" == "testnet" ]]; then
+            ( sleep "${WATCHDOG_DELAY:-1200}"; \
+              bash scripts/ci/quickstart-hang-watchdog.sh \
+                "$DIAG_DIR/testnet-hang-watchdog" "$STEP_PID" ) &
+            WATCHDOG_PID=$!
+            # Reap the watchdog on any step exit so a passing run never leaks
+            # the backgrounded sleep.
+            trap '[[ -n "$WATCHDOG_PID" ]] && kill "$WATCHDOG_PID" 2>/dev/null || true' EXIT
+          fi
+
           for probe_file in ${{ matrix.probes }}; do
             probe_name="${probe_file%.go}"
             probe_name="${probe_name#test_}"
@@ -536,8 +581,20 @@ jobs:
               -- go run "quickstart/tests/$probe_file"
           done
 
+          # Healthy path: probes finished, so reap the watchdog now (explicit
+          # kill, belt-and-suspenders with the EXIT trap) — no lingering sleep,
+          # no spurious dump on a passing run.
+          if [[ -n "$WATCHDOG_PID" ]]; then
+            kill "$WATCHDOG_PID" 2>/dev/null || true
+          fi
+
       - name: Upload diagnostics on failure
-        if: failure()
+        # DIAGNOSTIC instrumentation (#3286): upload on BOTH a step-timeout
+        # failure (the primary path — a testnet step_timeout_minutes kill marks
+        # the step failed, so failure() fires) AND a residual job-level cancel
+        # (belt-and-suspenders). The watchdog writes its dump before the step
+        # kill, so the artifact exists for either disposition.
+        if: ${{ failure() || cancelled() }}
         uses: actions/upload-artifact@v4
         with:
           name: quickstart-diagnostics-${{ matrix.network }}-${{ matrix.enable }}
--- a/scripts/ci/quickstart-hang-watchdog.sh
+++ b/scripts/ci/quickstart-hang-watchdog.sh
@@ -1,0 +1,82 @@
+#!/usr/bin/env bash
+# quickstart-hang-watchdog.sh — DIAGNOSTIC instrumentation for the testnet
+# quickstart "Run probes through wrapper" step hang (#3286).
+#
+# This script does NOT fix the hang. Its deliverable is DATA from the next
+# hang: a process-tree + open-fd snapshot written to the diagnostics dir
+# BEFORE the testnet step's tight `timeout-minutes` kill fires, so the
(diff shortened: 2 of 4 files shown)


tomerweller:
## Validation finding: instrumentation did NOT deliver — needs revision before merge

#3289's own testnet shard hung again (run 27586108310). Two defects observed:

**1. The `step_timeout_minutes: 25` did NOT fire on the hanging shard.** Attempt-1 step 8 "Run probes through wrapper" (started 00:56:21Z) was still `in_progress` at ~01:36Z — well past the 25-min mark (01:21Z). So the step-level timeout was NOT applied to the `test (testnet, core,horizon, test_core.go test_horizon_up.go test_horizon_core_up.go test_horizon_ingesting.go)` shard. Likely the `step_timeout_minutes` matrix key was added to a DIFFERENT testnet matrix entry than this one (there are multiple testnet variants — the main list vs the "Additional testnet shard" include). Verify which matrix entry this specific hanging shard comes from and attach the key THERE.

**2. No diagnostic artifact was captured** (no watchdog/process-tree artifact on the run — only image tarballs). Root cause is defect #1: because the step-timeout didn't fire, step 8 ran to the **job-level** 45-min cancel — and a GitHub **job-timeout cancellation does NOT reliably run subsequent steps**, even with `if: failure() || cancelled()`. So the "Upload diagnostics" step never ran and the watchdog's dump had nowhere to go. The entire data-capture chain depends on the STEP-level timeout firing (step FAILS → `if: failure()` upload runs) — which is exactly the part that's broken.

**Revision needed (focused):** wire `step_timeout_minutes: 25` so it actually applies to THIS testnet shard variant (confirm via the rendered `timeout-minutes` on the step). Once the step-level timeout fires at 25 min, step 8 fails (not job-cancels), the `if: failure()` upload runs, and the watchdog dump uploads. Consider a harness/CI assertion that the rendered step `timeout-minutes` for the testnet shard is 25 (not the 360 default), so this wiring can't silently regress.

Bouncing to ready-for-doing. Do NOT merge until a run demonstrates the step fails fast at ~25 min AND the process-tree artifact uploads.

tomerweller:
## Revision: correct-by-construction step-timeout assertion (de637922)

Addressed the validation bounce. Key finding while investigating run **27586108310**: that run was **attempt 2**, and the step-timeout wiring **already worked** there — step 8 "Run probes through wrapper" started 01:31:31 and ended **failure** at 01:59:04 (~27 min ≈ the 25-min `step_timeout_minutes` bound + grace), the step FAILED (not a job cancel), the `if: failure()` "Upload diagnostics" step ran (success), and the artifact `quickstart-diagnostics-testnet-core,horizon` contains `testnet-hang-watchdog/dump.txt` — a 690-line process-tree dump captured at 01:52:22Z (≈21 min in, after the 20-min `WATCHDOG_DELAY`). The bounce comment's "did NOT fire / job-cancelled" diagnosis described **attempt 1**; attempt 2 proves the `${{ matrix.step_timeout_minutes || 360 }}` form resolves to 25 for this exact shard in practice.

So the wiring in `quickstart.yml` was correct and is **left untouched** (and `scripts/ci/run-quickstart-test.sh` untouched per scope). The real gap the bounce identified — no guard so it can't silently regress / doesn't depend on a flaky hang — is now closed:

- **New:** `scripts/ci/render-quickstart-step-timeout.py` parses the workflow and evaluates the per-shard `timeout-minutes` expression exactly as GitHub Actions does (GHA `||` falsy semantics: unset matrix key → null → 360; explicit `25` → 25). Renders:
  ```
  local|core|360   local|rpc|360   local|core,rpc,horizon|360   local|galexie|360
  testnet|core,horizon|25
  pubnet|core,rpc,horizon|360
  ```
- **New harness test 29** asserts (a) testnet/core,horizon renders **25**, (b) every non-testnet shard renders **360**, (c) exactly one shard carries a tight (<360) bound. Harness: **92/92 pass**.
- **Mutation-proven non-vacuous:** moving the key to the pubnet entry (the prior-attempt failure mode) renders testnet→360 and turns the assertion red.

Watchdog + `if: failure() || cancelled()` upload kept as-is (correct; STEP_PID pinned via `$$` in the step body).

Validation: `bash scripts/test-quickstart-harness.sh` → 92/92; `bash -n` clean; workflow YAML parses; clippy/fmt pre-commit hooks passed.

Note: the testnet shard remains the chronic external-liveness flake — a future run may still hang, but now fail-fasts at ~25 min WITH the process-tree artifact uploaded (the deliverable). Merge does not depend on that shard passing; it depends on the other checks + this rendered-timeout assertion.

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Purely-additive diagnostic instrumentation (workflow step-timeout + watchdog + correct-by-construction renderer/harness). Reviewed against current `origin/main` (PR head already up-to-date). Harness 92/92 green; render assertions verified non-vacuous by a real mutation test; healthy-path watchdog reap verified to leave no dump and no leaked process.

<details>

**Linked issue kind:** `kind: bug-fix` for the chronic testnet hang (#3286), but this PR is explicitly **instrumentation, not a fix** — its deliverable is observability/data. The "regression test" here is the correct-by-construction renderer assertion that pins the diagnostic wiring, plus the watchdog-emits-dump test. Both are present and verified below.

- `test-coverage` — VERIFIED. Ran `bash scripts/test-quickstart-harness.sh` myself → **92/92 ok, 0 failed** (matches task expectation; PR body's "88" predates the second commit `de637922` which added render tests 89–92).
- `test-coverage` (non-vacuous render assertion) — VERIFIED by mutation. Moved `step_timeout_minutes: 25` from the testnet entry to a pubnet entry in a throwaway copy and re-rendered: testnet/core,horizon flips 25→360 (assertion (a) goes red) AND pubnet renders 25 (assertion (b) "non-testnet must be 360" goes red), and assertion (c) "exactly one tight shard" still finds 1 but on the wrong shard. The agent's reported mutation test is confirmed real, not a tautology.
- `test-coverage` (watchdog) — the renderer (`render-quickstart-step-timeout.py`) faithfully mirrors GHA `||` falsy semantics (null/false/0/''/NaN falsy), refuses to guess on unrecognized expression forms (fails loudly), and is exercised by tests 89–92. The watchdog test simulates a hang under an outer `timeout` and asserts the dump lands BEFORE the kill, with a unique-sentinel `pkill` reap so nothing leaks.
- `error-handling` — watchdog wraps every capture in `|| true` under `set -euo pipefail`; renderer `fail()`s loudly on structural drift. Good.
- `readability` — extensive WHY-comments; the step-body captures `STEP_PID=$$` in the step shell (not a subshell, not `$BASHPID`) so the fd dump targets the right shell. Correct.
- run-quickstart-test.sh is untouched (the twice-failed-fix surface), consistent with the PR's thesis.

No correctness concerns. APPROVE.
</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Change is additive, testnet-scoped, no Rust, `run-quickstart-test.sh` untouched. Non-testnet shards render the generous 360 default (proven by the renderer + my mutation test) and are byte-identical. Healthy-path watchdog reap verified to leave no leaked process and no spurious dump. NOT self-modifying.

<details>

- `regression-risk` — Other shards: the only workflow-body change they see is the shared `timeout-minutes: ${{ matrix.step_timeout_minutes || 360 }}` on the "Run probes through wrapper" step. Since no non-testnet shard sets `step_timeout_minutes`, they render 360, which is a no-op under the job's own 45-min `timeout-minutes` cap. Confirmed via the renderer for all local/pubnet shards (all 360). No behavior change to non-testnet shards.
- `regression-risk` (healthy testnet) — the watchdog only backgrounds when `NETWORK == testnet`, sleeps 1200s (20 min) < the 25-min step bound, and is reaped on both the EXIT trap and an explicit post-loop `kill`. I reproduced the healthy path: probes finish → watchdog killed before it fires → no `dump.txt`, no leaked process. The 25-min bound (>20-min healthy budget) does not trip a healthy run.
- `config-risk` — `WATCHDOG_DELAY` env override defaults to 1200; only the harness sets it short. No new required secrets/config.
- `api-break` / `migration-risk` — none; no data formats, no Rust, no public surface.
- `perf-regression` — none; a single backgrounded sleep on the testnet shard only.
- `self-modifying` — checked: changed paths are `.github/workflows/quickstart.yml`, `scripts/ci/quickstart-hang-watchdog.sh`, `scripts/ci/render-quickstart-step-timeout.py`, `scripts/test-quickstart-harness.sh`. None match the self-mod patterns (`.claude/skills/...`, `scripts/lib/...`, `.github/skills/shared/...`). **NOT self-modifying.**
- `operational` — the PR's own testnet shard may go RED when the node hangs (by design: fail-fast at 25 min WITH the diagnostic artifact). That red is the de-gated chronic environmental flake (#3272/#3286), not a defect in this diff.

No blocking risk. APPROVE.
</details>

Issue #3290, opened 2026-06-16
CI testnet quickstart: cap cumulative probe wall-clock / decide whether the shard should gate at all (operator policy)


tomerweller:
Follow-up from #3286 (option **C**, flagged for operator triage — a gating-policy decision, not a pure code fix).

## Context
#3286 root cause: a resource-starved CI runner where testnet captive-core never reports synced in time; #3286's fix (option A) bounds the `capture_diagnostics()` docker calls with `timeout` so the diagnostics-capture wedge no longer hangs `run_probe`, which lets the already-correct `--soft-on-timeout` 124/137 soft-skip fire. A+B fully bound the *diagnostics* wedge and de-gate the timeout.

## Residual concern (this issue)
Even after A, a legitimately-never-syncing node still consumes roughly `4 probes × 240s + one 240s retry` of wall-clock before the soft-skip resolves the step. If that stacked budget alone still risks the testnet shard's 25-min step bound (introduced by #3289), we may want to either:
- **Cap cumulative probe wall-clock** for the testnet shard so 4× full-budget timeouts cannot stack past the step bound, and/or
- **Decide whether the testnet shard should gate at all** — i.e. soft-skip the whole "node never came up" condition (the #3280 History-Publish precedent: distinguish never-synced = environmental-soft from a genuine failure), rather than only soft-skipping the per-probe timeout disposition.

## Why this is operator-policy, not a code fix
"Should the testnet shard gate at all?" / "how much environmental degradation do we tolerate before soft-skipping the whole shard?" is a CI gating-policy decision for the operator, not something the planner should decide unilaterally. #3286 deliberately leaves it out of the fix PR. Operator should triage:
- close as won't-fix if A's diagnostics-bound + existing soft-skip is sufficient in practice,
- or schedule a follow-up PR (cumulative wall-clock cap and/or whole-shard environmental soft-skip) if the stacked-timeout budget still risks the step bound.

tomerweller:
## Closing — #3286's fix (A) is sufficient; cumulative-budget cap not needed

[...]

PR #3291, opened 2026-06-16
Bound capture_diagnostics docker calls with timeout 30 to fix testnet quickstart hang

### LINKED ISSUE #3286: CI: Quickstart testnet shard STILL hangs at step 8 despite #3273 -k fix — second hang path (orphaned docker-exec holds tee pipe)

tomerweller:
Closes #3286

## Summary

The testnet quickstart `test` job hangs to a ~55-min job cancel. The #3289 instrumentation captured the process tree on a real hung run and pinned the root cause: `capture_diagnostics()` runs **inside** `run_probe()` — after the probe's `exit_code` is captured but **before** `run_probe` returns. On a resource-starved CI runner the docker daemon goes into uninterruptible D-state, and the previously-unbounded `docker ps -a` (guarded only by `|| true`, which can never bound a syscall hang) wedges forever. `run_probe` then never returns, so the **already-correct** 124/137 `--soft-on-timeout` soft-skip (which lives in `main`, after `run_probe`) is never reached and the step runs to its budget until the job is cancelled.

This bounds each docker call in `capture_diagnostics()` with `timeout 30` (option **A**), keeping `|| true` so a timeout-induced 124 is swallowed and the probe's already-captured disposition stays byte-identical. With the diagnostics phase bounded, `run_probe` returns promptly and the existing soft-skip fires. `should_soft_skip_timeout` and the soft-skip routing are **untouched** (the **B** finding: the soft-skip was already correct; broadening it would risk masking a genuine assertion failure). Option **C** (cumulative wall-clock cap / testnet de-gate) is filed as follow-up #3290.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3286#issuecomment-4714558219)

## Test plan

- [x] `bash -n` clean on both scripts
- [x] shellcheck — no new findings in the changed regions
- [x] `bash scripts/test-quickstart-harness.sh` — 94/94 passing

- **Test:** `scripts/test-quickstart-harness.sh::test_capture_diagnostics_docker_calls_are_time_bounded` (Test 30). A fake `docker` first on PATH makes `docker ps -a` wedge (a real `sleep 999` tagged via `exec -a` with a unique sentinel for precise reaping), standing in for a D-state daemon. The wrapper runs on the retryable testnet/core,horizon shard with `--soft-on-timeout` and an always-timing-out probe, inside an outer watchdog. Asserts (a) the wrapper RETURNS (outer rc ≠ 124) and (b) the probe disposition is preserved (exit 0 + SOFT-SKIP marker).
- **Pre-fix:** committed as `7351b32d` — verified FAILED with: outer watchdog fires (rc 124), no SOFT-SKIP (`capture_diagnostics` wedges, `run_probe` never returns).
- **Post-fix:** verified PASSES after `e4ea3866` — wrapper returns, soft-skip fires, 94/94.
- `test_soft_on_timeout_still_fails_real_assertion` (the genuine-assertion-stays-red guard) is untouched and passing.

## Deviations from plan

- The regression test's outer watchdog is **90s**, not the plan's 30s. With the production `timeout 30` bound, the fixed wrapper makes two retried `capture_diagnostics` calls each capping one `docker ps -a` at ~30s (~60s total), so a 30s outer watchdog would trip on the *fixed* wrapper. 90s lets the fixed path complete while the unbounded main path (`sleep 999`) still trips it — preserving the fail-before / pass-after contract. The assertions ((a) rc ≠ 124, (b) exit 0 + SOFT-SKIP) are exactly as the plan specified.

### DIFF
--- a/scripts/ci/run-quickstart-test.sh
+++ b/scripts/ci/run-quickstart-test.sh
@@ -192,13 +192,31 @@ capture_diagnostics() {
     local attempt_dir="$DIAGNOSTICS_DIR/attempt-$attempt"
     mkdir -p "$attempt_dir"
 
-    # Docker container state (best-effort)
+    # Docker container state (best-effort).
+    #
+    # Each docker call is bounded with `timeout 30` (#3286). capture_diagnostics
+    # runs INSIDE run_probe — after the probe's exit_code is captured but BEFORE
+    # run_probe returns. On a resource-starved CI runner the docker daemon can go
+    # into uninterruptible D-state, and an unbounded `docker ps -a` then wedges
+    # forever. The `|| true` below does NOT bound such a hang — a process stuck
+    # in a syscall never reaches `|| true` — so without `timeout` the call wedges,
+    # run_probe never returns, and the 124/137 --soft-on-timeout soft-skip (which
+    # lives in main, AFTER run_probe) is never reached: the step runs to its
+    # budget and the job is cancelled (the #3286 hang signature, confirmed by the
+    # #3289 instrumentation's captured process tree). `timeout 30` lets the
+    # diagnostics phase fail-open quickly so run_probe returns and the soft-skip
+    # fires. The `|| true` is KEPT so a timeout-induced 124 from a wedged docker
+    # call is swallowed and never contaminates the probe's own exit_code (already
+    # captured before capture_diagnostics runs) — the probe disposition stays
+    # byte-identical to before this fix. This bounds only the diagnostics-capture
+    # phase; it does NOT change any pass/fail/soft-skip logic.
     if command -v docker &>/dev/null; then
-        docker ps -a > "$attempt_dir/docker-ps.txt" 2>&1 || true
-        # Capture logs from any running quickstart containers
-        for cid in $(docker ps -aq --filter "name=quickstart" 2>/dev/null || true); do
-            docker logs "$cid" > "$attempt_dir/container-$cid.log" 2>&1 || true
-            docker inspect "$cid" > "$attempt_dir/container-$cid-inspect.json" 2>&1 || true
+        timeout 30 docker ps -a > "$attempt_dir/docker-ps.txt" 2>&1 || true
+        # Capture logs from any running quickstart containers. The enumeration is
+        # bounded too — the same D-state daemon wedges `docker ps -aq` identically.
+        for cid in $(timeout 30 docker ps -aq --filter "name=quickstart" 2>/dev/null || true); do
+            timeout 30 docker logs "$cid" > "$attempt_dir/container-$cid.log" 2>&1 || true
+            timeout 30 docker inspect "$cid" > "$attempt_dir/container-$cid-inspect.json" 2>&1 || true
         done
     fi
 
--- a/scripts/test-quickstart-harness.sh
+++ b/scripts/test-quickstart-harness.sh
@@ -1735,7 +1735,9 @@ test_testnet_shard_renders_step_timeout_25_others_360() {
 # `sleep 999` (the D-state stand-in — `command -v docker` then finds the fake,
 # so the docker block in capture_diagnostics runs). It runs the wrapper on the
 # retryable testnet/core,horizon shard with --soft-on-timeout and an
-# always-timing-out probe, INSIDE an outer 30s watchdog. It asserts:
+# always-timing-out probe, INSIDE an outer watchdog (sized above the FIXED
+# wrapper's bounded diagnostics cost: two retried attempts each cap one
+# `timeout 30 docker ps -a` at ~30s, ~60s total). It asserts:
 #   (a) the wrapper RETURNS within the watchdog (outer rc != 124 — NOT killed),
 #       proving capture_diagnostics no longer wedges, AND
 #   (b) the probe disposition is preserved: wrapper exit 0 with the grep-able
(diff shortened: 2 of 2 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Surgical CI-infra fix — the four `docker` calls in `capture_diagnostics()` gain `timeout 30` with `|| true` retained; soft-skip logic and the probe `timeout -k`/PIPESTATUS line are byte-identical to `origin/main`. The regression test is non-vacuous (fails on main, passes after) and the genuine-failure guard is untouched. Harness 94/94.

<details>

Cycle 1 (no prior Correctness comment). Complete change-list — no concerns; all classes clear.

**Diff scope (`test-coverage`, `regression-risk`):**
- `scripts/ci/run-quickstart-test.sh`: full-file `diff origin/main..head` shows the ONLY change is the `capture_diagnostics()` docker block — `docker ps -a`, `docker ps -aq --filter`, `docker logs`, `docker inspect` each prefixed with `timeout 30`, `|| true` kept on all four. Plus an explanatory comment. Nothing else differs.
- `should_soft_skip_timeout()` extracted by name (awk function-scope) and `diff`'d main↔head → **byte-identical**.
- Probe execution line `timeout -k "$KILL_GRACE" "$TIMEOUT" "${PROBE_CMD[@]}" ... | tee` + `exit_code=${PIPESTATUS[0]}` present and unchanged on head (only line offset shifted by the added comment).

**Regression test verification (`kind: bug-fix`, `test-coverage`):**
- PR body has a `## Regression test` section; matches the diff (Test 30 `test_capture_diagnostics_docker_calls_are_time_bounded`).
- Ran `bash scripts/test-quickstart-harness.sh` on PR head → **94/94**, Test 30 assertions `ok 93` / `ok 94`.
- **Verified fails-before:** swapped the production wrapper for `origin/main`'s unbounded version (0 `timeout 30 docker`), re-ran → `not ok 93` (outer watchdog rc 124, capture_diagnostics wedged on `sleep 999`) and `not ok 94` (no SOFT-SKIP). Restored bounded wrapper → green again. Test genuinely captures the bug.
- The test is not vacuous: it puts a fake `docker` whose `ps -a` does `exec -a <unique-marker> sleep 999`, reaped via `pkill -f "$marker"` in a RETURN trap — no cross-run leakage.

**90s outer watchdog reasoning (`regression-risk`):** sound. Fixed wrapper makes two retried `capture_diagnostics` calls, each capping one `timeout 30 docker ps -a` (~60s total); 90s leaves headroom so the fixed path completes while the unbounded main path (999s) still trips it. A 30s watchdog would false-trip the fixed wrapper. Deviation from plan documented in PR body.

**Genuine-failure guard untouched (`regression-risk`):** `test_soft_on_timeout_still_fails_real_assertion` (Test 23, lines 1280-1301) is entirely outside both diff hunks (hunks at 1709 and 1741); only a comment reference to it appears in new Test 30. It passed in the 94/94 run. The fix does not mask real assertion failures.

**Error handling / readability:** `|| true` retained means a `timeout`-induced 124 is swallowed and never contaminates the probe's already-captured `exit_code`. Comment clearly explains the D-state mechanism. `bash -n` clean on both scripts.
</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Pure CI-infra hardening with strictly-narrowing blast radius — bounding best-effort diagnostics with `timeout 30` can only make `capture_diagnostics()` return *sooner*, never later. No Rust, no API/data-format/migration impact, no production-node behavior change. The probe disposition is provably byte-identical. Not self-modifying.

<details>

Cycle 1 (no prior Risk comment). Complete change-list — no blocking concerns.

**`regression-risk` — none.** The change only adds an upper time bound to four best-effort diagnostic shell-outs. Behaviorally:
- On a healthy runner, `docker ps -a` returns in <1s, so `timeout 30` is a no-op — identical output, identical files written.
- On a wedged daemon (the bug case), it now returns in 30s instead of never. The `|| true` swallows the resulting 124 exactly as it already swallowed real docker errors. The probe's `exit_code` was captured *before* `capture_diagnostics` runs, so diagnostics can't change pass/fail/soft-skip. Confirmed: `should_soft_skip_timeout` byte-identical, PIPESTATUS line unchanged.
- Worst realistic edge: a slow-but-not-wedged daemon taking 20-29s still completes; one taking >30s loses that one diagnostic file (acceptable for best-effort capture, and strictly better than hanging the whole CI step to a 55-min cancel).

**`perf-regression` — none / positive.** Caps a previously-unbounded wait; reduces worst-case CI step time from job-cancel-budget to bounded.

**`api-break` / `migration-risk` / `config-risk` — none.** No public surface, no schema, no config, no migrations. `timeout` is GNU coreutils, already used by the wrapper (`timeout -k` on the probe path), so no new tool dependency.

**`security` — none.** No new external input; `timeout 30 docker ...` is the same command with a bounding prefix.

**Self-modifying check:** changed paths are `scripts/ci/run-quickstart-test.sh` and `scripts/test-quickstart-harness.sh` — neither matches the self-mod patterns (`.claude/skills/{...}`, `scripts/lib/{review-pr-merge,...}.sh`, `.github/skills/shared/scripts/`). **NOT self-modifying.** No extra-scrutiny gate required.

**Operational note (non-blocking, no follow-up needed):** `timeout 30` x 4 calls x 2 retries bounds diagnostics-capture worst case at ~120s in the pathological case; negligible vs the step budget. Option C (cumulative wall-clock cap / testnet de-gate) is already filed as follow-up #3290 per the PR body — correct scoping, not this PR's job.
</details>

Issue #3218, opened 2026-06-06
monitor-tick: archive-behind cadence ACCELERATING on b7e8190b (155→63→31min over 4 episodes)


tomerweller:
## Symptom

The archive-behind near-tip recovery episodes on `b7e8190b` are **accelerating** — the inter-episode interval has halved on each recurrence. This is a worsening trajectory, distinct from the stable/occasional phenomenon characterized (and closed) in #3208 / #3213 / #3217. Filing because the *rate of change*, not any single episode, is the new and actionable signal.

**Related to #3208, #3213, #3217 (all closed)** — same archive-behind near-tip recovery mechanism; this issue tracks the cadence acceleration.

## Episode cadence on b7e8190b (this session)

| # | time (UTC) | interval since prev | recovery path | self-heal |
|---|---|---|---|---|
| 1 | 11:15 | — | escalation→catchup + peer-SCP residual (peer-ahead hard reset) | ~30s |
| 2 | 13:50 | 155 min | peer-SCP back-fill (no hard reset) | ~26s |
| 3 | 14:53 | 63 min | peer-SCP back-fill (no hard reset) | ~30s |
| 4 | ~15:24 | 31 min | peer-SCP back-fill (no hard reset) | ~30s |

Intervals: **155 → 63 → 31 min** — roughly halving each time. Every episode still self-heals in ~26-30s via `near_tip_peer_scp_recovery` (the #3204/#3205 path); `archive_behind_peer_ahead_hard_reset` has stayed flat at 31 since episode #1.

## Counter deltas (tick 64 → tick 65, ~23 min — one full episode)

| metric | t64 | t65 | delta |
|---|---|---|---|
| `henyey_recovery_stalled_tick_total{reason="forcing_catchup_behind"}` | 1699 | 2038 | +339 |
| `henyey_recovery_stalled_tick_total{reason="near_tip_peer_scp_recovery"}` | 1677 | 2011 | +334 |
| `henyey_recovery_stalled_tick_total{reason="archive_behind_peer_ahead_hard_reset"}` | 31 | 31 | +0 |
| `henyey_post_catchup_hard_reset_total` | 15 | 18 | +3 |
| `stellar_herder_lost_sync_total` | 40 | 48 | +8 |

## Assessment

[...]

Issue #3227, opened 2026-06-08
monitor-tick: [needs-decision] recalibrate memory guardrails + surface startup peak for 32 GB validator host


tomerweller:
## Symptom

The monitor-tick memory guardrails and the daily summary are calibrated for the current 61 GB host. To run a mainnet validator on a **32 GB / no-swap** box, the thresholds need recalibration and the (new) startup-peak metric needs surfacing, otherwise the operator has no early warning before a hard OOM-kill.

This is the operational follow-up to the startup-peak instrumentation issue (Phase 0). It should land *after* that metric exists.

## Evidence

Current monitor-tick HIGH-MEMORY restart rule (`.claude/skills/monitor-tick/SKILL.md:805–815`, check 4) fires only when ALL hold:
1. `RSS > 16 GB`, AND
2. system `available` (`free -m`) `< 8 GB`, AND
3. latest two `heap_components_mb` both grew > 500 MB.

On a 61 GB box, `avail < 8 GB` means RSS ≈ 53 GB — appropriate there. On a 32 GB box, `avail < 8 GB` means RSS ≈ 24 GB, which is plausibly *normal* steady state once you account for OS + page cache, so the rule would either never fire or fire constantly depending on drift. The thresholds are host-size-relative and need a 32 GB profile.

Steady state today is ~18 GB anon / ~20 GB RSS; on 32 GB / no-swap the binding risk is the **catchup peak**, not steady drift.

## Suspected root cause

Guardrail thresholds are hardcoded for the 61 GB host and there is no host-memory-aware scaling. The daily summary does not surface `startup_peak_anon_rss_mb` (it does not exist yet — see Phase 0 issue).

## Concrete fix sketch

1. **Host-aware thresholds.** Derive the HIGH-MEMORY / restart thresholds from total system RAM rather than absolute GB, or add an explicit `monitor-loop.env` knob (e.g. `MONITOR_HOST_RAM_GB`) and express the rule as fractions:
   - report-only HIGH MEMORY at `RSS > ~0.6 * host_ram`
   - restart only at `RSS > ~0.75 * host_ram` AND `avail < ~0.12 * host_ram` AND heap growing.
   Update check 4 in `.claude/skills/monitor-tick/SKILL.md:805–815` accordingly.
[...]

Issue #3215, opened 2026-06-06
monitor-tick: [needs-decision] binary-relevance allowlist forces needless restart for test-only crates/ changes


tomerweller:
## Symptom

The deploy gate's binary-relevance allowlist treats **any** path under `crates/` as rebuild-requiring, with no carve-out for **test-only** changes. A commit that touches only `#[cfg(test)]` / `mod tests` code (or `crates/*/tests/`) produces a **byte-identical release binary** (`cargo build --release` does not compile test modules), yet a literal reading of the allowlist forces a full validator rebuild + restart — pure downtime + restart risk for zero behavioral change.

## Evidence (tick 55, 2026-06-06 11:46Z)

origin/main advanced `12cd0312` → `4f4449bb` = single commit #3212 "De-flake custom_bucket_directory_survives_genesis_publish_and_restore", which closed the flaky-test #3211 I filed in tick 53.

```
$ git show 4f4449bb --stat -- crates/app/src/app/publish.rs
 crates/app/src/app/publish.rs | 30 +++++++++++++++++++++++++++++-
```

The entire diff is inside `mod tests`: a new `reopen_app_after_drop` test helper plus its call site. Commit message: *"Production App::new / acquire_db_lock stay single-shot fail-fast … Test-only; the de-flaked test is the deliverable."*

Per the allowlist (`SKILL.md` §10 step 2: "changes to … `crates/` … require a rebuild + restart"), `needs_rebuild=yes`. I **deviated** and skipped the rebuild/restart this tick, because the release binary is provably identical to the deployed `b7e8190b` and the node had just self-recovered from an archive-behind episode (tick 54) — restarting to deploy an identical binary is the exact validator-thrash the cool-down rationale exists to prevent.

**Consequence of the gap:** because I (correctly) did not rebuild, `build_sha` stays `b7e8190b`; every subsequent tick will re-diff `b7e8190b..origin/main`, still see `publish.rs`, and re-trip `needs_rebuild=yes`. The gate will report a deploy-eligible state each tick until either (a) a genuinely binary-relevant commit lands and triggers a real rebuild, or (b) this carve-out is added.

## Suspected root cause

`SKILL.md` §10 step 2 binary-relevance check + its allowlist enumerate path *prefixes* only. `crates/` is correctly rebuild-requiring for production code, but the allowlist has no way to express "this `crates/` diff is confined to test code."

## Concrete fix sketch (needs a design decision — hence not-ready)

[...]

PR #3305, opened 2026-06-16
Add (3e) stuck-but-alive SYNC FAILURE auto-restart to monitor ladder

### LINKED ISSUE #3219: monitor-tick: no restart remediation for stuck-but-alive SYNC FAILURE (lcl frozen, not a 3b wedge)
### Gap
The monitor-tick remediation ladder has **no automatic restart trigger for a stuck-but-alive SYNC FAILURE**: the node process is alive and responsive (admin/RPC answer), but `lcl` is frozen, `age` climbs unbounded, and RPC reports `unhealthy`. Existing restart triggers cover only:
- dead process (crash-recovery relaunch),
- 3a (DB-corruption wipe),
- 3b (hung/wedged — process unresponsive),
- 3c, and
- OOM (RSS thresholds).

The stuck-but-alive case slips between 3b (process *responds*, so not a wedge) and the OOM gate (RSS is well under threshold — see below). The tick currently relies on **operator/monitor manual judgment** to restart, which I have now done 4× this session (15:48, 16:01, 16:19, 16:43Z).

### Symptom signature (reproduced 4×)
- `state=Validating` but `lcl` frozen across multiple polls (≥24s)
- RPC `getHealth` → `status: "unhealthy"`, `latestLedger` not advancing
- log flood: `henyey_overlay::manager: Broadcast backpressure: messages dropped due to full peer channels msg_type="SCP_MESSAGE"` (1786/2000 tail lines)
- `henyey_app::app::consensus: Recovery stalled for too long — forcing catchup ... gap=25 urgent=true archive_behind=true` in a tight loop
- counters climbing: `forcing_catchup_behind`, `near_tip_peer_scp_recovery`

### Why the existing gates don't catch it
- **3b (wedge):** process answers HTTP, so the unresponsive-wedge detection never fires.
- **OOM gate:** RSS at stick time was 9.6–10.5GB across episodes — under the 16GB restart floor. Memory is **not** the trigger (no positive RSS-vs-episode correlation; worst stick was at highest RSS). Root cause is the overlay SCP broadcast backpressure path — tracked on #3218 (urgent).

### Proposed remediation (skill change, Blocked on design)
Add a check-4 sub-branch: when `state ∈ {Validating, Synced}` AND `lcl` unchanged across the STUCK window (>600s, or a shorter near-tip window) AND RPC `unhealthy` AND backpressure/forcing-catchup flood present → classify as **SYNC-FAILURE-stuck**, rotate `monitor.log.stuck-*`, and restart (band-aid) with a cooldown + streak escalation to urgent, *without* waiting for the OOM floor or a 3b wedge.

Blocked because: a band-aid auto-restart masks the real overlay-backpressure root cause (#3218). The restart policy should land **after or alongside** the #3218 fix, or be explicitly gated so the auto-restart streak escalates to urgent rather than silently papering over the defect. Needs operator design sign-off on the trigger thresholds + cooldown.

### Evidence
- `~/data/74535976/telemetry-3218.log` — RSS-vs-episode series
- Logs rotated `monitor.log.stuck-*` (tick 66, 68) + `.preredeploy-*`/`.stuck-*` (deploy + tick 67)
- Related: #3218 (root cause, urgent), #3205 (consensus arc), #3216 (OOM/catchup — deploy-verify FAILED for this symptom)

tomerweller:
Closes #3219

## Summary

Adds a new remediation rung **(3e) stuck-but-alive SYNC FAILURE auto-restart** to the monitor-tick ladder, driven by a pure, unit-tested `classify_stuck_alive_sync` appended to `scripts/lib/monitor-decisions.sh`. It auto-restarts a node that is alive and *responsive* (admin `/info` answers) but stuck — frozen local `lcl`, climbing RPC `age`, RPC `unhealthy`, and RSS under the OOM floor — the residual failure mode left after (3c) soft-fail-wipe and (3b) wedge. It is a band-aid for root cause #3218 (overlay SCP broadcast backpressure); the escalate-on-streak guard surfaces a restart loop as `urgent` rather than masking the unfixed defect.

`STUCK_ALIVE_SYNC=yes` fires only when ALL six hold: state matches `valid|synced|track` (case-insensitive substring — verified against the real `/info` enum in `crates/app/src/compat_http/handlers/info.rs` where both healthy states serialize to `"Synced!"`, and `"Catching up"`/`"Booting"` never match); `PROC_RESPONSIVE=yes` (the mirror of the (3b) wedge — mutually exclusive); RPC `unhealthy`; `age > 600s`; **frozen-lcl wall-clock dwell `>= 600s`** (cadence-independent, mirrors check (2)'s `<ledger>|<ts>` STUCK idiom); and `RSS < 16384 MB`. Plus a **900s cooldown** (→`cooldown`) and a **max-3-restarts/2h** guard (→`escalate`, file an urgent issue pointing at #3218). The function is the sole reader/writer of the per-session streak state file; `NOW_EPOCH` is injectable for deterministic tests. On `yes`, the rung reuses the existing Stop-PID / Rotate-log(`stuck`) / Relaunch machinery.

Strictly localized to the remediation-ladder region — **no edits to §10, check (4), or check (7)** — and append-only to the two shared files.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3219#issuecomment-4722481345)

## Test plan

- [x] `bash -n scripts/lib/monitor-decisions.sh` — clean
- [x] `bash -n scripts/test-monitor-skill-snippets.sh` — clean
- [x] `bash scripts/test-monitor-skill-snippets.sh` — all 406 pass (was 395; +11 new cases)

## New coverage (kind: feature)

11 new TAP cases in `scripts/test-monitor-skill-snippets.sh` (T63a–T63k):
- **Pre-implementation:** committed as `01d92fcc` — verified FAILED (harness aborted at `classify_stuck_alive_sync: command not found`, exit 127, function absent).
- **Post-implementation:** verified PASS after `ea0d138c` — all 406 cases green.
- Covers: full reproduced signature → `yes`; transient sync (`Catching up`) → `no`; (3b) wedge (`PROC_RESPONSIVE=no`) → `no`; healthy validating → `no`; RSS over OOM floor → `no`; dwell not met → `no`; frozen counter resets on lcl advance → `no` (state hygiene); cooldown → `cooldown`; max-restarts → `escalate`; case-insensitive state match (`Validating!`/`Tracking`) → `yes`; and the SKILL.md consistency assertion.

## Deviations from plan

None. `/info` `state` enum verified per the plan's explicit requirement: healthy/validating serialize to `"Synced!"` (matched by the `synced` substring), `"Catching up"`/`"Booting"`/`"Stopping"` never match — the gate is correct against the live enum. #3218 is referenced as the escalate target per the converged plan (the original issue noted it may be closed; the rung text also falls back to escalating to the operator).

### DIFF
--- a/.claude/skills/monitor-tick/SKILL.md
+++ b/.claude/skills/monitor-tick/SKILL.md
@@ -811,6 +811,73 @@ Board-route to Backlog: `bash .github/skills/shared/scripts/move-issue-status.sh
 Recurrence-after-fix → NEW issue, not a comment on a closed one. Known prior
 incidents: #1904, #1873, #1921, #1949.
 
+**(3e) Stuck-but-alive SYNC FAILURE auto-restart** — a node can be alive AND
+responsive (admin `/info` answers) yet make no ledger progress: a frozen local
+`lcl`, climbing RPC `age`, RPC `unhealthy`, and RSS under the OOM floor. This is
+the residual mode left after (3c) (owns the fatal-state-wipe case) and (3b)
+(owns the UNRESPONSIVE / frozen-event-loop case). (3e) is the **mirror of (3b)**:
+(3b) fires when the admin port is dead/timed-out; (3e) requires a *live,
+answering* port — mutually exclusive by construction. It is a band-aid for root
+cause #3218 (overlay SCP broadcast backpressure); the escalate-on-streak guard
+surfaces a restart loop as `urgent` rather than silently masking the defect.
+
+Evaluate (3e) **strictly last** in the alive-process path — order is
+(3c) soft-fail-wipe → (3b) wedge → **(3e) stuck-alive-sync** — so it only catches
+the residual responsive-but-frozen state. The decision is made by the pure
+function `classify_stuck_alive_sync` (`scripts/lib/monitor-decisions.sh`, already
+sourced), which is the **sole reader/writer** of the per-session streak state
+file; the caller passes only the live `lcl` and the file path.
+
+```bash
+# Skip unless the process is alive AND this is NOT the wedge case (3b owns
+# unresponsive). PROC_RESPONSIVE: admin /info answered non-empty within timeout.
+[ -z "$PID" ] && : # skip — dead-process path (3a/3d) handles this
+INFO_BODY=$(curl -s -m 3 "http://localhost:$MONITOR_ADMIN_PORT/info" 2>/dev/null)
+PROC_RESPONSIVE=$([ -n "$INFO_BODY" ] && echo yes || echo no)
+# NODE_STATE: the /info `state` field — "Synced!" (healthy/validating),
+# "Catching up", "Booting", "Stopping" (crates/app/.../handlers/info.rs).
+NODE_STATE=$(printf '%s' "$INFO_BODY" | grep -oP '"state"\s*:\s*"\K[^"]+' | head -1)
+# CURRENT_LCL: local last-closed-ledger number (from /info `ledger.num`).
+CURRENT_LCL=$(printf '%s' "$INFO_BODY" | grep -oP '"num"\s*:\s*\K[0-9]+' | head -1)
+# AGE_SEC: RPC getHealth wall-clock `age` (the check-(2) staleness signal).
+# RPC_STATUS: "healthy" | "unhealthy" from validator-mode getHealth.
+# RSS_MB: the RSS measured in check (4) (reused; OOM gate owns RSS-over-floor).
+STUCK_STATE_FILE="/home/tomer/data/$MONITOR_SESSION_ID/metrics/stuck_restart_state"
+mkdir -p "$(dirname "$STUCK_STATE_FILE")"
+classify_stuck_alive_sync "$NODE_STATE" "$CURRENT_LCL" "$AGE_SEC" "$RPC_STATUS" \
+  "$RSS_MB" "$PROC_RESPONSIVE" "$STUCK_STATE_FILE"
+```
+
+`classify_stuck_alive_sync` sets `STUCK_ALIVE_SYNC`. It is `yes` only when ALL
+six hold: state matches `valid|synced|track` (case-insensitive substring; never
+`Catching up`/`Booting`); `PROC_RESPONSIVE=yes`; `RPC_STATUS=unhealthy`;
+`AGE_SEC > 600`; frozen-lcl wall-clock dwell `>= 600s` (cadence-independent,
+mirroring check (2)'s `<ledger>|<ts>` STUCK idiom — fires at ~10 min of freeze
+regardless of tick spacing); and `RSS_MB < 16384`. Thresholds are named
+constants at the top of the function for operator retuning. Guards: a **900s
+cooldown** (`NOW - last_restart < 900` → `cooldown`, no restart this tick) and a
+**max-3-restarts / 2h** guard (`>= 3` restarts in a 7200s rolling window →
+`escalate`, no restart).
+
+- On `STUCK_ALIVE_SYNC=yes`: **Stop-PID**, **Rotate-log** with suffix `stuck`,
+  then **Relaunch** (verbatim reuse of the Common-procedures machinery). The
+  `stuck` suffix is already a recognized rotation type (Crash-recovery signal 1
+  and check (5) retention), so the next tick sees `CRASH_RECOVERY=yes` and
+  extends the sync deadline to 60m — it will not false-fire SYNC FAILURE during
+  the legitimate replay that follows the restart.
+- On `STUCK_ALIVE_SYNC=escalate`: do **NOT** restart. File a single
+  `urgent`-labeled issue (reusing the (3b)-wedge filing idiom) noting the
+  restart streak and pointing at the unfixed root cause #3218 (overlay SCP
+  broadcast backpressure / near-tip-stall recovery); if #3218 is closed,
+  escalate to the operator. A restart loop is itself the signal that the root
+  cause is unaddressed.
+- On `STUCK_ALIVE_SYNC=cooldown` or `no`: report-only, no action this tick.
+
+This trigger fires ONLY on the genuine stuck-but-alive signature and never
+during legitimate transient sync/catch-up — a false restart on a consensus
+validator is harmful. The 600s frozen-lcl dwell + six-way AND-gate + 900s
+cooldown + max-3/2h guard enforce that conservatism.
+
 **(4) Memory** — `ps -o rss= -p $(_find_session_process "$HOME/data" "/proc" "$MONITOR_SESSION_ID")`, convert to MB.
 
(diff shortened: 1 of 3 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Pure, deterministic decision function with a complete 6-way AND-gate, injectable clock, and 11 new TAP cases (406/406 green). Logic, error handling, and test coverage are all sound. No blocking concerns.

<details>

**Cycle:** 1 (no prior Correctness verdict on this PR) — complete change-list below.

**kind: feature** — verified every new public surface is tested.

- `classify_stuck_alive_sync` (the sole new function) is exercised by 11 new cases T63a–T63k covering: full signature → `yes`; `Catching up` → `no`; wedge mirror (`PROC_RESPONSIVE=no`) → `no`; healthy/age-low → `no`; RSS over floor → `no`; dwell-not-met → `no`; lcl-advance reset → `no`; cooldown; max-restarts → `escalate`; case-insensitive `Validating!`/`Tracking` → `yes`; SKILL.md consistency. All green (`bash scripts/test-monitor-skill-snippets.sh` → 406, was 395, +11).

**correctness:**
- The 6-way AND-gate short-circuits to `no` on the first failing condition and persists state on every early return via `_write_stuck_state` — so frozen-lcl bookkeeping and window-pruning are never lost regardless of which gate trips. Verified.
- Frozen-lcl bookkeeping is correct: on `current_lcl != frozen_lcl` it resets `frozen_lcl` and `frozen_since_epoch=now`; otherwise dwell accumulates as `now - frozen_since`. The `frozen_lcl_seconds < 0` clamp guards against a stale/future state file. T63g confirms the reset-on-advance hygiene.
- Cooldown (`now - last_restart < 900`) and max-restarts (`>= 3` in a 7200s pruned window) guards are evaluated only after all six conditions hold, and both persist state before returning. The window prune drops stale epochs before counting (T63i confirms `escalate`).

**error-handling:**
- Verified empty/garbage inputs fail-safe to `no`, never crash: empty `AGE_SEC`, empty `RSS_MB`, and empty `NODE_STATE` all return `STUCK_ALIVE_SYNC=no rc=0` (tested directly). Bash arithmetic treats empty as 0; the state-gate and dwell-gate catch the residual cases. A restart trigger failing toward "no restart" is the correct conservative direction.
- Non-numeric `restart_epochs` entries are filtered with `[[ "$e" =~ ^[0-9]+$ ]]` before the window math — corrupt state files can't break the prune.
- State-file write uses `2>/dev/null || true` so a transient FS error can't abort the monitor tick.

**readability (non-blocking inline-class notes, not gating):**
- `_write_stuck_state` is a nested function that leaks into the global namespace after the first call (bash has no function-local functions). Harmless here (idempotent redefinition each call, closes over the same locals), but worth a one-line comment. → inline, follow-up at most.
- `restart_count` is computed with a `for` loop incrementing a counter; a word-count would be terser, but the explicit loop is clear and correct.

No regressions: the function is append-only after `diff_is_test_only`; no existing function bodies are touched; `bash -n` and `source` both clean.
</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** The dominant risk for this change is a false auto-restart of a live consensus validator. The 6-way AND-gate + 600s wall-clock dwell + 900s cooldown + max-3/2h→escalate are genuinely conservative and the trigger cannot fire during legitimate transient sync. No restart-loop, no scope bleed. Approve.

<details>

**Cycle:** 1 (no prior Risk verdict) — complete change-list, labeled by class.

**regression-risk — false-restart during legitimate transient sync (the primary danger):** REFUTED as a risk.
- The state gate matches `valid|synced|track` (case-insensitive substring). Confirmed against `crates/app/src/compat_http/handlers/info.rs`: `Synced`/`Validating` → `"Synced!"` (match), `CatchingUp` → `"Catching up"`, `Initializing` → `"Booting"`, `ShuttingDown` → `"Stopping"` — none of the catch-up/boot/stop strings contain `valid`/`synced`/`track`, so the gate **cannot** fire mid-catchup. The PR's claim is independently re-confirmed.
- A node legitimately catching up reports `Catching up` → gate 1 fails → `no`. A node that just restarted replays under `CRASH_RECOVERY=yes` (the `stuck` rotation suffix is a recognized crash signal) which extends the sync deadline to 60m, so the post-restart replay never re-trips the trigger. Self-consistent.
- Even in the (rare) window where a synced node briefly reports `unhealthy` with stale age, the 600s frozen-lcl wall-clock dwell means the lcl must be frozen ~10 minutes before any restart — far longer than any normal SCP round or ledger-close cadence. Conservative.

**regression-risk — mutual exclusivity with (3b) wedge:** confirmed. (3b) owns `PROC_RESPONSIVE=no` (dead/timed-out admin port); (3e) requires `PROC_RESPONSIVE=yes`. The two predicates are complementary booleans on the same signal — they cannot both fire. T63c asserts the wedge case → `no`.

**regression-risk — restart loop:** confirmed bounded. The 900s cooldown blocks back-to-back restarts (T63h → `cooldown`), and ≥3 restarts in a 7200s rolling window routes to `escalate` (T63i) which does **NOT** restart — it files an `urgent` issue pointing at root cause #3218 (or escalates to operator if #3218 is closed). The escalate path is surface-only; verified in SKILL.md the `escalate` branch explicitly says "do NOT restart." No unbounded loop is reachable.

**operational-risk — RSS gate ordering:** the `RSS_MB < 16384` floor cedes the over-floor case to the OOM gate (check (4)), so (3e) and the memory guardrail don't double-act on the same tick. Sensible separation.

**migration/config-risk:** none. No config schema change; thresholds are in-function named constants. State lives in a per-session metrics file the function solely owns — no cross-session or shared-state contention.

**scope/serialization:** the SKILL.md edit is confined to the remediation-ladder region (inserted between (3b) and (4) Memory) — disjoint from §10 (#3215) and check-4/7 (#3227). `monitor-decisions.sh` change is append-only (function defined exactly once, after `diff_is_test_only`; no collision with `eval_memory_guardrail`/`classify_path_binary_relevance`/`diff_is_test_only`). NOT self-modifying per the pipeline's gated-path set (monitor-tick skill dir and monitor-decisions.sh are outside it).

**api-break / security:** none — pure shell decision function, no process I/O, no network, no new external surface.
</details>

tomerweller:
## ✅ Merged

**Commit:** 7b89fe7b4b9c6e668fb2c2a4138e2d17aaa4a4dd

**Follow-up issues filed for unaddressed inline review comments:** none (no inline review threads were left unresolved).

Triple-green: Reviewer A (Correctness) APPROVE, Reviewer B (Risk) APPROVE, CI green (38 pass / 1 skip / 0 fail). No agent CHANGES_REQUESTED findings, so no refute pass was required. Not self-modifying per the pipeline's gated-path set.

PR #3375, opened 2026-06-17
Clean up crate:work: dead accessor, dead WorkEvent re-export, select! dedup

### LINKED ISSUE #3368: crates/work: remove dead `cancel_token`/`event_tx` API and dedupe scheduler select arms
Workspace `/cleanup` audit surfaced **3 medium-severity findings** for `henyey-work`. Medium = clearly-correct, behavior-neutral cleanup; not as urgent as the high-severity batch (closed via #3319-#3331) but worth landing as a maintenance pass.

### DEAD CODE
- **Location**: `crates/work/src/types.rs:136`
- **Lane**: C
- **Evidence**: `rg --type rust 'cancel_token\(\)' /Users/tomer/dev/henyey/crates/` returns only the definition itself in types.rs:136 and one internal use of `ctx.is_cancelled()` (the sibling method). No production code, no test, and no doctest constructs or calls `.cancel_token()`. The method exists as a 'for advanced cancellation patterns' escape hatch (per its docstring) but is unused.
- **Suggested fix**: Either remove the method, or annotate with `#[allow(dead_code)]` and document the intended future use. Conservative choice: keep but downgrade visibility to `pub(crate)` if the advanced pattern is not actually needed externally.

### OVERLY BROAD VISIBILITY / UNUSED PUBLIC API
- **Location**: `crates/work/src/types.rs:146`
- **Lane**: C
- **Evidence**: `rg --type rust 'event_tx:' /Users/tomer/dev/henyey/crates/` outside this crate shows only `event_tx: None` in `crates/app/src/app/catchup_impl.rs` and `crates/historywork/tests/history_work.rs` — no one ever provides a real sender. `WorkEvent` is re-exported (lib.rs:104) but never constructed or matched on outside `scheduler.rs::emit_event`. The whole event-emit feature is effectively dead at the workspace level despite being part of the public surface.
- **Suggested fix**: Either (a) wire the event channel to a real consumer (catchup progress tracker is the obvious candidate), (b) narrow `WorkEvent` and `event_tx` to `pub(crate)` and remove the re-export until a consumer exists, or (c) document the field as intentional future API. Preferred: file a follow-up to wire it to catchup metrics; meanwhile keep the surface but note it as unused.

### DUPLICATION
- **Location**: `crates/work/src/scheduler.rs:466`
- **Lane**: C
- **Evidence**: Lines 466-507 contain two `tokio::select!` blocks that differ only by the presence of the `_ = cancel.cancelled() => { ... }` arm in the non-cancelled branch. The join_next and retry_timers arms are byte-identical. This could be a single `tokio::select!` using a biased/guarded arm such as `_ = cancel.cancelled(), if !cancel_requested => { ... }`. Suppression note: `scheduler.rs` is parity-mapped to `WorkScheduler.h/.cpp` per PARITY_STATUS.md File Mapping, and the cleanup rules suppress DUPLICATION on parity-mapped files. Flagging anyway because the duplication is a Rust-async expression detail, not a C++ structure mirror.
- **Suggested fix**: Unify into a single `tokio::select!` with a `_ = cancel.cancelled(), if !cancel_requested` guarded arm. This [PARITY+] move actually clarifies that the cancellation path is one transition, matching the upstream's single `shutdown()` entry point.

---
_Generated from the same full-workspace `/cleanup` audit that produced #3319-#3331. Run `/cleanup crates/work --mode=plan` to verify; `--mode=apply` to land._

tomerweller:
Closes #3368

## Summary

Three behavior-neutral cleanups in `henyey-work` (net runtime diff = 0):

- **F1** — delete the zero-caller `WorkContext::cancel_token()` accessor. All real uses touch the `cancel_token` field directly in `scheduler.rs` or the sibling `is_cancelled()`.
- **F2** — drop `WorkEvent` from the `lib.rs` public re-export so the dead public surface (`henyey_work::WorkEvent`) is no longer nameable from outside the crate.
- **F3** — unify the two near-identical `tokio::select!` blocks in `run_until_done_with_cancel` into one `biased` select with an `if !cancel_requested`-guarded cancel arm and a defensive `else => break`. Preserves cancel-precedence, the `cancel_all()` + `retry_timers.clear()` first-cancel side effects, and the quiescence exit.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3368#issuecomment-4732719861)

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo build --all` — clean (proves the F1/F2 visibility changes keep crate:app + crate:historywork compiling)
- [x] `cargo test -p henyey-work` — 12/12 pass, including the new guard test

## Guard test

`crates/work/tests/scheduler.rs::test_cancel_during_run_preserves_transition` (two sub-cases):
1. Cancel mid-run → the in-flight item reaches `Cancelled` while a sibling that finished before the cancel keeps `Success` (cancel-precedence + `cancel_all()` + no clobber of completed work), returning under `tokio::time::timeout`.
2. `max_concurrency = 1`, one running + one queued, cancel the token so the post-cancel loop drains `running` + the queued item.

Negative-verified: removing the `if !cancel_requested` guard makes the scheduler loop forever on the always-ready cancel arm — the guard test then hangs (test process killed at >60s) instead of passing. This pins the load-bearing guard.

## Deviations from plan

- **F2 narrowing mechanism.** The plan specified making `WorkEvent` itself `pub(crate)`. That is not viable: the `pub WorkSchedulerConfig.event_tx: Option<EventSender>` field exposes `WorkEvent` transitively, so a `pub(crate)` struct trips the `private_interfaces` lint (`-D warnings`), and the field cannot be narrowed either without breaking the external struct-literal sites (`event_tx: None` in crate:app and crate:historywork tests, which require the field to be `pub`). Resolution kept within the plan's stated intent (remove the dead *public* surface, keep behavior, keep external sites compiling): `WorkEvent` stays `pub` on the struct but is **dropped from the `lib.rs` re-export**, so `henyey_work::WorkEvent` is no longer publicly nameable. `WorkEvent` is now referenced internally via `crate::types::WorkEvent`.
- **F3 `else => break` reachability.** The plan/critic described `else => break` as guarding a *reachable* all-branches-disabled panic. In the current code it is actually **unreachable**: reaching the select with all arms disabled requires `cancel_requested && running.is_empty() && retry_timers.is_empty()` with `queue` non-empty, but the inner fill loop only leaves the queue non-empty when `running.len() >= max_concurrency` (i.e. `running` is non-empty), and a queued cancelled item is dropped by `start_work` before the quiescence check. The arm is therefore kept as **defensive** (a future `start_work`/queue change could make it reachable, and dropping it is a latent panic) per the plan's explicit instruction, but no test can force it to fire today. The guard test consequently pins the `if !cancel_requested` guard (verified) rather than the `else => break` arm.

### DIFF
--- a/crates/work/README.md
+++ b/crates/work/README.md
@@ -43,7 +43,6 @@ budget, cancellation tokens, optional events, and dependency blocking.
 | `WorkContext` | Execution context containing the work ID, attempt number, and cancellation token. |
 | `WorkOutcome` | Result of one execution attempt: success, retry, failure, or cancellation. |
 | `WorkState` | Scheduler-visible lifecycle state for a registered work item. |
-| `WorkEvent` | Optional state transition notification sent over a Tokio channel. |
 | `WorkId` | Scheduler-local `u64` identifier used for dependencies and inspection. |
 
 ## Usage
--- a/crates/work/src/lib.rs
+++ b/crates/work/src/lib.rs
@@ -101,4 +101,4 @@ mod scheduler;
 mod types;
 
 pub use scheduler::{WorkScheduler, WorkSchedulerConfig, WorkSchedulerMetrics};
-pub use types::{Work, WorkContext, WorkEvent, WorkId, WorkOutcome, WorkState};
+pub use types::{Work, WorkContext, WorkId, WorkOutcome, WorkState};
--- a/crates/work/src/scheduler.rs
+++ b/crates/work/src/scheduler.rs
@@ -10,8 +10,8 @@ use tracing::{debug, error, info, warn};
 
 use futures::StreamExt as _;
 
-use crate::types::EventSender;
-use crate::{Work, WorkContext, WorkEvent, WorkId, WorkOutcome, WorkState};
+use crate::types::{EventSender, WorkEvent};
+use crate::{Work, WorkContext, WorkId, WorkOutcome, WorkState};
 
 /// Configuration for the work scheduler.
 ///
@@ -463,47 +463,39 @@ impl WorkScheduler {
 
             // Poll JoinSet, retry timers, and cancellation concurrently.
             // This ensures retry delays never block completion handling.
-            if cancel_requested {
-                tokio::select! {
-                    result = join_set.join_next(), if !running.is_empty() => {
-                        let Some(join_result) = result else { break };
-                        self.handle_join_result(
-                            join_result, &mut running, &mut queue, &mut retry_timers,
-                        );
-                    }
-                    expired = retry_timers.next(), if !retry_timers.is_empty() => {
-                        if let Some(expired) = expired {
-                            let id = expired.into_inner();
-                            if self.state_or_pending(id) == WorkState::Pending {
-                                queue.push(id);
-                            }
-                        }
-                    }
-                    else => break,
+            //
+            // `biased;` polls the arms top-down, keeping cancellation
+            // highest-priority ("cancel takes precedence"). The cancel arm is
+            // guarded by `if !cancel_requested`, so once cancellation has been
+            // observed it is disabled. The `else => break` arm is therefore
+            // reachable only post-cancel — when the cancel arm is disabled and
+            // both remaining arms are also disabled (`running` and
+            // `retry_timers` both empty). Without it, `tokio::select!` panics
+            // when every branch is disabled. Do NOT drop it.
+            tokio::select! {
+                biased;
+
+                _ = cancel.cancelled(), if !cancel_requested => {
+                    cancel_requested = true;
+                    self.cancel_all();
+                    retry_timers.clear();
+                    continue;
                 }
-            } else {
-                tokio::select! {
-                    _ = cancel.cancelled() => {
-                        cancel_requested = true;
-                        self.cancel_all();
-                        retry_timers.clear();
-                        continue;
-                    }
-                    result = join_set.join_next(), if !running.is_empty() => {
-                        let Some(join_result) = result else { break };
-                        self.handle_join_result(
-                            join_result, &mut running, &mut queue, &mut retry_timers,
-                        );
-                    }
-                    expired = retry_timers.next(), if !retry_timers.is_empty() => {
-                        if let Some(expired) = expired {
-                            let id = expired.into_inner();
-                            if self.state_or_pending(id) == WorkState::Pending {
-                                queue.push(id);
-                            }
+                result = join_set.join_next(), if !running.is_empty() => {
+                    let Some(join_result) = result else { break };
+                    self.handle_join_result(
+                        join_result, &mut running, &mut queue, &mut retry_timers,
+                    );
+                }
+                expired = retry_timers.next(), if !retry_timers.is_empty() => {
+                    if let Some(expired) = expired {
+                        let id = expired.into_inner();
+                        if self.state_or_pending(id) == WorkState::Pending {
+                            queue.push(id);
                         }
                     }
                 }
+                else => break,
             }
         }
 
--- a/crates/work/src/types.rs
+++ b/crates/work/src/types.rs
@@ -128,14 +128,6 @@ impl WorkContext {
     pub fn is_cancelled(&self) -> bool {
         self.cancel_token.is_cancelled()
     }
-
-    /// Returns a reference to the cancellation token.
-    ///
-    /// This can be used for more advanced cancellation patterns, such as
-    /// passing the token to async operations that support it directly.
-    pub fn cancel_token(&self) -> &CancellationToken {
-        &self.cancel_token
-    }
 }
 
 /// Event emitted by the scheduler when work state changes.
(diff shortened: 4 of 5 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Three behavior-neutral crate:work cleanups (F1 dead accessor, F2 dead re-export, F3 select! dedup) plus a load-bearing guard test. Build/test/clippy all green on a worktree off current origin/main; the guard test is negative-verified.

<details>

**Cycle:** 1 (no prior Correctness comment on this PR). Complete change-list below — no concerns survive; APPROVE.

`kind: refactor` (from #3368 Triage Report). Per the refactor contract: existing tests must still pass + the merged form must be behavior-neutral. Verified directly:

- **Build** (`test-coverage` / `regression-risk`): `cargo build --all` clean (2m31s) on `origin/do/issue-3368` merged with current `origin/main`. This is the load-bearing proof for F1/F2 — `henyey-app` and `henyey-historywork` (the two external `event_tx: None` sites) compile after dropping the `WorkEvent` re-export and the `cancel_token()` accessor.
- **Tests**: `cargo test -p henyey-work` → 12/12 pass incl. the new `test_cancel_during_run_preserves_transition` and 4 ignored doctests. Finished in 0.10s — no hangs.
- **Clippy**: `cargo clippy --all -- -D warnings` clean (no `private_interfaces` trip, confirming the F2 deviation rationale).
- **F1** (`api-shape`): `git grep cancel_token()` on `origin/main` returns zero callers — the accessor is genuinely dead. The `cancel_token` *field* remains used directly inside scheduler.rs (lines 360/570/588/798) + the sibling `is_cancelled()`. Deletion safe.
- **F2** (`api-shape`): no consuming crate names `WorkEvent` (empty grep across app/historywork/ledger/herder/history). README row removed to match. The deviation (keep struct `pub`, drop re-export) is sound and documented.
- **Guard test** (`test-coverage`): negative-verified myself — stripping `if !cancel_requested` makes the test hang >60s (timeout-killed), proving it pins the guard. Sub-case 1 pins cancel-precedence + no-clobber-of-completed-work; sub-case 2 exercises the post-cancel drain through `else => break`. Returns under `tokio::time::timeout`, so a regression manifests as a hang the test catches.

Logic, error handling, readability: the merged `biased` select is clearer than the duplicated two-block form; the inline comment block accurately documents the guard + `else` reachability. No concerns.

**Self-modifying:** No (only `crates/work/` paths).
</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** crate:work is parity-adjacent (the async scheduler drives catchup/history work). Reviewed F3 deeply for behavioral equivalence against the pre-PR two-block form: net runtime diff = 0, with one strictly-safe strengthening (`biased;` cancel-precedence). All first-cancel side effects and the quiescence exit are preserved verbatim.

<details>

**Cycle:** 1 (no prior Parity comment). Complete change-list; no concerns. Lens = behavioral/correctness-deep on the scheduler state machine (crate:work has no direct stellar-core C++ counterpart — it's a henyey-internal async driver — so "parity" here means *no observable behavior change* in the consensus/catchup-driving scheduler).

**F3 equivalence — verified by direct diff of `origin/main` (two blocks) vs merged form:**

- **State A (`!cancel_requested`):** original `else`-branch arms = {cancel, join `if !running.is_empty()`, timer `if !retry_timers.is_empty()`}, no `else`. Merged form with guard true = same three arms enabled + `else => break`. The cancel arm (`cancel.cancelled()`) is an always-pollable future, so it's never simultaneously disabled with the others → the `else` arm is unreachable in State A, exactly matching the original's no-`else` form. **Equivalent.**
- **State B (`cancel_requested`):** original `if`-branch arms = {join, timer, `else => break`}, no cancel arm. Merged form with guard false disables the cancel arm → identical {join, timer, `else => break`}. **Equivalent**, including the quiescence exit.
- **Cancel side effects:** `cancel_requested = true; self.cancel_all(); retry_timers.clear(); continue;` — byte-identical between original non-cancel branch and merged cancel arm. The pre-loop `if !cancel_requested && cancel.is_cancelled()` head-start check (lines 441-446) is untouched.
- **`biased;` — the one intentional change (`sequencing`):** the original non-cancel `select!` had no `biased`, so when a completion and a cancel were both ready, the choice was pseudo-random. The merged form makes cancel strictly highest-priority. This is a *strengthening* of the documented "cancel takes precedence" contract, consistent with the existing pre-loop cancel head-start — not a regression. No path observes a different terminal `WorkState` as a result.

**`else => break` reachability (deviation 2, `edge-case`):** I independently agree it is unreachable in current code (queue stays non-empty only while `running.len() >= max_concurrency`, i.e. `running` non-empty; a queued cancelled item is dropped by `start_work` before the quiescence check). But it must stay: in State B with the cancel arm disabled, an all-arms-disabled `select!` *panics*. It guards a latent panic against future `start_work`/queue changes, was present in the original cancel-branch, and is net-neutral. Acceptable as a dead defensive arm — not a behavior change.

**Guard test:** negative-verified (removing `if !cancel_requested` → hang). Genuinely pins the load-bearing guard.

No parity/sequencing/edge-case concerns.
</details>

tomerweller:
## ✅ Merged

**Commit:** 33cff4aaf65b97477358225a20eb87fa836aacea

Triple-green: Reviewer A (Correctness) APPROVE, Reviewer B (Parity-deep) APPROVE, required CI green. Merged with the non-required Quickstart test shards (local/testnet) still pending per the de-gated-flake policy (#3273). No inline review threads to preserve.

PR #3417, opened 2026-06-17
Remove dead herder cleanup: stale doc bullet, time_call re-export, spawn shim

tomerweller:
Closes #3392

## Summary

Behavior-preserving `crate:herder` low-cleanup — pure subtraction (6 insertions / 19 deletions), zero observable behavior change. Three findings:

- **F1 (docs):** Delete the stale `tx_broadcast` bullet from the `# Modules` docstring in `lib.rs`. The module was already removed (removal note lives a few lines below) and no `tx_broadcast*` file exists.
- **F2 (dead code):** Drop the genuinely-dead `time_call` re-export from `tracked_lock.rs` (zero references in lib or `--tests`) and remove the now-unnecessary `#[allow(unused_imports)]`. `PhaseTimer` (consumed at `herder.rs`) and `LOCK_SLOW_THRESHOLD` (11 in-module consumers) are **kept** — the audit's literal "reduce to just PhaseTimer" would not compile. All lock-timing / `WARN lock=…` instrumentation is byte-identical.
- **F3 (vestigial shim, cross-crate):** Delete the `spawn.rs` re-export shim and its `mod`/`pub use` in `lib.rs`, repointing all 3 callers — `herder.rs` ×2 (in-crate) and `crates/app/src/app/ledger_close.rs` ×1 (app crate) — directly at the identical `henyey_common` functions. The app edit also aligns the lone outlier with the prevailing `henyey_common::*` style.

Instrumentation / plumbing / doc only — no consensus behavior change vs stellar-core v26.0.1. `LOCK_SLOW_THRESHOLD` and every `WARN lock=…` emission site are untouched, so consensus timing/logging is byte-identical.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3392#issuecomment-4735614568)

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy --all -- -D warnings` — clean across the whole workspace (proves the shim deletion + repoints are complete)
- [x] `cargo build --all` under `-Dwarnings` — clean
- [x] `cargo test -p henyey-herder` — 1185 lib tests pass (incl. `tracked_lock` unit tests that would fail to compile if `LOCK_SLOW_THRESHOLD` were dropped) + integration/doc tests pass

## Deviations from plan

None. (Net diff is 6/-19 vs the plan's stated 10/-23 — the `tracked_lock.rs` comment rewrite is slightly tighter; functionally identical.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3418, opened 2026-06-17
Dedup Stage-E archive counter emits behind emit_archive_counter helper

### LINKED ISSUE #3393: crates/historywork: avoid header clone-out-of-mutex and dedupe Stage E counter boilerplate
Workspace `/cleanup` audit surfaced **2 low-severity findings** for `henyey-historywork`. Low = touch-up-when-nearby quality cleanup; not blocking, but worth landing the next time someone is in the file. This issue is intentionally lower priority than the high (#3319-#3331) and medium (#3352-#3368) batches; don't dedicate a planning cycle to it — fold into another change or `/cleanup crates/historywork --mode=apply` when convenient.

### RUST API ANTI-PATTERN (allocation)
- **Location**: `crates/historywork/src/download.rs:456:456`
- **Lane**: B
- **Evidence**: In `DownloadTransactionsWork::run` (line 456-459) and `DownloadTxResultsWork::run` (line 530-533), the code does `let headers = { let guard = self.state.lock().await; guard.headers.clone() };` cloning ~64 LedgerHeaderHistoryEntry values, then only uses `&headers` for read-only `find_header` lookups inside subsequent sync verify calls. The clone is unnecessary because (a) verify_tx_set / verify_tx_result_set are sync — holding the MutexGuard across them does not cross an `.await`, and (b) downstream work items wait for this one to complete, so no contention.
- **Suggested fix**: Acquire the guard at the top, run verification against `&guard.headers` while still holding the lock (no .await between lock and finish), then write `guard.transactions = entries;` and return. This drops the 64-entry header clone per checkpoint while preserving observable behavior.

### DUPLICATION (within crate)
- **Location**: `crates/historywork/src/download.rs:267:267`
- **Lane**: C
- **Evidence**: Each download work item repeats the same shape: `metrics::counter!("stellar_history_<event>_total", "archive" => archive_name.clone()).increment(1);` — at lines 75, 84, 90, 100, 273, 278, 288, 294, 371, 381, 388, 447, 463, 481, 556, 568, 583, 595, 604. Per-bucket success emits both download and verify counters in a tight loop (lines 273-281), each cloning `archive_name`. A small helper like `fn emit_archive_counter(name: &'static str, archive: &str)` would consolidate the .clone() boilerplate and reduce string allocations to a single `Cow`/`Arc<str>` per work item.
- **Suggested fix**: Introduce a small helper `fn emit_archive_counter(name: &'static str, archive: &str) { metrics::counter!(name, "archive" => archive.to_owned()).increment(1); }` and have callers pass `&archive_name`. Reduces ~18 sites of `archive_name.clone()` to single borrowed args.

---
_Generated from the same full-workspace `/cleanup` audit that produced #3319-#3368._

tomerweller:
Closes #3393

## Summary

Dedupes the 19 archive-labeled `metrics::counter!` emit sites in `crates/historywork/src/download.rs` behind one new `emit_archive_counter(metric: &'static str, archive_name: &str)` helper. The helper performs the `.to_owned()` internally (matching the `metrics` by-value label requirement), so the 5 sites that previously did `archive_name.clone()` collapse to `&archive_name` borrows. Net diff is negative for the source code. This is behavior-neutral: the same metric names, the same `"archive"` label key/value, and the same increment counts are emitted as before — only the textual factoring changed.

This is the Finding-2-only deliverable from the converged plan.

## Finding 1 — no code change (scoped out)

The plan scoped out Finding 1 (the header `headers.clone()` under the mutex in `DownloadTransactionsWork::run` / `DownloadTxResultsWork::run`). On current `origin/main` the clone is already outside the critical section: the `MutexGuard` drops at the close of the `let headers = { let guard = ...; guard.headers.clone() };` block, with no `.await` between lock and drop, and the entire verify loop runs lock-free. Removing the clone entirely would either lengthen the lock hold (hold across the verify loop) or require a non-behavior-neutral DAG data-flow refactor to thread `headers` ownership out of the shared state. Per "prefer correctness over the micro-optimization", the clone stays — no edit for Finding 1.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo build --all`
- [x] `cargo test -p henyey-historywork` (6 lib + 3 integration tests pass, incl. the rewritten label test)

## Test change

The existing `download.rs::tests::test_stage_e_historywork_archive_label_present` scanned a 200-byte window after each metric literal for `"archive"`, which breaks once the label moves into the helper. It is rewritten (non-vacuously) to assert structurally that:
- each Stage-E metric literal is the first argument of an `emit_archive_counter(` call,
- the `"archive"` label key lives in the helper body, and
- no bare `metrics::counter!(` carrying a Stage-E metric survives outside the helper,

with a `found_any` guard so it cannot pass vacuously. Verified non-vacuous by temporarily reverting one emit site to a bare `metrics::counter!` — the rewritten test fails as expected. The companion `test_stage_e_historywork_metric_literals_present` is unchanged (literals remain at call sites as the helper's first arg).

## Deviations from plan

None.

### DIFF
--- a/crates/historywork/src/download.rs
+++ b/crates/historywork/src/download.rs
@@ -66,36 +66,32 @@ impl Work for GetHistoryArchiveStateWork {
         match self.archive.fetch_checkpoint_has(self.checkpoint).await {
             Ok(has) => {
                 if let Err(err) = verify::verify_has_structure(&has) {
-                    metrics::counter!(
+                    emit_archive_counter(
                         "stellar_history_download_history_archive_state_failure_total",
-                        "archive" => archive_name,
-                    )
-                    .increment(1);
+                        &archive_name,
+                    );
                     return WorkOutcome::Failed(format!("HAS structure invalid: {err}"));
                 }
                 if let Err(err) = verify::verify_has_checkpoint(&has, self.checkpoint) {
-                    metrics::counter!(
+                    emit_archive_counter(
                         "stellar_history_download_history_archive_state_failure_total",
-                        "archive" => archive_name,
-                    )
-                    .increment(1);
+                        &archive_name,
+                    );
                     return WorkOutcome::Failed(format!("HAS checkpoint mismatch: {err}"));
                 }
-                metrics::counter!(
+                emit_archive_counter(
                     "stellar_history_download_history_archive_state_success_total",
-                    "archive" => archive_name,
-                )
-                .increment(1);
+                    &archive_name,
+                );
                 let mut guard = self.state.lock().await;
                 guard.has = Some(has);
                 WorkOutcome::Success
             }
             Err(err) => {
-                metrics::counter!(
+                emit_archive_counter(
                     "stellar_history_download_history_archive_state_failure_total",
-                    "archive" => archive_name,
-                )
-                .increment(1);
+                    &archive_name,
+                );
                 WorkOutcome::Failed(format!("failed to fetch HAS: {err}"))
             }
         }
@@ -263,33 +259,29 @@ impl Work for DownloadBucketsWork {
             for result in results {
                 match result {
                     Ok(()) => {
-                        metrics::counter!(
+                        emit_archive_counter(
                             "stellar_history_download_bucket_success_total",
-                            "archive" => archive_name.clone(),
-                        )
-                        .increment(1);
-                        metrics::counter!(
+                            &archive_name,
+                        );
+                        emit_archive_counter(
                             "stellar_history_verify_bucket_success_total",
-                            "archive" => archive_name.clone(),
-                        )
-                        .increment(1);
+                            &archive_name,
+                        );
                     }
                     Err(failure) => {
                         match &failure {
                             BucketDownloadFailure::Download(_)
                             | BucketDownloadFailure::Persist(_) => {
-                                metrics::counter!(
+                                emit_archive_counter(
                                     "stellar_history_download_bucket_failure_total",
-                                    "archive" => archive_name.clone(),
-                                )
-                                .increment(1);
+                                    &archive_name,
+                                );
                             }
                             BucketDownloadFailure::Verify(_) => {
-                                metrics::counter!(
+                                emit_archive_counter(
                                     "stellar_history_verify_bucket_failure_total",
-                                    "archive" => archive_name.clone(),
-                                )
-                                .increment(1);
+                                    &archive_name,
+                                );
                             }
                         }
                         if first_failure.is_none() {
@@ -362,28 +354,25 @@ impl Work for DownloadLedgerHeadersWork {
         let headers = match self.archive.fetch_ledger_headers(self.checkpoint).await {
             Ok(headers) => headers,
             Err(err) => {
-                metrics::counter!(
+                emit_archive_counter(
                     "stellar_history_download_ledger_failure_total",
-                    "archive" => archive_name,
-                )
-                .increment(1);
+                    &archive_name,
+                );
                 return WorkOutcome::Failed(format!("failed to download headers: {err}"));
             }
         };
 
         if let Err(err) = verify::verify_header_chain_from_entries(&headers) {
-            metrics::counter!(
+            emit_archive_counter(
                 "stellar_history_verify_ledger_chain_failure_total",
-                "archive" => archive_name.clone(),
-            )
-            .increment(1);
+                &archive_name,
+            );
             return WorkOutcome::Failed(format!("header chain verification failed: {err}"));
         }
-        metrics::counter!(
+        emit_archive_counter(
             "stellar_history_verify_ledger_chain_success_total",
-            "archive" => archive_name,
-        )
-        .increment(1);
+            &archive_name,
+        );
 
         let mut guard = self.state.lock().await;
         guard.headers = headers;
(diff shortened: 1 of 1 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Clean, behavior-neutral refactor: 19 archive-labeled Stage-E `metrics::counter!` emit sites in `crates/historywork/src/download.rs` collapse behind one `emit_archive_counter` helper. Build + clippy + the full historywork test suite (incl. the rewritten label test) are green on the PR head merged with current `origin/main`. The rewritten test is non-vacuous.

<details>

Linked issue #3393 is `kind: refactor` (dedup duplicated counter emits). Verified against current `origin/main` (PR head merged cleanly; main now includes #3417).

**Behavior-neutrality (the core property):**
- Helper body: `metrics::counter!(metric, "archive" => archive_name.to_owned()).increment(1)`. `metric: &'static str` is passed straight through, so each of the 10 distinct Stage-E metric names is emitted unchanged; the `"archive"` label key is unchanged; the value is `archive_name.to_owned()` — identical to the prior direct `archive_name` / `archive_name.clone()` value; `.increment(1)` unchanged. No metric dropped, renamed, or relabeled.
- Emit-site count: 19 (sum of metric-literal occurrences in main code: state 3+1, bucket 1+1+1+1, ledger 8+1, chain 1+1). Matches the PR claim exactly. The lone real `metrics::counter!` invocation is now line 662 (the helper); all other matches are doc-comment/test text.
- The 5 `archive_name.clone()` → `&archive_name` collapse is a caller-side ownership change only — the owning `.to_owned()` moved into the helper. Net negative source diff.

**Test verification (refactor → existing tests must still pass + the rewritten test must be sound):**
- `cargo build --all` green; `cargo clippy --all -- -D warnings` green; `cargo test -p henyey-historywork` green (6 lib + 1 integration + doctests). `test_stage_e_historywork_metric_literals_present` unchanged and passing.
- `test_stage_e_historywork_archive_label_present` rewritten. It asserts three structural facts on the pre-`#[cfg(test)]` slice: (a) `"archive" => archive_name.to_owned()` lives in the helper; (b) every Stage-E metric literal is the first arg of an `emit_archive_counter(` call (backward whitespace-trim to the opener token); (c) no bare `metrics::counter!(` window carries a Stage-E literal. A per-metric `found_any` guard fires if any literal is missing, so it cannot pass vacuously on a dropped/renamed metric.
- **Non-vacuity reconfirmed independently:** reverted the `download_ledger_success_total` site to a bare `metrics::counter!(...)`; the test FAILED with `metric literal "stellar_history_download_ledger_success_total" is not the first argument of an emit_archive_counter(...) call`. Restored; test green again.

**Finding 1 (header clone) correctly left as-is:** at `DownloadTransactionsWork::run` (L438-441) and `DownloadTxResultsWork::run` (L510-512) the `guard.headers.clone()` is inside a scoped block whose `MutexGuard` drops at the close brace, with no `.await` between lock acquisition and drop; the verify loop runs lock-free. The clone is already outside the critical section — removing it would either lengthen the lock hold or require a non-neutral data-flow refactor. PR body rationale is accurate; no code change is correct here.

Readability/error-handling: helper is `#[inline]`, documented, single responsibility; no error paths changed.
</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Minimal-risk internal refactor confined to `crates/historywork/src/download.rs`. No public API, data format, config, or metric-contract change; the emitted Prometheus metric names, the `"archive"` label, and increment semantics are byte-for-byte preserved. No regression, perf, or operational risk identified.

<details>

Risk lens (non-parity — only `crates/historywork/` touched, which is catchup-download but not in the parity-critical prefix set). Reviewed against the PR head merged with current `origin/main`.

**API-break risk — none.** `emit_archive_counter` is a private `fn` (no `pub`/`pub(crate)`); a repo-wide grep finds zero references outside `download.rs`. No exported surface changes; nothing downstream can break.

**Observability/metric-contract risk — none.** The 19 emit sites keep their exact `&'static str` metric names and the `"archive"` label key. The same metric names are declared in `crates/app/src/metrics.rs` (the registry: `HISTORY_VERIFY_BUCKET_SUCCESS_TOTAL`, `HISTORY_DOWNLOAD_LEDGER_SUCCESS_TOTAL`, etc.) and referenced in `crates/history/src/catchup/buckets.rs` — **none of which this PR touches.** So the Prometheus scrape/registry contract and any dashboards/alerts keyed on these series are unaffected. Label cardinality (one `archive` dimension) is unchanged.

**Regression risk — none.** Pure textual factoring; control flow, `WorkOutcome` return paths, and the failure-classification match arms are byte-identical aside from the macro→helper substitution. All existing historywork tests pass.

**Perf risk — negligible/neutral.** `#[inline]` helper; the `.to_owned()` allocation that previously happened at each call site (directly or via `.clone()`) now happens once inside the helper — same number of allocations, same hot-path behavior. These emits fire on per-checkpoint download/verify outcomes, not in any tight inner loop.

**Migration/config risk — none.** No config keys, no schema, no on-disk format, no CLI surface touched.

**Finding 1 left as-is — correct, low-risk.** Not removing the `headers.clone()` avoids lengthening a mutex hold across the verify loop; the existing scoped-guard pattern already drops the lock before the loop. Leaving it is the conservative, behavior-neutral choice.
</details>

tomerweller:
## ✅ Merged

**Commit:** f5cba2cbe2c38c43dc13a2e9f8833dd5925783a4

Issue #3419, opened 2026-06-17
Overlay: inbound authentication broken on mainnet after 2d24c6f6 (suspect #45a97265 SSC interop fix) — 0 inbound peers


tomerweller:
## Symptom

After deploying `2d24c6f6` (at 21:05Z), the mainnet validator accepts **zero inbound authenticated peers**. Sustained 44+ min across two monitor ticks:

```
stellar_overlay_inbound_attempt_total      111
stellar_overlay_inbound_establish_total      9
stellar_overlay_inbound_reject_total       102
stellar_overlay_inbound_drop_total           9
stellar_overlay_inbound_authenticated        0   <-- broken
stellar_overlay_outbound_authenticated      20
```

Inbound peers attempt (111) and a few establish (9), but **none complete authentication** (0), and all establishes drop. `max_inbound_peers=64` (no limit pressure). The `inbound-auth-low` alarm (threshold 3) fires.

## Correlation / root-cause suspect

- On the **previous binary `d2fd2202`** (running 18:53–21:05Z, identical config/port), inbound was healthy: peer_count 31–33 (inbound + outbound), `inbound-auth-low` never fired across 6 ticks.
- Immediately after the `2d24c6f6` deploy, inbound authenticated dropped to **0** and stayed there (peer_count fell to 20 = outbound-only).
- Config is unchanged across the transition → the regression is in the **binary**.
- The only runtime overlay change in `d2fd2202..2d24c6f6` is **`45a97265` "Fix critical SSC overlay interoperability bugs found during integration testing"** (the other overlay commit, #3405, is test-only). An SSC (Supercluster) interop fix plausibly altered the handshake/auth path in a way that **breaks authentication with real mainnet peers** (classic interop-fix-breaks-mainnet).

## Impact / severity

- **Consensus is currently healthy** — node is Validating at tip, `heard_from_quorum=true`, `is_v_blocking=true`, `peer_gap=0`, carried by 20 outbound peers. NOT validator-blocking → filed non-urgent.
- However, a validator accepting **0 inbound** is a degraded network participant: no other node can peer *to* it, reducing network topology resilience and the validator's contribution to quorum reachability. The `inbound-auth-low` alarm notes: "Healthy fleet has 50+ inbound. Aggregate peer_count can be >=8 from outbound while inbound starves consensus."

## Suggested investigation / fix

[...]

PR #3420, opened 2026-06-18
Refs #3419: add inbound-peer drop diagnostics to overlay

tomerweller:
Refs #3419

**Diagnostic instrumentation only — not a fix.** Per the operator's decision on #3419, this adds observability to the live mainnet validator to capture the real inbound-auth failure mode. The prior investigation (see issue comments) **exonerated the suspect commit `45a97265`**: henyey's inbound auth path provably works (471 overlay lib tests pass; `inbound_establish` increments; the gauge climbs slowly over hours), and the symptom is that non-tier1 remotes reset henyey-as-acceptor shortly after a successful handshake. This PR adds the instrumentation that investigation recommended to see *why*.

## Observability-only — zero behavior change

No change to handshake/auth sequencing, MAC/codec semantics, message framing, dispatch, or any control flow. All additions are:
- A cheap additive per-connection field (last-sent message wire-type) and a read-only accessor.
- `tracing` events (warn-level) at existing drop sites.
- One new monotonic counter.

The existing 471 overlay lib tests pass unchanged precisely because no behavior changed.

## What was instrumented

1. **Last message before reset.** `Peer` now tracks `last_sent_msg_type` (the wire name of the most-recent `StellarMessage` sent), updated on every send path (handshake `send_raw`/`send_auth` and post-auth `send`). On an **inbound** peer drop/reset in the peer loop — recv error (e.g. "Connection reset by peer" / errno 104), remote close, or send error — we emit `warn!(target: "overlay_inbound_diag", ...)` capturing: peer id, addr, direction (inbound), authenticated state at drop, **last-sent message type**, messages sent/received, and the drop reason + errno/detail. Gated to inbound only (the #3419 symptom) so it does not spam outbound churn.

2. **Per-direction authenticated-over-time signal.**
   - A `warn!(target: "overlay_inbound_diag", event="authenticated")` fires when an inbound peer transitions to authenticated (timestamped transition signal).
   - New monotonic counter `stellar_overlay_inbound_authenticated_total`, incremented at the inbound establish/register point. The existing `stellar_overlay_inbound_authenticated` is an instantaneous gauge; this counter lets an operator distinguish **"never authenticated"** (stays 0) from **"authenticated then churned"** (climbs while the gauge stays near 0).

## What to look for after redeploy

The operator will redeploy this build to mainnet to capture data. In the logs/metrics, look for:
- `overlay_inbound_diag` warn lines with `drop_reason=recv_error` / `remote_closed` and the `last_sent_msg_type` field — **this is the last message henyey sent the acceptor before the remote reset it**. The prior investigation flagged the post-auth `SEND_MORE_EXTENDED` grant / first flood/SCP reply as a candidate; this confirms (or refutes) it directly.
- `stellar_overlay_inbound_authenticated_total` climbing while `stellar_overlay_inbound_authenticated` (gauge) stays near 0 → confirms "authenticated then churned" rather than "never authenticated".
- Compare `messages_sent`/`messages_received` at drop to see how far into the session the reset happens (prior investigation saw ~15-17 msgs post-auth).

## Test plan

- [x] `cargo build -p henyey-overlay`
- [x] `cargo build --all`
- [x] `cargo test -p henyey-overlay` — all pass (471 lib + integration + doctests; behavior unchanged)
- [x] `cargo fmt --check` — clean

## Deviations from plan

None. This issue stays **open** — it is diagnostic instrumentation, not a fix, so the PR intentionally does not `Closes` it.

### DIFF
--- a/crates/app/src/metrics.rs
+++ b/crates/app/src/metrics.rs
@@ -627,6 +627,11 @@ metric_catalog! {
             => "Total inbound peer disconnections (run_peer_loop returned)";
         OVERLAY_INBOUND_REJECT_TOTAL = "stellar_overlay_inbound_reject_total"
             => "Total inbound connections rejected before establishment";
+        // #3419 diagnostic: monotonic count of inbound peers that reached
+        // authenticated state. Climbing while the `inbound_authenticated` gauge
+        // stays near 0 means "authenticated then churned", not "never authenticated".
+        OVERLAY_INBOUND_AUTHENTICATED_TOTAL = "stellar_overlay_inbound_authenticated_total"
+            => "Total inbound peers that transitioned to authenticated (monotonic)";
         OVERLAY_OUTBOUND_ATTEMPT_TOTAL = "stellar_overlay_outbound_attempt_total"
             => "Total outbound connection attempts (dial initiated)";
         OVERLAY_OUTBOUND_ESTABLISH_TOTAL = "stellar_overlay_outbound_establish_total"
@@ -1067,6 +1072,7 @@ pub(crate) async fn refresh_gauges(state: &ServerState) {
         OVERLAY_INBOUND_ESTABLISH_TOTAL.absolute(ov.inbound_establish);
         OVERLAY_INBOUND_DROP_TOTAL.absolute(ov.inbound_drop);
         OVERLAY_INBOUND_REJECT_TOTAL.absolute(ov.inbound_reject);
+        OVERLAY_INBOUND_AUTHENTICATED_TOTAL.absolute(ov.inbound_authenticated_total);
         OVERLAY_OUTBOUND_ATTEMPT_TOTAL.absolute(ov.outbound_attempt);
         OVERLAY_OUTBOUND_ESTABLISH_TOTAL.absolute(ov.outbound_establish);
         OVERLAY_OUTBOUND_DROP_TOTAL.absolute(ov.outbound_drop);
--- a/crates/overlay/src/manager/connection.rs
+++ b/crates/overlay/src/manager/connection.rs
@@ -481,6 +481,19 @@ impl OverlayManager {
             shared.pending_connections.release_peer_id(&peer_id);
         }
         shared.metrics.inbound_establish.inc();
+        // #3419 diagnostic (observability-only): an inbound peer transitioned to
+        // authenticated. The monotonic counter lets an operator distinguish
+        // "never authenticated" (stays 0) from "authenticated then churned"
+        // (climbs while the instantaneous gauge stays near 0). The warn event
+        // gives a timestamped, per-direction authenticated-transition signal.
+        shared.metrics.inbound_authenticated_total.inc();
+        warn!(
+            target: "overlay_inbound_diag",
+            peer_id = %peer_id,
+            direction = "inbound",
+            event = "authenticated",
+            "inbound peer transitioned to authenticated"
+        );
 
         let generation = register_result.generation;
         Self::run_peer_loop(
--- a/crates/overlay/src/manager/peer_loop.rs
+++ b/crates/overlay/src/manager/peer_loop.rs
@@ -909,6 +909,40 @@ impl OverlayManager {
         }
     }
 
+    /// #3419 diagnostic (observability-only): emit a `warn!` describing how an
+    /// inbound peer connection dropped/reset, capturing the last message henyey
+    /// sent on that connection before the drop. This does NOT alter any control
+    /// flow, handshake, auth, codec, or dispatch behavior — it only reads
+    /// already-tracked per-peer state and logs it.
+    ///
+    /// Gated to inbound connections only: the mainnet symptom in #3419 is
+    /// `stellar_overlay_inbound_authenticated` sustained near 0 while inbound
+    /// peers establish then get reset by the remote post-auth. Outbound drops
+    /// are not the subject of this investigation, so we don't spam them here.
+    ///
+    /// `reason` is a short, low-cardinality classification of the drop site
+    /// (e.g. "recv_error", "remote_closed", "send_error"); `detail` carries the
+    /// errno / error string when available (empty otherwise).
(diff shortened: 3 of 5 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Cycle 1. Clean, purely-additive diagnostic instrumentation for #3419. Build/test/clippy all green on a fresh worktree off current `origin/main`; the new diagnostic state is write-on-send / read-only-on-drop and cannot affect control flow.

<details>

**Linked issue kind:** diagnostic instrumentation (not a bug-fix). #3419 stays open by design — the PR uses "Refs #3419", not "Closes". `closingIssuesReferences` is empty (verified). No regression test is required because there is no behavior change to regress; the contract is "existing tests still pass," which they do.

**Verified locally (worktree off current origin/main, PR head already up-to-date with main):**
- `cargo build -p henyey-overlay` — OK
- `cargo build --all` — OK
- `cargo test -p henyey-overlay` — OK (lib + `overlay_tx_integration` + doctests; 0 failures across all result lines)
- `cargo clippy --all -- -D warnings` — clean

**Correctness of the diff:**
- `Peer::last_sent_msg_type: &'static str` is initialized to `"none"` in every constructor and test builder (all 8 struct-literal sites updated — no missing-field compile gap), set on the three send paths (`send_raw`, `send_auth`, `send`) via the already-existing `helpers::message_type_name`, and read only by the new `last_sent_msg_type()` accessor, which is consumed only by `log_inbound_drop_diag`. `grep` confirms zero reads in any auth/handshake/dispatch path. `&'static str` is `Copy`, so the accessor is allocation-free.
- `log_inbound_drop_diag` early-returns for non-inbound peers, reads `peer.stats()` / `is_ready()` / `remote_addr()` (all pre-existing accessors) and emits one `warn!`. It is invoked at four drop sites, each immediately before a pre-existing `break`, so it cannot change loop iteration, ordering, or the break decision.
- `inbound_authenticated_total` Counter is `.inc()`'d once, directly after the pre-existing `inbound_establish.inc()` at the register/establish point, and surfaced through the snapshot + app metric catalog (`error-handling`/`api-shape`: snapshot struct, `to_snapshot`, the `&self.<field>` registration list, and the app-side `.absolute(...)` are all consistently wired).

**Error handling / readability:** `drop_detail` carries `e.to_string()` at error sites and `""` at clean-close — correct. Comments are clear and label the observability-only intent at every site.

**test-coverage:** No new public *logic* surface — the one new `pub fn` is a trivial field accessor and the new field is asserted indirectly by the existing peer-construction tests still compiling/passing. Acceptable for a log/metric-only change. (Inline note only, non-blocking: a unit test asserting `last_sent_msg_type()` flips from `"none"` after a send would be nice-to-have, but the value is purely a log field and adds no logic risk.)
</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** crate:overlay is consensus-critical, so this lens applied the CRITICAL check: confirm the change is OBSERVABILITY-ONLY with ZERO behavior change. It is. No handshake/auth/MAC/codec/dispatch control flow was touched; nothing on the wire changed; parity with stellar-core is preserved trivially.

<details>

**Parity stance:** A diagnostic-only change preserves stellar-core parity iff it introduces no observable behavior, wire, or sequencing difference. I verified each surface:

- **Handshake / auth / MAC / codec — UNCHANGED.** `send_raw`/`send_auth`/`send` add exactly one statement each: `self.last_sent_msg_type = helpers::message_type_name(&message)`, placed *after* the existing `OverlayMessageKind::from_stellar_message(&message)` line and *before* any wrapping/wire-send. It does not touch `auth.wrap_unauthenticated` / `auth.wrap_auth_message` / sequence numbers / MAC computation / framing. `message_type_name` is a pure name lookup already used in `send`. No new bytes on the wire, no reordering of the send.
- **Dispatch / peer-loop control flow — UNCHANGED.** `log_inbound_drop_diag` is a read-only logger inserted immediately before four pre-existing `break` sites (send_error, flood_send_error, remote_closed, recv_error). The break decisions, their conditions, and their ordering are identical to before; the function early-returns for non-inbound peers and only reads `stats()`/`is_ready()`/`remote_addr()`. It cannot change which messages are processed, in what order, or when the loop exits.
- **`last_sent_msg_type` is additive state never read by logic** (`parity-gap` check): grep across `crates/overlay/src` shows the only reader is the new accessor, consumed only by the diag logger. It gates no branch, feeds no auth/dispatch decision. Removing it would change nothing but the log line.
- **New counter is monotonic-additive** (`sequencing`/`edge-case` check): `inbound_authenticated_total.inc()` sits directly after the existing `inbound_establish.inc()` at the same register/establish point — same trigger condition, no new branch, monotonic Counter. The existing instantaneous gauge `stellar_overlay_inbound_authenticated` is untouched.
- **`warn!` emissions don't change timing/ordering observably.** They are `tracing` events (in-process logging); they do not block, await, or send on the wire. stellar-core has no equivalent metric/log surface, but metrics/logs are explicitly outside the parity contract (observable *protocol* behavior is what must match) — and they fire at points that already had `info!`/`debug!` logging.

**No protocol/wire/behavior change → parity preserved trivially.** No stellar-core C++ path needs to mirror this; it adds no semantics, only observability. The existing overlay test suite passing unchanged (0 failures in my run: lib + integration + doctests) is the positive evidence that behavior is byte-identical.

**Self-modifying:** No. No pipeline skills/scripts touched (5 files, all `crates/{overlay,app}/src/`).
</details>

tomerweller:
## ✅ Merged

**Commit:** e52cf996966230b43cba64a7f29bdb6ffe423f40

**Follow-up issues filed for unaddressed inline review comments:** none (no inline review threads).

Both reviewer lenses APPROVE (Correctness + Parity). Required CI green; merged past the non-required Quickstart testnet shard per operator approval. This is diagnostic instrumentation (Refs #3419) — #3419 stays OPEN for the redeploy + data capture.

PR #3470, opened 2026-06-18
Fix Survey aggregate-drop double-count in PeerRateLimiter

### LINKED ISSUE #3428: crates/overlay: 2 high-severity cleanup findings
Second full-workspace `/cleanup` audit (re-run after the #3319-#3400 sweep). **2 high-severity findings** for `henyey-overlay`, verified against current `main`.

_High = correctness-adjacent or confirmed-dead code worth fixing soon._

### clippy::identity_op
- **Location**: `crates/overlay/src/codec.rs:549`
- **Lane**: A · **Parity**: parity-mapped (`TCPPeer.cpp (XDR framing)`)
- **Evidence**: cargo clippy -p henyey-overlay --all-targets fails with: error: this operation has no effect\n   --> crates/overlay/src/codec.rs:549:24\n549 |         let len: u32 = 0 | 0x80000000;\n     |                        ^^^^^^^^^^^^^^ help: consider reducing it to: `0x80000000`. Introduced in commit 45a97265 (codec semantics rewrite). Lib-only clippy is clean; --all-targets fails on this test-code lint, blocking the workspace clippy gate.
- **Suggested fix**: Replace `let len: u32 = 0 | 0x80000000;` with `let len: u32 = 0x80000000;` (line 549). Test intent (zero-length body, final-fragment bit set) is preserved; matches the form used at line 567 (`let len: u32 = 4097 | 0x80000000;`) which clippy accepts because that has a non-zero operand.

### POTENTIAL BUG (flagged as cleanup-adjacent)
- **Location**: `crates/overlay/src/manager/peer_loop.rs:226`
- **Lane**: C · **Parity**: not parity-mapped
- **Evidence**: peer_loop.rs:210-228 (aggregate-exhausted branch): for class==Survey, the match at 220-225 increments `self.dropped_aggregate += 1` (line 224), then line 226 unconditionally increments `self.dropped_aggregate += 1` again. Net effect: Survey drops contribute +2 to the aggregate-drop telemetry where every other class contributes +1 to its per-class counter AND +1 to aggregate. This is observable in metrics scrapes that join dropped_aggregate against per-class totals. Reading order: line 224 sets `Survey => self.dropped_aggregate += 1` (inside the per-class match), and line 226 re-increments unconditionally.
- **Suggested fix**: Remove the `TrafficClass::Survey => self.dropped_aggregate += 1` arm at line 224 (let line 226 handle the single aggregate increment for Survey), OR drop the unconditional `self.dropped_aggregate += 1` at line 226 and explicitly fan it out per class. Add a regression unit test that floods Survey messages past the aggregate cap and asserts `dropped_aggregate == cap_overshoot`, not `2 * cap_overshoot`. Routes through /code-review since it has observable behavior — flagged here because the body of `allow()` reads as a cleanup miss.

---
_Generated by the full-workspace `/cleanup` audit (18 crates, 169 findings). Run `/cleanup crates/overlay --mode=plan` to re-verify before applying._

tomerweller:
Closes #3428

## Summary

In `PeerRateLimiter::allow()` (`crates/overlay/src/manager/peer_loop.rs`), the aggregate-exhausted branch matched `TrafficClass::Survey => self.dropped_aggregate += 1` and then ran the unconditional `self.dropped_aggregate += 1`, so each dropped Survey message added **+2** to `dropped_aggregate` while every other traffic class adds +1 (its per-class counter) and +1 (aggregate). This is an observable telemetry-correctness bug: Survey was double-counted in the aggregate-drop counter. The fix makes the Survey match arm a no-op (`Survey => {}`), so the single unconditional increment records the one aggregate-cap drop, matching the other three classes. Survey is aggregate-only by design (no `dropped_survey` per-class field, no consumer), so none is added.

This is a metric-correctness fix — it changes an observable telemetry value (1 instead of 2 per dropped Survey). It touches **only** the `dropped_aggregate` counter; `allow()`'s return value (the actual drop decision) is unchanged, so live message-dispatch/flooding is untouched.

**Finding 1 was stale/refuted.** The issue's Finding 1 (clippy `identity_op` at `codec.rs:549`, `let len: u32 = 0 | 0x80000000;`) does not exist on current `origin/main` — the codec was rewritten since the cleanup audit captured it; `grep` finds no such line and `clippy --all-targets` reports no `identity_op` at codec.rs. No edit to `codec.rs` was made.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3428#issuecomment-4744090557)

## Test plan

- [x] cargo fmt --check
- [x] cargo build --all
- [x] cargo test -p henyey-overlay — 472 lib tests pass (471 prior + new regression test), all integration tests green

- **Test:** `crates/overlay/src/manager/peer_loop.rs::test_peer_rate_limiter_survey_drop_single_aggregate_count`
- **Pre-fix:** committed as `6252645ba106308d6adba5c216611a65b1e2c7a7` — verified FAILED with `assertion left == right failed: left: 10, right: 5` (dropped Survey double-counted: 2*N instead of N).
- **Post-fix:** verified PASSES after `08ee0b0c323334ce938f03a25128efd6d0420499`.

## Parity

n/a — non-parity path. `PeerRateLimiter` and its `dropped_aggregate` telemetry are henyey-specific hardening with no stellar-core analog.

## Deviations from plan

None.

### DIFF
--- a/crates/overlay/src/manager/peer_loop.rs
+++ b/crates/overlay/src/manager/peer_loop.rs
@@ -2040,6 +2040,33 @@ mod tests {
         assert!(limiter.allow(TrafficClass::ControlFetch));
     }
 
+    /// Regression test for #3428: a Survey message dropped at the aggregate
+    /// cap must increment `dropped_aggregate` by exactly 1, matching the other
+    /// traffic classes. Previously the aggregate-exhausted match arm did
+    /// `Survey => self.dropped_aggregate += 1` AND then the unconditional
+    /// `self.dropped_aggregate += 1` ran, so each dropped Survey added +2.
+    #[test]
+    fn test_peer_rate_limiter_survey_drop_single_aggregate_count() {
+        let mut limiter = PeerRateLimiter::new();
+
+        // Fill the aggregate window with Survey messages (all allowed).
+        for _ in 0..DEFAULT_PEER_RATE_LIMIT {
+            assert!(limiter.allow(TrafficClass::Survey));
+        }
+
+        // Drop N more Survey messages past the aggregate cap (all rejected).
+        const N: u64 = 5;
+        for _ in 0..N {
+            assert!(!limiter.allow(TrafficClass::Survey));
+        }
+
+        // Each aggregate-cap drop must count exactly once, not twice.
+        assert_eq!(
+            limiter.dropped_aggregate, N,
+            "each dropped Survey must add +1 to dropped_aggregate (not +2)"
+        );
+    }
+
     #[test]
     fn test_peer_rate_limiter_telemetry_counters() {
         let mut limiter = PeerRateLimiter::new();
--- a/crates/overlay/src/manager/peer_loop.rs
+++ b/crates/overlay/src/manager/peer_loop.rs
@@ -221,7 +221,10 @@ impl PeerRateLimiter {
                 TrafficClass::TxAndDemand => self.dropped_tx_demand += 1,
                 TrafficClass::Advert => self.dropped_advert += 1,
                 TrafficClass::ControlFetch => self.dropped_control_fetch += 1,
-                TrafficClass::Survey => self.dropped_aggregate += 1,
+                // Survey has no per-class drop counter (aggregate-only by
+                // design); the unconditional increment below records the
+                // single aggregate-cap drop, matching the other classes.
+                TrafficClass::Survey => {}
             }
             self.dropped_aggregate += 1;
             return false;


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Minimal, correct telemetry fix. The Survey aggregate-exhausted match arm becomes a no-op so the single unconditional `dropped_aggregate += 1` records exactly one aggregate-cap drop, matching the other three classes. The regression test genuinely pins the bug (verified failing pre-fix, passing post-fix).

<details>

**Issue kind:** bug-fix (telemetry-correctness). Per Reviewer-A test-verification protocol, a regression test that demonstrably catches the bug is required.

**Diff correctness (peer_loop.rs:~220-230):**
- The aggregate-exhausted branch is `if self.aggregate_count >= self.aggregate_limit { ... match class { ... } self.dropped_aggregate += 1; return false; }`. On `origin/main` the arm was `Survey => self.dropped_aggregate += 1`, so a dropped Survey hit `+1` (arm) and `+1` (unconditional) = `+2`. The other three classes have no `dropped_aggregate` touch in their arms (they bump their per-class `dropped_*`), so they correctly recorded `+1`. The fix `Survey => {}` makes Survey consistent with the others: exactly `+1`.
- Survey is aggregate-only by design — there is no `dropped_survey` field on `PeerRateLimiter` (confirmed: only `dropped_tx_demand`, `dropped_advert`, `dropped_control_fetch`, `dropped_aggregate`), and the non-exhausted `match` arm for Survey (`TrafficClass::Survey => { true }`) bumps only `aggregate_count`. So a no-op arm is the right shape; there is no per-class counter to feed.

**Return value / drop decision unchanged (`class:` regression-risk):**
- In the exhausted branch the `match` body only mutates counters; the function returns `false` unconditionally afterward, regardless of class. Changing the Survey arm from `dropped_aggregate += 1` to `{}` cannot alter the `return false`.
- The ControlFetch reserved-capacity early-return above the match (`return true`) is untouched.
- The non-exhausted path (`allowed` match → `if allowed { aggregate_count += 1 }`) is untouched.
- Therefore the actual rate-limit/drop behavior — what `allow()` returns, and hence live message dispatch / flooding — is byte-for-byte identical. Only the observable `dropped_aggregate` telemetry value changes (1 instead of 2 per dropped Survey).

**Regression test (`class:` test-coverage):**
- `test_peer_rate_limiter_survey_drop_single_aggregate_count` fills the aggregate window with `DEFAULT_PEER_RATE_LIMIT` allowed Surveys, drops `N=5` more, then asserts `dropped_aggregate == 5`.
- Verified the test was committed before the fix (`6252645b` test, `08ee0b0c` fix per PR body) and that it **FAILS on a clean `origin/main` + test** with `assertion left == right failed: left: 10, right: 5` (the 2×N double-count), and **PASSES with the fix**. The test captures the exact bug, not an incidental property.
- Test asserts the all-allowed phase (`assert!(limiter.allow(...))` ×200) and the all-dropped phase (`assert!(!limiter.allow(...))` ×5) too, so it also pins that the drop decision itself is preserved.

**Build/clippy/test (`class:` ci-failure):** On a worktree off current `origin/main` merged with the PR head: `cargo build --all` green, `cargo clippy --all -- -D warnings` clean, `cargo test -p henyey-overlay` green (incl. the new guard test).

No error-handling, readability, or API-shape concerns. The added comment accurately documents intent.
</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** No parity concern. `PeerRateLimiter` and its `dropped_aggregate` counter are henyey-specific overlay hardening with no stellar-core analog, and the diff touches only that telemetry counter — not any consensus/flooding/dispatch path that has a parity contract.

<details>

**Parity-mapping (`class:` parity-gap):**
- `crates/overlay/` is on the parity-critical list, so this gets the parity lens. However, `PeerRateLimiter` is a henyey-local per-peer rate-limit shim. stellar-core's overlay flow control uses a different mechanism (`FlowControl` / `mAdvertQueue` + capacity-based send-more), and there is no stellar-core `dropped_aggregate` metric to mirror. So there is no upstream value this counter must match — this is internal telemetry, not protocol-observable state.
- The drop *policy* (aggregate cap, per-class sub-budgets, ControlFetch reserved capacity) is unchanged by this diff; only the bookkeeping for one already-counted drop changes. Even if one wanted to argue the rate-limit policy itself diverges from core, that's pre-existing and out of scope — this PR neither widens nor narrows it.

**Sequencing / edge-case (`class:` sequencing):**
- The change is inside the aggregate-exhausted branch and is a pure counter no-op replacement. It does not reorder any check (window reset → aggregate cap → ControlFetch reservation → per-class budget) and does not change which branch returns `true`/`false`. No sequencing impact.
- Survey traffic classification (the `StellarMessage::*Survey*` → `TrafficClass::Survey` mapping at the top of the file) is untouched, so which messages are subject to the aggregate cap is unchanged.

**Consensus-adjacency check:** Survey messages are diagnostic/telemetry overlay traffic (time-sliced survey collecting/reporting), not SCP. Dropping or counting them differently has zero effect on consensus correctness. The actual drop decision for Survey is unchanged (verified: the exhausted branch returns `false` regardless of the arm body), so live SCP/tx flooding is unaffected.

Confirmed `cargo clippy --all -- -D warnings` clean and `cargo test -p henyey-overlay` green against current `origin/main`. No divergence from stellar-core to flag.
</details>

tomerweller:
## ✅ Merged

**Commit:** 46326610a7605eeb5aadabaf724aaac8652c6c3e

**Follow-up issues filed for unaddressed inline review comments:** none (no inline comments)

Triple-green: Reviewer A (Correctness) APPROVE, Reviewer B (Parity) APPROVE, required CI green. The only pending check at merge was the non-required `Quickstart / test (testnet, ...)` shard (`mergeStateStatus: UNSTABLE`), which does not gate per branch protection. Drop-decision in `PeerRateLimiter::allow()` verified unchanged; guard test verified FAILS on origin/main (left:10 right:5) and PASSES with the fix.

PR #3523, opened 2026-06-19
Consolidate scheduler lookups and hoist cancelled guard in work crate

### LINKED ISSUE #3466: crates/work: 4 low-severity cleanup findings
Second full-workspace `/cleanup` audit (re-run after the #3319-#3400 sweep). **4 low-severity findings** for `henyey-work`, verified against current `main`.

_Low = touch-up-when-nearby quality cleanup; do not dedicate a planning cycle._

### DUPLICATION / repeated lookup regression
- **Location**: `crates/work/src/scheduler.rs:562`
- **Lane**: B · **Parity**: parity-mapped (`WorkScheduler.h/WorkScheduler.cpp`)
- **Evidence**: Lines 562 (`entries.get_mut(&id).map(|entry| entry.attempts)`), 565-568 (`entries.get(&id).is_some_and(|entry| entry.cancel_token.is_cancelled())`), and 574 (`entries.get_mut(&id).expect("entry must exist")`) all key on the same `id`. Commit f8a3f282 explicitly title 'Reduce start_work to single HashMap lookup' consolidated this; the subsequent bae2e933 / 0582b85a churn re-introduced multiple lookups.
- **Suggested fix**: Restore the single-lookup form: take one `let Some(entry) = self.entries.get_mut(&id) else { return false; }`, extract `attempts` and check `cancel_token.is_cancelled()` from that one borrow, then drop the borrow before calling `self.finish_terminal_state(...)` / `self.transition_state(...)`. The `mem::replace(&mut entry.work, ...)` and field reads (`entry.name.clone()`, `entry.cancel_token.clone()`, `entry.attempts += 1`) all run inside the same borrow scope.

### DUPLICATION
- **Location**: `crates/work/src/scheduler.rs:705`
- **Lane**: B · **Parity**: parity-mapped (`WorkScheduler.h/WorkScheduler.cpp`)
- **Evidence**: Lines 706-708 (`WorkOutcome::Cancelled => ...`), 710-713 (`WorkOutcome::Success if cancelled => ...`), 718-721 (`WorkOutcome::Retry { delay: _ } if cancelled => ...`), and 723-726 (`WorkOutcome::Failed(_) if cancelled => ...`) all run the same two lines: `self.finish_terminal_state(id, WorkState::Cancelled, attempt); CompletionAction::None`. The `WorkOutcome::Cancelled` arm overlaps with `cancelled` semantically.
- **Suggested fix**: Hoist the cancelled check ahead of the match: `if cancelled || matches!(completion.outcome, WorkOutcome::Cancelled) { self.finish_terminal_state(id, WorkState::Cancelled, attempt); return CompletionAction::None; }`. Then the match shrinks to three arms (Success / Retry / Failed) with no `if cancelled` guards.

### DEAD CODE (variant write never observed)
- **Location**: `crates/work/src/scheduler.rs:237`
- **Lane**: C · **Parity**: parity-mapped (`WorkScheduler.h/WorkScheduler.cpp`)
- **Evidence**: grep WorkSlot in crates/work/src/scheduler.rs: only consumer is the destructure `let WorkSlot::Idle(mut work) = std::mem::replace(...) else { return false; };` at line 580, which treats Lost and InFlight identically. No code path distinguishes Lost from InFlight, and the existing protection against re-spawning a panicked work comes from the state machine (Failed/Cancelled are terminal and never re-enqueued via pending_queue/enqueue_dependents). Workspace-wide rg 'WorkSlot::Lost' confirms only the one write site.
- **Suggested fix**: Two options: (a) Remove the `Lost` variant entirely — let handle_join_error leave the slot in its existing `WorkSlot::InFlight` state set by start_work; the semantics are unchanged because both InFlight and Lost are rejected by the line-580 destructure. This deletes ~5 lines and one enum variant. (b) Keep `Lost` for documentation but add a `debug_assert!` in start_work that catches if the slot is Lost on a re-spawn attempt, to make the marker observable. Prefer (a); the variant earns nothing today.

### VERBOSE / unnecessary let-binding
- **Location**: `crates/work/src/scheduler.rs:376`
- **Lane**: B · **Parity**: parity-mapped (`WorkScheduler.h/WorkScheduler.cpp`)
- **Evidence**: Line 376: `let _ = self.cancel(id);`. `WorkScheduler::cancel` at line 351 returns `bool` with no `#[must_use]`. The `let _ =` is decorative — a bare `self.cancel(id);` compiles warning-free.
- **Suggested fix**: Drop the `let _ =` binding: `for id in ids { self.cancel(id); }`. Trivial, but it removes a misleading must-use hint and the parity column reads as cleanup.

---
_Generated by the full-workspace `/cleanup` audit (18 crates, 169 findings). Run `/cleanup crates/work --mode=plan` to re-verify before applying._

tomerweller:
Closes #3466

## Summary

Behavior-preserving cleanup of four low-severity findings in `crates/work/src/scheduler.rs`. Net observable scheduler behavior diff = 0; the existing `crates/work/tests/scheduler.rs` suite is the behavior-preservation oracle and continues to pass.

- **F1 (`start_work`):** folded the duplicate `attempts` read and the `is_cancelled` check into a single immutable `get(...).map(...)` that copies `attempts` (u32) + the cancel bool into locals, dropping the borrow before `finish_terminal_state` (which needs `&mut self`). The trailing `get_mut().expect()` stays as a separate `&mut` borrow for the mutate-and-spawn phase. 3 lookups → 2; ordering and side-effects unchanged.
- **F2 (`handle_completion`):** hoisted the four cancelled arms into one early guard `if cancelled || matches!(completion.outcome, WorkOutcome::Cancelled) { finish Cancelled; return None }` before the residual match. The `|| matches!` clause is load-bearing (preserves the old unconditional `Cancelled` arm). `warn!("work failed")` still fires only on the genuine non-cancelled `Failed` path. A guarded-unreachable `Cancelled => None` arm keeps the match exhaustive (Rust cannot prove the runtime guard removes the variant); behavior identical. Truth-table verified identical.
- **F3 (`WorkSlot::Lost`):** kept the variant (full removal deliberately scoped out as a separate design call). Added `debug_assert!(matches!(entry.work, WorkSlot::InFlight), …)` at the sole write site, and clarified the ownership doc + variant doc that `Lost` is a write-only diagnostic marker never re-read to drive behavior.
- **F4 (`cancel_all`):** `let _ = self.cancel(id);` → `self.cancel(id);` (`cancel` returns `bool`, no `#[must_use]` — warning-free).

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3466#issuecomment-4751635383)

## Test plan

- [x] cargo fmt --check
- [x] cargo build -p henyey-work --all-targets (RUSTFLAGS=-Dwarnings)
- [x] cargo test -p henyey-work — 12/12 scheduler tests pass (panic/abort tests exercise the F3 debug_assert under the debug test profile)
- [x] cargo build --all

## Deviations from plan

- F2's residual match keeps an explicit `WorkOutcome::Cancelled => CompletionAction::None` arm. The plan's pseudocode showed a 3-arm match, but Rust's exhaustiveness check cannot prove the runtime early-guard removes the `Cancelled` variant, so the arm is required to compile. It is unreachable at runtime (guarded) and returns `None`, so behavior is identical.

### DIFF
--- a/crates/work/src/scheduler.rs
+++ b/crates/work/src/scheduler.rs
@@ -105,7 +105,11 @@ impl Default for WorkSchedulerConfig {
 /// [`WorkSlot::InFlight`]). On successful completion, ownership is returned and
 /// the work item can be retried. On panic or task abort, the work item is
 /// permanently lost ([`WorkSlot::Lost`]) — this is safe because panics and aborts
-/// transition work to terminal states that are never re-run.
+/// transition work to terminal states that are never re-run. `Lost` is a
+/// write-only diagnostic marker: it is set once on the panic/abort path and is
+/// never re-read to drive behavior (the sole reader treats any non-`Idle` slot
+/// identically), so it is kept purely to distinguish "moved into a panicked or
+/// aborted task" from `InFlight` = "moved into a still-running task".
 ///
 /// # Example
 ///
@@ -235,6 +239,12 @@ enum WorkSlot {
     /// Work has been moved into a spawned task.
     InFlight,
     /// Work was irrecoverably lost due to panic or task abort (terminal).
+    ///
+    /// Write-only diagnostic marker: set once on the panic/abort path (from a
+    /// prior `InFlight` state) and never re-read to drive behavior — the sole
+    /// reader treats any non-`Idle` slot identically. Kept to distinguish a
+    /// panicked/aborted task from one that is still running (`InFlight`);
+    /// collapsing it into `InFlight` is a deliberately deferred design decision.
     Lost,
 }
 
@@ -373,7 +383,7 @@ impl WorkScheduler {
     pub(crate) fn cancel_all(&mut self) {
         let ids: Vec<WorkId> = self.entries.keys().copied().collect();
         for id in ids {
-            let _ = self.cancel(id);
+            self.cancel(id);
         }
     }
 
@@ -559,14 +569,18 @@ impl WorkScheduler {
             return false;
         }
 
-        let Some(attempts) = self.entries.get_mut(&id).map(|entry| entry.attempts) else {
-            return false;
-        };
-        if self
+        // Read both `attempts` and the cancel-token state out of a single
+        // immutable lookup into `Copy` locals, dropping the borrow before the
+        // `finish_terminal_state` call below (which needs `&mut self`). The
+        // mutate-and-spawn phase reacquires a `&mut` borrow via `get_mut`.
+        let Some((attempts, is_cancelled)) = self
             .entries
             .get(&id)
-            .is_some_and(|entry| entry.cancel_token.is_cancelled())
-        {
+            .map(|entry| (entry.attempts, entry.cancel_token.is_cancelled()))
+        else {
+            return false;
+        };
+        if is_cancelled {
             self.finish_terminal_state(id, WorkState::Cancelled, attempts);
             return false;
         }
@@ -702,33 +716,31 @@ impl WorkScheduler {
 
         self.finalize_entry(id, completion.work);
 
+        // Any cancellation — whether signalled via the cancel token/state
+        // (`cancelled`) or returned directly by the work item
+        // (`WorkOutcome::Cancelled`, which the unconditional arm below used to
+        // handle even when `cancelled` was false) — terminates as Cancelled
+        // with no further action. The `|| matches!(...)` clause is load-bearing:
+        // it preserves the old unconditional `Cancelled` arm so a genuine
+        // `Cancelled` outcome does not fall through to the match below.
+        if cancelled || matches!(completion.outcome, WorkOutcome::Cancelled) {
+            self.finish_terminal_state(id, WorkState::Cancelled, attempt);
+            return CompletionAction::None;
+        }
+
         match completion.outcome {
-            WorkOutcome::Cancelled => {
-                self.finish_terminal_state(id, WorkState::Cancelled, attempt);
-                CompletionAction::None
-            }
-            WorkOutcome::Success if cancelled => {
-                self.finish_terminal_state(id, WorkState::Cancelled, attempt);
-                CompletionAction::None
-            }
             WorkOutcome::Success => {
                 self.transition_state(id, WorkState::Success, attempt);
                 CompletionAction::Done { completed_id: id }
             }
-            WorkOutcome::Retry { delay: _ } if cancelled => {
-                self.finish_terminal_state(id, WorkState::Cancelled, attempt);
-                CompletionAction::None
-            }
             WorkOutcome::Retry { delay } => self.schedule_retry(id, attempt, delay),
-            WorkOutcome::Failed(_) if cancelled => {
-                self.finish_terminal_state(id, WorkState::Cancelled, attempt);
-                CompletionAction::None
-            }
             WorkOutcome::Failed(err) => {
                 warn!(work_id = id, error = %err, "work failed");
                 self.finish_terminal_state(id, WorkState::Failed, attempt);
                 CompletionAction::None
             }
+            // Cancelled is handled by the early guard above.
+            WorkOutcome::Cancelled => CompletionAction::None,
         }
     }
 
@@ -814,6 +826,10 @@ impl WorkScheduler {
 
         // Work item is permanently lost — it was moved into the panicked/aborted task.
         if let Some(entry) = self.entries.get_mut(&id) {
+            debug_assert!(
+                matches!(entry.work, WorkSlot::InFlight),
+                "Lost is only reachable from InFlight"
+            );
             entry.work = WorkSlot::Lost;
         }
     }


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Behavior-preserving cleanup of 4 low-severity findings in `crates/work/src/scheduler.rs`. F1/F2 are faithful factorings with identical ordering, side-effects, and borrow semantics; F3 is keep-with-clarification (no state-transition change); F4 is a no-op tidy. The existing 12-test scheduler suite (the behavior-preservation oracle) passes under the debug profile, exercising the F3 debug_assert via the panic/abort tests.

<details>

Issue #3466 is `kind: refactor` (net observable behavior diff = 0), so no regression test is required; the existing `crates/work/tests/scheduler.rs` suite is the oracle. Verified locally on the PR head merged with current `origin/main` (clean — already up to date):

- `cargo build -p henyey-work --all-targets` under `RUSTFLAGS=-Dwarnings` → clean (would have caught F4 unused-result and any F1/F2 unused binding).
- `cargo test -p henyey-work` → 12/12 scheduler tests pass under the debug profile (F3 debug_assert exercised by `test_panic_recovery`, `test_panic_no_retry`, `test_cancel_then_panic`, `test_panic_blocks_dependents`).

**F1 (`start_work`) — behavior-identical.** The fold copies `attempts` (u32) + `is_cancelled` (bool) out of a single immutable `get(&id).map(...)` into `Copy` locals; the borrow ends at the `let ... else` binding, so `finish_terminal_state` (needs `&mut self`) runs with no live borrow. The trailing `get_mut().expect()` is a separate `&mut` borrow for the mutate-and-spawn phase. 3 lookups → 2. Ordering/side-effects unchanged.

**F2 (`handle_completion`) — truth-table identical.** `WorkOutcome` has exactly 4 variants. The `|| matches!(completion.outcome, WorkOutcome::Cancelled)` clause is load-bearing — it preserves the old unconditional `Cancelled` arm so a genuine `Cancelled` outcome (cancelled=false) does not fall through. `warn!("work failed")` fires only on the non-cancelled `Failed` path. The residual `WorkOutcome::Cancelled => None` arm is required for exhaustiveness and is guarded-unreachable at runtime, returning None — no behavior difference.

**F3 (`WorkSlot::Lost`) — sound.** Variant kept (full removal scoped out as a separate design call). The `debug_assert!` at the sole write site is sound: `start_work` sets the slot to `InFlight` and the work is moved into the spawned task, so the slot is necessarily `InFlight` when the join error fires. Doc-only otherwise.

**F4 (`cancel_all`)** — `let _ =` dropped; `cancel` returns `bool` with no `#[must_use]`. Warning-free (confirmed by -Dwarnings build).

No concerns.
</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** No regression, performance, API, security, or operational risk. Single-file change to `crates/work/src/scheduler.rs`, all-internal (`pub(crate)`/private), no public API or data-format surface, no config/migration. The scheduler drives the catchup/publish state machines so it is correctness-critical, but every change is a behavior-preserving factoring — net observable diff = 0 — and the change reduces work rather than adding it (3→2 lookups in `start_work`).

<details>

Risk lens (regressions / perf / API breaks / security / operational), reviewed on the PR head merged with current `origin/main` (clean).

- **regression-risk:** none. F1/F2 are pure factorings (verified borrow-drop ordering and the 4×bool truth-table are identical to the prior arm-by-arm form). F3 keeps the `Lost` variant — no state-machine variant removed on the panic/abort path, so no behavioral risk on catchup/publish recovery. The 12-test scheduler suite (cancel/panic/retry/dependency) passes, including the panic/abort tests that drive the catchup-relevant join-error recovery path.
- **perf:** mild improvement — `start_work` goes from 3 HashMap lookups to 2 on the same `id`. No new allocations, no added work on hot paths.
- **api-break:** none. The only touched symbols are `pub(crate)`/private (`start_work`, `handle_completion`, `cancel_all`, `handle_join_error`, the private `WorkSlot` enum). No public surface, no XDR/wire/data-format change.
- **security:** none. No untrusted input, no unsafe, no new external interaction. The `debug_assert!` is debug-only and cannot affect release behavior.
- **operational:** none. No config, no migration, no on-disk format, no log-format change that ops depends on (`warn!("work failed")` still fires on exactly the same path).

Self-modifying check: the only changed file is `crates/work/src/scheduler.rs` — not under any pipeline skill/script path, so this PR is NOT self-modifying.

No concerns.
</details>

tomerweller:
## ✅ Merged

**Commit:** 1e2c69fa997d4eb5b399afb5901ce5b35bbcccc4

**Follow-up issues filed for unaddressed inline review comments:** none (no inline comments; both reviewers APPROVED at top level)

Triple-green: Reviewer A (Correctness) APPROVE, Reviewer B (Risk) APPROVE, CI required-green (Test / Doctest / Build (release-ci) / Clippy / Format all pass). The only remaining pending check is the non-required testnet quickstart shard (chronic flake, de-gated).

Issue #3501, opened 2026-06-19
Near-tip event-loop stall (phase=33) starves ledger-close/persist — trigger behind #3497 DB-lock abort


tomerweller:
## Near-tip event-loop stall starves ledger-close/persist (trigger behind #3497 DB-lock abort)

Split out from #3497 (the DB-lock fatal was the *symptom*; this stall is the *trigger*).

**Evidence (mainnet, 2026-06-19T03:27, build `61e7c2fa`), the ~40s before #3497's abort:**
```
03:27:09 WARN handle_close_complete slow total_ms=267 tx_queue_background_wait_ms=219
03:27:28 WARN WATCHDOG: Event loop slow (>15s) stale_secs=19 phase=33 fetch_channel_depth=101 fetch_channel_depth_max=609
03:27:38 WARN WATCHDOG: Event loop slow (>15s) stale_secs=29 phase=33 fetch_channel_depth=139 fetch_channel_depth_max=609
03:27:39 ERROR Fatal persist failure ... database is locked
```

A ~29s event-loop stall at `phase=33` with overlay fetch backpressure (`fetch_channel_depth` 101→139, max 609) held the close/persist path long enough to blow past the SQLite `busy_timeout=30s` window, so the ledger-close commit hit `SQLITE_BUSY`. This is the **near-tip-stall family** — cf. closed #3218 and operator-gated #3318 (peer-SCP routing).

**Contributing contention (from #3497 investigation):**
- The **maintainer runs DELETE writes on a separate pool connection** (`crates/app/src/maintainer.rs:390-412`) concurrently with ledger-close — a real second writer for the WAL write lock.
- Henyey leaves **`wal_autocheckpoint` at SQLite's default (1000 pages)**; stellar-core sets **10000** (`Database.cpp`). Default-frequency inline auto-checkpoints during a large close commit add contention. A small parity-adjacent hardening: set `wal_autocheckpoint=10000` to match core.

**Asks:**
1. Investigate the `phase=33` event-loop stall + overlay fetch-channel backpressure (depth max 609) — what blocks the loop for ~29s near tip? Relationship to #3318 / #3218.
2. Consider `wal_autocheckpoint=10000` (core parity) + whether the maintainer's concurrent-DELETE connection should be coordinated with the close path to avoid write-lock contention.

Labels: bug, urgent (the stall drives validator outages). Related: #3497 (symptom), #3318, #3218.

tomerweller:
## Triage Report

[...]

Issue #3497, opened 2026-06-19
OFFLINE: validator aborts on transient SQLite "database is locked" in ledger-close persist (persist.rs:439)


tomerweller:
## OFFLINE: validator aborted on transient SQLite "database is locked" during ledger-close DB write

**Symptom** — mainnet validator (session 74535976, build `61e7c2fa`) self-aborted at **2026-06-19T03:27:39Z**:
```
ERROR henyey_app::app::persist: Fatal persist failure, aborting context="ledger close DB write" error=SQLite error: database is locked
```
~10min outage until the monitor caught the dead process and relaunched (clean restore from persisted lcl 63094950 — on-disk state intact).

**Lead-up (the ~40s before the abort):**
```
03:27:09 WARN app.handle_close_complete slow total_ms=267 tx_queue_background_wait_ms=219
03:27:28 WARN WATCHDOG: Event loop slow (>15s) stale_secs=19 phase=33 fetch_channel_depth=101 fetch_channel_depth_max=609
03:27:38 WARN WATCHDOG: Event loop slow (>15s) stale_secs=29 phase=33 fetch_channel_depth=139 fetch_channel_depth_max=609
03:27:39 ERROR Fatal persist failure ... database is locked
```
So: an event-loop stall (watchdog firing, ~29s stale) + overlay fetch-channel backpressure (depth climbing 101→139, max 609) immediately preceded the persist DB write hitting `SQLITE_BUSY/locked`. The stall let another connection (WAL checkpointer or a concurrent reader) hold the write lock past the implicit busy window.

**Root cause** — `fatal_persist_error()` (`crates/app/src/app/persist.rs:439`) treats **all** persist failures as unrecoverable and calls `std::process::abort()`. The header comment claims "the node's on-disk state would diverge from in-memory state" — but that is **false for `database is locked`**: SQLite write transactions are atomic, so a `SQLITE_BUSY`/`SQLITE_LOCKED` means the txn **never committed**. On-disk state is consistent (one ledger behind), and a plain restart recovers it cleanly — exactly what happened here. Aborting the consensus validator on a *transient* lock is over-aggressive (same class as the ENOSPC bucket-merge panic, #3478).

**Suggested fix** (two complementary):
[...]

PR #3537, opened 2026-06-22
Offload tx_set_gc purge off the event loop via spawn_blocking

### LINKED ISSUE #3532: Event-loop freeze (39s, watchdog_freeze) during tx_set_gc — synchronous SQLite purge blocks tokio loop
## Symptom

A **39-second event-loop freeze** on the mainnet validator (session 74535976, binary `898ab456`) during transaction-set garbage collection.

```
01:16:41Z WARN  WATCHDOG: Event loop slow (>15s since last tick) stale_secs=19 phase=33 phase_sub=0 fetch_channel_depth=101 fetch_channel_depth_max=202
01:16:51Z WARN  WATCHDOG: Event loop slow (>15s since last tick) stale_secs=29 phase=33 phase_sub=0 fetch_channel_depth=141 fetch_channel_depth_max=202
01:17:01Z ERROR WATCHDOG: Event loop appears frozen! ... watchdog_freeze=true stale_secs=39 phase=33 phase_sub=0 fetch_channel_depth=180 fetch_channel_depth_max=202 pid=3306764
01:17:01Z ERROR WATCHDOG: Thread state summary count=2 state="R (running)"
01:17:01Z ERROR WATCHDOG: Thread state summary count=41 state="S (sleeping)"
01:17:01Z ERROR WATCHDOG: Thread wchan summary count=39 wchan="futex_wait_queue"
```

**phase=33 = `tx_set_gc`** (purge unreferenced persisted tx sets).

The node **self-recovered** — by 01:17:02Z it was back at `current_ledger=63139735 gap=1 herder_state=Tracking`, and is now Validating at tip. Consensus impact was minimal *this time* (gap stayed at 1). The watchdog did NOT auto-abort, so no restart occurred. This is a **latent wedge risk**, not an active outage.

## Suspected root cause

The `tx_set_gc` timer arm runs a **synchronous, blocking SQLite write transaction directly in the tokio `select!` event loop**:

`crates/app/src/app/lifecycle.rs:1182`
```rust
_ = tx_set_gc_interval.tick() => {
    self.set_phase(33); // 33 = tx_set_gc
    self.herder.purge_persisted_tx_sets();
}
```

`crates/herder/src/herder.rs:934`
```rust
pub fn purge_persisted_tx_sets(&self) {
    let Some(manager) = self.scp_persistence.get() else { return; };
    if let Err(e) = manager.purge_unreferenced_tx_sets() { error!(...); }
}
```

`crates/db/src/scp_persistence.rs` (`purge_unreferenced_tx_sets_atomic`)
```rust
// BEGIN IMMEDIATE so the purge holds a RESERVED write lock for the
// entire read-then-delete sequence ... (see #2770)
self.db.transaction_immediate(|tx| tx.purge_unreferenced_tx_sets_atomic())
```

`purge_unreferenced_tx_sets()` is **not awaited and not offloaded** — it executes inline on the event-loop thread. It opens a `BEGIN IMMEDIATE` (RESERVED write lock) transaction and does a read-then-delete scan over the persisted tx-set table. When that table has grown large (or the DB is under write contention from `save_tx_set`/`save_scp_state` on other pool connections), the scan+delete takes tens of seconds and **freezes the entire event loop** — SCP message processing, broadcast, and fetch-response draining all stall. The climbing `fetch_channel_depth` (101→141→180, max 202) is the secondary symptom: inbound fetch responses were not drained while the loop was blocked.

39 sustained `futex_wait_queue` threads + 2 running confirms the runtime was blocked on a lock, consistent with a synchronous DB transaction holding the connection.

## Why this matters

A consensus validator that freezes its event loop for 39s pauses SCP participation for that window. This recovered at gap=1, but a larger persisted-tx-set table or heavier DB contention could push the freeze past the watchdog auto-abort threshold → forced restart / dropped out of consensus. Frozen event loop is an enumerated urgent symptom.

## Concrete fix sketch

Offload the GC purge off the event-loop thread so the blocking SQLite transaction never freezes the async runtime. Either:

1. **`spawn_blocking`** — in `lifecycle.rs:1182`, run the purge on the blocking pool:
   ```rust
   _ = tx_set_gc_interval.tick() => {
       self.set_phase(33);
       let herder = self.herder.clone(); // or an Arc handle to scp_persistence
       tokio::task::spawn_blocking(move || herder.purge_persisted_tx_sets());
   }
   ```
   (fire-and-forget is acceptable — GC is idempotent and periodic; the `MissedTickBehavior::Skip` already tolerates overlap).

2. **Batch/incrementalize** the delete inside `purge_unreferenced_tx_sets_atomic` (delete in bounded chunks, releasing the RESERVED lock between batches) so no single transaction holds the write lock for tens of seconds.

Option 1 is the minimal, lowest-risk change and matches how other DB-heavy work should be kept off the event loop. Verify with a unit test that seeds a large persisted-tx-set table and asserts the event-loop arm returns promptly (purge runs off-thread).

## Evidence
- Binary: `898ab456dae265cb77d37096b17dd02638936a5b`
- Single occurrence in current log (since 23:49Z deploy); `grep -c watchdog_freeze=true` = 1
- Ledger at freeze: ~63139735; node now Validating at L63139818+

tomerweller:
Closes #3532

## Summary

The periodic tx-set GC purge (`Herder::purge_persisted_tx_sets` → a synchronous `BEGIN IMMEDIATE` SQLite write transaction in `db/scp_persistence.rs`) ran inline in the tokio `select!` event loop. With a large persisted-tx-set table or WAL write contention it blocked the loop for tens of seconds (observed **39s, `watchdog_freeze`, phase=33** on the mainnet validator), stalling SCP processing, broadcast, and fetch-response draining (the climbing `fetch_channel_depth` symptom).

This moves the purge onto the blocking thread pool via a thin `dispatch_tx_set_gc` helper guarded by a loop-local in-flight `JoinHandle`: a tick is skipped while a prior purge is still running, so purges stay **strictly serial and non-overlapping** — mirroring stellar-core rescheduling `startTxSetGCTimer()` only after `purgeOldPersistedTxSets()` returns (HerderImpl.cpp:2440-2487). `set_phase(33)` stays on the loop thread so the watchdog phase label is preserved, and the shutdown arm `abort()`s any in-flight handle (harmless: the purge is one atomic, idempotent transaction per #2770). The purge *contents* are unchanged, so behavior and parity are identical beyond no longer freezing the loop. App-crate-only; herder/db purge logic untouched.

This also addresses **#3501**'s phase=33 near-tip event-loop stall — the synchronous `purge_unreferenced_tx_sets` at phase=33 is the blocker that issue was tracking. It is independent of #3501's separate `wal_autocheckpoint` / maintainer-coordination asks and of #3497's operator-gated DB-lock policy decision.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3532#issuecomment-4764356675)

## Test plan

- [x] cargo fmt --check
- [x] cargo build -p henyey-app --all-targets (RUSTFLAGS=-Dwarnings)
- [x] cargo build --all (RUSTFLAGS=-Dwarnings)
- [x] cargo test --all — 7553 passed, 0 failed
- [x] Parity tripwires byte-identical: ledger_close_meta_vectors, tx_meta_hash_vectors, validation_parity, scp_parity_tests, txset_mixed_equivalence / scp_split_equivalence
- [x] Preserved: herder persistence_race (purge atomicity), app watchdog_shutdown

- **Test:** `crates/app/src/app/lifecycle.rs::tx_set_gc_offload_tests::{test_tx_set_gc_purge_runs_off_event_loop_thread, test_tx_set_gc_serial_no_overlap}`
- **Pre-fix:** committed as `86e427d3` — verified FAILED on origin/main: `unresolved import super::dispatch_tx_set_gc` (no offload helper exists; the arm runs the purge inline on the loop thread).
- **Post-fix:** verified PASSES after `945f26f5` — the purge runs on a `spawn_blocking` thread (thread id differs from the dispatcher), the dispatch returns <100ms, and a second tick while a purge is in-flight is coalesced (only one purge runs).

## Deviations from plan

None. The helper is a private free fn `dispatch_tx_set_gc` generic over the work closure (the arm passes `move || herder.purge_persisted_tx_sets()`), with in-module `#[cfg(test)]` unit tests driving the real helper — the plan's preferred non-public seam.

### DIFF
--- a/crates/app/src/app/lifecycle.rs
+++ b/crates/app/src/app/lifecycle.rs
@@ -82,6 +82,47 @@ pub(crate) fn query_rate_limit_window(close_duration: Duration) -> Duration {
     henyey_overlay::query_policy::query_rate_limit_window(close_duration)
 }
 
+/// Offload one periodic tx-set GC purge onto the blocking thread pool, keeping
+/// purges strictly serial via a loop-local in-flight handle (#3532).
+///
+/// The purge (`Herder::purge_persisted_tx_sets` → `BEGIN IMMEDIATE` read-then-
+/// delete over the persisted-tx-set table) is a synchronous SQLite write
+/// transaction. Running it inline in the tokio `select!` event loop froze the
+/// loop for tens of seconds (observed 39s, `watchdog_freeze`) when the table
+/// was large or the WAL write lock was contended, stalling SCP processing,
+/// broadcast, and fetch-response draining (the climbing `fetch_channel_depth`
+/// symptom). This moves the blocking work to `spawn_blocking` so the loop arm
+/// returns immediately.
+///
+/// Parity: stellar-core's `HerderImpl::purgeOldPersistedTxSets()`
+/// (HerderImpl.cpp:2448-2487) runs in an `async_wait` callback and reschedules
+/// `startTxSetGCTimer()` (HerderImpl.cpp:2440-2444) only AFTER the purge
+/// returns — i.e. purges never overlap. The `is_finished()` guard reproduces
+/// that serial, non-overlapping cadence across the async boundary (the timer's
+/// `MissedTickBehavior::Skip` only coalesces ticks, not in-flight tasks). The
+/// purge *contents* are unchanged (same atomic SQL, #2770), and GC only deletes
+/// orphaned/unreferenced persisted tx-sets — never ledger/consensus/bucket
+/// state — so the thread it runs on is not observable and parity is preserved.
+///
+/// `slot` is the loop-local in-flight handle. If a prior purge is still
+/// running, the tick is skipped (coalesced); otherwise a fresh blocking task is
+/// spawned and its handle stored. The handle is fire-and-not-awaited: GC is
+/// idempotent and periodic, so a skipped tick simply re-runs next interval, and
+/// on shutdown the loop `abort()`s any in-flight handle (harmless — the purge
+/// is a single atomic transaction).
+fn dispatch_tx_set_gc<F>(work: F, slot: &mut Option<tokio::task::JoinHandle<()>>)
+where
+    F: FnOnce() + Send + 'static,
+{
+    // Serial cadence: skip if a prior purge is still running (mirrors
+    // stellar-core rescheduling startTxSetGCTimer() only after the purge
+    // returns — purges never overlap).
+    if slot.as_ref().is_some_and(|h| !h.is_finished()) {
+        return;
+    }
+    *slot = Some(tokio::task::spawn_blocking(work));
+}
+
 /// Whether a transaction queue result requires removal of the FloodGate record,
 /// allowing re-delivery to be treated as new.
 ///
@@ -370,6 +411,11 @@ impl App {
             tx_set_gc_period,
         );
         tx_set_gc_interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+        // Loop-local in-flight handle for the offloaded tx-set GC purge (#3532).
+        // The purge runs on the blocking pool via `spawn_blocking`; this handle
+        // keeps purges strictly serial (skip a tick if the prior purge is still
+        // running) and is aborted on shutdown. See `dispatch_tx_set_gc`.
+        let mut tx_set_gc_task: Option<tokio::task::JoinHandle<()>> = None;
 
         // Get mutable access to SCP envelope receiver
         let mut scp_rx = self.scp_envelope_rx.lock().await;
@@ -1175,18 +1221,36 @@ impl App {
 
                 // Periodic GC of unreferenced persisted SCP tx sets.
                 // Parity: stellar-core HerderImpl::startTxSetGCTimer() at
-                // HerderImpl.cpp:2440-2444. Called inline (no spawn_blocking)
-                // to match stellar-core's strictly serial reschedule cadence
-                // and to avoid a fire-and-forget task surviving select! exit
-                // on shutdown.
+                // HerderImpl.cpp:2440-2444. The purge is a synchronous
+                // BEGIN IMMEDIATE SQLite write transaction; running it inline
+                // froze the event loop for tens of seconds (39s watchdog_freeze,
+                // #3532). It is offloaded to the blocking pool via
+                // `dispatch_tx_set_gc`, which keeps purges strictly serial
+                // (mirroring stellar-core's serial reschedule cadence) through
+                // the loop-local in-flight handle. `set_phase(33)` stays on the
+                // loop thread so the watchdog phase label is preserved.
                 _ = tx_set_gc_interval.tick() => {
                     self.set_phase(33); // 33 = tx_set_gc
-                    self.herder.purge_persisted_tx_sets();
+                    let herder = Arc::clone(&self.herder);
+                    dispatch_tx_set_gc(
+                        move || herder.purge_persisted_tx_sets(),
+                        &mut tx_set_gc_task,
+                    );
                 }
 
                 // Shutdown signal (lowest priority)
                 _ = shutdown_rx.recv() => {
                     tracing::info!("Shutdown signal received");
+                    // Abort any in-flight tx-set GC purge so the offloaded
+                    // blocking task does not survive loop exit (#3532). The
+                    // purge is a single atomic, idempotent transaction (#2770),
+                    // so an abort mid-flight is harmless — it either completes
+                    // (spawn_blocking closures are not force-cancelled) or the
+                    // process is exiting; a skipped purge simply re-runs next
+                    // interval on the next startup.
+                    if let Some(h) = tx_set_gc_task.take() {
+                        h.abort();
+                    }
                     break;
                 }
 
--- a/crates/app/src/app/lifecycle.rs
+++ b/crates/app/src/app/lifecycle.rs
@@ -3709,3 +3709,143 @@ mod tx_flood_forget_tests {
         assert!(should_forget_tx_flood_record(&TxQueueResult::TryAgainLater));
     }
 }
+
+#[cfg(test)]
+mod tx_set_gc_offload_tests {
+    //! Regression tests for #3532 — the tx_set_gc event-loop freeze.
+    //!
+    //! On `origin/main` the `tx_set_gc_interval.tick()` arm called
+    //! `self.herder.purge_persisted_tx_sets()` inline on the tokio event-loop
+    //! thread. A large persisted-tx-set table (or write contention) made that
+    //! synchronous `BEGIN IMMEDIATE` purge block the loop for tens of seconds
+    //! (observed 39s, watchdog_freeze). The fix offloads the purge via
+    //! `spawn_blocking` behind a loop-local in-flight guard so the loop arm
+    //! returns promptly and purges stay strictly serial (mirroring
+    //! stellar-core's serial reschedule cadence).
+    use super::dispatch_tx_set_gc;
+    use std::sync::atomic::{AtomicUsize, Ordering};
+    use std::sync::{Arc, Barrier};
+    use std::thread::ThreadId;
+    use std::time::{Duration, Instant};
+    use tokio::task::JoinHandle;
+
+    /// The dispatch must return promptly (loop not blocked) and the work must
+    /// run on a DIFFERENT thread than the dispatcher (the event-loop thread).
+    ///
+    /// Fails on `origin/main`: the arm runs the purge inline, so the dispatch
+    /// would block for the full ~1s barrier and the work thread id would equal
+    /// the dispatcher's thread id.
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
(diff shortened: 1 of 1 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The offload is correctly implemented: `Arc::clone(&self.herder)` + `move || herder.purge_persisted_tx_sets()` is `Send + 'static` (Herder is `Arc<Herder>`, purge takes `&self`, returns `()`, logs its own errors), the in-flight guard correctly coalesces overlapping ticks, and shutdown `abort()` is safe. Regression test verified: fails to compile at test-commit `86e427d3` (`unresolved import super::dispatch_tx_set_gc`), both assertions pass at fix-commit `945f26f5`.

<details>

This is **cycle 1** (no prior Correctness comment). Complete change-list below; no concerns rise to CHANGES_REQUESTED.

**Test verification (kind: bug-fix) — PASS:**
- PR body has a `## Regression test` section matching the diff (`tx_set_gc_offload_tests::{test_tx_set_gc_purge_runs_off_event_loop_thread, test_tx_set_gc_serial_no_overlap}`).
- Walked the commit list: test committed first (`86e427d3`), fix second (`945f26f5`). Checked out `86e427d3` and confirmed `cargo test -p henyey-app tx_set_gc_offload --no-run` FAILS with `error[E0432]: unresolved import super::dispatch_tx_set_gc`. At `945f26f5` both tests pass (0.06s). The failing-test-first discipline is satisfied — the test cannot pass without the offload helper, and its behavioral assertions (work runs on a different thread id; peak concurrency == 1; total purges == 1 when a 2nd tick arrives in-flight) directly encode the bug.

**Correctness of the offload — PASS:**
- `Send + 'static`: `self.herder: Arc<Herder>`; `Arc::clone` then `move` into the closure is `Send` (Herder is shared across the app via Arc, so already `Send + Sync`). `purge_persisted_tx_sets(&self)` returns `()` and internally logs errors via `error!` — nothing to propagate, so the `JoinHandle<()>` carrying no result is appropriate.
- In-flight guard: `slot.as_ref().is_some_and(|h| !h.is_finished())` — on the first tick `slot` is `None` → spawns. On a later tick, if the prior handle is unfinished → returns (skip). `is_finished()` flips true once the blocking closure has run to completion, so a finished handle is correctly replaced by a fresh spawn. This is a correct, race-free coalescing because the slot is loop-local (single-threaded access from the event loop only). No leak: the finished `JoinHandle` is dropped (detached) when overwritten; a detached finished `spawn_blocking` handle frees its slot cleanly.
- `set_phase(33)` deliberately stays on the loop thread so the watchdog phase label is preserved — good; the phase is now a transient marker (arm returns immediately) rather than a 39s freeze.

**Shutdown abort — PASS (test-coverage note, non-blocking, inline):**
- `tx_set_gc_task.take()` + `h.abort()` on the shutdown arm. `abort()` on a `spawn_blocking` task does NOT force-cancel an already-running closure (tokio cannot preempt a blocking thread) — it only prevents a not-yet-started task from starting and detaches the handle. The purge is one atomic idempotent SQLite txn (#2770), so a mid-flight purge running to completion during shutdown is harmless, and an aborted not-yet-started purge re-runs next startup (the GC also runs in `restore_scp_state`). Correct.
- Minor (inline, non-blocking): the shutdown abort path is not directly unit-tested. The existing `watchdog_shutdown` integration test exercises shutdown but not specifically the in-flight-purge-abort interleaving. Acceptable as a follow-up — the abort is a 2-line, well-reasoned path.

**Readability / error handling — PASS:**
- The doc comment on `dispatch_tx_set_gc` and the inline comments on the arm and shutdown are thorough and cite stellar-core line numbers and the relevant issues (#2770, #3532). Error handling is unchanged (purge logs internally). The generic `F: FnOnce() + Send + 'static` seam keeps the helper testable without exposing Herder internals — matches the plan's non-public seam.

No blocking concerns. APPROVE.
</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** No regression in observable behavior, parity, data format, or API. The purge contents are unchanged and GC remains strictly serial, so determinism/parity is preserved. One residual-risk observation (`regression-risk`: a slow purge can still transiently block the loop *via the `last_slot_saved` write lock* the persist path shares) is real but materially smaller than the bug being fixed and does not warrant blocking — recorded as a non-blocking inline + follow-up.

<details>

This is **cycle 1** (no prior Risk comment). Complete change-list, concerns labeled by class.

**`determinism-parity` — PASS (no change):**
- The GC effect is identical: same `purge_unreferenced_tx_sets_atomic` SQL, same serial cadence (in-flight guard reproduces stellar-core rescheduling `startTxSetGCTimer()` only after the purge returns — HerderImpl.cpp:2440-2487). GC only deletes orphaned/unreferenced persisted tx-sets, never ledger/consensus/bucket state, so the thread it runs on is not observable. No consensus, ledger-close, or hash-affecting path is touched. Parity tripwires byte-identical per PR body (scp_parity 124/124 etc.). Confirmed app-crate-only (single file `crates/app/src/app/lifecycle.rs`).

**`api-break` / `migration-risk` / `config-risk` — PASS (none):**
- No public API change (helper is a private free fn), no data-format change (purge SQL untouched), no config/migration. `--all -Dwarnings` builds clean (verified locally: app crate + tests compile, 2/2 regression tests pass).

**`regression-risk` (residual, NON-BLOCKING — inline + follow-up):**
- The offloaded purge calls `ScpPersistenceManager::purge_unreferenced_tx_sets` (persistence.rs:638), which takes `self.last_slot_saved.write()` and holds it for the **entire** read-then-delete SQLite transaction. The event-loop persist path — `process_verified` → `emit` → persist_callback → `persist_scp_state` (persistence.rs:510) — takes the **same** `last_slot_saved.write()` synchronously on the event-loop thread (`process_verified` is not itself offloaded; lifecycle.rs:2378/2434).
- Consequence: moving the purge to `spawn_blocking` removes the *unconditional* per-tick freeze, but if a `persist_scp_state` happens to fire while a slow purge holds the write lock, the event loop still blocks on `parking_lot::RwLock::write()` for up to the remaining purge duration. The worst-case freeze is therefore reduced in **probability** (only the persist↔purge overlap window, not every tick) but the worst-case **duration** is still bounded by the slow purge.
- Why this is non-blocking, not CHANGES_REQUESTED: (1) it is a strict improvement over `origin/main`, where the purge froze the loop on *every* tick regardless of any persist; (2) eliminating it entirely requires changing the herder/db locking model (e.g. moving the persist write off-loop too, or coarsening/replacing the manager-level `last_slot_saved` lock), which is explicitly out of this PR's app-crate-only scope and overlaps #3497's operator-gated DB-lock policy decision; (3) the persist write critical section is short (a few INSERTs) so the contended window is narrow. Filing as backlog rather than blocking the freeze fix.

**`regression-risk` — shutdown task lifetime — PASS:**
- The in-flight handle is `abort()`ed on the shutdown arm and the loop breaks; no detached `spawn_blocking` survives loop exit beyond its own (harmless, atomic) completion. No fire-and-forget leak across runtime teardown.

No blocking risk. APPROVE.
</details>

tomerweller:
## ✅ Merged

**Commit:** 9781c5378f63ade93997169a8045033064e1ed14

**Reviewer A (Correctness):** APPROVE — offload is `Send + 'static`-correct; in-flight guard coalesces; regression test verified fails-at-test-commit / passes-at-fix-commit.
**Reviewer B (Risk):** APPROVE — no parity/determinism/API/format change; one residual non-blocking risk filed as follow-up.

**Offload-safety determination (operator REFUTE focus):**
1. In-flight guard genuinely prevents overlapping purges — CONFIRMED (loop-local `Option<JoinHandle>`, single-threaded access, `is_finished()` gate; `test_tx_set_gc_serial_no_overlap` asserts peak==1, total==1). Matches stellar-core serial reschedule.
2. `abort()` on shutdown is safe — CONFIRMED. `abort()` cannot preempt a running `spawn_blocking` closure (runs to completion); the purge is one atomic idempotent txn (#2770), so completion-or-skip is harmless.
3. No determinism/parity change — CONFIRMED. Same SQL, same serial cadence, GC only deletes orphaned tx-sets; thread is unobservable.

**Follow-up issues filed for unaddressed inline review comments:** none (no inline threads).

**Follow-up issues filed for non-blocking reviewer concerns:** #3538 (offloaded purge can still transiently block the loop via the shared `last_slot_saved` write lock — out of this PR's app-crate scope; overlaps #3497).

**Note:** required CI green (Test/Build (release-ci)/Clippy/Doctest/Format/gitlink/Soroban Pin all SUCCESS). The first Test run hit an unrelated flake (`henyey-ledger::manager::tests::test_scan_and_merge_pending_buffer_bounded`, an untouched crate; passed 3/3 locally on the merged tree); a `--failed` re-run went green. Non-required testnet Quickstart shard still pending (UNSTABLE) — merged per policy.

This also resolves the root cause of #3501 (the synchronous phase=33 `purge_unreferenced_tx_sets` near-tip event-loop stall). #3501's separate `wal_autocheckpoint` / maintainer-coordination asks remain open for triage.

Issue #3543, opened 2026-06-22
overlay: remove the dead, never-constructed SurveyManager shell (superseded by app/survey.rs)


tomerweller:
Discovered during the #3517 plan (and relevant to #3500).

## Finding

`crates/overlay/src/survey.rs` defines a full `SurveyManager` (with `CollectingNodeData`, `CollectingPeerData`, local `TimeSlicedNodeData`, `start_collecting`/`stop_collecting`/`get_node_data`/`stats`/`record_added_peer`/`record_dropped_peer`/`modify_node_data`). Verified against origin/main @ 89af0aa0: **this `SurveyManager` is never constructed outside its own `#[cfg(test)]` block.** All production survey behavior — including the wire `TimeSlicedNodeData` response, SCP-latency p75, and added/dropped-peer counters — lives in the **app crate** `crates/app/src/survey.rs` (`SurveyDataManager`) and `crates/app/src/app/survey_impl.rs`. The overlay copy is a stale, superseded duplicate.

`crates/overlay/PARITY_STATUS.md:47` currently misrepresents this: `SurveyDataManager.* → survey.rs \"Merged into SurveyManager\"`. The live SurveyDataManager is the app-crate one.

## Proposed work

Evaluate removing the entire overlay `SurveyManager` (and its `pub use` re-exports in `crates/overlay/src/lib.rs`) as dead code, OR — if any of it is intended as a future overlay-layer home — document why it exists unused. Update PARITY_STATUS.md/README to point at the app layer as the survey owner.

## Relationship to siblings

- #3517 (this issue's parent) does the narrow drop of just the SCP-latency fields from this shell.
- **#3500** proposes *wiring* `record_dropped_peer`/`record_added_peer` in this same shell into production — that premise is in tension with this finding (it would be wiring dead code). #3500 should be re-evaluated against this discovery. **/do MUST serialize #3517, #3500, and this issue — same module.**

Parent: #3517 (carved from #3461). Crate is parity-sensitive (P2P wire protocol).

tomerweller:
## Triage Report

**Verdict:** ACCEPT

**Type:** enhancement
**Kind:** refactor
**Severity:** low
**Crate:** crate:overlay
**Size:** M

[...]

PR #3545, opened 2026-06-22
Suppress duplicate added-peer count on mutual-dial replacement

### LINKED ISSUE #3500: overlay: wire survey recordDroppedPeer/node-data counters into authenticated-peer lifecycle (parity decision)
Parity-decision follow-up from #3444 (finding O6).

## Problem

`SurveyDataManager::record_dropped_peer()` (crates/overlay/src/survey.rs:610) is implemented with correct internal semantics (phase-gated to `Collecting`, removes peer from inbound/outbound data, increments `dropped_authenticated_peers`) but has **zero production callers**. As a result, a henyey survey response always reports `dropped_peers: 0` (read at survey.rs:509), diverging from stellar-core.

The symmetric add path is *also* production-unwired: `record_added_peer()` / `modify_node_data()` (survey.rs:631 / :577) are only test-called. henyey's `added_authenticated_peers` AtomicU64 (manager/mod.rs:570) is a **separate** metric, not the survey node-data counter.

## stellar-core reference (v26.0.1)

- `SurveyManager::recordDroppedPeer` is fired from `OverlayManagerImpl::PeersList::removePeer` **only when an authenticated peer is dropped** (OverlayManagerImpl.cpp:154), after erasing from `mAuthenticated`.
- It increments `mDroppedAuthenticatedPeers` only during `COLLECTING` phase (SurveyDataManager.cpp:248-268).
- The symmetric add path fires from `moveToAuthenticated` via `modifyNodeData([&]{ ++mAddedAuthenticatedPeers; })` (OverlayManagerImpl.cpp:200).

## Parity decision needed (operator input)

This is a **behavior-changing** wire-output fix, deliberately scoped OUT of the #3444 behavior-neutral cleanup PR (which only corrected the false `Full` PARITY_STATUS rows to `None`/`Partial`). To resolve:

1. Identify henyey's authenticated-peer-drop site (manager/connection.rs cleanup/disconnect path) and thread the survey handle to call `record_dropped_peer` there — matching stellar-core's authenticated-only, phase-gated semantics.
2. Decide whether to also wire the symmetric add path (`record_added_peer`/`modify_node_data`) for consistency, vs. reconciling with the existing `added_authenticated_peers` AtomicU64.
3. Add a regression test asserting the survey dropped-peer counter increments on authenticated-peer drop during `Collecting`.
4. Restore the PARITY_STATUS rows to `Full` once wired.

## Why a follow-up, not part of #3444

Per #3444's triage CRITICAL FLAG, O6 is a parity decision (observable survey wire output), not mechanical cleanup, and must not be bundled into the behavior-neutral O1-O5 removal commit.

tomerweller:
Closes #3500

## Summary

The mutual-dial "new wins" tiebreaker in `OverlayManager::try_register_peer` (`crates/overlay/src/manager/connection.rs`) replaces an existing authenticated `PeerHandle` in place but then unconditionally bumped `added_authenticated_peers` again. Because the displaced peer's slot was already counted at its original registration, a single surviving steady-state peer was reported as `+2 add / 0 drop`, diverging from stellar-core v27.0.0, which reports `+1 add / 0 drop` (it resolves the same collision pre-authentication at HELLO — `Peer::recvHello` `ERR_CONF` "already-connected", Peer.cpp:1883-1896 — so the loser is dropped while still pending: no `++mAddedAuthenticatedPeers`, no `recordDroppedPeer`).

Fix: gate the `added_authenticated_peers.fetch_add(1, ..)` on `replaced.is_none()`, so a replacement (which reuses an already-counted slot) does not double-count. Suppressing the duplicate add (rather than bumping a drop) is the absolute-counter match to stellar-core: `+1 add / 0 drop`. The drop-counter path (`cleanup_peer`'s authenticated-only `removed.is_some()` gate) is untouched; the stale-generation cleanup of the displaced peer remains a no-op.

This is non-consensus, observability-only: `added_authenticated_peers` feeds only the survey-response XDR (`TimeSlicedNodeData.added_authenticated_peers`) via the app delta path in `crates/app/src/app/survey_impl.rs` (confirmed sole consumer). No consensus, ledger, herder, or SCP path changes.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3500#issuecomment-4765460760)

## Test plan

- [x] cargo fmt --check
- [x] RUSTFLAGS="-Dwarnings" cargo build -p henyey-overlay -p henyey-app --all-targets
- [x] cargo test --all (green)
- [x] cargo test -p henyey-overlay tiebreaker:: (11 passed)

- **Test:** `crates/overlay/src/manager/connection.rs::test_mutual_dial_replacement_does_not_double_count_added_peers`
- **Pre-fix:** committed as `5a8d8a412df0cffe89536d102e4dd63204ad50d4` — verified FAILED with `assertion left == right failed ... left: 1, right: 0` (the replacement double-counted the add atomic).
- **Post-fix:** verified PASSES after `d1bdf2619df128037c577c3bd74e41e3a27e8048`.
- Two guard tests also added: `test_normal_registration_still_counts_added_peer` (vacant entry still +1) and `test_mutual_dial_loser_does_not_count` (rejected duplicate stays +0).

## Deviations from plan

- The converged plan also re-pointed two rows in `crates/overlay/PARITY_STATUS.md`. Per operator direction (#3543 owns ALL survey-row PARITY_STATUS edits and is being implemented in parallel), this PR is scoped to `connection.rs` + tests ONLY to remain file-disjoint from #3543. No PARITY_STATUS.md edits here.

### DIFF
--- a/crates/overlay/src/manager/connection.rs
+++ b/crates/overlay/src/manager/connection.rs
@@ -323,9 +323,20 @@ impl OverlayManager {
             }
         };
         shared.peer_info_cache.insert(peer_id.clone(), peer_info);
-        shared
-            .added_authenticated_peers
-            .fetch_add(1, Ordering::Relaxed);
+        // Only count a NEWLY-occupied authenticated slot. When this registration
+        // REPLACED an existing authenticated handle (the mutual-dial "new wins"
+        // tiebreaker above), the slot was already counted at the displaced peer's
+        // original registration, so re-counting would double-bump (+2 add / 0 drop
+        // for a single surviving peer). stellar-core v27.0.0 resolves the same
+        // collision pre-auth at HELLO (Peer::recvHello ERR_CONF "already-connected"),
+        // dropping the loser while still pending — never ++mAddedAuthenticatedPeers,
+        // never recordDroppedPeer. Suppressing the duplicate add (NOT bumping a drop)
+        // matches core's absolute counters: +1 add / 0 drop. (#3500)
+        if replaced.is_none() {
+            shared
+                .added_authenticated_peers
+                .fetch_add(1, Ordering::Relaxed);
+        }
         Ok(RegisterResult {
             outbound_rx,
             flow_control,
--- a/crates/overlay/src/manager/connection.rs
+++ b/crates/overlay/src/manager/connection.rs
@@ -2849,5 +2849,130 @@ mod tests {
             shared.cleanup_peer(&remote, new_gen);
             assert!(!shared.peers.contains_key(&remote));
         }
+
+        #[tokio::test]
+        async fn test_mutual_dial_replacement_does_not_double_count_added_peers() {
+            // Regression for #3500: the mutual-dial "new wins" replacement branch
+            // reuses an already-counted authenticated slot, so it must NOT bump the
+            // `added_authenticated_peers` atomic again. stellar-core v27.0.0 resolves
+            // the same collision pre-auth at HELLO (Peer::recvHello ERR_CONF
+            // "already-connected") so the loser is dropped while still pending —
+            // never counted as added, never recordDroppedPeer'd. Net core: +1/0.
+            // henyey must match: a winning replacement adds 0 (the slot was counted
+            // at the displaced peer's original registration) and the stale-generation
+            // cleanup drops 0.
+            let local = low_id();
+            let remote = high_id();
+            let shared = make_shared(local);
+            // Seed an existing authenticated outbound peer at generation 0. NOTE:
+            // insert_fake does NOT bump the add atomic, so we sample the baseline
+            // immediately before the replacing call and assert the *delta* across it.
+            insert_fake(&shared, &remote, ConnectionDirection::Outbound);
+
+            let added_before = shared.added_authenticated_peers.load(Ordering::Relaxed);
+            let dropped_before = shared.dropped_authenticated_peers.load(Ordering::Relaxed);
+
+            // Winning cross-direction inbound registration (local < remote ⇒ accept
+            // inbound ⇒ new wins ⇒ replaced.is_some()).
+            let peer = crate::peer::Peer::new_test_inbound(
+                remote.clone(),
+                false,
+                Arc::clone(&shared.metrics),
+            );
+            let reg = OverlayManager::try_register_peer(
+                &peer,
+                &remote,
+                peer.info().clone(),
+                &shared,
+                1024,
+            )
+            .expect("replacement should succeed");
+            assert!(
+                reg.replaced.is_some(),
+                "this case must be a replacement to exercise the double-count path"
+            );
+
+            // The replacement reused an already-counted slot: the add atomic must
+            // not have moved. On origin/main this is +1 (the bug) — the test fails.
+            assert_eq!(
+                shared.added_authenticated_peers.load(Ordering::Relaxed),
+                added_before,
+                "mutual-dial replacement must not bump added_authenticated_peers (double count)"
+            );
+
+            // Stale-generation cleanup of the displaced peer (the caller drops the
+            // old handle; its peer_loop calls cleanup_peer with the old generation):
+            // the new generation is installed, so this is a no-op and must not bump
+            // the drop atomic.
+            shared.cleanup_peer(&remote, 0);
+            assert_eq!(
+                shared.dropped_authenticated_peers.load(Ordering::Relaxed),
+                dropped_before,
+                "stale-generation cleanup must not bump dropped_authenticated_peers"
+            );
+            assert!(shared.peers.contains_key(&remote));
+        }
+
+        #[tokio::test]
+        async fn test_normal_registration_still_counts_added_peer() {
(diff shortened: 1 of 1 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Cycle 1. bug-fix kind. Failing-test-first verified (test fails at the test-only commit, passes at the fix); fix is a one-line, well-scoped gate; two guard tests prevent over-correction. No concerns.

<details>

- PR body has a `## Regression test` section naming `test_mutual_dial_replacement_does_not_double_count_added_peers`, committed alone in `5a8d8a41` (before the fix `d1bdf261`). Verified by walking the commit list and checking out the parent of the fix.
- Checked out `5a8d8a41` (test-only): production `fetch_add(1, Relaxed)` is unconditional there; ran the test → **FAILS** with `assertion left == right failed ... left: 1, right: 0` at connection.rs:2897. The regression test genuinely captures the bug.
- Checked out `d1bdf261` (fix): ran `cargo test -p henyey-overlay tiebreaker::` → 11 passed, 0 failed, including the regression + both guards.
- Guard 1 `test_normal_registration_still_counts_added_peer`: vacant entry (`replaced.is_none()`) still bumps +1 — confirms the gate does NOT suppress legitimate adds.
- Guard 2 `test_mutual_dial_loser_does_not_count`: rejected duplicate (existing-wins → early `Err`) leaves the add atomic unchanged — pre-existing behavior preserved.

**Logic / readability:**
- The gate `if replaced.is_none() { fetch_add(1) }` is placed after the `peer_info_cache.insert`, identical position to the original bump, so no ordering change for the new-slot case. `replaced` is computed by the `match` on the dashmap entry: `None` for Vacant, `Some(ReplacedPeer)` only on the new-wins replacement branch; the existing-wins branch early-returns `Err` before reaching the bump. Gate is exactly correct.
- Comment is accurate and cites the stellar-core reference.
- Single file (`connection.rs`); no PARITY_STATUS.md edits (file-disjoint from #3543/#3544 per the deviation note in the PR body).

**Error handling:** unchanged. No new error paths.

No blocking concerns.
</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** The fix matches stellar-core v27.0.0 absolute counters (+1 add / 0 drop for a surviving mutual-dial peer). Suppressing the duplicate ADD — rather than bumping a DROP — is the correct match, confirmed against the pre-auth-drop path in `Peer::recvHello` + `PeersList::removePeer`/`moveToAuthenticated`.

<details>

**class: parity-gap (verification, not a defect)**

Traced the three counter-mutation sites in `stellar-core/src/overlay/`:

- `++mAddedAuthenticatedPeers` occurs in EXACTLY one place: `PeersList::moveToAuthenticated` (OverlayManagerImpl.cpp:201), on the pending→authenticated transition.
- `recordDroppedPeer` occurs in EXACTLY one place: `PeersList::removePeer` (OverlayManagerImpl.cpp:154), and ONLY in the `mAuthenticated` branch. If the peer is still in `mPending`, `removePeer` takes the pending branch (cpp:131-141) and `return`s BEFORE reaching `recordDroppedPeer` — no drop counted.
- The mutual-dial duplicate is resolved PRE-AUTHENTICATION in `Peer::recvHello` (Peer.cpp:1883-1896): when the incoming peer's id is already in `getAuthenticatedPeers()`, core `sendErrorAndDrop(ERR_CONF, "already-connected ...")`. The losing connection is dropped while STILL PENDING.

Net stellar-core for a single surviving steady-state peer in a mutual-dial collision: the survivor was counted once at its own `moveToAuthenticated` (+1 add); the duplicate is dropped pending (no add, no drop). Result: **+1 add / 0 drop**.

**Why suppress-add is the correct match (refuting the alternative "bump a drop"):**
A reviewer might argue henyey should instead bump `dropped_authenticated_peers` to balance the extra add. That would produce +2 add / +1 drop, which does NOT match core's absolute counters (+1/+0). Core never records a drop for the mutual-dial loser because it never authenticated (dropped while pending). henyey's model authenticates first then replaces in-place, so the only divergence is the spurious second ADD; the correct fix is to suppress that ADD. The PR does exactly this. The drop path (`cleanup_peer`'s authenticated-only `removed.is_some()` gate) is correctly left UNTOUCHED — the displaced handle's stale-generation cleanup is a no-op (verified by the regression test: drop stayed +0). This is the absolute-counter match. **REFUTED the bump-a-drop alternative.**

**Sequencing / edge cases:** The gate keys on `replaced.is_none()`, which is `Some` only on the new-wins replacement branch; the existing-wins branch early-returns `Err(AlreadyConnected)` before the bump (mirrors core rejecting the loser). Vacant entry → `None` → still +1. All branches map correctly to core semantics.

**Scope:** observability-only counter (sole consumer is the time-sliced survey XDR via survey.rs → survey_impl.rs). No consensus/ledger/herder/SCP behavior change. No PARITY_STATUS.md edits (file-disjoint from #3543/#3544 per operator direction).

</details>

Issue #3384, opened 2026-06-17
pipeline: /plan + critics mis-reason about dead-code reachability under -Dwarnings (cfg(test) callers don't keep items alive)


tomerweller:
## Anomaly (self-reflection — recurring, 2 occurrences)

Two `/plan`-stage plans converged with all 3 critics APPROVE, then got bounced at `/do` for the **same root-cause reasoning error**: the plan asserted a symbol was safe to narrow (`pub`→`pub(crate)`) based on a 'zero external callers' grep, but the symbol had **zero non-test callers at all** — so under this repo's global `-Dwarnings` (`.cargo/config.toml`), demoting it surfaced hard dead-code compile errors.

### Occurrences
1. **#3363 (overlay)** — plan claimed `MessageDispatcher`/`ConnectionPool::release` were 'reached via the OverlayManager facade'. False: the entire `message_handlers` module was vestigial (zero production reachability); `release` had only a `#[cfg(test)]` caller. `/do` bounced; re-plan correctly switched to *delete the dead code* (option b).
2. **#3365 (scp)** — plan claimed the narrowed info-types/methods were 'kept green by in-crate/`#[cfg(test)]` callers'. False: they had zero non-test callers (genuinely public-API-only), so `pub(crate)` produced **15 dead-code errors across 6 files**. `/do` bounced (F1 was fine and is preserved; F2/F3 need re-decision).

### Root cause
The plan + critics reason about visibility-narrowing/dead-code findings by **grepping for callers** and distinguishing only 'external vs internal'. But under `-Dwarnings`:
- A `#[cfg(test)]`-only caller does **not** keep an item alive in a normal `cargo build` — narrowing/keeping it then trips `dead_code`.
- 'Zero external callers' does not distinguish 'has in-crate **non-test** callers' (safe to narrow) from 'genuinely dead' (narrowing → dead-code error; the right move is *delete* or keep-pub).

A grep can't see this; only **building with the narrowing applied** can.

### Proposed fix (operator-gated — touches pipeline skills)
[...]

PR #3547, opened 2026-06-22
Route stateless fresh-genesis recovery to bucket-apply, never FATAL at genesis

### LINKED ISSUE #3410: SSC: Henyey enters prolonged catchup recovery loop when peers disconnect before first sync
## Observed behavior

During the mixed-image SSC mission, when Henyey starts from genesis and stellar-core peers begin closing ledgers, if Henyey fails to sync before the consensus stall timeout fires, it enters a recovery catchup cycle that loses time and may produce a fatal state error.

## Sequence

1. Henyey starts, force-bootstraps SCP from genesis (ledger 1)
2. Overlay eventually connects to stellar-core peers (~10s)
3. Henyey enters Validating state, tracking slot 2
4. If peers connect but Henyey hasn't externalized slot 2's EXTERNALIZE before the stall timeout (~35s), sync is lost
5. Recovery catchup fires: Henyey targets the latest archive checkpoint (ledger N from old-core)
6. First catchup attempt downloads ledgers from the archive but hits `knit-to-LCL failed at LCL` because Henyey's genesis (ledger 1 hash) differs from the archive's genesis
7. Henyey self-heals by forcing a full bucket-apply rebuild, which succeeds, but the process is slow
8. By this point, the SSC mission timeout often fires before Henyey rejoins consensus

## Impact

Henyey's cold-start sync is fragile under SSC timing. The knit-to-LCL mismatch is expected (different genesis from peers that started earlier), but the recovery is slow and doesn't reliably beat the consensus stall timeout. This manifests as the Henyey pod eventually crashing with `FATAL: unrecoverable local state failure` after repeated failed catchup attempts.

## Suggested approach

- Reduce the initial consensus stall timeout for fresh-genesis validators in SSC test environments (they have no state to protect).
- Consider skipping the first knit-to-LCL check when the LCL is genesis (ledger 1) and the archive has a later checkpoint — just accept the archive's state directly.
- Shorten the inter-catchup-retry backoff for the first few attempts on a fresh node.

## Environment

- Henyey running under Stellar Supercluster (SSC) on Namespace Kubernetes
- Mixed-image mission: 2x `stellar/stellar-core:latest` + 1x `henyey-ssc`
- Henyey starts from genesis while stellar-core peers are already at ledger 60+

tomerweller:
Closes #3410

## Summary

A fresh-genesis Henyey under SSC (LCL == ledger 1) joining already-advanced stellar-core peers fails to externalize slot 2 before the consensus-stall timeout, falls into recovery catchup, replays from genesis, and runs `verify_knit_to_lcl` against its force-bootstrapped genesis snapshot. Under the mixed-image mission that snapshot legitimately differs from the earlier-started peers' archive history chain, so the knit hits a `KnitLcl*` error. That class is `is_fatal_catchup_failure()`, so `handle_catchup_result` escalated straight to `FATAL: unrecoverable local state failure` — bypassing the #3288 self-heal latch — and the pod crash-looped.

This PR implements the operator-approved **genesis-gating** approach (two narrowly-scoped, `LCL == genesis`-gated changes in `crates/app/src/app/catchup_impl.rs`):

1. **Load-bearing — route genesis recovery to bucket-apply.** `genesis_recovery_catchup_plan()` forces the effective catchup depth to `Minimal` when LCL is exactly genesis, so `CatchupRange::calculate` selects `BucketsOnly` / `BucketApplyAndReplay` (archive-authoritative bucket-apply, mirroring core's fresh-node `CatchupWork::downloadApplyBuckets`) instead of replay-from-genesis. Local genesis state is not seeded. The knit-to-LCL check then runs only in the post-checkpoint replay phase (LCL > genesis), never against genesis.
2. **Defense-in-depth — don't FATAL at genesis.** In the `Err` arm, `at_genesis` is read at the top (before any state mutation); a knit-to-LCL error (`HistoryError::is_knit_to_lcl_failure`) is reclassified as a retryable bucket-apply (`catchup_needs_full_reset`) instead of `trigger_fatal_shutdown` — **only when `at_genesis`**.

Also: the inter-catchup-retry backoff is shortened for a fresh-genesis node (3s vs the 10s steady-state cooldown) so the bucket-apply re-attempt races the SSC consensus-stall timeout.

**Operator signed off the genesis-gating approach** (archive-authoritative bucket-apply + do-not-FATAL-at-genesis), not the narrower timeout-only lever.

**For `LCL > genesis`, the `CatchupRange` selection and the knit-to-LCL fatal classification are byte-for-byte unchanged** — the #3282/#3288 terminal-wipe semantics for a node with committed state are fully preserved. The deviation is documented in `crates/app/PARITY_STATUS.md` (Architectural Difference 5).

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3410#issuecomment-4737099707)

## Test plan

- [x] cargo fmt --check
- [x] cargo clippy -p henyey-app -p henyey-history --all-targets -- -D warnings
- [x] RUSTFLAGS=-Dwarnings cargo build -p henyey-app -p henyey-history --all-targets (#3384 build gate)
- [x] cargo build --all
- [x] cargo test --all green (one pre-existing parallel-tracing-capture flake, `verify::tests::test_verify_bucket_hash_mismatch_emits_tracing_no_ledger_seq`, passes in isolation and on retry — unrelated to this diff which touches no tracing/verify code)

- **Test:** `crates/app/src/app/catchup_impl.rs::test_genesis_knit_mismatch_does_not_trigger_fatal_shutdown`
- **Pre-fix:** verified FAILED on `origin/main` with: `a knit-to-LCL mismatch at genesis must NOT be fatal` (the node FATALs/loops).
- **Post-fix:** verified PASSES after the fix commit (`ecb93f98`).
- **Guard tests (preserve #3288):** `test_nongenesis_knit_mismatch_still_fatal` (knit above genesis still FATALs/wipes) and the adjusted `test_full_bucket_apply_knit_mismatch_still_wipes` both PASS.
- **Helper/classifier coverage:** `test_genesis_recovery_forces_bucket_apply_mode` (genesis → Minimal + no seed; above genesis → unchanged) and `crates/history/src/error.rs::test_is_knit_to_lcl_failure`.

Note: the tests and fix landed in a single commit because the new `genesis_recovery_catchup_plan` helper is unused (dead code) without its live call site, which the `-D warnings` pre-commit gate (#3384) rejects.

## Deviations from plan

- The plan's two `crates/history/tests/replay_integration.rs` integration tests were not added. Their behavioral assertion — "genesis recovery selects bucket-apply, not the doomed replay-from-genesis knit" — is captured deterministically by the `test_genesis_recovery_forces_bucket_apply_mode` unit test over `genesis_recovery_catchup_plan` (which drives `CatchupRange` mode selection) plus the existing `catchup_range.rs::test_minimal_from_genesis_to_checkpoint` / `test_case3_minimal_non_checkpoint` cases, avoiding a heavy/flaky HTTP-fixture archive with a synthetic divergent genesis. The required failing-test-first and #3288-guard tests (operator directive) are the app-level ones above.

### DIFF
--- a/crates/app/PARITY_STATUS.md
+++ b/crates/app/PARITY_STATUS.md
@@ -265,6 +265,32 @@ Features not yet implemented. These ARE counted against parity %.
    - **Rust**: `henyey-app` owns restart/catchup policy while low-level persistence APIs live in `henyey-db`.
    - **Rationale**: The workspace factors storage concerns into a dedicated crate instead of keeping them inside the app layer.
 
+5. **Stateless fresh-genesis recovery (`src/app/catchup_impl.rs`, #3410)** — operator-approved deviation
+   - **stellar-core**: a fresh node joining a network with existing history
+     adopts archive state via `CatchupWork::downloadApplyBuckets`; a knit-to-LCL
+     disagreement is surfaced as a retryable `CatchupWork::fatalFailure()`
+     boolean, and core has no self-wipe-on-corruption mechanism at all.
+   - **Rust**: when a node's LCL is **exactly genesis** (ledger 1, i.e. no
+     committed local state), a forced recovery catchup is routed to an
+     archive-authoritative bucket-apply (effective `Minimal` depth, so
+     `CatchupRange::calculate` selects `BucketsOnly` / `BucketApplyAndReplay`
+     instead of replay-from-genesis), and a knit-to-LCL mismatch at genesis is
+     reclassified as a retryable bucket-apply (`catchup_needs_full_reset`)
+     rather than escalating to `FATAL: unrecoverable local state failure`. The
+     inter-catchup-retry backoff is also shortened at genesis (3s vs the 10s
+     steady-state cooldown) so a fresh node re-attempts the bucket-apply quickly.
+   - **Rationale**: under the SSC mixed-image mission a fresh Henyey's
+     force-bootstrapped genesis hash legitimately differs from the
+     earlier-started peers' archive history chain, so a replay-from-genesis
+     would always fail `verify_knit_to_lcl` and the resulting FATAL was a
+     cold-start crash, not real corruption (#3410). The carveout is gated
+     **strictly on `LCL == genesis`**: for `LCL > genesis` the `CatchupRange`
+     selection and the knit-to-LCL fatal classification are byte-for-byte
+     unchanged, fully preserving the #3282/#3288 terminal-wipe semantics for a
+     node that has real state to protect. Change 2 (don't-FATAL-at-genesis) is
+     additive henyey-specific robustness with no analogue or divergence for
+     core's non-genesis path.
+
 ## Test Coverage
 
 | Area | stellar-core Tests | Rust Tests | Notes |
--- a/crates/app/src/app/catchup_impl.rs
+++ b/crates/app/src/app/catchup_impl.rs
@@ -3,6 +3,48 @@
 use super::archive_cache::CacheResult;
 use super::*;
 
+/// Decide how a forced **recovery catchup** should run given the node's
+/// current LCL and the operator's configured catchup mode.
+///
+/// This is the single safety hinge for the stateless fresh-genesis carveout
+/// (#3410). A node whose LCL is exactly genesis (ledger 1) has, by
+/// definition, no externalized/committed ledger state of its own. Under the
+/// SSC mixed-image mission its force-bootstrapped genesis hash legitimately
+/// differs from the (earlier-started) peers' archive history chain, so a
+/// replay-from-genesis catchup would compare the archive's genesis-chain
+/// header against the local genesis snapshot and fail `verify_knit_to_lcl`
+/// with a `KnitLcl*` error — the structurally-doomed first attempt described
+/// in #3410.
+///
+/// The fix: when `current == GENESIS_LEDGER_SEQ`, force the effective catchup
+/// depth to [`CatchupMode::Minimal`] (count = 0). At genesis,
+/// `CatchupRange::calculate` then yields `BucketsOnly` (checkpoint target) or
+/// `BucketApplyAndReplay` (non-checkpoint target) — i.e. it adopts the
+/// archive's state via bucket-apply, exactly as stellar-core's fresh-node
+/// `CatchupWork::downloadApplyBuckets` does. The knit-to-LCL check then runs
+/// only in the post-checkpoint replay phase (LCL > genesis at that point),
+/// never against the genesis snapshot. `seed_local_state == false` so the
+/// genesis branch does NOT clone the local genesis bucket state for replay.
+///
+/// For `LCL > genesis` the operator's configured mode is returned unchanged
+/// and local state is seeded as before — `CatchupRange` selection and the
+/// knit-fatal classification are byte-for-byte unchanged (preserving the
+/// #3282/#3288 semantics).
+///
+/// Returns `(effective_mode, seed_local_state)`.
+pub(crate) fn genesis_recovery_catchup_plan(
+    current: u32,
+    configured_mode: CatchupMode,
+) -> (CatchupMode, bool) {
+    if current == GENESIS_LEDGER_SEQ {
+        // Stateless fresh node: route to archive-authoritative bucket-apply.
+        (CatchupMode::Minimal, false)
+    } else {
+        // Has committed local state: unchanged behavior.
+        (configured_mode, true)
+    }
+}
+
 impl App {
     pub(crate) fn should_skip_externalized_catchup_cooldown(
         target_checkpoint: u32,
@@ -193,6 +235,34 @@ impl App {
             );
         }
 
+        // #3410 change 1 (load-bearing): a stateless fresh-genesis node
+        // (LCL == ledger 1) forced into recovery catchup must NOT replay from
+        // genesis. Under the SSC mixed-image mission its force-bootstrapped
+        // genesis hash differs from the (earlier-started) peers' archive
+        // history chain, so a replay-from-genesis would compare the archive's
+        // genesis-chain header against the local genesis snapshot and fail
+        // `verify_knit_to_lcl` with a `KnitLcl*` error — the structurally
+        // doomed first attempt. Force the effective depth to `Minimal` (count
+        // = 0) so `CatchupRange::calculate` yields `BucketsOnly` /
+        // `BucketApplyAndReplay` — archive-authoritative bucket-apply,
+        // mirroring core's fresh-node `CatchupWork::downloadApplyBuckets`. The
+        // knit-to-LCL check then runs only in the post-checkpoint replay phase
+        // (LCL > genesis at that point), never against genesis.
+        //
+        // For LCL > genesis the configured mode and local-state seeding are
+        // returned unchanged — `CatchupRange` selection is byte-for-byte
+        // identical (preserves #3282/#3288). `force_full` already ignores
+        // local state, so the seeding decision below also short-circuits on it.
+        let (mode, seed_local_state) = genesis_recovery_catchup_plan(current, mode);
+        if current == GENESIS_LEDGER_SEQ {
+            tracing::info!(
+                target_ledger,
+                "Fresh-genesis recovery catchup — routing to archive-authoritative \
+                 bucket-apply (effective Minimal depth), skipping replay-from-genesis \
+                 knit (#3410)"
+            );
+        }
+
         // #3282: track whether this catchup is seeded from CLONED LOCAL state.
         // Only the near-tip fast path below (initialized ledger manager, no
         // force_full) clones the live LCL header and sets `override_lcl`,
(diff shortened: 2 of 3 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The genesis carveout is correctly gated, the #3288 non-genesis FATAL path is byte-for-byte preserved, and the regression test is a genuine failing-test-first (verified FAILS on origin/main, PASSES on HEAD). One non-blocking scope observation about the helper applying to all `catchup_with_run_mode` callers is filed as an inline note, not a blocker.

<details>

**Cycle:** 1 (no prior Correctness comment). Complete change-list below; nothing held back.

### Failing-test-first (kind: bug-fix) — VERIFIED
- `test_genesis_knit_mismatch_does_not_trigger_fatal_shutdown` uses only pre-existing APIs (`App::new`, `KnitLclHashMismatch`, `handle_catchup_result`, `fatal_state_failure`, `catchup_needs_full_reset`). I ported its body verbatim onto `origin/main` (2fc81fa1) and ran it: it **FAILS** with `a knit-to-LCL mismatch at genesis must NOT be fatal` (panic at the first assert — main FATALs at genesis). On PR HEAD it **PASSES**. Genuine regression capture.
- Classifier test `crates/history/src/error.rs::test_is_knit_to_lcl_failure` PASSES (positive: all four Knit* variants; subset-of-fatal check; negatives: `VerificationFailed` stays fatal-but-not-knit, `ArchiveUnreachable` neither).
- Helper test `test_genesis_recovery_forces_bucket_apply_mode` PASSES (genesis → `(Minimal, false)` for every configured mode; above-genesis → `(configured, true)` unchanged).

### #3288 preservation (class: `regression-risk`) — CONFIRMED
- `at_genesis` is read at the **top** of the `Err` arm (catchup_impl.rs:2814), before any state mutation/cooldown write. Correct ordering.
- The new genesis branch (2897) is gated `at_genesis && is_knit_to_lcl` and `return false`s **before** the `is_fatal` classification block (2924). For `LCL > genesis`, `at_genesis == false`, so the branch is skipped entirely and `is_fatal → trigger_fatal_shutdown` runs exactly as on main. Verified by reading the full arm: no non-genesis flow can reach the new retry path.
- Guard test `test_nongenesis_knit_mismatch_still_fatal` (LCL advanced to 100) PASSES: still FATAL, `fatal_state_failure` set, shutdown signalled.
- `test_full_bucket_apply_knit_mismatch_still_wipes` adjustment is **legitimate, not a weakening**: on main this test used a fresh `App::new` (which is *at genesis*), so post-fix it would no longer FATAL. The diff advances LCL to 100 so it continues to pin the #3288 wipe intent for a node with committed state. The original genesis-FATAL assertion was load-bearing only by accident of `App::new` starting at genesis; the new `test_nongenesis_knit_mismatch_still_fatal` now pins the above-genesis FATAL explicitly. Intent preserved.

### #3282 interaction — CORRECT
- The new genesis branch sits *after* the `seeded_from_local_clone && is_local_vs_archive_divergence` self-heal (2853) and *before* the fatal block. At genesis a recovery catchup is not seeded-from-local-clone (the helper returns `seed_local_state=false` and `force_full` paths don't clone), so the #3282 branch is not taken at genesis; the #3410 branch handles it. No double-handling.

### Error handling / readability
- `is_knit_to_lcl_failure` is a clean `matches!` over the four Knit* variants, well-documented with §11.2 case mapping. Subset-of-fatal is asserted in tests.
- The genesis branch mirrors the #3282 branch's tail exactly (`catchup_needs_full_reset`, `restore_operational_state`, cooldown write, stuck-state reset, `return false`) — consistent.
- 3s genesis cooldown is bounded and gated strictly on genesis; once bucket-apply advances LCL past genesis the 10s steady-state cooldown resumes. No unbounded log-spam: even if the archive is unreachable, the retry cadence is a fixed 3s and the node is stateless, so this is acceptable (residual #3).

### Non-blocking inline note (filed, not a blocker)
- `genesis_recovery_catchup_plan` is invoked in `catchup_with_run_mode` (line 256), the **single** catchup entry point shared by recovery (lifecycle/consensus), offline-CLI `catchup` (catchup_cmd.rs:259), and startup (run_cmd.rs:482). At `current == genesis` ALL of these now force `Minimal`/bucket-apply. For startup/recovery to a far network tip this is benign (a far target already selects bucket-apply). The only true behavior change is offline-CLI `catchup --mode complete <small-target-in-first-checkpoint>`, which on main is Case-0 replay-from-genesis and now becomes bucket-apply. This is the operator-approved direction (archive-authoritative at genesis) and the offline replay-from-genesis-to-a-tiny-target is not a supported production scenario, so I treat it as a documentation/scope note rather than a blocker. See Parity reviewer for the parity angle.

Pre-existing flake `verify::tests::test_verify_bucket_hash_mismatch_emits_tracing_no_ledger_seq` lives in `crates/history/src/verify.rs` (NOT in this diff) and the diff touches no tracing/verify code — unrelated, not a blocker.
</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** The genesis-gating + no-FATAL-at-genesis is the operator-approved deviation for stateless fresh-genesis nodes; it moves the genesis path *toward* stellar-core's fresh-node posture (archive-authoritative bucket-apply, retryable knit failure, no self-wipe). For LCL > genesis the path is byte-for-byte unchanged, and no parity vectors are touched. Documented correctly in PARITY_STATUS.md.

<details>

**Cycle:** 1 (no prior Parity comment). Complete change-list below.

### stellar-core cross-check (class: `spec-divergence`) — FAITHFUL
Verified the PR's three cited upstream anchors against the pinned `stellar-core/` submodule:
- `CatchupWork.h:80` — `fatalFailure()` is a **boolean** surfaced from a future; a knit disagreement is a *retryable* Work failure, not a state wipe. Core has **no** self-wipe-on-corruption mechanism.
- `ApplyCheckpointWork.cpp:236` — the "replay of {} at LCL {} disagreed on hash" throw is exactly the knit-to-LCL disagreement the four `Knit*` variants model.
- `CatchupWork.cpp:198` — `downloadApplyBuckets()` is the fresh-node archive-authoritative bucket-apply the genesis carveout routes to.

The henyey self-wipe-on-knit-mismatch (#3282/#3288) is *already* a documented henyey-specific addition with no core analogue. The #3410 genesis carveout makes the **genesis** case behave more like core (retryable bucket-apply, no wipe), so it narrows, not widens, the divergence from upstream. Change 1 (route genesis recovery to `Minimal`/bucket-apply) directly mirrors core's fresh-node `downloadApplyBuckets` rather than a genesis replay.

### Genesis-gating correctness (class: `edge-case`) — CONFIRMED
- `genesis_recovery_catchup_plan(current, mode)` is a pure function: `current == GENESIS_LEDGER_SEQ (1)` → `(Minimal, false)`; else `(configured, true)`. Unit-tested across `Complete`/`Recent(_)`/`Minimal` at genesis and at seqs {2,64,100,61M}.
- The knit-no-FATAL branch is gated `at_genesis && is_knit_to_lcl`, evaluated before the fatal classifier, with `at_genesis` snapshotted at the top of the Err arm. For LCL > genesis the `CatchupRange` selection and the fatal classification are unchanged — confirmed by reading the full arm.

### Parity tripwires (class: `parity-gap`) — UNAFFECTED
- The diff touches only `crates/app/PARITY_STATUS.md`, `crates/app/src/app/catchup_impl.rs`, `crates/history/src/error.rs`. **No** `ledger_close_meta_vectors`, `validation_parity`, or `scp_parity_tests` fixtures/sources are touched. The change is confined to the recovery/catchup *control flow* and does not alter ledger application, header hashing, SCP, or meta emission — so those byte-for-byte vectors are unaffected by construction. CI's Test job exercises them.

### PARITY_STATUS.md (class: `doc-drift`) — ADEQUATE
- New "Architectural Difference 5" documents the deviation, the strict `LCL == genesis` gate, the core anchors, and the #3282/#3288 preservation for LCL > genesis. Accurate.
- Minor (non-blocking, inline): the doc frames the carveout as applying to "a forced recovery catchup", but the helper is invoked in the shared `catchup_with_run_mode` entry point, so offline-CLI/startup catchup at genesis are also routed to bucket-apply. In practice this is the same archive-authoritative direction and a far/network-tip target already selects bucket-apply; the only changed scenario (offline `catchup --mode complete` to a tiny first-checkpoint target) is not a supported production parity scenario. Noted as a doc-precision follow-up, not a divergence that gates merge.

No parity blocker.
</details>

tomerweller:
## ✅ Merged

**Commit:** d4454e4ddef1f516af2f4e62b73682ce7496967c

**Follow-up issues filed for unaddressed inline review comments:** none (no inline review threads; both reviewers' notes were non-blocking observations in their top-level verdicts)

Triple-green: Reviewer Correctness APPROVE, Reviewer Parity APPROVE, required CI green (Test / Build (release-ci) / Clippy / Doctest / Format / gitlink / Soroban Pin). Non-required Quickstart shard red was the known testnet flake and is not a merge gate.

PR #3551, opened 2026-06-22
Track CURRENT_LEDGER_PROTOCOL_VERSION in default_protocol_version()

### LINKED ISSUE #3550: Compat /info under-advertises max protocol: default_protocol_version() hardcoded 25 while henyey supports 27
## Summary

Henyey advertises a **max supported ledger protocol of 25 by default**, even though the binary supports **27** (`CURRENT_LEDGER_PROTOCOL_VERSION = 27`). stellar-core, by contrast, advertises its compiled max (`Config::LEDGER_PROTOCOL_VERSION`, which defaults to `CURRENT_LEDGER_PROTOCOL_VERSION`). So a stock henyey node under-reports the protocol it can actually run — a latent stellar-core parity gap, and a hard blocker for any protocol-27 Supercluster (SSC) mission.

## Root cause

`default_protocol_version()` is hardcoded to `25`:

```rust
// crates/app/src/config.rs:1716
fn default_protocol_version() -> u32 {
    25
}
```

This feeds `config.network.max_protocol_version` (`config.rs:419-420`), which the compat `/info` handler reports verbatim as the node's supported protocol:

```rust
// crates/app/src/compat_http/handlers/info.rs:39
protocol_version: app.config().network.max_protocol_version,
```

## stellar-core parity

stellar-core reports its **compiled max** supported protocol, which defaults to the current protocol constant:

```cpp
// stellar-core src/main/ApplicationImpl.cpp:468
info["protocol_version"] = getConfig().LEDGER_PROTOCOL_VERSION;

// stellar-core src/main/Config.cpp:34,160
uint32 const Config::CURRENT_LEDGER_PROTOCOL_VERSION = 27;
...
LEDGER_PROTOCOL_VERSION = CURRENT_LEDGER_PROTOCOL_VERSION;   // default
```

Henyey already defines the matching constant:

```rust
// crates/common/src/protocol.rs:137
pub const CURRENT_LEDGER_PROTOCOL_VERSION: u32 = 27;
```

So the parity-correct default for henyey's advertised max protocol is `CURRENT_LEDGER_PROTOCOL_VERSION`, not a hardcoded literal that drifts behind it every protocol bump.

## Impact

- **SSC protocol-27 mission (immediate):** Supercluster picks the network's protocol-upgrade target by reading `/info`'s `protocol_version` (`StellarCoreHTTP.fs:572` `GetSupportedProtocolVersion() = GetInfo().ProtocolVersion`). With henyey reporting 25, a mixed network against a protocol-27 stellar-core image either caps the upgrade at 25 or, if the target is driven by the core majority (27), henyey can't follow and diverges. The `MixedImageLoadGenerationWithOldImageMajority` loadgen phase needs both images at the same protocol → henyey must advertise 27.
- **General:** any non-pinned henyey deployment under-advertises its protocol support, unlike a stellar-core peer of the same version. stellar-core nodes advertise support for the next protocol ahead of a network upgrade; henyey defaulting low is off-parity.

## Proposed fix

Make the default track the constant instead of a literal:

```rust
fn default_protocol_version() -> u32 {
    henyey_common::protocol::CURRENT_LEDGER_PROTOCOL_VERSION   // 27
}
```

**Mainnet is unaffected** — `configs/validator-mainnet.toml:91` pins `max_protocol_version = 25` explicitly, so the validator keeps advertising 25 until the operator deliberately raises it. Only nodes that *don't* pin the value (compat-mode SSC nodes, defaults) pick up the parity-correct 27.

(Alternative if a code change is undesirable: teach `compat_config` to honor the real stellar-core `LEDGER_PROTOCOL_VERSION` cfg key — which stellar-core does read — mapping it to `max_protocol_version`, and have the SSC cfg generator emit `LEDGER_PROTOCOL_VERSION = 27` for the henyey node. The default-tracking fix above is simpler and fixes the latent gap for everyone.)

## Tests

- Unit: assert `default_protocol_version() == CURRENT_LEDGER_PROTOCOL_VERSION`, so the default can never silently drift behind the supported protocol again.
- Unit/handler: assert the compat `/info` handler reports `CURRENT_LEDGER_PROTOCOL_VERSION` when the config doesn't pin `max_protocol_version`, and the pinned value when it does (mirror the mainnet-pins-25 case).

## Context

Found while preparing the protocol-27 Henyey × Supercluster mixed-image mission (building a BUILD_TESTS stellar-core `v27.0.0` `--old-image`). This is the henyey-side gating item for that run.

tomerweller:
Closes #3550

## Summary

`default_protocol_version()` (crates/app/src/config.rs) returned a hardcoded `25` while henyey compiles at protocol 27, so a non-pinned node under-advertised its max protocol in compat `/info` and capped operator upgrade proposals one version below what it can actually apply. It now returns `henyey_common::protocol::CURRENT_LEDGER_PROTOCOL_VERSION` (the same import path already used by `default_testing_upgrade_protocol_version()`), so the default auto-tracks the compiled current protocol — mirroring stellar-core's `LEDGER_PROTOCOL_VERSION = CURRENT_LEDGER_PROTOCOL_VERSION` default (`Config.cpp:160`). The constant is tracked, not a literal 27, so future protocol bumps carry automatically.

This is **advertisement + operator-proposal-cap only, NOT live consensus**. The only consumers of `config.network.max_protocol_version` are: (1) compat `/info` advertisement (`info.rs:39`, the bug), (2) the herder operator `set_upgrade_parameters` cap (`herder.rs:948`/`:2875` → `upgrades.rs:388`, byte-identical to stellar-core `Upgrades.cpp:263`), and (3) a cosmetic `PROTOCOL_VERSION` metrics gauge. Re-grepped on the fresh branch: the live SCP nomination/apply validity path (`scp_driver.rs:1926`) already hardcodes `CURRENT_LEDGER_PROTOCOL_VERSION`, independent of config — so this REMOVES a henyey-only internal inconsistency rather than altering consensus. Shipped mainnet `*.toml` pin `max_protocol_version = 25` explicitly and override the default, so mainnet advertising is unaffected.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3550#issuecomment-4769844170)

## Test plan

- [x] cargo fmt --check
- [x] cargo build -p henyey-app --all-targets (RUSTFLAGS=-Dwarnings)
- [x] cargo build --all
- [x] cargo test --all (exit 0, 0 failed across all summary lines)

- **Tests:** `crates/app/src/config.rs`::`test_default_protocol_version_tracks_current_constant`, `test_network_config_default_max_protocol_is_current` (invariant), plus `test_explicit_max_protocol_version_overrides_default` (override guardrail).
- **Pre-fix:** committed as `50374bd2` — the two invariant tests verified FAILED with `assertion left == right failed: left: 25, right: 27`; the override-guard test passed (pins explicit TOML `max_protocol_version = 25` still overrides).
- **Post-fix:** all three verified PASS after `b6ff831d`.

## Parity considerations

No parity tripwire vector reads `config.network.max_protocol_version` — `validation_parity`, `ledger_close_meta_vectors`, and `scp_parity_tests` live in crates that do not depend on henyey-app's config; byte-identity unaffected (confirmed by green `cargo test --all`).

## Deviations from plan

None.

### DIFF
--- a/crates/app/src/config.rs
+++ b/crates/app/src/config.rs
@@ -4936,4 +4936,64 @@ name = "test"
         assert_eq!(meta.commit_hash(), Some("a".repeat(40).as_str()));
         assert_eq!(meta.build_timestamp(), Some("2024-01-01T00:00:00Z"));
     }
+
+    // #3550: `default_protocol_version()` must track the compiled
+    // `CURRENT_LEDGER_PROTOCOL_VERSION` constant, not a stale literal. A
+    // non-pinned node otherwise under-advertises its max protocol in compat
+    // `/info` and caps operator upgrade proposals below what it can actually
+    // apply, mirroring stellar-core's
+    // `LEDGER_PROTOCOL_VERSION = CURRENT_LEDGER_PROTOCOL_VERSION` default
+    // (Config.cpp:160).
+    #[test]
+    fn test_default_protocol_version_tracks_current_constant() {
+        assert_eq!(
+            default_protocol_version(),
+            henyey_common::protocol::CURRENT_LEDGER_PROTOCOL_VERSION,
+            "default_protocol_version() must track CURRENT_LEDGER_PROTOCOL_VERSION, not a literal"
+        );
+        // Anti-drift: pin the current expected value so a stale literal is caught.
+        assert_eq!(default_protocol_version(), 27);
+    }
+
+    #[test]
+    fn test_network_config_default_max_protocol_is_current() {
+        // The non-pinned default (used by compat `/info` advertisement, a
+        // verbatim passthrough at compat_http/handlers/info.rs:39) must equal
+        // the compiled current protocol.
+        let config = AppConfig::default();
+        assert_eq!(
+            config.network.max_protocol_version,
+            henyey_common::protocol::CURRENT_LEDGER_PROTOCOL_VERSION
+        );
+
+        // A TOML that names `[network]` but omits `max_protocol_version` must
+        // deserialize to the current constant via `#[serde(default = ...)]`.
+        let toml_str = r#"
+[network]
+passphrase = "Test SDF Network ; September 2015"
+"#;
+        let parsed: AppConfig = toml::from_str(toml_str).unwrap();
+        assert_eq!(
+            parsed.network.max_protocol_version,
+            henyey_common::protocol::CURRENT_LEDGER_PROTOCOL_VERSION,
+            "absent max_protocol_version must default to CURRENT_LEDGER_PROTOCOL_VERSION"
+        );
+    }
+
+    #[test]
+    fn test_explicit_max_protocol_version_overrides_default() {
+        // Override guardrail: an explicit pin (as in all shipped mainnet *.toml)
+        // must win over the now-current default. This is what keeps mainnet
+        // advertising the pinned 25 while the default auto-tracks the constant.
+        let toml_str = r#"
+[network]
+passphrase = "Public Global Stellar Network ; September 2015"
+max_protocol_version = 25
+"#;
+        let config: AppConfig = toml::from_str(toml_str).unwrap();
+        assert_eq!(
+            config.network.max_protocol_version, 25,
+            "an explicit TOML max_protocol_version must override the default"
+        );
+    }
 }
--- a/crates/app/src/config.rs
+++ b/crates/app/src/config.rs
@@ -1714,7 +1714,12 @@ fn default_base_reserve() -> u32 {
 }
 
 fn default_protocol_version() -> u32 {
-    25
+    // Track the compiled current protocol so a non-pinned node advertises (and
+    // caps operator upgrade proposals at) its real max, mirroring stellar-core's
+    // `LEDGER_PROTOCOL_VERSION = CURRENT_LEDGER_PROTOCOL_VERSION` default
+    // (Config.cpp:160). Shipped mainnet *.toml pin this explicitly and override
+    // the default. See #3550.
+    henyey_common::protocol::CURRENT_LEDGER_PROTOCOL_VERSION
 }
 
 fn default_db_path() -> PathBuf {


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Clean one-line fix (`default_protocol_version()` now returns `henyey_common::protocol::CURRENT_LEDGER_PROTOCOL_VERSION` instead of literal `25`) backed by correct failing-test-first tests. Cycle 1 — complete change-list, no blocking concerns.

<details>

Linked issue kind: `bug-fix`. Verified per the test-verification checklist.

**`test-coverage` (failing-test-first) — PASS:**
- Regression test commit `50374bd2` ("Regression test for #3550 — fails on current main") was committed BEFORE the fix commit `b6ff831d`. Confirmed by static inspection at `50374bd2`: `default_protocol_version()` still returns `25` there, while `test_default_protocol_version_tracks_current_constant` asserts it equals `CURRENT_LEDGER_PROTOCOL_VERSION` (= 27 on origin/main and on the PR head, confirmed via `git show origin/main:crates/common/src/protocol.rs` → `:137 = 27`). `25 != 27` ⇒ the two invariant tests FAIL at the test-commit; they PASS after `b6ff831d`. This is a genuine failing-test-first capture.
- `test_explicit_max_protocol_version_overrides_default` passes at both commits by design (TOML pin `max_protocol_version = 25` ⇒ field is 25 regardless of the default) — correctly exercises the serde override guardrail that keeps shipped mainnet `*.toml` advertising the pinned 25.

**`api-shape` / `constant-tracking` — PASS:** The fix tracks the CONSTANT, not a literal 27, so future protocol bumps auto-carry. Same import path already used by `default_testing_upgrade_protocol_version()`, so no new dependency surface. The one literal in the diff is the deliberate anti-drift tripwire `assert_eq!(default_protocol_version(), 27)` inside the test — that is intentional belt-and-suspenders (catches a silent constant rename), not a parity literal; it will require a one-line bump at protocol 28, which is acceptable as a tripwire.

**`regression-risk` — none:** `NetworkConfig::default()` now returns 27 programmatically. No `*_config_default` assertion in the suite hardcodes the old 25 for this field (the only field consumers are info.rs advertisement, the metrics gauge, and the herder operator-cap — none assert a literal default). The known-flaky `manager::tests::test_ledger_manager_config_default` (asserts 1024 != 0) is unrelated to this field.

**`error-handling` / readability — PASS:** comment cites `Config.cpp:160` and `#3550`; clear and accurate.

No blocking concerns.
</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** Raising the non-pinned default to `CURRENT_LEDGER_PROTOCOL_VERSION` is parity-correct: it is **advertisement + operator-proposal-cap only, NOT live consensus**. Independently re-audited every consumer of `config.network.max_protocol_version` on the PR head — none thread into the live SCP nomination/apply validity path.

<details>

Audited fresh against the PR head `b6ff831d` (origin/main `CURRENT_LEDGER_PROTOCOL_VERSION = 27`).

**`parity-gap` — none. Consumer audit (the load-bearing claim):**
`git grep` for `network.max_protocol_version` on `b6ff831d` returns exactly three non-test consumers plus the three new test asserts:
1. `crates/app/src/compat_http/handlers/info.rs:39` — compat `/info` `protocol_version` advertisement (the bug being fixed). stellar-core advertises `LEDGER_PROTOCOL_VERSION`, which defaults to `CURRENT_LEDGER_PROTOCOL_VERSION` (`Config.cpp:160` / `ApplicationImpl.cpp`). Raising the default to the constant makes henyey advertise its true compiled max — matching core.
2. `crates/app/src/metrics.rs:1195` — cosmetic `PROTOCOL_VERSION` gauge. No consensus effect.
3. `crates/app/src/app/mod.rs:1820` → `HerderConfig.max_protocol_version`. This is the only path that could plausibly reach consensus, so I traced it fully. `HerderConfig.max_protocol_version` is read at exactly two sites — `herder.rs:948` and `herder.rs:2875` — and **both** call `RuntimeUpgrades::set_parameters(params, max_protocol)`. That is the **operator runtime upgrade-proposal cap**: `upgrades.rs:388` rejects `version > max_protocol_version` with "supported is up to N, passed is M" — byte-identical in shape to stellar-core `Upgrades.cpp:263`. `herder.rs:2875` merely re-validates already-accepted params after `removeUpgrades` (with an "already validated when set" comment). Neither is a consensus acceptance path.

**`spec-divergence` — none. Live SCP path is independent of the config field (verified):**
The live SCP value-validation/apply path is `crates/herder/src/scp_driver.rs:1925-1931` (`upgrade_valid_for_apply`), which passes `henyey_common::CURRENT_LEDGER_PROTOCOL_VERSION` as the cap to `upgrade_valid_for_apply_non_config(other, current_version, CURRENT_LEDGER_PROTOCOL_VERSION)` — hardcoded to the constant, NOT read from `config.network.max_protocol_version`. (The operator's reference to "scp_driver.rs:1926" resolves to this site in `crates/herder/src/scp_driver.rs`.) So SCP nomination/ballot/apply validity is unchanged by this PR. Before the fix the config default (25) was *below* the constant the live path uses (27) — a henyey-only internal inconsistency where a non-pinned node would advertise/cap at 25 while its consensus path accepted up to 27. This PR REMOVES that inconsistency; it does not alter consensus.

**`edge-case` — mainnet unaffected:** all shipped mainnet `*.toml` pin `max_protocol_version = 25` explicitly; the serde `#[serde(default = ...)]` override path (proven by `test_explicit_max_protocol_version_overrides_default`) keeps the pinned value. Only a *non-pinned* node changes behavior, and only in advertisement + operator-cap, never in consensus.

**Parity tripwires byte-unaffected:** `validation_parity`, `ledger_close_meta_vectors`, `scp_parity_tests` do not reference `max_protocol_version` (grep-confirmed) and live in `crates/ledger`/`crates/scp`, which do not depend on henyey-app config. No fixture value changes.

Parity-correct. No blocking concerns.
</details>

Issue #3553, opened 2026-06-22
Land #3552 L16 bucketListHash fix + fail-on-main regression from captured core L16 level-2 bucket


tomerweller:
Phase-3 follow-up from #3552. The #3552 instrumentation PR (per-level `bucketListHash` logging + real-core-bucket comparator/serialize/hash guard) does NOT fix the consensus divergence — it refutes the comparator hypothesis (proven byte-correct against real core buckets `f0c6394a`/`0cf535ed`) and lands the infrastructure to capture the missing oracle.

## What this issue covers
1. Run the SSC v27 mixed-image mission with henyey per-level `bucketListHash` logging enabled (and core per-level hashes via core instrumentation or a full-checkpoint archive).
2. Capture core's authoritative L16 level-2 bucket (its file/entries) and henyey's divergent L16 level-2 bucket.
3. Decode both; diff the entry SET / CONTENT / INIT-vs-LIVE classification / `last_modified_ledger_seq` — the comparator/serialize/hash are already proven equal, so the divergence is in WHICH entries (or their bodies) land in the level-2 bucket.
4. Land the minimal parity fix mirroring stellar-core v27, plus a fail-on-main `crate:bucket` regression built from the captured core entries (asserting henyey reproduces core's L16 level-2 bucket hash).

## Tripwire note
The eventual fix changes bucket hashes. Parity tripwires (`ledger_close_meta_vectors`, `validation_parity`, `tx_meta_hash_vectors`, `scp_parity_tests`) may legitimately shift to the now-correct values — they must be validated AGAINST CORE (the captured L16 oracle), never silently regenerated from buggy henyey.

Keep #3552 OPEN as the tracking issue for the consensus divergence until this lands.

tomerweller:
## Triage Report

**Verdict:** BLOCKED

**Reason:** needs-human-decision (mission-gated)

**Detail:** The L16 level-2 bucket entry-set/content oracle required to diagnose and fix the consensus divergence exists **only** from a live, instrumented SSC v27 mixed-image mission re-run (the #3552 instrumentation enables capture, but the capture itself requires operator execution). This fix is not offline-codeable: without the operator running the SSC mission and sharing the L16 level-2 bucket data from both core and henyey, no regression test or fix can be written.

Blocked awaiting: operator SSC re-run with per-level bucketListHash logging + L16 level-2 bucket capture from both core and henyey. Keep #3552 open as the consensus-divergence tracker until this lands.

[...]

PR #3554, opened 2026-06-22
Add L16 bucketListHash diagnostics + real-core-bucket parity guard (#3552)

tomerweller:
Refs #3552
References #3553

## Summary

**This is a diagnostic-advancement PR — it does NOT fix the #3552 L16 `bucketListHash` divergence.** Per the converged (re-)plan and CLAUDE.md (*"a wrong speculative consensus fix is worse than none"*), the re-plan **proved** — against two real stellar-core v27 merge-output buckets archived from the mission — that henyey's bucket comparator, entry sort, XDR serialization, record-marking, and hashing are **byte-identical to core**. The machinery is correct; the root cause is in the L16 entry SET/CONTENT (INIT-vs-LIVE classification), whose live oracle is **absent from the single-checkpoint archive** and is tracked in follow-up **#3553**. `#3552` stays **open** as the tracking issue — this PR advances it but does not resolve it.

What this PR lands:

1. **Per-level `bucketListHash` instrumentation hook** in `crates/bucket/src/bucket_list.rs` `add_batch_internal`: a gated, read-only hook (tracing target `bucket_list_hash`) emitting each level's curr/snap hashes + the top-level hash after every `add_batch`. Enable via `RUST_LOG=bucket_list_hash=debug`. **Provably non-perturbing** — it only reads `level_hashes()`/`hash()` and short-circuits to a no-op (reading/allocating nothing) when the target is disabled. This lets the next SSC v27 mission re-run pinpoint the divergent level live (feeds #3553).
2. **Permanent real-core-bucket parity guard** `oracle_genesis_bucket_roundtrip_matches_core`: vendors two real core v27 buckets (`f0c6394a`, 127 entries inc. 19 INIT ConfigSetting + Soroban; `0cf535ed`, 56 entries) under `crates/bucket/tests/fixtures/issue3552/` and asserts henyey reproduces core's published hashes byte-for-byte via both `from_xdr_bytes` AND `from_entries` (comparator re-sort). **PASSES on main** — locks in the verified-correct machinery; guards against regression.
3. The two prior structural diagnostics (from `do/issue-3552` `19996a6a`) are kept and superseded by this branch (no dangling duplicate).
4. `crates/bucket/PARITY_STATUS.md` records the open L16 divergence + the verified machinery.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3552#issuecomment-4771560324)

## Why no fail-on-main regression test

This is a **bug-fix issue reclassified to diagnostic/instrumentation for this PR**, with verified justification (see the converged plan): the only offline-reproducible lead (the comparator) is **proven correct** against real v27 buckets, so no synthetic input reproduces the divergence, and the real oracle (core's L16 level-2 bucket) is unavailable from preserved artifacts. The fail-on-main regression lives in **#3553** once core's L16 entries are captured live. Please do not bounce for "missing bug-fix regression test."

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo build -p henyey-bucket -p henyey-ledger --all-targets` with `-Dwarnings` (build gate #3384)
- [x] `cargo build --all`
- [x] `cargo test --all` (all green)
- [x] `level1_to_level2_spill_parity` (3 tests) + `test_per_level_hash_trace_is_non_perturbing` pass
- [x] Parity tripwires byte-identical: `ledger_close_meta_vectors`, `validation_parity`, `tx_meta_hash_vectors`, `scp_parity_tests` (zero hashed-byte changes; the only src change is the gated read-only hook)

## Deviations from plan

None.


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Diagnostic-advancement PR (#3552, fix tracked in #3553). The gated per-level hash hook is provably non-perturbing and its non-perturbing test is sound; the real-core-bucket parity guard genuinely feeds two real v27 buckets and asserts byte-for-byte reproduction. All claims verified by running the tests.

<details>

Cycle 1 — complete change-list (no concerns block).

**Linked issue kind:** `bug` reclassified to diagnostic/instrumentation for this PR (verified justification in PR body + converged plan; #3553 carries the fail-on-main regression once core's L16 oracle is captured live). Per the operator brief this is NOT expected to fail-on-main — it locks in verified-correct machinery. No "missing bug-fix regression test" bounce.

**`crates/bucket/src/bucket_list.rs` — the instrumentation hook (correctness verified):**
- `emit_per_level_hash_trace(&self, ...)` takes `&self` (no mutation possible). It first calls `tracing::enabled!(target: ..., DEBUG)` and `return`s early if disabled — a true no-op reading/allocating nothing when the target is off.
- When enabled it only calls `self.level_hashes()` and `self.hash()`. Confirmed `level_hashes()` (bucket_list.rs:1673) is `self.levels.iter().enumerate().map(|(idx, level)| (idx, level.hash(), level.curr.hash(), level.snap.hash()))` — pure read, no `&mut`. `hash()` is `&self`. The hook therefore cannot alter a single bucket byte whether on or off.
- Call site is placed AFTER all level mutations in `add_batch_internal`, immediately before `Ok(())` — it observes final state, never feeds back.

**Test verification — `test_per_level_hash_trace_is_non_perturbing` (RAN, passes):**
- Runs the identical 16-ledger `add_batch` chain twice (hook off via no subscriber; hook on via a `CountingSubscriber` enabling only the `bucket_list_hash` target at DEBUG) and asserts `chain_off == chain_on` over the full per-ledger `bucket_list_hash` chain — byte-identical, proving non-perturbing in BOTH states.
- Also asserts the hook actually emitted `BUCKET_LIST_LEVELS * 16` per-level records and `16` list-level records — proving the enabled path genuinely exercised the read, so the equality isn't vacuous (both branches were covered).
- Ran locally on the PR head: `test ... test_per_level_hash_trace_is_non_perturbing ... ok`.

**Test verification — `oracle_genesis_bucket_roundtrip_matches_core` (RAN against MAIN machinery, passes):**
- Fixtures are real binary core v27 BucketEntry streams (20796 + 8384 bytes), content-addressed by filename hash (the oracle).
- Asserts `from_xdr_bytes(raw).hash() == expected` (framing + decode + SHA-256) AND `from_entries(parsed).hash() == expected` (comparator re-sort + re-serialize) AND `resorted.entries() == parsed` (comparator reproduces core's on-disk order).
- Verified it PASSES against `origin/main`'s bucket crate (copied the new test + fixtures onto a main worktree and ran): `test oracle_genesis_bucket_roundtrip_matches_core ... ok`. Confirms it banks verified-correct machinery and is intentionally NOT a fail-on-main reproduction.

**Readability / error handling:** test code uses `expect`/`unwrap_or_else(panic!)` with descriptive messages — appropriate for tests. Hook docs accurately describe the no-op semantics and point at the proving test. PARITY_STATUS.md records the open L16 divergence (→ #3553) and the new guard.

**Tripwires:** only src change is the gated read-only hook; no parity-tripwire test file is touched. The change is byte-invisible to all hashing paths, so `ledger_close_meta_vectors` / `validation_parity` / `tx_meta_hash_vectors` / `scp_parity_tests` cannot shift.

No blocking concerns.
</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** The only src change is a gated read-only tracing hook that is byte-invisible to all hashing/serialization paths, so it cannot shift any parity output. The new guard test genuinely feeds REAL core v27 merge-output buckets and asserts byte-for-byte hash reproduction via both raw decode and comparator re-sort — strengthening parity, not regressing it.

<details>

Cycle 1 — complete parity change-list (no concerns block).

**Parity surface of the src change (`bucket_list.rs`):**
- The hook reads `level_hashes()`/`hash()` only and short-circuits via `tracing::enabled!` when disabled. It is `&self`, placed after all mutations, and emits to a tracing target only. There is no path by which it alters bucket bytes, entry order, record-marking, or the computed `bucketListHash` — on or off. Hence **zero divergence from stellar-core** introduced: stellar-core has no equivalent emission, but emission is not an observable consensus output.
- `class: parity-gap` — none. The hook does not change any bucket-list semantics or sequencing vs core.

**Parity tripwires byte-IDENTICAL:** confirmed no tripwire test file is in the diff (`git diff --name-only`: only PARITY_STATUS.md, bucket_list.rs, the 2 fixtures, the new test). The src change being a gated read-only emission means `ledger_close_meta_vectors`, `validation_parity`, `tx_meta_hash_vectors`, `scp_parity_tests` are unaffected by construction — no hashed byte can move.

**Cross-implementation parity guard (`oracle_genesis_bucket_roundtrip_matches_core`):**
- Feeds REAL core v27 buckets (binary BucketEntry XDR streams, 20796 + 8384 bytes from the mission archive), NOT synthetic — verified via `git cat-file -s` and `wc -c`. Content includes genesis ConfigSetting (19 INIT) + Soroban ContractCode/ContractData/Ttl + LIVE/INIT Accounts, exactly the genesis-shaped content implicated in the L16 divergence.
- Asserts byte-for-byte hash reproduction against core's content-addressed (filename) hash via `from_xdr_bytes` (framing + XDR decode + SHA-256) AND `from_entries` (henyey's `compare_entries` comparator + entry sort + re-serialize + record-mark + hash). This directly verifies henyey's comparator/sort/serialize/record-mark/hash parity with core v27 for this content.
- Verified it PASSES on `main` (ran against the main bucket crate) — correctly locks in verified machinery rather than reproducing the bug. This matches the diagnostic framing: the divergence is in the L16 entry SET/CONTENT (INIT-vs-LIVE), whose live oracle is unavailable from the single-checkpoint archive and is tracked in #3553.

**Tracking disposition:** PARITY_STATUS.md "Known Divergences (open)" row correctly marks L16 as Open → #3553, and #3552 remains the open tracking issue. Confirmed #3552 OPEN and #3553 OPEN + linked.

**Risk angle (regression/perf/api):** no API break (hook is a new private method + a new public `const`; `level_hashes()` already public). No perf concern — the only cost when disabled is one `tracing::enabled!` check per `add_batch`, which is the documented zero-cost-when-off path. No migration/config risk.

No blocking concerns.
</details>

PR #3562, opened 2026-06-22
Emit per-ledger EvictionIterator update on every V20+ ledger (#3552 L16 divergence)

### LINKED ISSUE #3553: Land #3552 L16 bucketListHash fix + fail-on-main regression from captured core L16 level-2 bucket

tomerweller:
Closes #3553
Refs #3552

## Summary

Fixes the lone L16 `bucketListHash` divergence from stellar-core v27 captured in #3552. henyey gated the per-ledger `CONFIG_SETTING_EVICTION_ITERATOR` update — including the `live_entries.push(eviction_iter_entry)` — behind the non-core conjuncts `prev_version >= V23 && has_hot_archive && eviction_settings.is_some()`. When any conjunct was false the per-ledger EvictionIterator live entry was dropped; the resulting level-1→2 spill output bucket came out empty with default `{ledgerVersion:0, ext:void}` metadata (the `InputDerived` merge of two empty inputs) instead of core's `{27, V1(Live)}` + the EvictionIterator entry, flipping `bucketListHash` → L16 ledger-hash divergence.

This change mirrors stellar-core exactly (no new semantics): load the Soroban network config inline for every `initialLedgerVers >= V20` ledger, run the eviction scan, and write the EvictionIterator config setting **unconditionally** (zero-eviction included), keeping the hot-archive `addHotArchiveBatch`/restore sub-step gated separately on V23 + hot-archive presence.

## Oracle-confirmed root cause

Decoded byte-for-byte from the SSC v27 mixed-image mission (cluster `rfh1c2qdatp96`, `~/data/9eb89c28/ssc-v27-diag/core-bucket-snaps/`):

- **henyey `d15ebeb4` (16 B):** `Metaentry(BucketMetadata { ledger_version: 0, ext: V0 })` — EMPTY.
- **core `0ebd5750` (60 B):** `Metaentry({ ledger_version: 27, ext: V1(BucketListType::Live) })` **+ one** `Liveentry` carrying `ConfigSetting(EvictionIterator { bucket_list_level: 6, is_curr_bucket: true, bucket_file_offset: 0 })` at `last_modified_ledger_seq: 17`.

The dropped entry (b) is the root cause; the `{0,void}` metadata (a) is a downstream symptom (the `InputDerived`-of-two-empty-inputs merge artifact once the entry is missing from the level-1 inputs). The bucket/merge machinery is verified-correct — feeding it the entry reproduces core's bucket byte-for-byte — so the fix is in the ledger crate, not bucket. henyey's divergent L16 set was confirmed to also carry the per-ledger entries it DID emit (`123224d3` @9, `a96ad9ce` @16, both byte-identical to core), pinning the divergence to the single dropped iterator update.

## Core parity

- `LedgerManagerImpl.cpp:2984-3060` (`sealLedgerTxnAndStoreInBucketsAndDB`): outer gate = V20 (`SOROBAN_PROTOCOL_VERSION`); loads `SorobanNetworkConfig::loadFromLedger` inline; runs `resolveBackgroundEvictionScan`. Inner V23 (`FIRST_PROTOCOL_SUPPORTING_PERSISTENT_EVICTION`) gate wraps ONLY the hot-archive restore / `addHotArchiveBatch`.
- `BucketManager.cpp:1290` (`resolveBackgroundEvictionScan`): `networkConfig.updateEvictionIterator(ltx, newEvictionIterator)` called unconditionally — the iterator is written every ledger, zero-eviction included.

- **Test:** `crates/ledger/tests/ledger_close_integration.rs::test_eviction_iterator_emitted_every_v20_ledger`
- **Pre-fix:** committed as `bd5ddce9` — verified FAILED: `EvictionIterator last_modified left: 0, right: 1` (the per-ledger update was dropped when the non-core `eviction_settings.is_some()` conjunct was false).
- **Post-fix:** verified PASSES after `335c7b86`.
- **Bucket-layer parity guard (passes on `main`):** `crates/bucket/tests/level1_to_level2_spill_parity.rs::oracle_l16_eviction_iterator_bucket_matches_core` — vendors core's `0ebd5750` (`crates/bucket/tests/fixtures/issue3552/bucket-0ebd5750.xdr`), asserts hash + METAENTRY `{27,V1(Live)}` + the single EvictionIterator entry. Locks the cross-impl bucket bytes.

## Parity-tripwire validation

Re-introducing the per-ledger EvictionIterator entry changes live bucket hashes only when a fixture exercises a V20+ ledger where the non-core conjuncts were previously false. **All four tripwires PASS unchanged — no shift:**

- `ledger_close_meta_vectors` — ✅ pass, no shift
- `tx_meta_hash_vectors` — ✅ pass, no shift
- `validation_parity` (henyey-tx) — ✅ pass, no shift
- `scp_parity_tests` (henyey-scp) — ✅ pass, no shift

No tripwire shifted because their fixtures all carry full Soroban config (hot archive + StateArchival present), so the non-core conjuncts were already true on those vectors and the entry was already emitted. No values were regenerated.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo build -p henyey-ledger -p henyey-bucket --all-targets` with `-Dwarnings` (#3384)
- [x] `cargo build --all` (no broken-main; clean on origin/main `1b29670f`)
- [x] `cargo test --all` — 0 failures
- [x] Parity tripwires (all pass, no shift)

## Deviations from plan

- The plan flagged a conditional `start_merge` `metadata_policy` hardening "only if `/do` proves a live henyey path emits `{0,void}`". Confirmed the `{0,void}` bucket is the correct `InputDerived`-of-two-empty-inputs merge result (a symptom of the dropped entry, not an independent bug), so per the plan it is NOT in scope — no metadata code changed.
- The fail-on-main test exercises the `eviction_settings.is_some()` conjunct (a synthetic StateArchival-absent state) rather than the version bound: henyey is protocol 24+ only (`MIN_LEDGER_PROTOCOL_VERSION = 24 >= V23 >= V20`), so the V23→V20 gate lowering has no observable effect on supported ledgers — exactly as the plan notes ("the version bound is not the trigger — the conjunct removal is"). The conjunct removal is the load-bearing change.


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Faithful, well-scoped fix. The regression test fails on the pre-fix commit (`left:0 right:1`) and passes after; oracle bucket guard + all four parity tripwires pass with no shift. No correctness or test-coverage gaps.

<details>

**Cycle 1** (no prior Correctness comment). Complete change-list below; all concerns resolved → APPROVE.

**kind: bug-fix — regression-test verification (`test-coverage`):** PASS.
- Walked the commit list: `bd5ddce9` (test) precedes `335c7b86` (fix), per the failing-test-first discipline.
- Checked out `bd5ddce9` ledger crate and ran `test_eviction_iterator_emitted_every_v20_ledger`: **FAILS** with `EvictionIterator must be re-stamped … left: 0, right: 1` — the test genuinely captures the dropped per-ledger update.
- Restored `335c7b86`: test **PASSES**. PR body's `## Regression test` section matches the diff exactly.

**Logic / control-flow (`correctness`):** PASS.
- The diff restructures the eviction block from `if V23 { if has_hot_archive && eviction_settings.is_some() { … } }` to `if V20 { if let Some(eviction_settings) { …real scan+push… } else { …re-stamp unchanged iterator… } }`. The `Some` arm is the unchanged prior body (background-scan reuse, inline fallback, resolve, `live_entries.push(eviction_iter_entry)`); the only structural change is the outer gate (V23→V20) and dropping the `has_hot_archive` conjunct.
- `archived_entries` is populated only inside the V20 block and consumed only in the separate V23 hot-archive `add_hot_archive_batch` block (lines 5863-5942), which remains gated on `prev_version >= V23`. At V20-V22 `archived_entries` stays empty and no hot-archive batch runs — correct wiring, no use-before-init.
- `else` fallback uses `if let Ok(iter) = load_eviction_iterator_from_bucket_list(...)` (non-`?`), so a degenerate bucket list without an iterator entry cannot turn a successful close into an error — it simply skips the re-stamp. Appropriate: this branch is only reachable on synthetic/degenerate state (real V20+ ledgers always carry StateArchival settings).

**Error handling (`error-handling`):** PASS. The `?` on `load_eviction_iterator_from_bucket_list` inside the real-scan path is unchanged; the new fallback deliberately swallows the load error (logged at debug) rather than failing the close. Consistent with "core writes it unconditionally; if we genuinely can't, don't fabricate, don't crash."

**Readability:** Comments cite the exact core call sites and explain the symptom-vs-root-cause relationship. Good.

**Test isolation note (non-blocking, inline):** the regression test closes ledger 1 from genesis specifically to avoid a level spill / Tokio runtime — well-reasoned and documented in the test. No concern.

No CHANGES_REQUESTED items.
</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** The V23→V20 gate move + conjunct removal is an exact mirror of stellar-core v27.0.0 `sealLedgerTxnAndStoreInBucketsAndDB` / `resolveBackgroundEvictionScan`. Core writes the EvictionIterator on every V20+ ledger close, zero-eviction included, loads settings inline, and reserves V23 for the hot-archive sub-step only. No new semantics, no over-emission, all four parity tripwires unchanged.

<details>

**Cycle 1** (no prior Parity comment). Complete parity change-list; all classes clear → APPROVE.

**`parity-gap` — outer gate (V20):** PASS. `stellar-core/src/ledger/LedgerManagerImpl.cpp:2872` — `sealLedgerTxnAndStoreInBucketsAndDB` wraps the eviction work in `if (protocolVersionStartsFrom(initialLedgerVers, SOROBAN_PROTOCOL_VERSION))` (= V20). The PR lowers both the `eviction_settings` load gate and the eviction-scan/iterator-push gate from V23 to V20, matching `initialLedgerVers >= V20` exactly. (Note: the PR body cites `LedgerManagerImpl.cpp:2984-3060`; in the pinned submodule the region is `2872-2945` — line numbers drift but the structure is identical. Non-blocking doc nit.)

**`parity-gap` — inline settings load:** PASS. Core line 2878: `auto sorobanConfig = SorobanNetworkConfig::loadFromLedger(ltx);` inside the V20 gate, then `resolveBackgroundEvictionScan(...)`. `resolveBackgroundEvictionScan` (`BucketManager.cpp:1205`) *also* loads `SorobanNetworkConfig::loadFromLedger(ls)` itself. henyey loads `load_state_archival_settings()` inline at V20 — same point, same semantics.

**`spec-divergence` — unconditional iterator write:** PASS (the load-bearing parity claim). `BucketManager.cpp:1304` — `networkConfig.updateEvictionIterator(ltx, newEvictionIterator)` is called **unconditionally** at the end of `resolveBackgroundEvictionScan`, after the eviction `while` loop. On a zero-eviction scan the loop body never runs and `newEvictionIterator` defaults to `evictionCandidates->endOfRegionIterator`, but the write still happens. The write is gated on **neither** a hot-archive bucket list **nor** a pre-loaded settings value. The PR's removal of the `has_hot_archive && eviction_settings.is_some()` conjuncts and unconditional `live_entries.push` mirrors this precisely. The conjunct removal — not the V23→V20 bound (a no-op on henyey's 24+ floor) — is the operative fix, and the regression test exercises exactly the `eviction_settings.is_some()==false` conjunct.

**`sequencing` — V23 reserved for hot-archive only:** PASS. Core's inner `if (protocolVersionStartsFrom(initialLedgerVers, LiveBucket::FIRST_PROTOCOL_SUPPORTING_PERSISTENT_EVICTION))` (V23, line ~2884) wraps ONLY the restored-hot-archive-keys collection + `addHotArchiveBatch` / `addHotArchiveBatchWithP23HotArchiveFix`. henyey keeps its hot-archive `add_hot_archive_batch` block separately gated on `prev_version >= V23` (manager.rs:5877) — unchanged by this PR. Correct separation.

**`edge-case` — V20-V22 ledgers (no hot archive):** PASS. In core, V20/V21/V22 ledgers enter the V20 gate, run `resolveBackgroundEvictionScan`, and write the iterator, but skip the V23 hot-archive batch. After this PR henyey does the same: iterator pushed, `archived_entries` empty, hot-archive block skipped. Previously henyey dropped the iterator on these ledgers (V23 gate + has_hot_archive conjunct). This is strictly MORE core-correct.

**`over-emission` check:** NONE found. The only new emission is the per-ledger EvictionIterator entry that core also emits. The `else` re-stamp branch is reachable only when settings fail to load — impossible on a real V20+ ledger (core `releaseAssert`s settings present); it is a fail-safe that matches core's "always write" intent without fabricating a moved cursor.

**Tripwire re-run (no shift):** I re-ran all four against the PR head — `ledger_close_meta_vectors` ✅, `tx_meta_hash_vectors` ✅, `validation_parity` (henyey-tx) ✅, `scp_parity_tests` (henyey-scp) ✅ — all pass against committed expected vectors. A silent hash shift would have failed these; none did. Confirms the change touches no already-correct ledger-close path.

**Oracle guard:** `oracle_l16_eviction_iterator_bucket_matches_core` decodes core's vendored `0ebd5750`, asserts the content hash, the `{ledger_version:27, ext:V1(Live)}` METAENTRY, and the single `EvictionIterator{level:6, is_curr:true, offset:0}` LIVEENTRY @ last_modified 17 — genuinely checks henyey reproduces core's bucket bytes. Passes.

No CHANGES_REQUESTED items.
</details>

tomerweller:
## ✅ Merged

**Commit:** 3ab671c8532ef889e17f0aaf56208436f58a2460

Both agent reviewers (Correctness, Parity) APPROVED. Required CI green (Build release-ci, Clippy, Test, Doctest, Format, Gitlinks, Soroban Pin, Skill-Snippets). Non-required testnet/galexie quickstart shards still in progress at merge — not gating. Parity verified as an exact stellar-core v27 mirror: V20 outer gate, inline settings load, unconditional per-ledger EvictionIterator write, V23 reserved for the hot-archive sub-step only. Failing-test-first confirmed (test fails on bd5ddce9 with left:0/right:1, passes after). All four parity tripwires re-run with no shift.

Issue #3563, opened 2026-06-22
CI: Quickstart local galexie shard times out (exit 124) on main — blocks deploy gate


tomerweller:
## Symptom

`Quickstart` workflow on `origin/main` (1b29670f) fails: the **`test (local, galexie, test_galexie.go)`** shard times out — `timeout -k 30s 480s go run quickstart/tests/test_galexie.go` hits the 480s wall and exits 124 ("=== Failed (exit 124), not retryable ==="). The run already auto-retried (**attempt 2**) and failed again, so it is persistent this cycle, not a one-off flake.

This blocks the monitor-tick deploy gate (check 10 step 4: any recent origin/main CI `failure` defers deploy), so 1b29670f and subsequent commits cannot auto-deploy until Quickstart goes green. The validator itself is healthy and unaffected (running 329938b0).

## Evidence

- Run: `Quickstart` id `27988069693`-class, headSha `1b29670f`, created `2026-06-22T22:22:20Z`, attempt=2, conclusion=failure.
- Failing step: `Run probes through wrapper` → `Command: timeout -k 30s 480s go run quickstart/tests/test_galexie.go` → `=== Failed (exit 124), not retryable ===` → `Process completed with exit code 1`.
- Test-suite `CI` workflow on the same sha is **green** (the earlier `test_core4_multi_step_payment_load` failure was a flake that cleared on rerun) — so this is isolated to the Quickstart galexie shard, not the validator/library tests.

## Suspected root cause

Recurrence of the galexie-probe-hang class. The local galexie shard's `test_galexie.go` probe is not completing within 480s and the wrapper does not mark it soft-on-timeout (only the testnet shard sets `matrix.soft_on_timeout: true`, per the wrapper comments referencing #3272). Either the galexie probe genuinely hangs at a step, or 480s is too tight for the local shard under current CI load.

## Related

- **Related to #3272 (closed)** — "CI: Quickstart testnet shard cancelled… probe wrapper hangs at step 8 (systematic, not flaky)". Same galexie-probe-timeout family, now recurring on the **local** shard. The #3272 fix added per-shard soft-on-timeout but did not cover this shard / this hang.

## Fix direction

Decode where `test_galexie.go` hangs on the local shard (add step-level logging or run with `--verbose`); either fix the probe hang or extend the local shard's `step_timeout_minutes` / set `soft_on_timeout` for it the way the testnet shard does. Validate against a clean Quickstart run on main.

tomerweller:
[...]

PR #3564, opened 2026-06-23
Gate hot-archive merge metaentry on output version to fix L16 bucketListHash divergence

### LINKED ISSUE #3552: Consensus divergence: henyey ledger-16 hash differs from stellar-core on an empty ledger (suspect bucketListHash at level-1→2 merge boundary) — SSC v27 loadgen
## Summary

During the protocol-27 Henyey × Supercluster mixed-image loadgen mission, **henyey's ledger 16 hash diverges from stellar-core's**, even though both nodes had byte-identical inputs for that ledger (same previous hash, empty tx set, same close time, no upgrades). The divergence is in the **ledger-close computation of an empty ledger** — not transaction or upgrade application. After diverging, henyey gets stuck at L16 (it can't externalize L17, which references the network's L16 hash it never computed), enters recovery catchup, and **FATALs on a knit-to-LCL mismatch** (the `#3410` failure path).

This is a **consensus/state-hash parity violation** (henyey ≠ stellar-core), surfaced exactly as the SSC mission is designed to catch it.

**Note — the loadgen blocker is cleared:** this run was the first to reach and complete the loadgen phase (100/100 txns applied), using a self-built BUILD_TESTS `stellar-core v27.0.0` `--old-image` (`nscr.io/k4jkul01t5rr0/stellar-core-testing:v27.0.0`). henyey advertised protocol 27 (#3550 fix) and the network formed correctly. The divergence below is the next blocker.

## Evidence: per-ledger hash comparison (henyey vs stellar-core authoritative)

Core's authoritative hashes are read from its consensus `prev=` field (the previous ledger's hash) in `core-old-0`'s log; henyey's are its logged `computed_hash`.

| Ledger | stellar-core (authoritative) | henyey computed | Match |
|---|---|---|---|
| L10 | `9bd715…` | `9bd7152e…` | ✅ |
| L11 (22 txs) | `ca4ed8…` | `ca4ed852…` | ✅ |
| L12 (24 txs) | `efc230…` | `efc230d7…` | ✅ |
| L13 (30 txs) | `fc06af…` | `fc06af48…` | ✅ |
| L14 (24 txs) | `b2e69a…` | `b2e69a3e…` | ✅ |
| L15 (0 txs) | `128613…` | `128613b9…` | ✅ |
| **L16 (0 txs)** | **`cb6a410d…`** | **`ea577a7e…`** | ❌ **DIVERGE** |

- All 100 loadgen transactions were applied across L11–L14 and **matched** stellar-core byte-for-byte (the prev-hash chain agrees through L15).
- Core's consensus for L16: `[seq=16, txs=0, ops=0, ct=1782148323, upgrades: [ ]]`. Henyey's L16 close: `tx_count=0, close_time=1782148323`, `prev` = L15 `128613b9…`. **Identical input, divergent output.**
- Confirmation of core's L16 hash: core's `seq=17` consensus carries `prev=cb6a41`. The knit-to-LCL error on henyey reported `expected ea577a7e… (henyey LCL), got cb6a410d… (downloaded core chain)`.

## Localization

The L16 ledger header differs while these fields are known-equal: `previousLedgerHash` (= L15, matched), `scpValue` (empty tx set + same close time + no upgrades), `txSetResultHash` (`df3f6198…`, the empty-set hash, on both), `ledgerSeq`, `totalCoins` (1e18), `feePool` (10000), and the fee/reserve/maxTxSetSize fields (the `maxtxsetsize=1000` upgrade applied at L10 and L10–L15 matched). The remaining unverified header field is **`bucketListHash`** — henyey's L16 `bucket_list_hash = 167a7373ec184de57a9ec3c10705df2015b8042b58e8c2cd184e9ab9d36ff849`.

**Prime suspect: the bucket-list merge at the ledger-16 boundary.** Ledger 16 is the first ledger that triggers a **deeper (level-1→level-2) bucket spill/merge** — ledgers 4, 8, 12 only trigger level-0→1 merges, and henyey matched all of those. The L11–L14 transactions populated level-0 buckets with real entries; the first time those propagate through a level-1→2 merge is L16. A divergence there (merge ordering, shadow/tombstone handling, or output-bucket hashing) would produce a different `bucketListHash` → different ledger hash, with zero transactions involved. This must be confirmed by an offline header-field diff (see repro).

## Reproduction (offline verify-execution — per CLAUDE.md hash-mismatch workflow)

Core's full published history archive (checkpoint 63, all headers 1–63) is snapshotted at:
`~/data/9eb89c28/ssc-v27-run/core-old-0-archive.tar.gz`

Steps:
1. Replay ledgers 1–16 from this archive through henyey's execution (offline verify-execution).
2. Diff henyey's L16 `LedgerHeader` against the archive's L16 header **field by field** to confirm `bucketListHash` is the divergent field (or identify the actual one).
3. If `bucketListHash`: diff the per-level bucket hashes at L16 to find which level/merge diverged; build a narrow `crate:bucket` unit test reproducing the level-1→2 merge with the L11–L14 entry set.

## Preserved artifacts (all under `~/data/9eb89c28/ssc-v27-run/`)

- `core-old-0-archive.tar.gz` — stellar-core's authoritative published chain (for offline replay)
- `henyey-prev.log` — henyey's full crash-instance log (all `computed_hash`/`bucket_list_hash` per ledger; the FATAL)
- `core-old-0.log` — stellar-core's log (consensus `prev=`/upgrade sequence)
- `console.log`, `ssc/*` — mission harness output + per-node metrics

## Related

- The FATAL recovery escalation (`knit-to-LCL failed at LCL … State wipe required. Node will shut down.`) is the **`#3410`** failure path. That recovery behavior is a secondary concern; the **root cause is this L16 ledger-close divergence**. Even with #3410's recovery fix, henyey would still produce a wrong L16 and be unable to stay in consensus.
- Image/run provenance: henyey `nscr.io/k4jkul01t5rr0/henyey:ssc-v27-8167588b` (main @ `8167588b`), core `nscr.io/k4jkul01t5rr0/stellar-core-testing:v27.0.0` (BUILD_TESTS, protocol 27).

## Severity

**High** — a consensus/state-hash divergence from stellar-core violates the core parity mandate. It is deterministic (reproducible from the preserved archive) and on the live ledger-close path (bucket merge at a level boundary), not a test-only artifact.

tomerweller:
Closes #3552

## Summary

Fixes the **residual** ledger-16 `bucketListHash` consensus divergence that survived the #3553 EvictionIterator-emission fix. At the L16 hot-archive level-1→2 spill, `merge_hot_archive_buckets` merges two empty inputs at the current protocol (27). It derived the output meta version from the input buckets (`max(curr, snap)` = 0 for empty/meta-less inputs, per `get_protocol_version`) and then **unconditionally pushed a `{ledgerVersion: 0, ext: V0}` metaentry**, producing the meta-only bucket `d15ebeb4…` where stellar-core produces an **empty** bucket.

stellar-core's `BucketOutputIterator` writes the METAENTRY only when `meta.ledgerVersion >= FIRST_PROTOCOL_SUPPORTING_INITENTRY_AND_METAENTRY` (V11) (`BucketOutputIterator.cpp:45-72`); for an empty-input merge the derived version is 0 (< V11) so **no metaentry is emitted** and `getBucket` returns an empty bucket (`mObjectsPut == 0`, `BucketOutputIterator.cpp:182-191`). The spurious `{v0,void}` bucket flipped henyey's hot-archive bucket-list hash, which combines `SHA256(live_hash || hot_archive_hash)` into the header `bucketListHash` — so henyey's L16 diverged (`759a1bbe…` vs core `9f58962b…`) and the node got stuck unable to externalize L17.

**Fix (parity-faithful, no new semantics):** emit the hot-archive merge metaentry only when `output_version >= V11`, and collapse a zero-entry merge result to the canonical empty bucket — mirroring core exactly. The #3553 EvictionIterator fix stays (necessary and correct); this is the disjoint residual.

## Pinned divergence (offline, deterministic)

Localization category: **(b/c)** — the tx-bearing level-1→2 boundary merge, specifically in the **hot-archive** sub-list, NOT the live bucket list and NOT the EvictionIterator value (candidate (a) was refuted: the iterator value is a stable `{level=6, is_curr=true, offset=0}` that matches core).

Decoded from the captured oracle (`ssc-v27-cap2` run, same bug): henyey's L16 bucket set = core's 14 buckets **+ one extra** `d15ebeb4f9249d16…` = a single `HOT_ARCHIVE_METAENTRY { ledgerVersion: 0, ext: V0 }` (16-byte payload `8000000c ffffffff 00000000 00000000`). Core's empty hot-archive bucket is `8559d8bc…` = `METAENTRY { ledgerVersion: 27, ext: V1(HotArchive) }` — present (shared, byte-identical) in both. The divergent bucket sits at hot-archive level-2 curr after the L16 spill.

## Validation — offline from-genesis close-ledger replay

Reproduced the plan's offline replay against the preserved SSC v27 archive (`ssc-3552-verify-archive.tar.gz`): 100-account genesis (passphrase `Private test network 'ssc-2346z-574ca2'`) **gated on L1 `bucketListHash == 1f2c1920…`** (matches core), then `close_ledger` L2→L16 with the archive's real generalized tx sets, decoded close times, `Version(27)@L6` + `MaxTxSetSize(1000)@L11`, disk-backed merges, inline eviction scan.

- **Pre-fix (origin/main `3ab671c8`):** L16 `bucketListHash = 759a1bbe…` ❌ (header `469d3f8f…` live).
- **Post-fix:** L16 `bucketListHash = 9f58962b4ef703fb…` ✅ and full header `header_hash = 46531748…` ✅ — matches core's authoritative L16. The **entire L1–L15 chain** matches core byte-for-byte.

- **Test:** `crates/bucket/src/hot_archive.rs::test_hot_archive_empty_input_merge_at_v27_yields_empty_not_v0_meta`
- **Pre-fix:** committed as `47b376cc` — verified FAILED with `left != right` producing the exact live-captured `d15ebeb4f9249d16…` `{v0,void}` meta-only bucket.
- **Post-fix:** verified PASSES after `bf4e823d` — the empty-input merge at protocol 27 yields the canonical empty bucket (zero hash).

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3552#issuecomment-4774385369)

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo build -p henyey-ledger -p henyey-bucket --all-targets` with `RUSTFLAGS=-Dwarnings` (#3384 gate)
- [x] `cargo build --all`
- [x] `cargo test --all` — 0 failures
- [x] **Parity tripwires (NO shift):** `ledger_close_meta_vectors` (2 ok), `tx_meta_hash_vectors` (1 ok/1 ignored), `validation_parity` (12 ok), `scp_parity_tests` (124 ok) — all byte-identical; the fix touches only the empty-input hot-archive merge path, which was producing the wrong (non-core) bucket, so no already-correct ledger-close hash shifts.

## Deviations from plan

- The plan placed the regression in `crates/ledger/tests/ledger_close_integration.rs` using `henyey_app::initialize_genesis`/`build_genesis_entries`. Those live in `henyey-app`, which depends on `henyey-ledger` — so they cannot be a dev-dep of `henyey-ledger` (circular). The full from-genesis L1→L16 replay was used to **localize and validate** the fix offline (reproducing core's `9f58962b…`/`46531748…`), but the committed regression is a **narrow `crate:bucket` unit test** that isolates the exact pinned mechanism (the empty-input hot-archive merge metaentry), per the plan's "narrow unit test isolating the pinned mechanism" clause. This is tighter, has no cross-crate fixture dependency, and fails on `3ab671c8` / passes after the fix.
- Localization category resolved to the **hot-archive** sub-list (the plan's candidate set spanned live-merge/(b) and a boundary field/(c)); the EvictionIterator value (a) was refuted.

### DIFF
--- a/crates/bucket/PARITY_STATUS.md
+++ b/crates/bucket/PARITY_STATUS.md
@@ -110,6 +110,7 @@ Corresponds to: `HotArchiveBucket.h`, `HotArchiveBucketList.h`, `HotArchiveBucke
 | `fresh()` | `HotArchiveBucket::fresh()` | Full |
 | `isTombstoneEntry()` / `maybePut()` | `is_hot_archive_tombstone()` / merge helpers | Full |
 | `mergeCasesWithEqualKeys()` | `merge_hot_archive_buckets()` | Full |
+| `BucketOutputIterator` metaentry gate (`meta.ledgerVersion >= V11`) | `merge_hot_archive_buckets()` output-version gate | Full — empty-input spill (output version 0) emits NO metaentry and collapses to the empty bucket, matching core's `getBucket` empty-output path; previously emitted a `{v0,void}` meta-only bucket (`d15ebeb4…`), flipping the hot-archive hash and the header `bucketListHash` at the L16 level-1→2 spill (#3552) |
 | `bucketEntryToLoadResult()` / `convertToBucketEntry()` | `get()` / `HotArchiveBucket::fresh()` conversion | Full |
 | `countOldEntryType()` / `countNewEntryType()` / `checkProtocolLegality()` | No-ops consistent with upstream | Full |
 | `HotArchiveBucketList::addBatch()` | `HotArchiveBucketList::add_batch()` | Full |
--- a/crates/bucket/src/hot_archive.rs
+++ b/crates/bucket/src/hot_archive.rs
@@ -1842,16 +1842,32 @@ pub fn merge_hot_archive_buckets(
     // memory for the HashMap AND a subsequent O((n+m) log(n+m)) sort.
     let mut result_entries = Vec::with_capacity(1 + curr.len() + snap.len());
 
-    // Add metadata - hot archive buckets always use V1 with BucketListType::HotArchive
-    // for Protocol 23+.
-    let mut meta = BucketMetadata {
-        ledger_version: output_version,
-        ext: BucketMetadataExt::V0,
-    };
-    if protocol_version_starts_from(output_version, ProtocolVersion::V23) {
-        meta.ext = BucketMetadataExt::V1(BucketListType::HotArchive);
+    // Emit the metaentry only when the derived output version supports it
+    // (>= V11), mirroring stellar-core's `BucketOutputIterator` constructor,
+    // which writes the METAENTRY only when `meta.ledgerVersion >=
+    // FIRST_PROTOCOL_SUPPORTING_INITENTRY_AND_METAENTRY` (V11)
+    // (BucketOutputIterator.cpp:45-72). When both inputs are empty/meta-less,
+    // `output_version` is 0 (BucketInputIterator defaults a meta-less bucket to
+    // protocol 0; BucketInputIterator.cpp:35-42) and core emits NO metaentry,
+    // so `getBucket` sees `mObjectsPut == 0` and returns an EMPTY bucket
+    // (BucketOutputIterator.cpp:182-191). Henyey previously pushed a
+    // `{ledgerVersion: 0, ext: V0}` metaentry unconditionally, producing the
+    // meta-only bucket `d15ebeb4…` where core has an empty bucket — flipping the
+    // hot-archive bucket-list hash and (via the live+hot-archive combine) the
+    // header `bucketListHash` (#3552 residual L16 divergence). Hot-archive
+    // buckets exist only at V23+, so a populated merge always uses the V1
+    // extension; the V0/no-meta branch is reachable only for the empty-input
+    // spill, which must collapse to the canonical empty bucket.
+    if protocol_version_starts_from(output_version, ProtocolVersion::V11) {
+        let mut meta = BucketMetadata {
+            ledger_version: output_version,
+            ext: BucketMetadataExt::V0,
+        };
+        if protocol_version_starts_from(output_version, ProtocolVersion::V23) {
+            meta.ext = BucketMetadataExt::V1(BucketListType::HotArchive);
+        }
+        result_entries.push(HotArchiveBucketEntry::Metaentry(meta));
     }
-    result_entries.push(HotArchiveBucketEntry::Metaentry(meta));
 
     // Sorted merge-join: iterate both iterators in lock-step.
     // Snap (newer) wins on equal keys.
@@ -1913,6 +1929,16 @@ pub fn merge_hot_archive_buckets(
                 }
             }
         }
+    }
+
+    // If the merge produced no entries at all (no metaentry because
+    // output_version < V11, and no surviving data entries), collapse to the
+    // canonical empty bucket (hash zero), mirroring stellar-core's
+    // `getBucket` empty-output path (BucketOutputIterator.cpp:182-191) rather
+    // than materializing a zero-entry bucket.
+    if result_entries.is_empty() {
+        tracing::trace!("hot_archive merge: empty result, returning empty bucket");
+        return Ok(HotArchiveBucket::empty());
     }
 
     let result = HotArchiveBucket::from_entries(result_entries)?;
--- a/crates/bucket/src/hot_archive.rs
+++ b/crates/bucket/src/hot_archive.rs
@@ -3283,4 +3283,64 @@ mod tests {
             err_msg
         );
     }
+
+    /// Regression for #3552 (residual L16 `bucketListHash` divergence).
+    ///
+    /// At the ledger-16 hot-archive level-1→2 spill, both merge inputs are
+    /// empty (no entries, no metaentry ⇒ `get_protocol_version()` == 0), but the
+    /// merge is invoked with the CURRENT protocol (e.g. 27) as `maxProtocolVersion`.
+    /// stellar-core's `BucketOutputIterator` gates the metaentry on
+    /// `meta.ledgerVersion >= FIRST_PROTOCOL_SUPPORTING_INITENTRY_AND_METAENTRY`
+    /// (V11) (`BucketOutputIterator.cpp:45-72`). With the derived output version
+    /// 0 (< V11) it emits NO metaentry, so `mObjectsPut == 0` ⇒ a truly EMPTY
+    /// bucket (`getBucket`, `BucketOutputIterator.cpp:182-191`).
+    ///
+    /// Henyey instead unconditionally pushed a `{ledgerVersion: 0, ext: V0}`
+    /// metaentry, producing the meta-only bucket `d15ebeb4…` where core has an
+    /// empty bucket. That flips the hot-archive bucket-list hash, which combines
+    /// with the live hash into the header `bucketListHash` ⇒ the L16 consensus
+    /// divergence (henyey `759a1bbe…` vs core `9f58962b…`).
+    ///
+    /// Pre-fix this produced a non-empty `{v0,void}` meta-only bucket
+    /// (`d15ebeb4f9249d16…`); post-fix it is the canonical empty bucket.
+    #[test]
+    fn test_hot_archive_empty_input_merge_at_v27_yields_empty_not_v0_meta() {
+        // The d15ebeb4 divergent bucket: a single `{ledgerVersion:0, ext:V0}`
+        // HOT_ARCHIVE_METAENTRY (the exact 16-byte payload captured live).
+        let v0_meta_only = HotArchiveBucket::from_entries(vec![HotArchiveBucketEntry::Metaentry(
+            BucketMetadata {
+                ledger_version: 0,
+                ext: BucketMetadataExt::V0,
+            },
+        )])
+        .unwrap();
+        let v0_meta_only_hash = v0_meta_only.hash();
+
+        // Spill of two empty hot-archive buckets at the current protocol (27),
+        // exactly as the L16 level-1→2 hot-archive spill invokes it.
+        let merged = merge_hot_archive_buckets(
+            &HotArchiveBucket::empty(),
+            &HotArchiveBucket::empty(),
+            27,
+            true,
+        )
+        .unwrap();
+
+        assert_ne!(
+            merged.hash(),
+            v0_meta_only_hash,
+            "empty-input merge must NOT produce the {{v0,void}} meta-only bucket (d15ebeb4)"
+        );
+        assert!(
+            merged.is_empty(),
+            "empty-input merge at protocol 27 must yield an EMPTY hot-archive bucket (parity \
+             with stellar-core: output version 0 < V11 ⇒ no metaentry ⇒ empty), got hash {}",
+            merged.hash().to_hex()
+        );
+        assert!(
+            merged.hash().is_zero(),
+            "empty hot-archive bucket hash must be zero, got {}",
+            merged.hash().to_hex()
+        );
+    }
 }


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** kind: bug-fix. Regression test committed before the fix (47b376cc), verified RED at the test-commit producing the exact live-captured divergent bucket d15ebeb4f9249d16…, and GREEN after the fix (bf4e823d). The diff is minimal, well-commented, and the empty-output collapse + V11 metaentry gate are correct.

<details>

Cycle 1 — complete change-list. No prior Correctness comment on this PR.

**Test verification (`test-coverage`) — PASS:**
- PR body has a `## Regression test` section. Test = `crates/bucket/src/hot_archive.rs::test_hot_archive_empty_input_merge_at_v27_yields_empty_not_v0_meta`.
- Commit order correct: test `47b376cc` precedes fix `bf4e823d`.
- Re-ran the test at the test-only parent `47b376cc` (no fix): **FAILED** with `left != right`, `left: Hash256(d15ebeb4f9249d16be3e6d76c31d07408f43461760f6876abe01ac690724db5c)` — the exact divergent `{v0,void}` meta-only bucket the issue pins. The test genuinely captures the bug.
- Re-ran at the fix commit merged with current `origin/main`: **ok. 1 passed**. (Merge was a no-op — PR head already contains current main.)
- The test asserts all three required properties: not-equal to the d15ebeb4 v0-meta bucket, `is_empty()`, and `hash().is_zero()`.

**Logic (`correctness`) — PASS:**
- Two changes: (1) the metaentry push is now wrapped in `if protocol_version_starts_from(output_version, V11)`; the inner `V23 ⇒ V1(HotArchive)` extension logic is preserved unchanged, so populated hot-archive merges (always V23+) are byte-identical to before. (2) after the merge loop, `if result_entries.is_empty() { return Ok(HotArchiveBucket::empty()) }` collapses a zero-entry result to the canonical empty bucket.
- The two branches compose correctly: the only way `result_entries` is empty is when output_version < V11 (no meta) AND no surviving data entries — exactly the empty-input spill. A populated merge always has ≥1 data entry plus a V23 metaentry, so the collapse never fires for real data.
- No error-handling regression: `from_entries` is still called on the non-empty path.

**Readability — PASS:** the inline comments cite the precise core file:line anchors (BucketOutputIterator.cpp:45-72, 182-191; BucketInputIterator.cpp:35-42) and explain the d15ebeb4 mechanism. PARITY_STATUS.md updated with the new gate row.

**Consistency note (non-blocking):** henyey's own *live* merge path (`crates/bucket/src/merge.rs:619-630` and the empty-collapse at `:354-355,417-425,709-710`) already does exactly this — V11-gated metaentry + collapse-empty-to-empty-bucket. This PR brings the hot-archive path into line with both core and the pre-existing live path. Strong internal corroboration.
</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** The fix mirrors stellar-core v27.0.0 `BucketOutputIterator` exactly — metaentry gated on `meta.ledgerVersion >= V11` and an empty merge returns an empty bucket — for the hot-archive output path, matching the live path core and henyey already implement. All four parity tripwires re-run with NO shift; the offline replay reproduces core's authoritative L16. No new semantics.

<details>

Cycle 1 — complete change-list. No prior Parity comment on this PR. CONSENSUS-CRITICAL bucket-output change; reviewed against the `stellar-core/` submodule (v26.0.1 pin, V11 semantics unchanged through v27).

**parity-gap — RESOLVED (the fix), verified against core source:**
- `BucketOutputIterator.cpp:45-72`: the constructor writes the METAENTRY (live or HOT_ARCHIVE_METAENTRY) only inside `if (protocolVersionStartsFrom(meta.ledgerVersion, FIRST_PROTOCOL_SUPPORTING_INITENTRY_AND_METAENTRY))`. Confirmed `FIRST_PROTOCOL_SUPPORTING_INITENTRY_AND_METAENTRY = ProtocolVersion::V_11` (LiveBucket.h:77-78). The PR's `if protocol_version_starts_from(output_version, ProtocolVersion::V11)` is the exact mirror.
- `BucketOutputIterator.cpp:182-191` (`getBucket`): when `mObjectsPut == 0` it deletes the file and `return std::make_shared<BucketT>()` — an empty bucket. The PR's `if result_entries.is_empty() { return Ok(HotArchiveBucket::empty()) }` mirrors this. An empty-input merge derives output version 0 (< V11) ⇒ no metaentry ⇒ zero entries ⇒ empty bucket. Exactly core's path.
- The metaentry is gated for BOTH output paths as core does: the live path was already V11-gated (`merge.rs:619`) with empty-collapse (`merge.rs:354,417,709`); this PR makes hot-archive symmetric. Both live and hot-archive now match core.

**no-suppression check — PASS (the critical safety question):** the fix does NOT suppress any metaentry core would emit. A populated hot-archive bucket only ever exists at V23+ (hot-archive is a V23 feature), so for any real merge `output_version >= V23 >= V11` and the metaentry (V1/HotArchive ext) is emitted exactly as before — the V23 inner branch is untouched. Only the empty/<V11 spill loses the (spurious) metaentry, which is precisely what core also omits. No legitimate metaentry is dropped.

**edge-case — PASS:** the only newly-reachable branch is output_version 0 < V11 (empty inputs). Verified the empty-collapse cannot mask a real data merge: a non-empty merge always carries ≥1 entry, so `result_entries.is_empty()` is false.

**Dispositive evidence (re-confirmed, not taken on trust):**
- Regression test RED at parent `47b376cc` produced the exact `d15ebeb4f9249d16…` `{v0,void}` bucket; GREEN at fix.
- All 4 parity tripwires re-run on the fix+main tree — counts byte-match the PR's claim: `ledger_close_meta_vectors` 2 ok, `tx_meta_hash_vectors` 1 ok/1 ignored, `validation_parity` 12 ok, `scp_parity_tests` 124 ok. **NO shift** — a bucket-output change is exactly the kind that could shift already-correct ledger-close hashes; none did, confirming the fix only corrects the empty hot-archive path.

**#3553 coexistence — PASS:** this PR touches only `merge_hot_archive_buckets` metaentry emission. It does not touch the EvictionIterator-emission path (#3553), so the live bucket list still carries the EvictionIterator entry. The two L16 fixes are disjoint and both remain in effect.

**Accepted deviation:** the regression is a narrow `crate:bucket` unit test rather than a from-genesis `ledger_close_integration` test — reasonable, since the genesis builders live in `henyey-app` (a reverse dep of `henyey-ledger`), making the integration form a circular dev-dep. The full offline from-genesis replay (L1 gated on `1f2c1920…`, post-fix L16 `bucketListHash 9f58962b…` + header `46531748…` matching core) was used to localize/validate. The unit test isolates the exact pinned mechanism. Accepted.
</details>

tomerweller:
## ✅ Merged

**Commit:** 74ac4c987eb13b20fc99ee7aa55c4300f0f514af

Both agent reviewers APPROVE (Correctness + Parity). Required CI green (Test, Build release-ci, Clippy, Doctest, Format, gitlink, Soroban Pin, Skill-Snippets). Non-required galexie shard red is the known-failing #3563 shard (explicitly non-blocking); testnet quickstart shard non-required.

Parity confirmed against stellar-core `BucketOutputIterator` (metaentry gated on `>= V11`, empty merge → empty bucket) for the hot-archive path, matching the live path. Regression test verified RED at the test-commit (producing the exact d15ebeb4 divergent bucket) and GREEN at the fix. All 4 parity tripwires re-run with NO shift.

PR #2797, opened 2026-05-18
Wire ScpPersistenceManager::restore_scp_state into startup

tomerweller:
Closes #2769

## Summary

Wires `ScpPersistenceManager::restore_scp_state()` into `init_herder()` so a node recovering from a crash can resume SCP tracking from persisted state rather than waiting for fresh network envelopes. Adds `Herder::apply_restored_scp_state()` which replays persisted tx sets, quorum sets, and SCP envelopes into the herder's in-memory caches, matching stellar-core's `restoreSCPState()` (HerderImpl.cpp:2189-2261).

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/2769#issuecomment-4474431893)

## Test plan

- [x] cargo fmt --check
- [x] cargo clippy -p henyey-herder -p henyey-app -- -D warnings
- [x] cargo test -p henyey-herder (1123 tests pass)

## Changes

- `crates/app/src/app/mod.rs`: Call `restore_scp_state()` + `apply_restored_scp_state()` in `init_herder()` after installing the persistence manager
- `crates/herder/src/herder.rs`: Add `apply_restored_scp_state()` method that replays tx sets, quorum sets, and envelopes in stellar-core order
- `crates/herder/src/scp_driver.rs`: Add `store_quorum_set_by_hash()` for hash-only quorum set storage during restore
- `crates/herder/src/quorum_set_tracker.rs`: Add `store_by_hash()` inserting into `by_hash` cache without NodeId association
- `crates/herder/PARITY_STATUS.md`: Update footnote [^txset-gc] recording restore wiring is complete

## New tests

- `test_apply_restored_scp_state_loads_envelopes_into_scp`
- `test_apply_restored_scp_state_loads_quorum_sets`
- `test_apply_restored_scp_state_loads_tx_sets`
- `test_apply_restored_scp_state_skips_corrupt_tx_set`

## Deviations from plan

- Test assertions use `scp.get_entire_current_state(slot)` instead of `get_slot_envelopes(slot)` because `set_state_from_envelope` stores into nomination's `latest_nominations` (not the slot-level `envelopes` map), and local-node envelopes are only visible via `get_entire_current_state` when the slot is not fully validated.
- Omitted the integration test in `crates/herder/tests/` — the unit tests already cover the full restore flow end-to-end without needing a separate file since `apply_restored_scp_state` is a pure in-memory operation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3603, opened 2026-06-24
Drive restored externalized lcl+1 slot to ledger close on restart

### LINKED ISSUE #3595: Investigate restored-slot live-replay restart stall — evidenced cause is recovery/re-emission, not slot purge
Spun out of #3555 (purge-parity re-scope). #3555 brings `SCP::purge_slots` to parity with stellar-core `SCP::purgeSlotsOutsideRange` (adds the missing upper/max-slot bound). During that plan it was **evidenced that the purge gap is NOT the cause** of the `Herder::RESTORE_LIVE_ENVELOPE_REPLAY` restart-rejoin stall — so live-replay re-enablement remains an OPEN investigation tracked here.

## Evidenced cause (why purge-parity does not enable live-replay)

In `crates/simulation/tests/app_simulation.rs::test_core3_restart_rejoin_over_loopback`: node0 is removed after ledger 2, the network closes **ledger 3** without it, node0 restarts with `lcl = 2` and a persisted **slot-3** local envelope (`slot=3 > lcl=2`, retained by the future-slot filter). With `RESTORE_LIVE_ENVELOPE_REPLAY` enabled, node0 installs its own slot-3 ballot via `set_state_from_envelope`.

Slot 3 is `<= nextConsensusLedgerIndex() + LEDGER_VALIDITY_BRACKET` (≈103), so it is **inside the retention window of BOTH** henyey and stellar-core — neither core's max-bound nor a fixed henyey max-bound evicts slot 3. Therefore the slot-purge upper-bound gap cannot be the mechanism that stalls (or would un-stall) slot 3.

The likely true wedge (from the #3555 round-1 investigation, accepted by the operator): the **restored-slot re-emission / recovery-routing interaction** — a restored not-fully-validated local slot sets `last_envelope_emit` and `fully_validated=false`, so `send_latest_envelope` (`crates/scp/src/ballot/envelope.rs:12`) returns early and node0 never re-drives the slot, while peers wait. Core tolerates a restored in-flight slot via replay-all + catchup advancing lcl, NOT via ballot supersession (which both core and henyey correctly refuse). The ballot machine is confirmed core-faithful; do NOT revive the PR #2797 supersession approach.

## Work needed (investigation-first)

1. With `RESTORE_LIVE_ENVELOPE_REPLAY` re-enabled behind a test path, reproduce the stall and capture the precise wedge state (node0's slot-3 phase + `fully_validated` + whether the network EXTERNALIZE is rejected-by-value vs accepted-but-silent).
2. Determine the parity-faithful remedy (candidate: ensure post-restore recovery / `restore_slot_fully_validated` + `request_scp_state_from_peers` + catchup drives or discards the silent restored slot exactly as core does — NO new ballot semantic).
3. Only then re-enable `RESTORE_LIVE_ENVELOPE_REPLAY` and gate on `test_core3_restart_rejoin_*` staying green ≥5 runs (the gate PR #2797 lacked).

## References
- Parent: #3555; original misdiagnosis + prior failed attempt PR #2797.
- henyey: `Herder::restore_persisted_scp_state` + `RESTORE_LIVE_ENVELOPE_REPLAY` (`crates/herder/src/herder.rs`), `crates/scp/src/ballot/envelope.rs:12`.
- stellar-core: `HerderImpl::restoreSCPState` (`HerderImpl.cpp:2239`), `purgeOldSlotsAndProcessSCPQueue` (`HerderImpl.cpp:260`).

tomerweller:
Closes #3595

## Summary

Re-enables `Herder::RESTORE_LIVE_ENVELOPE_REPLAY` and makes
`restore_persisted_scp_state` complete the externalization of a restored,
SCP-externalized next-to-close slot (`slot == lcl + 1`), fixing the
restart-rejoin consensus stall.

**The wedge.** With replay on, restore installs node0's persisted slot-`lcl+1`
**Externalize** via `set_state_from_envelope` — the ballot is pinned to
`phase=Externalize` and the externalized value is recorded in the ballot — but,
exactly like stellar-core's `BallotProtocol::setStateFromEnvelope`
(`BallotProtocol.cpp:1898-1910`, which sets `mLastEnvelope=mLastEnvelopeEmit` and
does NOT call `valueExternalized`), this does **not** close the ledger. The
matching network EXTERNALIZE for that slot is then accepted-but-silent
(value-match → `EnvelopeState::Valid` → herder maps to `Duplicate`,
`herder.rs:2588`), so node0 wedged at `phase=Externalize` with LCL stuck. Core
escapes at the ledger layer via its recovery/catchup machinery; the 3-node
restart-rejoin simulation has no history archive to catch up from.

**The fix (parity-faithful, no new SCP/ballot semantics).** After the replay
loop, for the **distinct** restored slot satisfying `slot == lcl + 1` AND
`scp.is_slot_externalized(slot)`, run the *same* sequence the live externalize
path uses (`request_tx_set` + `record_externalized` + `cleanup_externalized` +
`complete_externalization`) so the externalized value reaches the app's
`process_externalized_slots` / `out_of_sync_recovery` close pipeline. This is the
faithful henyey analog of core's `restoreSCPState → processExternalized →
mLedgerManager.valueExternalized` (`HerderImpl.cpp:2239` / `311-396`). PR #2797
ballot supersession is **not** revived. Higher restored slots (`> lcl+1`) are not
force-closed; restored Prepare/Confirm slots are no-ops.

The SCP layer is unchanged (byte-faithful). The change is herder-only plus a
restart-rejoin sim assertion move (see Deviations).

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3595#issuecomment-4785691431)

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo build -p henyey-herder -p henyey-scp -p henyey-simulation --all-targets` with `RUSTFLAGS=-Dwarnings`, `cargo build --all`
- [x] `cargo test -p henyey-herder` — 1194 passed (incl. the #2769 auto-tracking `test_restore_persisted_scp_state_replays_future_slot_and_deps`, now `!is_empty` with the flag on)
- [x] `cargo test -p henyey-scp` + `scp_parity_tests` (124), `validation_parity` (12), `ledger_close_meta_vectors` (2), `tx_meta_hash_vectors` (1) — all GREEN and **unshifted** (no SCP-layer change)
- [x] **Race gate:** `test_core3_restart_rejoin_over_loopback` **10/10**, `test_core3_restart_rejoin_over_tcp` **10/10** (timeout-wrapped, 180s each)

- **Tests:** `herder::tests::test_restore_externalized_next_slot_drives_ledger_close` (positive) and `::test_restore_prepare_next_slot_does_not_publish_externalized` (negative)
- **Pre-fix:** committed as `cd538409` — positive verified FAILED (`restored Externalize must mark slot lcl+1 externalized in SCP`: with the flag off, restore never installs/completes the slot)
- **Post-fix:** both verified PASS after `64ae79be`
- **Integration gate:** reproduced the stall on main+flag (FAIL correlated exactly with restore installing both the slot-3 Nominate *and* Externalize); GREEN 10/10 loopback + 10/10 tcp after the fix

## Deviations from plan

- The plan named `complete_externalization(slot)` as the terminal. In practice that call only publishes `latest_externalized`; the **live** externalize path it pointed to (`herder.rs:2596-2616`) also runs `request_tx_set` + `record_externalized` + `cleanup_externalized` first, and the app close pipeline (`process_externalized_slots`, which reads `get_externalized(next_slot)`) requires the recorded value. The fix therefore mirrors that **full** live sequence, not just `complete_externalization`. Same terminal, same end-state — just the complete sequence rather than the single final call.
- The restart-rejoin sim test's `assert_scp_intake_reached` check is moved to **after** `manual_close_until(4)`. With restored-Externalize completion, node0 can now legitimately close ledger 3 directly from its own persisted, SCP-externalized slot **without** any live SCP intake, so the pre-ledger-4 checkpoint is no longer a reliable witness of intake. Closing ledger 4 still requires node0's live SCP participation, so the assertion's intent (prove the intake pipeline works post-restart) is preserved, not weakened.

### DIFF
--- a/crates/herder/PARITY_STATUS.md
+++ b/crates/herder/PARITY_STATUS.md
@@ -150,16 +150,42 @@ of two intentional, architecture-forced adaptations (NOT upstream parity):
    rejects superseding live envelopes — `ballot/mod.rs:854`). This is the
    regression guard for the consensus stall that closed PR #2797.
 
-2. **Live-envelope replay is currently gated OFF**
-   (`Herder::RESTORE_LIVE_ENVELOPE_REPLAY == false`). #2769 ships the
-   dependency hydration (decode + cache the referenced tx-sets and quorum-sets
-   for retained `slot > lcl` envelopes) and the transitive-quorum rebuild, but
-   does NOT replay the future-slot envelopes via `set_state_from_envelope`,
-   because doing so reproduces the PR #2797 stall (the simulation restart suite
-   `test_core3_restart_rejoin_over_loopback` stalls ~4/5 runs with replay on,
-   and is deterministically green with it off). Re-enabling live replay is
-   tracked by #3555 (allow a restored ballot slot to be superseded by a later
-   network EXTERNALIZE); once that lands the flag flips to `true`.
+2. **Live-envelope replay is enabled with a proactive restore-time
+   externalization completion** (`Herder::RESTORE_LIVE_ENVELOPE_REPLAY == true`,
+   re-enabled in #3595). Beyond the dependency hydration (decode + cache the
+   referenced tx-sets and quorum-sets) and transitive-quorum rebuild, restore
+   replays the future-slot envelopes via `set_state_from_envelope`. Replaying a
+   restored **Externalize** installs the ballot in `phase=Externalize` and
+   records the externalized value but — exactly like core's
+   `BallotProtocol::setStateFromEnvelope` (`BallotProtocol.cpp:1898-1910`, which
+   sets `mLastEnvelope=mLastEnvelopeEmit` and does NOT call `valueExternalized`)
+   — does NOT itself close the ledger. The matching network EXTERNALIZE for that
+   slot is then accepted-but-silent (value-match → `EnvelopeState::Valid` →
+   herder `Duplicate`), so without an explicit completion node0 would wedge at
+   `phase=Externalize` with LCL stuck (the PR #2797 / #3595 restart-rejoin
+   stall). The fix is **not** ballot supersession: after the replay loop,
+   restore drives `complete_externalization(slot)` for the **distinct** restored
+   slot satisfying `slot == lcl + 1` **and** `scp.is_slot_externalized(slot)` —
+   the faithful henyey analog of core's `HerderImpl::processExternalized →
+   mLedgerManager.valueExternalized` (`HerderImpl.cpp:311-396`), reached from
+   `restoreSCPState` (`HerderImpl.cpp:2239`). This publishes `latest_externalized`
+   and the app's `out_of_sync_recovery` loop closes ledger `lcl+1`.
+
+   **Trigger-point divergence (intentional).** Core reaches that terminal for a
+   restored slot via its recovery/catchup machinery
+   (`setTrackingSCPState`+`processSCPQueue`+`outOfSyncRecovery`/`getMoreSCPState`,
+   then catchup advancing LCL); henyey triggers it directly from the restore
+   hook. The end-state is identical (LCL → `lcl+1`) and the terminal call is the
+   same — only the trigger site differs. The proactive close is required because
+   henyey's 3-node restart-rejoin simulation has no history archive to catch up
+   from. The `slot == lcl+1` gate mirrors core's "only the next-to-close slot
+   applies immediately" invariant; higher restored slots wait for their
+   predecessor, and restored Prepare/Confirm slots are not externalized so the
+   completion is a no-op for them (never force-closed). The SCP layer itself is
+   byte-faithful (`setStateFromEnvelope`/`sendLatestEnvelope`/slot ctor verified)
+   and is NOT modified. Gated on the simulation restart suite
+   `test_core3_restart_rejoin_over_{loopback,tcp}` staying green across ≥10
+   consecutive runs (the race the gate PR #2797 lacked).
 
    The restore-time quorum-tracker rebuild follows upstream's
    local-node-preserving closure shape (`PendingEnvelopes.cpp:924-955`): the
--- a/crates/herder/src/herder.rs
+++ b/crates/herder/src/herder.rs
@@ -9921,6 +9921,226 @@ mod tests {
         )
     }
 
+    /// Build the StellarValue + its referenced deps for a restore test
+    /// envelope: returns `(value, tx_set_hash, stored_tx_set_bytes,
+    /// quorum_set_hash, quorum_set)`. Shared by the Externalize / Prepare
+    /// helpers below so they reference identical deps to the Nominate helper.
+    fn make_restore_value_with_deps(
+        herder: &Herder,
+        secret: &SecretKey,
+    ) -> (
+        Value,
+        stellar_xdr::Hash,
+        Vec<u8>,
+        stellar_xdr::Hash,
+        ScpQuorumSet,
+    ) {
+        use stellar_xdr::WriteXdr;
+
+        let lcl_hash = herder.scp_driver.current_header_hash();
+        let tx_set = TransactionSet::new(lcl_hash, Vec::new());
+        let tx_set_hash = stellar_xdr::Hash(tx_set.hash().0);
+        let stored_bytes = tx_set
+            .to_xdr_stored_set()
+            .to_xdr(Limits::none())
+            .expect("encode StoredTransactionSet");
+
+        let close_time = TimePoint(1);
+        let network_id = herder.scp_driver.network_id();
+        let mut sign_data = network_id.0.to_vec();
+        sign_data.extend_from_slice(&EnvelopeType::Scpvalue.to_xdr(Limits::none()).expect("xdr"));
+        sign_data.extend_from_slice(&tx_set_hash.to_xdr(Limits::none()).expect("xdr"));
+        sign_data.extend_from_slice(&close_time.to_xdr(Limits::none()).expect("xdr"));
+        let value_sig = secret.sign(&sign_data);
+
+        let value_node_id = XdrNodeId(stellar_xdr::PublicKey::PublicKeyTypeEd25519(
+            stellar_xdr::Uint256(*secret.public_key().as_bytes()),
+        ));
+        let stellar_value = StellarValue {
+            tx_set_hash: tx_set_hash.clone(),
+            close_time,
+            upgrades: vec![].try_into().unwrap(),
+            ext: StellarValueExt::Signed(LedgerCloseValueSignature {
+                node_id: value_node_id,
+                signature: stellar_xdr::Signature(
+                    value_sig.0.to_vec().try_into().unwrap_or_default(),
+                ),
+            }),
+        };
+        let value = Value(
+            stellar_value
+                .to_xdr(Limits::none())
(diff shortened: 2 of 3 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Cycle 1. Tight, well-scoped herder-restore-orchestration fix. Failing-test-first verified RED→GREEN, positive+negative unit tests present, idempotency/double-close guarded at the app layer, race gate reproduced green. No blocking concerns.

<details>

**kind: bug-fix** (consensus stall) — regression-test discipline verified:

- Commit order correct: regression tests `cd538409` land BEFORE the fix `64ae79be`.
- **Pre-fix RED confirmed:** checked out `cd538409` and ran `test_restore_externalized_next_slot_drives_ledger_close` → `FAILED` at `restored Externalize must mark slot lcl+1 externalized in SCP`. The test genuinely captures the bug.
- **Post-fix GREEN confirmed:** at PR head, both `test_restore_externalized_next_slot_drives_ledger_close` (positive) and `test_restore_prepare_next_slot_does_not_publish_externalized` (negative) PASS; full `henyey-herder` lib suite 1194/1194.
- The #2769 auto-tracking test `test_restore_persisted_scp_state_replays_future_slot_and_deps` now asserts `RESTORE_LIVE_ENVELOPE_REPLAY == !get_current_state_for_slot(lcl+1).is_empty()` — flag-coupled, flips to the installed branch automatically with the constant; PASS.

**Logic correctness:**

- The restore completion sequence (`request_tx_set` + `record_externalized` + `cleanup_externalized` + `complete_externalization`, herder.rs:1130-1142) is a byte-for-byte mirror of the live externalize block (herder.rs:2658-2678). The deviation from the plan's `complete_externalization`-alone is correct: `complete_externalization` only publishes `latest_externalized`; the app close pipeline (`process_externalized_slots` → `get_externalized`) needs the value `record_externalized` writes, so the single call would leave the value unrecorded and the close would not fire. The fuller sequence is required, not gold-plating.
- **Gating is airtight:** completion fires only when `slot == lcl+1` AND `completed_slots.insert(slot)` (dedupe, so once per distinct slot) AND `is_slot_externalized(slot)` (SCP-layer truth) AND `get_externalized_value(slot)` is `Some`. A restored Prepare/Confirm/Nominate-only slot is not externalized → no-op (proven by the negative test). Higher slots (`> lcl+1`) are skipped. Cannot fire for the wrong slot.
- **Error handling:** `StellarValue::from_xdr` failure is a graceful `if let Ok` (tx_set request skipped) — but `record_externalized`/`complete_externalization` still run on the raw value; consistent with the live path which also tolerates the parse failing only for the tx_set request. `get_externalized_value` returning `None` is a clean `else { continue }`.

**Readability:** the inline comment block (herder.rs:1078-1117) and the PARITY_STATUS footnote thoroughly document the wedge, the fix, and the deviation. Excellent.

**Verification I ran:** herder lib 1194/1194; positive RED at parent / GREEN at head; race gate loopback 4/4 + tcp 3/3 green (timeout-wrapped 180s, on top of the /do's 10/10+10/10).

No blocking concerns; no inline follow-ups.

</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** Cycle 1. The full restore-completion sequence is the faithful henyey analog of core's `restoreSCPState → processExternalized → valueExternalized`. SCP layer byte-unmodified; all four parity tripwires unshifted. The trigger-point divergence is an honest, well-documented archive-forced adaptation, not a semantic divergence. No parity gaps.

<details>

Verified against `stellar-core/` (v26.0.1 submodule).

**`setStateFromEnvelope` does NOT close — confirmed.** `BallotProtocol::setStateFromEnvelope` (BallotProtocol.cpp:1898-1910) sets `mLastEnvelope = e; mLastEnvelopeEmit = mLastEnvelope;` and does NOT call `valueExternalized` (the only `valueExternalized` call is in the emit/attemptBump path, BallotProtocol.cpp:1693). So a restored Externalize in core installs the ballot at `phase=Externalize` without closing — exactly henyey's behavior. The wedge analysis in the PR is correct.

**The full sequence IS the faithful analog of `processExternalized` — DEVIATION VERDICT: FAITHFUL.** `HerderImpl::processExternalized` (HerderImpl.cpp:311-396) does, in order:
1. `mPendingEnvelopes.getTxSet(value.txSetHash)` — fetch the tx_set ⇔ henyey `request_tx_set(sv.tx_set_hash, slot)`.
2. builds `LedgerCloseData ledgerData(slotIndex, externalizedSet, value)` from the externalized **value** ⇔ henyey `record_externalized(slot, value, None)` (records the value so the close pipeline can construct close data — this is precisely the step `complete_externalization`-alone would skip).
3. `mLedgerManager.valueExternalized(ledgerData, isLatestSlot)` — advance/close ⇔ henyey `complete_externalization(slot)` → `advance_tracking_slot` + `publish_externalized` → app `process_externalized_slots` closes lcl+1.
   (`cleanup_externalized` is henyey-local bookkeeping, no core counterpart, harmless.)
   ORDER and SEMANTICS match: tx_set → record value → advance/close. The /do's deviation from the plan is the correct call.

**Trigger-point divergence is intentional and documented (not a parity violation).** Core's `restoreSCPState` (HerderImpl.cpp:2239-2311) itself only does `setStateFromEnvelope` + `rebuildQuorumTrackerState` — it does NOT call `processExternalized`. Core reaches `processExternalized` for the restored slot via its **separate** recovery/catchup machinery (setTrackingSCPState + processSCPQueue + outOfSyncRecovery/getMoreSCPState, then catchup advancing LCL from the archive). Henyey's 3-node restart-rejoin sim has no history archive, so it triggers the *same terminal* directly from the restore hook. End-state identical (LCL→lcl+1), terminal call identical; only the trigger site differs. PARITY_STATUS.md §2 documents this as a deliberate architecture-forced adaptation. This is a trigger-site change, not a change to the close/externalize semantics themselves.

**Value correctness (safe):** `get_externalized_value(slot)` is the node's OWN persisted Externalize value for a slot it had already SCP-externalized (committed) — not an unvalidated network value. `is_slot_externalized` is the SCP ballot-machine witness. Recording its own committed value is safe.

**Idempotency / no double-apply (consensus-safe):** after restore closes lcl+1, `current_ledger` advances. When the live network EXTERNALIZE for lcl+1 later arrives it is a value-match → SCP `Valid` (not-new) → `process_scp_envelope_with_tx_set` returns `Duplicate` at herder.rs:2650 BEFORE the externalize block. Even if `latest_externalized` is re-set, `process_externalized_slots` skips any `slot <= current_ledger` (ledger_close.rs:1546, `skipped_stale`) and tracks `last_processed_slot`. No second close/apply of lcl+1.

**SCP layer unchanged:** diff touches only herder.rs + app_simulation.rs + PARITY_STATUS.md. `set_state_from_envelope`/ballot/nomination untouched. Confirmed tripwires GREEN and unshifted by running them: scp_parity_tests **124**, validation_parity **12**, ledger_close_meta_vectors **2**, tx_meta_hash_vectors **1** — exactly the claimed counts.

No parity gaps; no inline follow-ups.

</details>

PR #3604, opened 2026-06-24
Fix flaky create_upgrade in henyey-majority SSC missions (#3602)

### LINKED ISSUE #3602: Flaky create_upgrade: applies but writes no resolvable ConfigUpgradeSet entry (~15-20%, henyey-majority SSC missions)
## Summary

In the supercluster mixed-image `MixedImageLoadGeneration` missions, the henyey-driven `create_upgrade` step intermittently (~15-20% of runs) **applies the create_upgrade soroban transaction but produces no resolvable `ConfigUpgradeSet` entry**. When this happens the arm is rejected (`configupgradesetkey did not resolve to a VALID upgrade set`), the Soroban config-settings upgrade never applies, and the mission hangs at `Waiting for LedgerMaxInstructions=250000000`.

This is a follow-up to the (now-closed) create_upgrade fix chain merged in #3601. Those fixes made the upgrade work in the common case; this issue is the residual flaky failure that survives them.

## Affected

- Both henyey-loadgen compositions: **1 core + 2 henyey** and **0 core + 3 henyey** (loadgen runs on the henyey majority, so henyey drives `create_upgrade`).
- ~15-20% of runs (rough rate from ~10 observed runs: ~2 failures).

## Observed behavior

- The create_upgrade tx **is included in a closed ledger** (`tx_count=1`) and consumes the source account's sequence number (so `wait_till_complete` / account-sync returns), but the `ConfigUpgradeSet` `ContractData` entry is **never resolvable** via `make_from_key` (`validate_config_upgrade_set_key` returns false; a manual re-arm minutes later still fails).
- **Consistent within a failing run**: the loadgen's resubmit logic (#3601) re-submits the idempotent create_upgrade up to 8×; every re-submission is *included* in a ledger (observed `tx_count=1` at the resubmit cadence ledgers) yet **none** yields a resolvable entry. So the retry does not rescue this mode.
- **Flaky across runs**: the same upgrade-set content hash (`c6ec918c…`) resolves fine in ~80% of runs (passing runs show the create_upgrade applying with `e2e_invoke OK n_changes=3`, `rent_fee≈7,983,102`, and the entry `FOUND` immediately).
- Genesis Soroban config caps soroban txns at **1 per ledger** (`ledgerMaxTxCount=1`), and the create_upgrade competes with the upload/deploy setup txns for that single slot.

## Hypotheses (unconfirmed)

The deterministic-looking invoke (same content) giving different results across runs points to a state-dependent at-apply outcome. Candidates:
- The create_upgrade invoke flakily **fails at apply** (writes nothing) — e.g. the hardcoded `CREATE_UPGRADE_RESOURCE_FEE = 50_000_000` / `write_bytes = 3100` is marginal for some ledger states (the contract extends two TTLs by ~1,036,800 ledgers; rent ≈ 8M in the passing case), and a `txFAILED`/insufficient-refundable-fee path would consume the seq without writing and without a host diagnostic event.
- Or it **applies and writes** but the TEMPORARY entry isn't visible in the snapshot the arm/nomination queries (a bucket-list/snapshot commit-ordering or temp-entry retention issue).

## How to root-cause (diagnostic harness)

The failure has **evaded capture** so far — every diagnostic-instrumented re-run happened to land in the passing 80%. To catch it, add temporary `warn!` instrumentation and loop runs until one fails:
- `crates/tx/src/soroban/host.rs` (P27 `invoke_host_function`): log `e2e_invoke` OK-vs-FAILED + `n_changes` (0 = no write vs 3 = entry written) + the host error on failure, and the computed `rent_fee` (vs the declared resource fee).
- `crates/ledger/src/config_upgrade.rs` `make_from_key`: log entry FOUND vs NOT-FOUND (and which reject branch: TTL / durability / decode).
- `crates/herder/src/tx_queue/selection.rs`: log soroban-lane selection summary (`soroban_accounts` vs `selected`, `had_not_fitting`).

Build: `nsc build -f Dockerfile --platform linux/amd64 --push -n nscr.io/k4jkul01t5rr0/henyey:ssc-diagN .`
Run loop (fresh **4h** clusters — 3h clusters expire mid-matrix and cause false hangs): `compare/run-matrix.sh` / `run-fin.sh` against `MixedImageLoadGenerationAllNewImage` (0c3h reproduces fastest).

Key question to answer from a failing run: does the create_upgrade `e2e_invoke` return **FAILED** (→ resource/fee fix on the create_upgrade tx), or **OK with `n_changes=0`** (→ contract/footprint issue), or **OK `n_changes=3` but the entry still isn't resolvable** (→ snapshot/temp-entry retention bug)?

## Parity note

stellar-core computes the create_upgrade resources/fee per-tx (`sorobanResourceFee(...) + 20'000'000`, TxGenerator.cpp:1251) rather than hardcoding; henyey hardcodes `instructions=2'500'000`, `disk_read=3'100`, `write=3'100`, `resource_fee=50'000'000`. If the fee/resource hypothesis holds, matching stellar-core's computed sizing is the fix.

## Refs

- Follows the merged fix chain in #3601 (closed #3600).
- Affects only the henyey-majority SSC mixed-image compositions; the with-core path (core drives create_upgrade) is unaffected.

tomerweller:
## Summary

Fixes #3602 — the intermittent (~50% in all-henyey runs) hang of the
`MixedImageLoadGeneration` config-upgrade step when henyey drives the load
generator.

Two changes, **reframed after review** (the host.rs change is *not* the
root-cause cure it was originally described as — see below):

1. **Gate Soroban-setup loadgen completion on contract-state sync** — the
   load-bearing fix.
2. **Withhold the module cache from `UploadContractWasm`** — a verified-inert
   defensive parity guard (not the cure), with accurate comments.

## Root cause (revised)

Under the genesis `MinimumSorobanNetworkConfig` (`ledgerMaxTxCount = 1`), the
upgrade-setup deploy competes for the single per-ledger Soroban slot and can be
surge-excluded for several ledgers. `wait_till_complete` previously gated only
on **account-sync**, which returns as soon as the source account's sequence
advances — and a *failed/never-applied* setup tx still leaves the seq where the
poll is satisfied. So the loadgen reported setup "complete" while the contract
code/instance was never materialized; the downstream `create_upgrade` then had
nothing to resolve and the SSC `/upgrades?configupgradesetkey` arm hung.

## Fix

**(1) Load-bearing — loadgen completion gate.** `wait_till_complete` now also
gates on `soroban_state_synced()` (uploaded code key + every deployed instance
key present on-ledger) for Soroban-setup modes, and marks the run **failed**
(`LoadResult::Failed`) on timeout-with-unsynced. This mirrors stellar-core
`LoadGenerator::checkSorobanStateSynced` (`LoadGenerator.cpp:1259`) +
`emitFailure`, so a setup whose instance never materialized fails fast and is
retried rather than silently proceeding to an unresolvable `create_upgrade`.
The pure decision is factored into `soroban_state_synced_inner` with a unit test.

**(2) Defensive parity guard — withhold the module cache from uploads.**
`upload_skips_module_cache()` returns true only for `UploadContractWasm`, and the
`execute_host_function_p24..p27` paths pass `None` for those.

> **Review note (corrects the original framing):** tracing the pinned
> soroban-env shows the upload path (`upload_contract_wasm`) parses with a
> throwaway engine and **never consults the module cache** — so for an upload,
> `Some(cache)` vs `None` is **byte-identical metering** (same instructions,
> output, footprint, refund, events) on every protocol (P27 routes through the
> p26 env `b351f88`). The originally-stated mechanism ("passing the cache charges
> ~1.5M native-compile against the 2.5M upload budget") does **not** occur in
> this host version; this change is therefore a **no-op on the consensus path**,
> not the cure. It is kept as a defensive guard that documents uploads have no
> need of the cache and pins upload metering parse-only (matching stellar-core,
> which also never charges uploads for cache interaction). Only
> `UploadContractWasm` is withheld — the executing paths (`InvokeContract`,
> `CreateContract`/`CreateContractV2`) keep the cache, since withholding it from
> those would itself diverge from core.

## Validation

Supercluster `MixedImageLoadGeneration`, protocol 27, fresh clusters:

| Composition | Before | After |
|---|---|---|
| All-henyey (0c3h) | ~50% pass | **8/8 pass** |
| Core-majority mixed (2c1h) | pass | **3/3 pass** (no consensus divergence) |

The 2c1h result confirms no state-hash divergence in a mixed cluster (consistent
with the host.rs change being inert). Note the 8/8 was observed with **both**
changes present; the loadgen gate is the parity-correct fix, and (per the review
note) confirming the gate alone reproduces 8/8 — vs the host.rs guard
contributing — would need a host.rs-free SSC run.

Tests: `henyey-tx` lib + `henyey-simulation` lib green; `test_upload_skips_module_cache`
and `test_soroban_state_synced_inner_gating`.

### DIFF
--- a/crates/simulation/src/loadgen.rs
+++ b/crates/simulation/src/loadgen.rs
@@ -1485,7 +1485,10 @@ impl LoadGenerator {
                     // the create-upgrade loadgen completes; if we reported "done"
                     // on submit, the `ConfigUpgradeSet` entry would not yet be in
                     // a closed ledger and the arm would be rejected (#3596).
-                    self.wait_till_complete().await;
+                    self.wait_till_complete(config).await;
+                    if self.failed {
+                        return LoadResult::Failed;
+                    }
                     // For create_upgrade, account-sync is necessary but not
                     // sufficient: the TEMPORARY ConfigUpgradeSet ContractData
                     // write can lag the source-account seq update in the
@@ -1555,19 +1558,37 @@ impl LoadGenerator {
     }
 
     /// Wait until every submitted transaction has been applied (accounts
-    /// synced) before declaring the run complete.
+    /// synced) — and, for Soroban setup modes, until the deployed contract
+    /// code + instance entries actually exist on-ledger — before declaring the
+    /// run complete.
     ///
     /// Parity: stellar-core `LoadGenerator::waitTillComplete()`
-    /// (`src/simulation/LoadGenerator.cpp:1345`) polls `checkAccountSynced`
-    /// each ledger until there are no inconsistencies (all txns applied),
-    /// timing out after `TIMEOUT_NUM_LEDGERS` (20). We poll ~once per second
-    /// and bound the wait by ledger advancement so a dropped/never-applied tx
-    /// can't hang the run forever.
-    async fn wait_till_complete(&mut self) {
+    /// (`src/simulation/LoadGenerator.cpp:1345`) gates completion on BOTH
+    /// `checkAccountSynced().empty()` AND `checkSorobanStateSynced(cfg).empty()`
+    /// each ledger, timing out after `TIMEOUT_NUM_LEDGERS`. The Soroban-state
+    /// gate is load-bearing for the create_upgrade flow (#3602): under the
+    /// genesis `MinimumSorobanNetworkConfig` (`ledgerMaxTxCount = 1`,
+    /// `ledgerMaxInstructions = 2_500_000`) the upgrade-setup `create_contract`
+    /// deploy competes for the single per-ledger Soroban slot and can be
+    /// surge-excluded for several ledgers. Account-sync alone returns as soon
+    /// as the source seq advances (a *failed* apply still bumps the seq), so
+    /// without the state gate the run could report "complete" while the
+    /// contract instance was never created — the subsequent `create_upgrade`
+    /// then writes nothing resolvable and the SSC arm hangs. We poll ~once per
+    /// second and bound the wait by ledger advancement.
+    ///
+    /// On timeout with the Soroban state still unsynced (setup modes), the run
+    /// is marked failed — parity with stellar-core `emitFailure` — so the
+    /// mission fails fast and is retried, rather than silently proceeding to a
+    /// create_upgrade that can never resolve.
+    async fn wait_till_complete(&mut self, config: &GeneratedLoadConfig) {
         const TIMEOUT_NUM_LEDGERS: u32 = 30;
+        let check_soroban = config.mode.is_soroban_setup();
         let start_ledger = self.tx_generator.app.current_ledger_seq();
         loop {
-            if self.tx_generator.check_accounts_synced() {
+            let accounts_synced = self.tx_generator.check_accounts_synced();
+            let soroban_synced = !check_soroban || self.soroban_state_synced();
+            if accounts_synced && soroban_synced {
                 return;
             }
             let elapsed = self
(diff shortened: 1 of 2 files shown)


tomerweller:
⚠️ **Not a complete fix — needs more work before merge.**

Re-validation on a fresh cluster (`iuool464d4c3c`) shows the create_upgrade **setup** flake recurs at ~50%, including with the *exact* image validated here (no other changes): **2/4 pass**. The original 8/8+3/3 was on a favorable cluster.

This PR genuinely fixes the **upload** half (parse-only metering, withholding the module cache from `UploadContractWasm` — a real metering parity bug worth keeping), but the **deploy** (`create_contract`) still flakily fails because it depends on timing-sensitive ledger-close module-cache warming for `write_bytes`. Details + remaining options in stellar-experimental/henyey#3602.

Suggest holding this PR (or merging the upload-metering part as a partial improvement while the deploy/warming fix lands separately). Deferring to reviewer.

tomerweller:
Follow-up #3607 root-causes and fixes the residual ~50% setup flake I flagged here: the create_contract deploy could apply before upload_wasm (Storage MissingValue). #3607 gates the deploy phase on the uploaded wasm being verified on-ledger (parity LoadGenerator.cpp:360); validated 10/10 on the previously-flaky cluster. This PR's upload-metering fix stands; #3607 completes the picture.

Issue #3612, opened 2026-06-25
Tx-queue: age-eviction caps queue below capacity under heavy load, surge fee-floor never engages (vs stellar-core)


tomerweller:
# Tx-queue: under heavy load age-eviction caps the queue below capacity, so the surge fee-floor never engages (vs stellar-core)

## Summary

Under sustained classic load that over-drives the network (submission rate > apply rate at the
1000-tx/ledger cap), henyey's transaction queue diverges from stellar-core in *which* eviction path
fires, with an observable downstream effect on loadgen account convergence (see #3611).

Both henyey and core are configured identically for this:
- queue pool capacity = `1000 * POOL_LEDGER_MULTIPLIER` = 2000 ops (henyey) ==
  `maxLedgerResources * TRANSACTION_QUEUE_SIZE_MULTIPLIER` = 2000 (core);
- `pending_depth` / `TRANSACTION_QUEUE_TIMEOUT_LEDGERS` = 4;
- both have the eviction fee-floor: a per-lane "min inclusion fee to beat evicted" set when the queue
  does a **capacity** eviction, after which lower-fee txns are rejected at admission
  (henyey `fee_below_lane_threshold` → `FeeTooLow`; core `TxQueueLimiter::canAddTx` →
  `mLaneEvictedInclusionFee`).

## The divergence (measured, mixed-image 1c2h heavy run)

- **core**: the queue saturates to its ~2000-op capacity → capacity-eviction fires → the lane fee
  floor is set → subsequent same-fee loadgen txns are **rejected at admission** (~8,259 loadgen
  retries → rollbacks). The backlog stays bounded and the loadgen's accounts stay synced.
- **henyey**: age-eviction (4-ledger) trims the queue every ledger and caps the depth at ~1,830 —
  just **below** the 2,000 capacity limit — so capacity-eviction never fires, the fee floor is never
  set, and low-fee txns keep being **admitted** (~812 loadgen retries). Those admitted txns then age
  out (~5,891, matching core's 5,839), leaving their accounts seq-ahead.

Net: both apply ~37.8k of 50k and age-evict ~5.9k, but core ends with ~0 unsynced accounts (its
backlog is rejected at admission) while henyey ends with ~4,840 unsynced (delta=1 each) — the
root of #3611's run-1 `wait_till_complete` timeout.

The configs are identical; the difference is **emergent eviction-timing**: when the queue is near
[...]

Issue #3611, opened 2026-06-24
Loadgen run-1 times out wait_till_complete from un-rolled-back rejected txns (residual after #3609)


tomerweller:
# Loadgen run-1 (classic payments) times out `wait_till_complete` from a few un-rolled-back rejected txns

## Context

Follow-up to #3609 (cross-run account-cache poisoning, fixed in #3610). After
that fix, the henyey-majority mixed-image mission still runs ~2m11s slower than
core-majority (11m2s vs 8m51s on identical config). The entire residual is run 1
(classic payments) timing out `wait_till_complete` once (~150s).

## Mechanism

In a fixed-image run, run 1 ends via the 30-ledger `wait_till_complete` timeout
with `accounts_synced=false`, despite **0 herder drops** and the mission passing
(`exit 0`, consistency OK). The loadgen metrics show `loadgen.txn.rejected ≈ 92`
out of 50000. Those rejected txns had their expected (in-memory) sequence number
advanced at submission but never applied, so ~92 accounts are left seq-ahead of
the ledger permanently → `check_accounts_synced()` stays false → run 1 burns the
full timeout.

core-majority does not hit this: core's loadgen run 1 drains and completes
without the timeout.

## Possible fixes (pick after confirming parity with stellar-core)

1. Roll back the expected sequence number when a submission is **rejected**
   (not added to the queue), so the account stays synced — analogous to the
   existing QueueFull rollback path.
2. Make `wait_till_complete` / `check_accounts_synced` tolerant of a small number
   of permanently-stuck accounts (e.g. complete once the unsynced set stops
   shrinking), matching stellar-core's behavior.

Confirm what the rejected txns actually are (capture the reject reason — current
logs don't surface it at default level) before choosing.

## Acceptance

henyey-majority `MixedImageLoadGenerationWithNewImageMajority` duration ≈
core-majority `...WithOldImageMajority` on the same config (no run-1 timeout).

tomerweller:
## Correction: original "rejected-seq" hypothesis is wrong

Deeper analysis of the fixed-image 1c2h run (henyey loadgen node log + metrics) refutes the
opening hypothesis (a few un-rolled-back rejected txns):

- `loadgen.txn.rejected = 92` out of ~50000, and **0** "Transaction queue full, dropped" log lines.
  92 rejections cannot account for a full 30-ledger (~150s) timeout.
[...]

PR #3614, opened 2026-06-25
Track live maxTxSetSize in classic tx-queue capacity

tomerweller:
Refs #3612

## Summary

The classic transaction queue's ops capacity (`max_queue_ops`) and tx-count cap (`max_size`) were frozen at app construction (`1000 * POOL_LEDGER_MULTIPLIER`, `crates/app/src/app/mod.rs`) and never re-derived when `UpgradeMaxTxSetSize` changed the live ledger header. Stellar-core's `TxQueueLimiter::reset(ledgerVersion)` rebuilds the classic generic-lane capacity from `maxScaledLedgerResources(false) = maxTxSetSize * mPoolLedgerMultiplier` (`TxQueueLimiter.cpp:61,244`) on **every** ledger close, so core's classic queue capacity tracks `maxTxSetSize`; henyey's did not — a latent deterministic parity gap (issue option 3).

This PR re-derives the classic ops capacity from the live header's `max_tx_set_size` on each ledger close (and at bootstrap/restore), scaled by `POOL_LEDGER_MULTIPLIER`, preferring it over the construction-time config:

- `TransactionQueue::update_classic_queue_capacity(scaled_max_queue_ops)` stores the live capacity, drops the persistent global-ops eviction queue so it rebuilds with the new limit, and clears the stale global fee floor.
- `effective_max_queue_ops()` / `effective_max_size()` prefer the dynamic value over `config.max_queue_ops` / `config.max_size`; the admission gates (`check_and_collect_evictions`, `ensure_queue_capacity`, fee-floor checks) now read the effective values.
- Wired into the per-ledger-close path (`ledger_close.rs`, inside the `spawn_blocking` closure where `result_header.max_tx_set_size` is available) and the bootstrap seed (`seed_validation_context`), mirroring the existing dynamic-Soroban-limits pattern.

**Parity scope:** only the classic ops capacity tracks `maxTxSetSize`, matching core. The DEX lane (`mMaxDexOperations`) and byte allowances are independent fixed config in core and are **not** scaled by `maxTxSetSize`, so they are left unchanged (avoiding the divergence that scaling them would introduce).

This closes the **deterministic capacity-tracking** parity gap only. The emergent age-vs-capacity eviction-timing root cause behind #3611's loadgen run-1 divergence (issue options 1/2) requires the mixed-image runtime comparison and stays **operator-gated** per the converged plan — hence `Refs #3612`, not `Fixes`.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3612#issuecomment-4800989945)

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy --all -- -D warnings` (the CI gate) passes
- [x] `cargo test -p henyey-herder --lib tx_queue` — 308 passed
- [x] `cargo build -p henyey-herder -p henyey-app --all-targets` clean

## Regression test

- **Test:** `crates/herder/src/tx_queue/mod.rs::test_classic_queue_capacity_tracks_live_max_tx_set_size`
- **Pre-fix:** committed as `019f2398` — verified FAILED (capacity frozen; `update_classic_queue_capacity` / `effective_max_queue_ops` / `effective_max_size` did not exist).
- **Post-fix:** verified PASSES after `ddf921d9` — effective capacity grows 50→200 (×2 → 400 ops) on upgrade and shrinks 200→50 on downgrade.

## Deviations from plan

The converged plan named `TxQueueLimiter::reset` (`crates/herder/src/tx_queue_limiter.rs:227`) as the fix site. During implementation I found `TxQueueLimiter` is **only constructed in its own unit tests** — it is not in the production admission path. The production classic-capacity gate is `TransactionQueue` / `QueueStore` in `crates/herder/src/tx_queue/mod.rs`, driven by `TxQueueConfig.max_queue_ops` / `max_size` (frozen at `mod.rs` construction, exactly as the issue's option 3 describes). Fixing only the dead `TxQueueLimiter::reset` would have been a no-op. The fix is therefore implemented in the real production component; the design (re-derive classic capacity from the live `maxTxSetSize` per ledger close) is unchanged from the plan's intent.

The plan/issue also mentioned scaling "DEX-lane / classic-byte-allowance × POOL_LEDGER_MULTIPLIER" with `maxTxSetSize`. Verified against core: only the generic-lane **ops** capacity comes from `maxScaledLedgerResources`; DEX/byte allowances are independent. Scaling them would diverge from core, so they are intentionally left unchanged.

### DIFF
--- a/crates/app/src/app/ledger_close.rs
+++ b/crates/app/src/app/ledger_close.rs
@@ -1086,6 +1086,15 @@ impl App {
             self.herder.ledger_close_duration(),
         );
 
+        // Seed the classic queue ops capacity from the restored header's
+        // `maxTxSetSize` so admission tracks the live capacity from the very
+        // first ledger, before the first post-restore close runs (#3612).
+        self.herder.tx_queue().update_classic_queue_capacity(
+            header
+                .max_tx_set_size
+                .saturating_mul(POOL_LEDGER_MULTIPLIER),
+        );
+
         // Seed Soroban per-tx limits and dynamic resource limits from network config.
         if let Some(soroban_info) = self.soroban_network_info() {
             self.herder.tx_queue().set_soroban_limits(SorobanTxLimits {
@@ -2769,6 +2778,17 @@ impl App {
             if let Some(s) = selection_limit {
                 herder.tx_queue().update_soroban_selection_limits(s);
             }
+
+            // Re-derive the classic queue ops capacity from the live ledger
+            // header's `maxTxSetSize` so the limiter tracks `UpgradeMaxTxSetSize`.
+            // Parity: stellar-core `TxQueueLimiter::reset` rebuilds capacity from
+            // `maxScaledLedgerResources(false) = maxTxSetSize * mPoolLedgerMultiplier`
+            // every ledger close (#3612). Henyey froze this at app construction.
+            herder.tx_queue().update_classic_queue_capacity(
+                result_header
+                    .max_tx_set_size
+                    .saturating_mul(POOL_LEDGER_MULTIPLIER),
+            );
 
             let shift_result = herder.tx_queue().shift();
             crate::metrics::CLOSE_TX_QUEUE_SHIFT_UPDATE_SECONDS
--- a/crates/herder/src/tx_queue/mod.rs
+++ b/crates/herder/src/tx_queue/mod.rs
@@ -1256,6 +1256,13 @@ impl EvictionThresholds {
     fn reset_soroban(&self) {
         self.soroban_lane_fees.write().clear();
     }
+
+    /// Reset only the global ops eviction threshold. Used when the classic ops
+    /// capacity changes (#3612) so a stale fee floor computed against the old
+    /// capacity does not reject admissions under the new one.
+    fn reset_global_ops(&self) {
+        *self.global_fees.write() = None;
+    }
 }
 
 /// ## Lock Ordering (MUST be followed by all methods)
@@ -1312,6 +1319,18 @@ pub struct TransactionQueue {
     /// Separate from queue-admission limits which apply
     /// `SOROBAN_TRANSACTION_QUEUE_SIZE_MULTIPLIER` (2x).
     dynamic_selection_soroban_resources: RwLock<Option<Resource>>,
+    /// Dynamic classic queue ops capacity (already scaled by
+    /// POOL_LEDGER_MULTIPLIER), re-derived from the live ledger header's
+    /// `maxTxSetSize` on every ledger close. Takes precedence over the
+    /// construction-time `config.max_queue_ops` / `config.max_size`.
+    ///
+    /// Parity: stellar-core `TxQueueLimiter::reset(ledgerVersion)` rebuilds the
+    /// classic generic-lane capacity from `maxScaledLedgerResources(false)` =
+    /// `maxLedgerResources(false) * mPoolLedgerMultiplier`
+    /// (`TxQueueLimiter.cpp:61,244`), reading the live `maxTxSetSize` each close.
+    /// Henyey froze the capacity at app construction, so `UpgradeMaxTxSetSize`
+    /// was never tracked. See #3612.
+    dynamic_max_queue_ops: RwLock<Option<u32>>,
     /// Arbitrage flood damper. Acquired after `store` lock in `broadcast_with_visitor`
     /// and `shift()`. Cleared by `shift()`, preserved by `reset_and_rebuild()`.
     arb_damper: parking_lot::Mutex<arb_flood_damping::ArbitrageFloodDamper>,
@@ -1384,6 +1403,7 @@ impl TransactionQueue {
             skip_fee_balance_check: std::sync::atomic::AtomicBool::new(false),
             dynamic_queue_soroban_resources: RwLock::new(None),
             dynamic_selection_soroban_resources: RwLock::new(None),
+            dynamic_max_queue_ops: RwLock::new(None),
             arb_damper: parking_lot::Mutex::new(arb_damper),
             arb_tx_seen: std::sync::atomic::AtomicU64::new(0),
             arb_tx_dropped: std::sync::atomic::AtomicU64::new(0),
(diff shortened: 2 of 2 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Capacity threading is sound, always-initialized, falls back correctly to config defaults, and is exercised by a passing regression test. No race surface introduced. One non-blocking test-depth note left inline.

<details>

Cycle 1 (no prior Correctness comment). Complete change-list below.

**Initialization / no zero-capacity window (`correctness`):** `dynamic_max_queue_ops` starts `None`; `effective_max_queue_ops()` falls back to `config.max_queue_ops` and `effective_max_size()` to `config.max_size`. App construction sets both to `1000 * POOL_LEDGER_MULTIPLIER` (`crates/app/src/app/mod.rs:1617,1621`), so there is no None/zero-capacity admission window before the first close. `seed_validation_context()` additionally seeds the dynamic value from the restored header at bootstrap (`ledger_close.rs:~1090`), and every per-close path updates it from `result_header.max_tx_set_size` before `shift()`. Verified the bootstrap call sits next to the existing `update_validation_context`/`set_soroban_limits` seeding.

**Fallback semantics (`correctness`):** `effective_max_queue_ops` = `dynamic.or(config.max_queue_ops)` — correct (dynamic wins, else config, else None → gate disabled). `effective_max_size` returns the scaled ops as the count cap when dynamic is set, else `config.max_size`. At construction both equal `1000*MULT`, so they track consistently; after an upgrade both grow to `maxTxSetSize*MULT`. Reasonable and internally consistent.

**Gate rewiring (`correctness`):** All 4 production admission/eviction gates that previously read `self.config.max_queue_ops`/`self.config.max_size` (PR head lines 2038, 2086, 2192, 2526) now read the `effective_*` accessors. Confirmed via grep on the PR head that no production gate still reads the frozen `config.*` for the ops/size capacity. The byte gate (`max_queue_classic_bytes`) is deliberately left on config — consistent with the parity reviewer's finding that bytes are not part of core's classic tx limiter.

**Stale-state handling (`correctness`):** `update_classic_queue_capacity` drops `store.global_ops_queue` (forcing a rebuild with the new limit via `ensure_global_ops_queue`) and clears the stale global-ops fee floor via `reset_global_ops()`. This prevents a fee floor computed against the old (smaller) capacity from rejecting admissions under a larger new capacity. `reset_global_ops` only clears `global_fees` (the ops floor), leaving the byte/dex floors untouched — correct scoping.

**Race (`correctness`):** The new field is an `RwLock<Option<u32>>` accessed through the same `tx_queue()` handle used by the surrounding close-path updates; no new cross-lock ordering. `update_classic_queue_capacity` takes `dynamic_max_queue_ops.write()`, `store.write()`, and the eviction-threshold write independently — same pattern as existing setters. No deadlock/ordering hazard observed.

**Test (`test-coverage`):** `test_classic_queue_capacity_tracks_live_max_tx_set_size` passes on PR head; the regression-test commit (`019f2398`) precedes the fix commit (`ddf921d9`) per test-first discipline. "Fails on current main" here is a compile failure (the API doesn't exist on main) rather than a runtime assertion against pre-existing code — acceptable because the gap being fixed is the *absence* of a capacity-update mechanism, so there is no prior code path to assert against. I ran the new test (pass) and the full 148-test tx_queue module (all pass) under a fresh build — no regression. Non-blocking inline note: the test asserts the getters + downgrade, but does not drive an end-to-end admission (accept-more-txs-after-upgrade) through the rewired gates. Filing as a follow-up rather than blocking, since the getters feed the gates directly and the gate rewiring is mechanically verified.

kind: this is a latent parity-gap fix (Refs #3612, partial); the deterministic capacity-tracking gap is closed here, runtime root-cause stays operator-gated as documented.
</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** Scaling ONLY the classic generic-lane ops capacity from the live header (leaving DEX and byte allowances fixed) matches stellar-core exactly. The per-close timing and live-header derivation are deterministic and identical across nodes.

<details>

Cycle 1 (no prior Parity comment). Verified against pinned stellar-core v26.0.1.

**ops-vs-DEX/byte scaling matches core (`parity-gap` — RESOLVED, no divergence):**
- `LedgerManagerImpl::maxLedgerResources(false)` returns `Resource(getLastMaxTxSetSizeOps())` — a **single-dimension OPS** resource read from the live LCL header (`LedgerManagerImpl.cpp:687-700`, `:671-684`). No byte dimension for classic.
- `TxQueueLimiter::maxScaledLedgerResources(false)` = `maxLedgerResources(false) * mPoolLedgerMultiplier` (`TxQueueLimiter.cpp:60-65`) — ops-only, scaled. This is exactly `max_tx_set_size * POOL_LEDGER_MULTIPLIER`, which is what the PR computes and passes to `update_classic_queue_capacity`.
- The DEX lane is set ONCE in the `TxQueueLimiter` ctor as `mMaxDexOperations = *maxDexOps * multiplier` (`TxQueueLimiter.cpp:~40-44`) and is NOT re-derived from the header in `reset()`/`resetBestFeeTxs()` (`:262`, `:114` pass it through unchanged). So the DEX lane being independently fixed in henyey matches core — leaving it on `config.max_dex_ops` is correct.
- Bytes: `DexLimitingLaneConfig` ctor's `releaseAssert(...getLaneLimits()[0].size() == NUM_CLASSIC_TX_RESOURCES)` ("Ensure byte limits aren't counted in tx limiter", `TxQueueLimiter.cpp:265-266`, `:115-116`) confirms the classic **tx limiter** has no byte dimension. The classic byte allowance is a separate flood/admission concern; the PR correctly leaves `max_queue_classic_bytes` untouched. No divergence.

**live-header derivation is deterministic (`determinism` — OK):** Core does NOT cache classic capacity — `canAddTx` (`TxQueueLimiter.cpp:183`) and `evictTransactions` (`:202`) call `maxScaledLedgerResources` live on every admission, reading the LCL header each time. Henyey caches it in `dynamic_max_queue_ops`, refreshed at every close from `result_header` (the just-closed ledger N). Since admission for ledger N+1 happens only between closes, henyey's cached value during that window equals what core computes live from the same LCL header. Both nodes derive the value from the identical agreed ledger header → identical capacity on every node for the same ledger. Admission itself is node-local mempool policy and does not feed nomination/validation determinism; this change cannot diverge tx-set nomination/validation.

**per-close timing matches core's reset point (`sequencing` — OK):** Core rebuilds the classic flood lane config in `shift()` via `resetBestFeeTxs(LCL.ledgerVersion)` (`TransactionQueue.cpp:976`) and the admission capacity is read live (always current). The PR updates `dynamic_max_queue_ops` from `result_header` immediately BEFORE `herder.tx_queue().shift()` in the close closure — so the capacity for ledger N+1 admission uses the header that closed ledger N, exactly as core. The bootstrap seed (`seed_validation_context`) covers the restore-before-first-close window, which core gets for free from the always-live read. Good.

**fee-floor / age-eviction untouched (`parity-gap` — correctly out of scope):** The plan's option-1/option-2 are correctly NOT implemented; the diff only touches the capacity-derivation path. `reset_global_ops()` clears only the stale ops fee floor on a capacity change (so an old-capacity floor can't reject under new capacity) — this mirrors that core recomputes the floor against the live `maxLimits` and never carries a floor across a capacity change. No new fee-floor semantics introduced.

Partial (Refs, not Fixes) is the right call: the deterministic capacity-tracking gap is closed and parity-correct; the emergent age-vs-capacity eviction-timing root cause remains operator-gated on the runtime comparison, as documented in the plan.
</details>

tomerweller:
## ✅ Merged

**Commit:** 86952ac9cb7cf8e8006ba9bbbceffe2c4409b736

Triple-green: Reviewer Correctness APPROVE, Reviewer Parity APPROVE (ops-only generic-lane scaling confirmed to match stellar-core; DEX/byte lanes correctly left fixed), all 6 required checks green (Test, Doctest, Clippy, Format, Build release-ci, Analyze rust).

**Follow-up issues filed for non-blocking concerns:** #3615 (end-to-end admission test coverage)

This PR is partial (Refs #3612). The emergent age-vs-capacity eviction-timing root cause remains operator-gated on a runtime comparison; #3612 stays open/blocked.

Issue #3625, opened 2026-06-25
Implement credit-based inbound flow control for SCP/overlay (full stellar-core parity)


tomerweller:
Follow-up from #3623 (OOM restart-loop fix). The #3623 fix bounds the overlay→event-loop SCP ingest channel with a fixed-capacity drop-on-full mpsc — a simple approximation of stellar-core's bounded inbound capacity.

stellar-core (`src/overlay/Peer.cpp` `recvMessage`/`recvSCPMessage`, `FlowControlCapacity.cpp` `FlowControlMessageCapacity::canRead` / `FlowControlByteCapacity::canRead`) uses **credit-based inbound flow control**: a sending peer cannot exceed the receiver's negotiated inbound message/byte capacity, and capacity is released as messages are consumed. This back-pressures senders (no send-more granted) rather than dropping.

This issue tracks implementing the full credit-based inbound flow-control model in henyey's overlay for closer parity (both the message-count and byte capacities), superseding the fixed-capacity drop-on-full approximation. Lower priority — the #3623 fix already removes the fatal unbounded-growth/OOM property.

Scope/parity reference: `stellar-core/src/overlay/FlowControl.cpp`, `FlowControlCapacity.cpp`.

tomerweller:
## 📝 Plan Draft (Round 1)

**Summary:** Wire henyey's already-ported inbound flow-control machinery (`maybe_throttle_read`/`can_read`/`stop_throttling`) into the live peer read loop and fix SEND_MORE-grant timing so credit is released only after the SCP consumer actually drains the message — replacing the #3626 drop-on-full backstop with genuine credit-based back-pressure. Phase 1 of a staged feature.

### Key finding (changes the issue's premise)

[...]

Issue #3582, opened 2026-06-23
phase=28 event-loop stall holds DB write lock >30s (root cause of #3497 recoverable-shutdowns)


tomerweller:
## Summary

Follow-up to #3497 (option A merged, `5ae6188b`): a transient `SQLITE_BUSY`/`DatabaseBusy` at the ledger-close persist write now triggers a clean **recoverable-shutdown** instead of a fatal abort — but the validator **still goes down (~9-13 min restart + catchup)** each time, because the underlying cause is unaddressed: a **~20 s event-loop stall at `phase=28`** holds the SQLite write lock past the (already-30 s) `busy_timeout`, so the ledger-close write on the other connection times out and busies.

This is the **root cause** that, if fixed, would eliminate the downtime entirely — option A only converts the FATAL abort into a clean shutdown.

## Evidence (from #3497, 4 aborts this session)

Each abort is immediately preceded by:
```
WATCHDOG: Event loop slow (>15s since last tick) stale_secs=20 phase=28 phase_sub=0 fetch_channel_depth=...
Failed to update peer record err=Sqlite(... DatabaseBusy ... "database is locked")
Fatal persist failure ... "ledger close DB write" error=SQLite error: database is locked
```
`phase=28` = the 1-second consensus-tick arm (`crates/app/src/app/lifecycle.rs:1106`, labelled `peer_maintenance`): `process_externalized_slots` → `try_start_ledger_close` → gap-detection → `request_pending_tx_sets` → `maybe_publish_history` → `try_trigger_consensus` (+ a concurrent peer-record DB update). One of these blocks the event loop ~20 s while holding/contending the DB write lock.

## Fix direction (same class as #3537/#3532, which fixed the phase=33 tx_set_gc stall)

1. **phase_sub-instrument** the phase=28 tick — time each sub-op (process_externalized / try_start_ledger_close / maybe_publish_history / request_pending_tx_sets / try_trigger_consensus / peer-record-update) and emit `phase_sub` so the watchdog log pinpoints the ~20 s sub-op (`phase_sub=0` today is uninformative). This instrumentation PR is parity-neutral/observability-only and autonomously implementable; it needs a **deploy** to capture the live data.
[...]

PR #3626, opened 2026-06-25
Bound the SCP ingest channel to prevent OOM under a stalled event loop

### LINKED ISSUE #3623: URGENT: validator OOM restart-loop — stalled event loop buffers SCP unbounded during post-catchup transition (escalation of #3582; NOT a b0231cc4 regression)
## Summary

Deploying **b0231cc4** (`Reset loadgen account-sequence cache between runs (#3610)`) to the mainnet validator (session 74535976) caused an immediate **post-restart consensus stall**. The node restored state from disk, transitioned Herder to `Tracking` at the restored lcl, and then **never externalized another ledger** — it sat wedged at the restored slot reporting `state=Synced` while mainnet advanced ~60+ slots beyond it.

Rolled back to last-good **70f280c2** (which had run 46h clean). b0231cc4 added to `deploy_quarantine.txt`.

## Observed behavior (b0231cc4)

- Launched 19:48:20Z, restored `lcl_seq=63195949`, logged `Restored state from disk, skipping full catchup`.
- `Transitioning to Tracking on advance_tracking_to slot=63195949` at 19:48:24Z.
- Overlay came up healthy: ~15+ peers authenticated within the first 2 minutes.
- **No further progress**: ledger frozen at `63195949` for >7 minutes (age climbed 110s → 305s+), zero `externalize` / `ledger close` / `catchup` log lines after bootstrap. Only overlay peer-auth churn.
- `/info` reported `state=Synced ledger=63195949` with an empty `quorum` object the entire time.
- RPC `getHealth` → `unhealthy` (latestLedger not advancing).
- Node never detected it was behind the live tip → never triggered catchup → wedged.

## Prime suspect

b0231cc4's range over the deployed 70f280c2 includes **#3599** — *"Match stellar-core purgeSlotsOutsideRange: add max-slot upper bound to SCP::purge_slots"* (`crates/scp/src/scp.rs`, `crates/scp/src/slot.rs`). A miscomputed max-slot upper bound could purge the very slots the node needs to process forward from the restored lcl, which matches the symptom precisely: bootstrapped at slot N, Tracking, but unable to advance to N+1.

Other runtime-affecting commits in the range (lower suspicion): #3603 (drive restored externalized lcl+1 to ledger close on restart — **also restart-path SCP, worth ruling in**), #3587 (/info maxSorobanTxSetSize), #3593 (configupgradesetkey arm-time validation), #3614 (tx-queue capacity).

**#3603 ("Drive restored externalized lcl+1 slot to ledger close on restart")** is arguably an even closer match than #3599 — it directly changes the restart-from-restored-lcl ledger-close path, which is exactly where the wedge occurred. Both should be investigated.

## Repro / next steps

- Targeted offline reproduction: start a node from a restored HAS at a recent lcl on b0231cc4 and confirm it fails to externalize lcl+1, vs. 70f280c2 which advances. This isolates the restart-path regression without touching the live validator.
- VE-green on b0231cc4 (2026-06-25 07:13) did **not** catch this — VE replays a fixed ledger range and does not exercise the live restart-from-restored-state + SCP-catchup path. Consider a restart-path regression test.

## Deployed binary status

- Mainnet validator rolled back to **70f280c2** (last-good, 46h prior clean uptime).
- b0231cc4 quarantined; deploy gate will not re-select it until its diff clears origin/main or it is resolved.

tomerweller:
Closes #3623

This is the **OOM-mechanism fix** for the fatal restart-loop in #3623. The #3582 stall trigger (SQLite write-lock contention during post-catchup maintenance) and full credit-based inbound flow control (#3625) remain separate, out of scope here.

## Summary

The overlay→event-loop SCP ingest channel (`scp_message_tx`/`scp_message_rx`) was an unbounded `mpsc::unbounded_channel`. When the main event loop stalls (#3582), ~24 tier-1 validators flood ~100+ SCP envelopes/slot with nothing draining the channel, so RSS grows ~4 GB/min until the validator is OOM-killed — a fatal, self-perpetuating restart loop. This was the **sole unbounded element** in the SCP ingest path (`PendingEnvelopes.max_slots`, `MAX_DRAIN_PER_TICK`, `max_externalized_slots`, `drain_verified_bounded` are all already bounded).

The fix converts that one channel to a bounded `mpsc::channel(SCP_CHANNEL_CAPACITY)` (8192). At the overlay send site it uses non-blocking `try_send()`; on a full channel it **drops** the envelope and increments the existing `messages_dropped` counter (never `.await` on the peer path — blocking the send path is its own event-loop hazard). Dropping over-capacity SCP is recoverable: peers re-flood every slot and the loop's gap-detection + `SyncRecoveryManager` backfill via `GetScpState`.

The dummy fallback SCP receiver in `run()` (taken only when there is no overlay) retains its paired sender (`_dummy_scp_tx`) for the loop lifetime so the bounded channel never closes and the `Some(scp_msg) = scp_message_rx.recv()` select arm cannot observe a spurious `None` (no busy-loop).

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3623#issuecomment-4804077047)

## Files changed

- `crates/overlay/src/manager/mod.rs` — add `pub const SCP_CHANNEL_CAPACITY = 8192` with sizing rationale; SCP channel `unbounded_channel()` → `channel(SCP_CHANNEL_CAPACITY)`; field/return types `UnboundedSender`/`UnboundedReceiver` → `Sender`/`Receiver`; send site → `try_send` drop-on-full + `messages_dropped`; updated 4 test-helper channel constructions; **regression test**.
- `crates/app/src/app/lifecycle.rs` — bounded dummy fallback SCP channel retaining `_dummy_scp_tx`; **companion regression test**.
- `crates/overlay/src/manager/connection.rs` — bounded the auxiliary test SCP channel construction.
- `crates/overlay/src/lib.rs` — re-export `SCP_CHANNEL_CAPACITY`.
- `crates/overlay/PARITY_STATUS.md` — document the bounded-channel approximation of core's `FlowControlMessageCapacity` (ref #3625).

(The threaded type-change call sites: SharedPeerState field, OverlayManager field+rx, `new()` construction, `subscribe_scp()` return, the send site, the lifecycle dummy, plus the test-helper constructions in `mod.rs`/`connection.rs`.)

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test -p henyey-overlay` — 459 lib + integration pass (incl. `test_broadcast_backpressure_*`, `test_try_send_to_backpressure_*`, `overlay_scp_integration`)
- [x] `cargo test -p henyey-app` SCP/recovery focus — `pump_tests`, `scp_dedup_pipeline_tests`, out-of-sync-recovery (75 tests), `test_scp_intake_channel_is_bounded` pass

- **Test:** `henyey_overlay::manager::tests::test_scp_channel_bounded_drops_when_receiver_stalled` — floods `SCP_CHANNEL_CAPACITY + 256` envelopes through the `route_to_subscribers` send path with the receiver never drained, asserts `rx.len() <= SCP_CHANNEL_CAPACITY` and `messages_dropped >= 256`.
- **Pre-fix:** verified FAILS against the unbounded channel — `bounded SCP channel must hold at most 8192 items, held 8448` with `messages_dropped == 0` (the exact unbounded-growth / OOM mechanism).
- **Post-fix:** verified PASSES (channel capped at 8192; overflow counted).
- **Companion:** `henyey_app::app::lifecycle::scp_intake_bound_tests::test_scp_intake_channel_is_bounded` — asserts the receiver wired via `subscribe_scp()` is the bounded type with `max_capacity == 8192`.

## Deviations from plan

- The plan cited `connection.rs:~2584` as the auxiliary construction site; it was indeed there, plus 4 additional test-helper SCP channel constructions in `mod.rs` that also needed threading. All updated.
- The drop counter reuses the existing `messages_dropped` Counter (the plan's stated preferred option) rather than adding a dedicated `scp_messages_dropped`.


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Bounding the sole unbounded SCP ingest channel is correct, non-blocking on the send path, and cannot busy-loop on close. The regression test genuinely reproduces the OOM mechanism (caps `rx.len() <= 8192`, asserts `messages_dropped >= 256`) and the companion test pins the bounded type.

<details>

Cycle 1 — complete change-list. kind: bug-fix; regression test present and verified.

**no-None-busy-loop (CRITICAL) — PASS.** The two `try_recv()` drain loops (lifecycle.rs:644, 1045) use `Err(_) => break`, which covers both `TryRecvError::Empty` (live senders, channel empty) and `TryRecvError::Disconnected` — neither spins. The `select!` arm (lifecycle.rs:848) is `Some(scp_msg) = scp_message_rx.recv() =>`; a refutable pattern in `tokio::select!` that fails to match (i.e. `recv()` yields `None`) **disables** that branch for the iteration rather than looping hot. Belt-and-suspenders: in production the sender (`scp_message_tx`) is owned by `OverlayManager`/`SharedPeerState` (held for the whole `run()` via `self.overlay()`), and `subscribe_scp()` only takes the *receiver* — so the channel never closes. In the dummy fallback path, `_dummy_scp_tx: Some(tx)` is retained for `run()`'s lifetime. No path yields `None` + spin.

**Non-blocking send (CRITICAL) — PASS.** The only method-call on `scp_message_tx` is `try_send()` at mod.rs:735 (line 1072 is a `.clone()`, not a send). Explicit `Ok` / `TrySendError::Full` (drop + `messages_dropped.add(1)` + debug log) / `TrySendError::Closed` (error log) arms. No `.send().await`, no `blocking_send()` anywhere in the tree.

**Regression test — PASS.** `test_scp_channel_bounded_drops_when_receiver_stalled` floods `SCP_CHANNEL_CAPACITY + 256` envelopes through the real `route_to_subscribers` path. The test msg is `ScpMessage(Externalize)` → `is_scp == true` → exercises the `try_send` branch. Asserts (a) `rx.len() <= 8192` and (b) `messages_dropped >= 256`. Pre-fix (unbounded) it structurally must fail: an unbounded receiver has no `max_capacity` and enqueues all 8448 with zero drops — matching the PR body's "held 8448, drops==0". `Counter::add` is atomic (`fetch_add`), correct under the concurrent peer-receive callers. Companion `test_scp_intake_channel_is_bounded` asserts `rx.max_capacity() == 8192`, which both pins the bounded type and the value (unbounded receivers lack `max_capacity`).

**Type threading — PASS.** All ~10 call sites updated consistently (field types, `new()`, `subscribe_scp()` return, send site, lifecycle dummy, 4 test-helper constructions in mod.rs + 1 in connection.rs). The two remaining `unbounded_channel` SCP-named channels (lifecycle.rs:404 fetching-relay bridge, mod.rs:1397 scp_timer) are NOT the peer-flood ingest path — relay is gated by `PendingEnvelopes.max_slots` upstream, timer is bounded by armed-timer count. Correctly out of scope.

**Error handling / readability — PASS.** Closed-channel case logs error (cannot happen in production given sender liveness, but defensive). Doc comments on `SCP_CHANNEL_CAPACITY`, the field, and `subscribe_scp()` are accurate and updated (no stale "never dropped" claims left). `messages_dropped` reuse (vs a dedicated `scp_messages_dropped`) is acceptable — it is the existing drop counter and the plan's stated preference; minor observability blur (SCP drops not distinguishable from other drops) is non-blocking.

No blocking concerns.
</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** Bounding the SCP ingest channel with drop-on-full is parity-positive: stellar-core bounds inbound SCP via `FlowControlMessageCapacity` and has no unbounded overlay→herder queue. Dropped over-capacity SCP is recoverable via peer re-flood + `GetScpState` backfill, and at capacity 8192 steady-state never drops. The prior unbounded channel had no core analog.

<details>

Cycle 1 — complete change-list, concern classes: `parity-gap`, `recoverability`, `sizing`.

**Drop-recoverability / parity (PASS).** stellar-core processes SCP synchronously on the main thread under bounded inbound flow control (`Peer::recvMessage` → `recvSCPMessage`, gated by `FlowControlMessageCapacity::canRead`): a peer cannot exceed the receiver's negotiated inbound capacity, and capacity is released only as messages are consumed — so a stalled receiver applies backpressure and peers stop being granted send-more. There is **no unbounded async queue** between overlay receive and herder in core. Henyey's prior `unbounded_channel` was therefore a divergence with no upstream analog. The bounded channel + drop-on-full is a count-based approximation of core's credit-based bound — it achieves the essential invariant (bounded inbound memory) and moves toward parity. SCP is inherently lossy under backpressure in core too, so drop-on-full is not a novel semantic. Recovery is real: SCP is re-flooded by ~24 validators every ~5s slot, and the loop's gap-detection + `SyncRecoveryManager` backfills missing slot state via `GetScpState` (the dedicated fetch-request channel, which remains reliable). No SCP gap/sequence invariant downstream assumes total delivery — `PendingEnvelopes` is slot-keyed and tolerates missing envelopes by design (it fetches/waits).

**Reliable→lossy semantic change (PASS).** All downstream stages are independently bounded: `PendingEnvelopes.max_slots`, `max_externalized_slots`, `MAX_DRAIN_PER_TICK` (200), `drain_verified_bounded`. The SCP channel was verifiably the **sole** unbounded element on the ingest path (confirmed by grep). Nothing downstream breaks on a dropped envelope — it is equivalent to a momentary network loss, which the consensus path already handles.

**Sizing 8192 (PASS).** At ~100 envelopes/slot over ~5s slots, 8192 buffers ~80 slots' worth of a full multi-validator flood. In steady state the event loop drains far faster than arrival, so the channel sits near-empty and drops effectively never occur — the change is invisible in normal operation and engages only during the pathological multi-second stall (#3582). Worst-case retained memory ~few MB (envelopes <1KB) vs the ~37.5GB unbounded growth observed in #3623.

**PARITY_STATUS.md (PASS).** New entry #7 accurately documents the approximation, cites `FlowControlMessageCapacity`, and references #3625 as the tracked follow-up for a faithful credit-based byte bound. No parity-column regression.

Non-blocking note (inline-worthy, not gating): the count-based bound is an approximation, not exact core parity — the byte-accurate credit-based port is correctly deferred to #3625. No action needed here.
</details>

tomerweller:
## Review: Verdict — APPROVE / APPROVE (awaiting CI)

Both reviewer lenses APPROVE (no CHANGES_REQUESTED → no refute pass required).

- **Correctness:** APPROVE — no-None-busy-loop and non-blocking-send both verified against code (see comment).
- **Parity:** APPROVE — bounded drop-on-full is parity-positive vs core `FlowControlMessageCapacity`; drop recoverable via peer re-flood + `GetScpState`.

**Critical determinations:**
1. **No-None busy-loop — SAFE.** `try_recv()` drains use `Err(_) => break` (covers Empty + Disconnected); `select!` arm `Some(..) = recv()` disables (not spins) on `None`; production sender held by OverlayManager for `run()` lifetime; dummy path retains `_dummy_scp_tx`. Channel never closes.
2. **Non-blocking send — SAFE.** Sole send site (mod.rs:735) is `try_send()` with Full→drop+counter, Closed→log. No `.await`/`blocking_send` anywhere.

**CI at decision time:** 5/6 required GREEN (Clippy, Format, Analyze-rust, Build release-ci, + Doctest/Test running). Required Test + Doctest still pending (long-running). Non-required Quickstart Harness pending (OK). No reds.

Not merging on pending required per policy. Re-pick next tick to complete merge once Test + Doctest finish green.

PR #3635, opened 2026-06-26
Reset loadgen meters on clearmetrics (#3630)

### LINKED ISSUE #3630: clearmetrics does not reset loadgen meters — breaks Supercluster max-TPS progress + binary search
# clearmetrics does not reset loadgen meters — breaks Supercluster max-TPS progress + binary search

## Summary

henyey's `clearmetrics` HTTP command is a no-op (`crates/app/src/app/mod.rs` `clear_metrics`
logs and returns; "Prometheus-style metrics don't have a clear operation"). stellar-core's
medida meters ARE reset by `clearmetrics`. Supercluster's max-TPS / min-block-time binary
search calls `clearMetrics` before **every** step and then reads the loadgen meters per run,
so against henyey:

1. `loadgen_txn_attempted` accumulates across steps. The harness progress line
   `Waiting for loadgen run N ... C/D txns` reads this cumulative meter (numerator) against
   the per-step target `txs` (denominator), so it shows a growing "overshoot"
   (observed 168% -> 221% -> 238% -> 329% across steps = the running sum of per-step targets).
   The actual per-run load is correct; the display is just misleading.
2. **More serious**: `loadgen_run_failed` is never reset. `IsLoadGenComplete` returns
   `Failure` whenever `loadgen_run_failed != 0`
   (`vendor/supercluster/src/FSLibrary/StellarCoreHTTP.fs:868`). So once henyey has ONE
   failing step, every subsequent step immediately reads Failure and the binary search
   converges to the wrong (too-low) max rate. (We avoided this only because henyey passed
   every step up to the max-tx-rate cap in the run where this was found.)

## Evidence

`MaxTPSClassic` against henyey (after fixing #3627/#3628/#3629): the harness reported
`264716/80485 txns` etc. for a single step — the numerator is the lifetime
`loadgen_txn_attempted`, not the per-step count, because `clearmetrics` didn't zero it.

## Root cause

`clear_metrics` performs no reset. The loadgen counters
(`crates/app/src/metrics.rs`: `LOADGEN_RUN_START/COMPLETE/FAILED`,
`LOADGEN_ACCOUNT_CREATED`, `LOADGEN_TXN_ATTEMPTED/REJECTED/BYTES`, the per-mode counters,
`LOADGEN_STEP_COUNT`) are monotonic `CounterMetric`s that are never zeroed.

## Fix

In `clear_metrics`, when the domain is empty or `loadgen`, reset the loadgen counters via
`CounterMetric::absolute(0)` (the type already supports `absolute`, used elsewhere to set
counters from snapshots). This mirrors stellar-core's resettable meters and fixes both the
cumulative progress display and the run_failed-lingers binary-search corruption. (A broader
clearmetrics that resets all domains like core's medida `clear()` is out of scope here.)

- Unit: increment `LOADGEN_TXN_ATTEMPTED` and `LOADGEN_RUN_FAILED`, call
  `clear_metrics("")`, assert both render as 0 afterward.
- Regression: simulate two loadgen runs with a `clearmetrics` between; assert the second
  run's `loadgen_txn_attempted` starts from 0 and a prior `loadgen_run_failed` does not
  persist.

## Context

4th drop-in gap found while running `MaxTPSClassic` core-vs-henyey on a Namespace 32x64
instance, after #3627 (pregenerate CLI), #3628 (peer-connection cap), #3629 (preload-file
default). This one specifically invalidates the max-TPS measurement methodology against
henyey until fixed.

tomerweller:
Closes #3630.

`clear_metrics` was a no-op. Supercluster's MaxTPS/MinBlockTime binary search calls `clearmetrics` before every step and reads the loadgen meters per run, so against henyey (a) `loadgen_txn_attempted` accumulated across steps (misleading X/Y progress) and (b) `loadgen_run_failed` from a failed step lingered, making every subsequent step read Failure. This resets the loadgen domain counters via `absolute(0)`, matching stellar-core's resettable medida meters.

Found while running MaxTPSClassic core-vs-henyey on Namespace. **Regression test still TODO** (see #3630).


tomerweller:
## 🔍 Reviewer: Correctness

**Summary:** The reset mechanism is a no-op for exactly the case it must handle. `CounterMetric::absolute(0)` bottoms out at `CounterFn for AtomicU64::absolute`, which is implemented as `fetch_max(value)` — so `absolute(0)` = `fetch_max(0)` = leaves any non-zero counter unchanged. The PR therefore does NOT reset the loadgen meters and does NOT fix #3630. There is also no regression test (PR body says "Regression test still TODO").

<details>

**class: correctness (BLOCKING)**

- `m.absolute(0)` does not reset the counter. In the `metrics` crate (v0.24.5), `impl CounterFn for AtomicU64::absolute(value)` is `self.fetch_max(value, ...)` (`metrics-0.24.5/src/atomics.rs:27-29`). The Prometheus exporter stores every counter as `Arc<AtomicU64>` (`metrics-exporter-prometheus-0.18.3/src/registry.rs:14`), and the generational wrapper delegates the write to that same `AtomicU64`. So `absolute(0)` = `fetch_max(0)` = `max(current, 0)` = `current`. For any counter that has accumulated a positive value — which is the entire scenario #3630 describes — the call is a silent no-op.
- Empirically verified against the PR head: I added a probe test that increments `LOADGEN_TXN_ATTEMPTED` to 100, renders, calls `LOADGEN_TXN_ATTEMPTED.absolute(0)`, and re-renders. The post-reset render still shows `loadgen_txn_attempted 100` (not `0`). Test FAILED. The probe was reverted; it is not part of the diff.
- Net effect: after `clearmetrics`, `loadgen_txn_attempted` still accumulates across Supercluster steps and `loadgen_run_failed` still lingers — i.e. #3630 is not fixed by this change. The PR description's claimed behavior ("This resets the loadgen domain counters via absolute(0)") is incorrect.
- The correct fix must make the underlying counter actually drop to 0. Options: (a) re-register / swap the counter handle so a fresh `Arc<AtomicU64>` at 0 backs the series; (b) track a per-meter baseline offset subtracted at render time; or (c) model these as resettable gauges if the exposition contract allows. Whatever the approach, it must produce a render of `loadgen_* 0` after `clearmetrics` on a previously-non-zero meter, and it must be covered by a test that asserts exactly that.

**class: test-coverage (BLOCKING)**

- Linked issue #3630 is a bug-fix; PR body explicitly states "Regression test still TODO." No test exists. A correct fail-before/pass-after test (increment a loadgen meter, call the clearmetrics path, assert the rendered value is 0) would have immediately caught the `fetch_max` no-op. This is the failing-test-first requirement in CLAUDE.md.

**Scope (non-blocking, confirmed OK):** `clear_metrics` is invoked only from the two HTTP handlers (`compat_http/handlers/plaintext.rs:77`, `http/handlers/admin.rs:143`) — not on any tick. The 15 meters listed match the full loadgen catalog in `metrics.rs` exactly, and the `domain.is_empty() || eq_ignore_ascii_case("loadgen")` gate is the right selector. The intent and selection are correct; only the reset primitive is wrong.

</details>

tomerweller:
## 🔍 Reviewer: Risk

**Summary:** Low operational/regression risk in isolation (observability-only, HTTP-admin-gated, no consensus state touched), but the change ships a fix that does not work: `absolute(0)` = `fetch_max(0)` is a no-op on non-zero counters, so #3630 remains broken and a future reader may believe it is fixed. Merging a non-functional fix that closes the issue is the risk.

<details>

**class: regression-risk** — None to consensus or ledger state. These are pure Prometheus counters scraped externally; `clear_metrics` is only reachable via the `clearmetrics` HTTP command (`plaintext.rs:77`, `admin.rs:143`), never from the event loop or a tick. No risk of corrupting counts via accidental periodic firing.

**class: correctness-of-fix (BLOCKING)** — The change is advertised (PR body, and it Closes #3630) as resetting the loadgen domain, but the primitive used cannot reset a non-zero monotonic counter (`fetch_max(0)`; see Reviewer Correctness for the full chain + empirical proof). Landing this would close #3630 while leaving the underlying Supercluster maxTPS binary-search bug fully intact — a worse state than today, because the issue would be marked resolved. The risk here is a false-positive fix, not a behavioral regression.

**class: parity-intent (non-blocking)** — Parity intent is correct: stellar-core's `ApplicationImpl::clearMetrics` runs a `MetricResetter` over the domain and medida meters genuinely reset to 0 (`ApplicationImpl.cpp:1375`). Henyey's Prometheus counters are monotonic and do not, which is precisely why `absolute(0)` is the wrong tool. The fix needs a primitive that yields a true zero in the rendered exposition.

No API/data-format/migration/config concerns.

</details>

tomerweller:
## 🥊 Refute — Correctness: `absolute(0)` does not reset the loadgen counters

**Outcome:** STANDS

Switched stance and tried to break the finding on all three grounds; none hold:

- **(a) wrong / reviewer misread?** No. `CounterMetric::absolute` → `counter!(name).absolute(v)` → `CounterFn for AtomicU64::absolute(v)` = `self.fetch_max(v, AcqRel)` (`metrics-0.24.5/src/atomics.rs:27-29`). The Prometheus exporter backs every counter with `Arc<AtomicU64>` (`metrics-exporter-prometheus-0.18.3/src/registry.rs:14`); the generational wrapper only bumps a recency generation and delegates the value mutation to that same `AtomicU64`. `fetch_max(0)` cannot lower a positive counter. `CounterMetric` exposes no `set`-style reset path. No misread.
- **(b) already-fixed on current origin/main?** No. Reproduced on the PR head merged state. The diff IS the change; there is no other reset path on main that would make this work.
- **(c) non-blocking?** No — this IS the fix. If the reset is a no-op, #3630 is not fixed and the PR closes the issue falsely.

Empirical proof: a probe test (incr `LOADGEN_TXN_ATTEMPTED` to 100, render, `absolute(0)`, re-render) renders `loadgen_txn_attempted 100` after the reset — FAILS. Reverted; not in the diff.

Finding STANDS → Reviewer Correctness CHANGES_REQUESTED is upheld.

tomerweller:

**Reason:** Reviewer A (Correctness) and Reviewer B (Risk) both requested changes — the fix does not work.

**Reviewer A:** CHANGES_REQUESTED — `absolute(0)` is `fetch_max(0)`, a no-op on non-zero counters; loadgen meters are not actually reset, so #3630 is not fixed. No regression test.
**Reviewer B:** CHANGES_REQUESTED — observability-only/HTTP-gated (low op risk), but ships a non-functional fix that would falsely close #3630.
**CI:** 4/6 required green; Test + Doctest pending (irrelevant — CHANGES_REQUESTED bounces on any CI state).

**Core defect (refute pass: STANDS):** `CounterMetric::absolute(0)` bottoms out at `CounterFn for AtomicU64::absolute = self.fetch_max(value)` (`metrics-0.24.5/src/atomics.rs:27`). The Prometheus exporter stores each counter as `Arc<AtomicU64>` (`metrics-exporter-prometheus-0.18.3/src/registry.rs:14`). So `absolute(0)` = `max(current, 0)` = `current` — a silent no-op for any non-zero meter, i.e. exactly the accumulated-value case #3630 is about. Empirically verified: a probe incrementing `LOADGEN_TXN_ATTEMPTED` to 100 then calling `absolute(0)` still renders `loadgen_txn_attempted 100`.

**What `/do` Mode B must do:**
1. Replace the reset primitive with one that yields a true `0` in the rendered exposition (e.g. swap/re-register a fresh `Arc<AtomicU64>` for each loadgen series, a render-time baseline offset, or another mechanism that actually drops the counter). `absolute(0)`/`fetch_max` cannot do this.
2. Add a fail-before/pass-after regression test (CLAUDE.md requires the failing test first): increment one or more loadgen meters, invoke the clearmetrics path, assert the rendered value is `0`. PR body's "Regression test still TODO" must be resolved.
3. Keep the (correct) scope: gate on `domain.is_empty() || eq_ignore_ascii_case("loadgen")`, only on the HTTP clearmetrics path. Parity intent matches core's `MetricResetter` (`ApplicationImpl.cpp:1375`); only the henyey primitive is wrong.

tomerweller:
## Addressed review feedback — reset now renders a true 0

Both reviewers (Correctness + Risk) were right: `CounterMetric::absolute(0)` = `AtomicU64::fetch_max(0)`, a no-op on any non-zero counter, so the loadgen meters were never reset. A monotonic Prometheus counter cannot be lowered through any public API of the installed `metrics-exporter-prometheus` recorder (the backing `Arc<AtomicU64>` is unreachable; the registry is `pub(crate)` inside `Inner`).

**Mechanism (reviewer option c):** the 15 loadgen meters are now pre-registered **gauges**, reset with `gauge.set(0.0)` (a true `swap`) in a new `metrics::reset_loadgen_meters()`, called from the same `clear_metrics` HTTP path with the unchanged `domain.is_empty() || eq_ignore_ascii_case("loadgen")` gate. This mirrors stellar-core's resettable medida meters (`MetricResetter`, `ApplicationImpl.cpp:1375`). Supercluster reads the value via the compat `/metrics` JSON (`LOADGEN_COMPAT_MAP` → `count`), which is type-agnostic, so the counter→gauge exposition change is invisible to the consumer.

**Empirical render-0 proof (regression test `metrics::tests::test_clear_loadgen_meters_resets_to_zero_rendered`):**
- Increment each loadgen meter (incl. `loadgen_txn_attempted` to 43), render → non-zero (e.g. `loadgen_run_start 1`, `loadgen_txn_attempted 43`).
- Call `reset_loadgen_meters()`, render → asserts every loadgen series renders exactly `0`.
- **Fail-before:** on the prior `absolute(0)` mechanism the test FAILED (`left: "loadgen_run_start 1"`, `right: "loadgen_run_start 0"`) — committed as the failing-test artifact `4371335e`.
- **Pass-after:** PASSES on the fix `a554413e`.

All 1123 `henyey-app` lib tests pass; `cargo fmt --check`, `cargo clippy --all -- -D warnings` clean. Ready for re-review (a554413ef9798d5fad8183a96bfd67882f1891de).

Issue #3574, opened 2026-06-23
Henyey loadgen aborts run on QueueFull (no back-off/retry) — soroban loadgen phase-2 fails; SSC henyey-majority compositions fail


tomerweller:
## Summary

With the loadgen-metrics fixes in (#3571 + #3573), SSC henyey-majority compositions now clear the **classic** loadgen, but **fail the soroban loadgen**: henyey's load generator treats a transient **`QueueFull`** transaction-queue result as a **fatal run failure** (sets `loadgen.run.failed`) instead of backing off and retrying. stellar-core's loadgen paces submission and retries on a full queue, so it does not fail. This blocks both henyey-majority SSC compositions (1 core + 2 henyey; 0 core + 3 henyey), where henyey drives the soroban loadgen.

## Verification (both henyey-majority runs, henyey `dacc262b`)

Identical failure on c1h2-rerun2 (cluster `gkv1bvavo84ri`) and c0h3-rerun2 (`fnhfu6t0dhm2u`):
```
INFO  henyey::loadgen_runner: Load generation complete submitted=100          # soroban setup phase 1 (WASM upload) OK
INFO  henyey_simulation::loadgen: Setup phase 1 complete, transitioning to instance deployment n_instances=1
WARN  henyey_simulation::loadgen: Transaction submission failed: QueueFull     # phase 2 (instance deploy) first submit
ERROR henyey::loadgen_runner: Load generation failed
```
Supercluster then reads `loadgen.run.failed != 0` and raises `Loadgen failed!` (`StellarCoreHTTP.fs:896`), exit 134. Henyey itself stays healthy and `Synced!` — this is purely a loadgen-runner robustness/parity gap, not a consensus issue.

(The classic-loadgen progress now advances correctly — c1h2 reached `95/100 txns` on run 1 before the phase-2 QueueFull — confirming #3571/#3573 work.)

## Root cause

`crates/simulation/src/loadgen.rs` ~1542-1576, the submit-transaction retry loop:
```rust
match result {
    TxQueueResult::Added => return true,
    TxQueueResult::Invalid(Some(TxResultCode::TxBadSeq)) => { /* retry up to TX_SUBMIT_MAX_TRIES */ }
    TxQueueResult::TryAgainLater | TxQueueResult::FeeTooLow if config.skip_low_fee_txs => { /* roll back + skip */ }
    other => {
        warn!("Transaction submission failed: {:?}", other);
[...]

PR #3636, opened 2026-06-26
Fail classic loadgen run on wait-till-complete timeout (#3631)

### LINKED ISSUE #3631: Loadgen wait_till_complete doesn't fail classic path on timeout — inflates max-TPS, masks dropped txns
# Loadgen wait_till_complete does not fail the classic path on timeout — inflates max-TPS, masks dropped txns

## Summary

`LoadGenerator::wait_till_complete` (`crates/simulation/src/loadgen.rs:1811-1845`) only
marks the run failed on timeout when `check_soroban && !soroban_synced`. For the **classic**
`Pay` / `PayPregenerated` path (`check_soroban == false`), a timeout where accounts are NOT
synced — i.e. submitted transactions were never applied — returns **without** setting
`self.failed`. The driver then maps the run to `LOADGEN_RUN_COMPLETE` and Supercluster reads
`loadgen.run.start == loadgen.run.complete` as **Success**. So a classic loadgen run whose
transactions never landed is reported as a successful run.

stellar-core does the opposite: `LoadGenerator::waitTillComplete`
(`sc-v27/src/simulation/LoadGenerator.cpp:1399-1404`) calls `emitFailure(!sorobanIsDone)`
**unconditionally** on timeout, and `emitFailure` (`:1430-1433`) marks `mLoadgenFail`
unconditionally (the bool arg only controls soroban-state reset). So core fails the run for
both classic and soroban timeouts.

## Impact (discovered via MaxTPSClassic core-vs-henyey on a Namespace 32x64 VM)

This silently inflates `MissionMaxTPSClassic` / `MaxTPSMixed` results for henyey. The binary
search declares a tx-rate "passed" purely on run-complete + pairwise-consistency + in-sync —
none of which catch empty ledgers. Measured ground truth on the live henyey nodes at a
"passing" rate of 2753 tx/s:

- `loadgen.txn.attempted` ≈ 1,028,781 vs a 152,742 target = ~6.7x over-attempt (the loadgen
  dumps, the network can't include it).
- Of 92 closed ledgers, only **7 were non-empty (8%)**; median/75th-pct txns-per-ledger = 0.
  `herder.pending.transactions` climbed into the thousands while ledgers closed empty.
- Node CPU stayed ~7/32 (near idle) at the "passing" rate — because almost no apply work was
  happening.

So henyey "passed" 2012/2506/2753 tx/s and only failed at 4376, vs stellar-core honestly
failing at ~1542 (ceiling **1531**) on the identical infra/harness. The henyey ceiling was an
artifact of this bug, not real throughput. Consistency checks don't catch it because empty
ledgers are trivially identical across all nodes.

## Fix

In the timeout branch of `wait_till_complete`, set `self.failed = true` **unconditionally**
(we only reach it when accounts or soroban state are unsynced), matching core's unconditional
`emitFailure` on timeout. (Applied locally; this is what this issue tracks for a proper PR +
regression test.)

- Unit/integration: run a classic Pay/PayPregenerated loadgen at a rate the (test) network
  cannot apply within `TIMEOUT_NUM_LEDGERS`; assert the run is marked FAILED
  (`loadgen_run_failed` incremented, `loadgen_run_complete` not), not completed.
- Keep the sustainable-rate case passing (accounts sync before timeout → complete).

## Context

5th drop-in / parity issue found during the max-TPS A/B, after #3627 (pregenerate CLI),
#3628 (peer-connection cap), #3629 (preload-file default), #3630 (clearmetrics loadgen
reset). This one is the most serious: it makes henyey's max-TPS numbers untrustworthy until
fixed. Found via an adversarial audit of the measurement.

tomerweller:
Closes #3631.

`wait_till_complete` only marked `self.failed` on timeout when `check_soroban && !soroban_synced`. The classic `Pay`/`PayPregenerated` path returned **without** failing even when accounts never synced (submitted txns never applied), so the run was reported as complete = success. stellar-core's `waitTillComplete` calls `emitFailure` **unconditionally** on timeout (`LoadGenerator.cpp:1399-1404,1430-1433`).

Impact: this silently inflated henyey's max-TPS — a 23-node single-VM run 'passed' 2753 tx/s with >90% empty ledgers while core honestly failed at ~1531. Fix: mark failed unconditionally on timeout (we only reach that branch when something is unsynced).

Found via an adversarial audit while running MaxTPSClassic core-vs-henyey on Namespace. **Regression test still TODO** (see #3631).

### DIFF
--- a/crates/simulation/src/loadgen.rs
+++ b/crates/simulation/src/loadgen.rs
@@ -1831,13 +1831,21 @@ impl LoadGenerator {
                     "loadgen wait-till-complete timed out; some submitted txns \
                      were not applied (likely dropped before inclusion)"
                 );
-                // Parity: emitFailure(!sorobanIsDone). A setup run whose
-                // contract code/instance never materialized must fail — a
-                // downstream create_upgrade built against the missing instance
-                // would never produce a resolvable ConfigUpgradeSet entry.
-                if check_soroban && !soroban_synced {
-                    self.failed = true;
-                }
+                // Parity: stellar-core `waitTillComplete` marks `mLoadgenFail`
+                // UNCONDITIONALLY on timeout via `emitFailure(!sorobanIsDone)`
+                // (`LoadGenerator.cpp:1399-1404` calls emitFailure; `:1430-1433`
+                // marks the fail meter regardless of the arg — the arg only
+                // controls whether soroban state is reset). We reach this branch
+                // only when accounts (or soroban state) are still unsynced, i.e.
+                // some submitted txns were never applied, so the run MUST fail —
+                // for the classic Pay/PayPregenerated path too, not just soroban.
+                // Previously henyey only failed the soroban-setup path, so a
+                // classic max-TPS step whose txns never landed (ledgers closing
+                // empty under over-rate) still reported run-complete = success,
+                // inflating the measured ceiling far past the real one (a
+                // 23-node single-VM run "passed" 2753 tx/s with >90% empty
+                // ledgers while core honestly failed at ~1531). (#3631)
+                self.failed = true;
                 return;
             }
             tokio::time::sleep(Duration::from_millis(1000)).await;


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Minimal, correct fix: the wait_till_complete timeout branch now marks `self.failed = true` unconditionally, which is only reachable when accounts/soroban are genuinely unsynced. Failure correctly surfaces as `LoadResult::Failed`. No regression test (acknowledged TODO in #3631) is the only gap — non-blocking, filed as follow-up.

<details>

Reviewed against current origin/main (loadgen.rs is the only changed file, +15/-7).

**Correctness — confirmed sound:**
- The timeout branch (`elapsed >= TIMEOUT_NUM_LEDGERS`) is reached only after the top-of-loop early-return `if accounts_synced && soroban_synced { return; }` fails. So `self.failed = true` fires only when something is genuinely unsynced (submitted txns never applied). No false-failure path on a genuinely-completed run.
- No completion race: completion is re-checked at the top of every poll iteration before the timeout test; a run that just-barely finished returns `Done` on that iteration rather than timing out. Same control structure as stellar-core's `waitTillComplete`.
- Failure surfacing: `generate_load` calls `self.wait_till_complete(config).await; if self.failed { return LoadResult::Failed; }` — previously the classic path left `self.failed = false` on timeout → `LoadResult::Done { submitted }` = false success. Now returns `Failed`. This is the bug fix.

**Soroban path unaffected (concern class: regression-risk):**
- Old code: `if check_soroban && !soroban_synced { self.failed = true; }`. For soroban-setup timeouts this already set failed; new code sets failed in the same case plus the classic case. So soroban-setup behavior is unchanged; only the classic (and any non-setup) timeout path changes from silent-ok to fail. The `soroban_synced` term collapses to `true` for non-setup modes (`!check_soroban || ...`), so reaching the timeout there means `accounts_synced` is false — correctly a real failure.

**Test coverage (concern class: test-coverage — NON-BLOCKING, inline + follow-up):**
- No unit/integration test asserts classic-timeout → `LoadResult::Failed`. The branch sets `self.failed` inline and requires a full `App` to drive, consistent with `test_stop_signal_contract` deferring behavioral checks to integration tests.
- The fix would be made unit-testable by extracting the timeout decision into a pure helper (the established `classify_submit_result` / `apply_load_overlay_gate_fails` pattern in this file) returning whether to fail. Recommended but not blocking; PR body already tracks the regression test under #3631. Will file a follow-up so this is not dropped.

**Comment accuracy:** the cited core references are correct — `emitFailure` (LoadGenerator.cpp ~1430) calls `mLoadgenFail.Mark()` unconditionally; the `resetSoroban` bool only gates `resetSorobanState()`. The PR comment describes this precisely.
</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Low blast radius (loadgen only — a test/benchmark harness, not consensus/ledger). The change makes a timeout that already meant "txns never applied" surface as a failure instead of a false success. Parity-with-core determination confirmed: core fails (not warns) on this timeout. No risk to soroban/setup modes or to genuinely-completed runs.

<details>

Re-read the diff fresh through the risk lens.

**Blast radius (concern class: regression-risk):**
- Single file `crates/simulation/src/loadgen.rs`, +15/-7, all in `wait_till_complete`. `simulation` is the load-generation/benchmark crate; it does not touch consensus, ledger close, bucket, or overlay correctness. No determinism/parity-hash surface. Worst case of the change is a loadgen run that previously reported a (false) success now reports a failure — which is the intended behavior and the whole point of #3631.

**Parity determination (vs stellar-core v26.0.1 LoadGenerator) — confirmed:**
- Verified `LoadGenerator.cpp`: in `waitTillComplete`, the `!classicIsDone || !sorobanIsDone` branch, on `++mWaitTillCompleteForLedgers >= TIMEOUT_NUM_LEDGERS`, calls `emitFailure(!sorobanIsDone); return;`.
- `emitFailure(bool resetSoroban)` does `mLoadgenFail.Mark(); reset(); if (resetSoroban) resetSorobanState();`. The fail meter is marked UNCONDITIONALLY; the bool only controls soroban-state reset. So core FAILS the run on this timeout for the classic path too — this is parity, not new semantics. The prior henyey code (`if check_soroban && !soroban_synced`) under-failed relative to core; the PR restores parity.

**Soroban-path-safe (concern class: regression-risk):**
- For non-setup modes `check_soroban` is false → `soroban_synced` is forced true, so a timeout there implies unsynced accounts = a real classic failure. For setup modes the old code already failed; behavior preserved. No mode that legitimately has different completion semantics is wrongly failed: every mode reaches this branch only with real unsynced state.

**False-failure / race (concern class: regression-risk):**
- Completion is checked at the top of each poll before the timeout test; a just-completed run returns early with success. The fail only triggers on a true 30-ledger timeout. No newly-introduced flakiness beyond core's own timeout sensitivity (and the prior #3574/#3611/account-cache-reset fixes already harden the not-synced-forever poisoning cases).

**Operational risk:** in CI/missions this can newly fail a loadgen step that was previously (wrongly) green. That is correct and is the stated motivation (the 2753 vs 1531 tx/s inflation). No mitigation needed.

Inline non-blocking note already raised by Correctness lens: a pure-helper extraction + unit test for the timeout decision; tracked as follow-up.
</details>

tomerweller:
## Review: Auto-merge armed

Both reviewer lenses (Correctness + Risk) APPROVE; refute pass did not overturn. All 6 required checks green (Test, Doctest, Clippy, Format, Build (release-ci), Analyze (rust)). No external CHANGES_REQUESTED.

Parity confirmed vs stellar-core v26.0.1: `waitTillComplete` timeout calls `emitFailure(!sorobanIsDone)`, and `emitFailure` marks `mLoadgenFail` UNCONDITIONALLY (the bool only gates `resetSorobanState`). The classic path must fail on timeout — this is parity, not new semantics. Soroban/setup paths unchanged; no false-failure on completed runs.

**Follow-up issues filed:** #3639 (unit test + pure-helper extraction for the timeout→fail decision; regression test from #3631).

Merging (squash, --admin).

PR #3644, opened 2026-06-26
Pace loadgen on submit attempts, fail run on PayPregenerated reject

### LINKED ISSUE #3638: Loadgen does not pace classic submission at the configured tx-rate (over-submits)
# Loadgen does not pace classic submission at the configured tx-rate (over-submits / dumps)

## Summary

henyey's classic `Pay` / `PayPregenerated` loadgen appears not to pace submission at the
configured `txrate`. Instead it submits as fast as it can read/build transactions, so
`loadgen_txn_attempted` overshoots the per-run target by large factors very quickly
(observed ~1.5x–6.7x within seconds, growing with rate). stellar-core paces submission so
that attempts track `rate * elapsed`.

## Why it matters

This corrupts Supercluster max-TPS measurement in both directions:
- Before the timeout-failure fix (#3631), the over-submitted flood + the masked failure made
  henyey "pass" absurd rates with empty ledgers.
- After #3631, the un-paced flood means the tx queue is swamped at ANY rate, so the network
  cannot drain it within `TIMEOUT_NUM_LEDGERS` and runs FAIL even at low rates (henyey failed
  293/537/1025 in an honest rerun, with ledgers only ~27–49% full and CPU idle). So the
  honest max-TPS number is artificially depressed by the loadgen, not by apply capacity.

Net: henyey's MaxTPSClassic result is untrustworthy until submission is paced to the
configured rate (and the per-run `txs` count is respected).

## Evidence

- `loadgen.txn.attempted` reached ~6.7x the per-step target and got there in seconds, not
  over the intended ~5 min window (live nodes during MaxTPSClassic on a Namespace 32x64 VM).
- Ledgers under-filled (~27–49% of maxTxSetSize) with node CPU ~7–13/32 (idle) — the box was
  not apply-bound; the bottleneck was queue admission / submission shape.

## What to check / fix

- `crates/simulation/src/loadgen.rs` submission loop: confirm whether it sleeps/schedules to
  honor `config.txrate` (e.g. submit `rate` txns per simulated second / per ledger) vs
  draining the pregenerated reader as fast as possible. Compare to stellar-core
  `LoadGenerator::scheduleLoadGeneration` / `mLoadgen` step pacing
  (`sc-v27/src/simulation/LoadGenerator.cpp`), which spreads `txCount` across the run at the
  target rate.
- Confirm the per-run `txs` (count) limit is enforced for PayPregenerated (attempts should
  not exceed it except for genuine resubmissions of rejected txns).

- Unit/integration: run a classic loadgen at rate R for N txns and assert the submission
  timeline tracks ~R/s (attempts at time t ≈ R*t, within tolerance), not a single burst.

## Context

6th issue from the MaxTPSClassic core-vs-henyey A/B (after #3627–#3631). Together with #3631
this is why henyey's max-TPS is currently unmeasurable; also relevant: under-filled ledgers +
idle CPU suggest a separate consensus/tx-set-inclusion throughput investigation.

tomerweller:
Closes #3638

## Summary

henyey's loadgen paces via a cumulative-target rate limiter (`get_tx_per_step`, matching core's `getTxPerStep`), but the pacing counter `total_submitted` advanced only on **accepted** txns, whereas stellar-core advances per **attempt** (`LoadGenerator.cpp:900,919` — `++count` regardless of accept, then `mTotalSubmitted += count`). Under queue rejection henyey's deficit never amortized, so each step burst harder to "catch up" — the observed 1.5–6.7x overshoot that grows with the rejection rate.

This change feeds the pacing counter per **attempt** instead. Work-remaining (`config.n_txs`) stays **accept**-gated, matching core's `--cfg.nTxs` (line 874). `submit_tx` now returns a three-way `SubmitOutcome` (`Accepted` / `NotAccepted`); the **not-attempted** case (`get_next_available_account → None` break) produces no outcome and never advances pacing. The per-step pacing update is extracted into a pure helper (`step_pacing_update`) for unit-testability, following the existing `classify_submit_result` precedent.

Same-bug amplifier also fixed: a `PayPregenerated` reject now sets `self.failed` and fails the run (`pay_pregenerated_reject_fails`), mirroring core's `submitTx` mFailed+abort (`LoadGenerator.cpp:958-963`), instead of silently draining the next tx from the file.

`get_tx_per_step` is left unchanged (already matches `getTxPerStep`).

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3638#issuecomment-4806307627)

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy -p henyey-simulation --all-targets -- -D warnings`
- [x] `cargo build -p henyey-simulation --all-targets`
- [x] `cargo test -p henyey-simulation --lib` (103 passed, incl. all loadgen + new pacing tests)

- **Tests:** in `crates/simulation/src/loadgen.rs` `#[cfg(test)]`:
  - `test_pacing_counter_counts_attempts_not_just_accepts` (5 attempts / 2 accepts → pacing +5, not +2)
  - `test_deficit_tracks_rate_under_rejection` (attempts-based deficit bounded; accepts-only balloons to 90)
  - `test_not_attempted_does_not_advance_pacing` (no-account break → pacing +0)
  - `test_pay_pregenerated_reject_fails_run` (any non-`Added` → fail run)
- **Pre-fix:** committed as `5bc4c023` — verified FAILED: tests do not compile on main (`SubmitOutcome`, `step_pacing_update`, `pay_pregenerated_reject_fails` do not exist — the missing parity surface).
- **Post-fix:** verified PASS after `883af261`.

## Deviations from plan

The plan offered two drivers for `test_pay_pregenerated_reject_fails_run`: the `app_simulation.rs` App harness, or a pure seam ("if a pure seam is feasible, assert via the classification logic instead. /do picks the drivable form"). I chose the **pure seam** (`pay_pregenerated_reject_fails`) — a deterministic queue reject is not cheaply reproducible through a live `App`, and the pure helper matches the `classify_submit_result` precedent the plan endorses.

### DIFF
--- a/crates/simulation/src/loadgen.rs
+++ b/crates/simulation/src/loadgen.rs
@@ -3434,4 +3434,129 @@ mod tests {
             SubmitAction::Accept,
         );
     }
+
+    // ---------------------------------------------------------------------
+    // #3638: per-step pacing counter must track ATTEMPTS (mirroring core's
+    // `++count` / `mTotalSubmitted += count`, LoadGenerator.cpp:900,919), not
+    // just accepts. `n_txs` (work remaining) stays accept-gated. These tests
+    // drive the pure accounting helpers so the pacing logic is deterministic
+    // and decoupled from a live `App`.
+    // ---------------------------------------------------------------------
+
+    /// #3638 key regression: the pacing counter advances by the number of
+    /// ATTEMPTS in a step, not the number of accepts. A step that attempts 5
+    /// txns and accepts 2 must advance `total_submitted` by 5.
+    ///
+    /// FAILS on origin/main: there is no `step_pacing_update` helper and the
+    /// loop adds only accepts (`submitted_this_step += 1` inside `if ok`), so
+    /// the counter would advance by 2.
+    #[test]
+    fn test_pacing_counter_counts_attempts_not_just_accepts() {
+        // Five outcomes: 2 accepted, 3 attempted-but-not-accepted. All five are
+        // attempts; none is "not attempted".
+        let outcomes = [
+            SubmitOutcome::Accepted,
+            SubmitOutcome::NotAccepted,
+            SubmitOutcome::Accepted,
+            SubmitOutcome::NotAccepted,
+            SubmitOutcome::NotAccepted,
+        ];
+        let attempts: i64 = outcomes.iter().filter(|o| o.is_attempt()).count() as i64;
+        let accepts: i64 = outcomes.iter().filter(|o| o.is_accepted()).count() as i64;
+        assert_eq!(attempts, 5, "all five outcomes are attempts");
+        assert_eq!(accepts, 2, "two of the five were accepted");
+
+        // Pacing counter advances by attempts (5), NOT accepts (2).
+        let total_submitted = 100;
+        assert_eq!(
+            step_pacing_update(total_submitted, attempts),
+            105,
+            "pacing counter must advance by attempts (mirrors mTotalSubmitted += count)"
+        );
+    }
+
+    /// #3638: under a rejection-heavy step the next step's deficit must track
+    /// `rate*elapsed - attempts_so_far` (bounded ≈ one step's worth), NOT the
+    /// inflated `rate*elapsed - accepts_so_far`. Driven by feeding scalars into
+    /// the pure pacing helper + the `get_tx_per_step` deficit math (no `Instant`
+    /// flakiness).
+    ///
+    /// FAILS on origin/main: with accepts-only accounting the deficit balloons.
+    #[test]
+    fn test_deficit_tracks_rate_under_rejection() {
+        // tx_rate = 100 tx/s. After 1s the cumulative target is 100.
+        let tx_rate: i64 = 100;
+        let elapsed_ms_after_step_1: i64 = 1000;
+        let target_after_step_1 = elapsed_ms_after_step_1 * tx_rate / 1000; // 100
+
+        // Step 1 attempted all 100 targeted txns but only 10 were accepted
+        // (90 rejected under queue pressure).
+        let attempts_step_1: i64 = 100;
+        let accepts_step_1: i64 = 10;
+
+        // Attempts-based pacing (the fix): total_submitted == attempts.
+        let total_attempts = step_pacing_update(0, attempts_step_1);
+        let deficit_attempts = (target_after_step_1 - total_attempts).max(0);
+
+        // Accepts-only pacing (the bug on main): total_submitted == accepts.
+        let total_accepts = step_pacing_update(0, accepts_step_1);
+        let deficit_accepts = (target_after_step_1 - total_accepts).max(0);
+
+        // The fix keeps the deficit bounded (≈ 0 here — we already attempted the
+        // full step's target), while the bug leaves a 90-tx deficit that the
+        // next step would burst to "catch up".
+        assert_eq!(deficit_attempts, 0, "attempts-based pacing fully amortizes");
+        assert_eq!(
+            deficit_accepts, 90,
+            "accepts-only pacing leaves an inflated deficit (the bug)"
+        );
+        assert!(
+            deficit_attempts < deficit_accepts,
+            "attempts-based deficit must be bounded relative to accepts-only"
+        );
+    }
+
+    /// #3638 (Critic A): a step whose inner loop hits the
+    /// `get_next_available_account -> None` break makes ZERO `submit_tx` calls
+    /// (a non-attempt), so the pacing counter must NOT advance. Guards against
+    /// the fix over-counting the no-account case.
+    #[test]
+    fn test_not_attempted_does_not_advance_pacing() {
+        // Zero attempts this step (the loop broke before any submit_tx call).
+        let total_submitted = 42;
+        assert_eq!(
+            step_pacing_update(total_submitted, 0),
+            42,
+            "a step with no attempts must not advance the pacing counter"
+        );
+        // SubmitOutcome's attempt predicate is the seam the loop uses to decide
+        // whether an iteration counts; only attempted outcomes are attempts.
+        assert!(SubmitOutcome::Accepted.is_attempt());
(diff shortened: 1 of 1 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The pacing-counter-per-attempt fix is correct and well-factored: `SubmitOutcome` cleanly encodes the attempt/accept distinction, the not-attempted (`get_next_available_account → None`) break correctly produces no outcome and never advances pacing, and the new pure helpers (`step_pacing_update`, `pay_pregenerated_reject_fails`) follow the existing `classify_submit_result` precedent. Regression tests genuinely fail-before / pass-after.

<details>

**kind: bug-fix — regression test verification (PASS):**
- Test commit `5bc4c023` precedes fix commit `883af261` (correct ordering).
- The three new symbols (`SubmitOutcome`, `step_pacing_update`, `pay_pregenerated_reject_fails`) do **not** exist on `origin/main` (grep count 0), so the regression tests committed at `5bc4c023` **fail to compile on main** — a legitimate fail-before. They pass after `883af261`. The PR body's `## Regression test` section accurately describes this.
- All four tests assert the intended behavior: attempts (5) vs accepts (2) → pacing +5; deficit bounded under rejection (0 vs 90); no-attempt step → pacing +0; any non-`Added` PayPregenerated result → fail.

**Logic correctness:**
- The inner-loop wiring is correct: `is_attempt()` advances `attempts_this_step` for both `Accepted` and `NotAccepted`; `is_accepted()` gates the `config.n_txs` decrement. The not-attempted break (no-account) occurs before `submit_tx`, so no outcome is produced — matches the doc and the `test_not_attempted_does_not_advance_pacing` guard.
- `submit_tx` return-type migration `bool → SubmitOutcome` is mechanically faithful: every `return true` → `Accepted`, every `return false` → `NotAccepted`. The two seqnum-rollback paths (SkipLowFee, RetryQueueFull) are byte-identical to the prior code.
- PayPregenerated arm now sets `self.failed = true` and returns `NotAccepted` on any non-`Added` result, replacing the prior silent `return false`. Matches the issue intent.

**Minor (non-blocking, inline-class `accounting-edge`):** On the terminal-failure path (lines 1818-1824), the failing attempt is folded into `total_submitted` before `return LoadResult::Failed`. This is harmless because the counter is never read after a Failed return (the run terminates), so it's observationally dead — noted for completeness, not a blocker.

**Error handling / readability:** Doc comments are thorough and cite the exact core line numbers. `i64` arithmetic in `step_pacing_update` cannot overflow at realistic loadgen scales. No clippy/fmt concerns (CI Clippy + Format green).

Not self-modifying (touches only `crates/simulation/src/loadgen.rs`).
</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** No regression risk to consensus or determinism — this is test/benchmark tooling. The core-parity claim is sound: pacing advances per-attempt (mirrors `++count`/`mTotalSubmitted += count`, LoadGenerator.cpp:900/919), work-remaining stays accept-gated (`--cfg.nTxs`:874), PayPregenerated-reject fails the run (`submitTx` mFailed:958-963). The deterministic-order question checks out: account selection stays fully deterministic, and it was never a core-parity surface to begin with.

<details>

**Parity determination — attempt-vs-accept (the central question):**
- Core `generateLoad` (LoadGenerator.cpp:869-919): `submitTx` true → `--cfg.nTxs`; then `if (mFailed) break;` (896); then `++count` (900); finally `mTotalSubmitted += count` (919). So the pacing counter advances **per attempt** while work-remaining decrements **per accept**. The PR mirrors this exactly: `is_attempt()` → `attempts_this_step`, `is_accepted()` → `n_txs.saturating_sub(1)`. **Parity: correct.**
- 3-way distinction is correct: Accept→`Accepted`; SkipLowFee/classic-reject (no mFailed)→`NotAccepted` with `self.failed==false` so the loop counts the attempt and continues — matches core, where a SkipLowFee `submitTx` returns false WITHOUT mFailed and still reaches `++count`. Not-attempted (no-account break before `submit_tx`)→no outcome, no pacing — matches core's `++count` being post-submit-only.
- **Minor parity divergence (`parity-edge`, non-blocking):** terminal-fail paths (build error / `SubmitAction::Fail` / PayPregenerated reject) set `self.failed` and the loop folds that attempt into `total_submitted` before returning Failed. Core, by contrast, breaks at `if(mFailed)` (896) **before** `++count` (900), so core does NOT count the failing attempt. This is observationally dead — the counter is read only by `get_tx_per_step` on a *subsequent* step, and there is no subsequent step after `LoadResult::Failed`. No behavioral divergence; flagging only for the record.

**PayPregenerated reject → fail run:** Matches core `submitTx` lines 945-963 (PAY_PREGENERATED non-PENDING → `mFailed = true; return false`), which propagates to the outer `if(mFailed) break`. The PR's `self.failed = true; return NotAccepted` reproduces this. Correct.

**Deterministic-order (focus #3) — no regression:**
- `get_next_available_account` seeds `deterministic_rand(self.total_submitted, ledger_num)`. Changing `total_submitted` from accept-gated to attempt-gated therefore shifts the account-selection *sequence* relative to old henyey. BUT: (a) core's `getNextAvailableAccount` (line 985) uses `rand_uniform`/the global engine — it does NOT seed off `mTotalSubmitted` at all, so account selection was never a core-parity surface; henyey's `deterministic_rand` seam is already a deliberate henyey-only determinism substitute. (b) Selection remains **fully deterministic** (same inputs → same output) — the change introduces no nondeterminism. The PR's scope claim ("affects submission RATE only, not the deterministic generation order") holds in the sense that matters: determinism is preserved and tx *generation* (seqnums, amounts, mode dispatch) is untouched. The CI Test shard exercises the loadgen unit tests; any selection-sequence assertion would surface there.

**Other risk vectors:** No API/data-format break (return-type change is crate-internal, `submit_tx` is private). No perf impact (one extra branch per inner-loop iteration). No config/migration concern. Not consensus-path. Not self-modifying.
</details>

Issue #3642, opened 2026-06-26
Phase 2: wire read-side socket throttle (can_read/maybe_throttle_read) into peer loop


tomerweller:
Phase 2 follow-up of #3625 (credit-based inbound flow control). Phase 1 (drain-gated SEND_MORE release) lands the headline property. Phase 2 wires the already-ported but dead-code read-throttle into the peer read loop.

## Scope
- Wire `FlowControl::can_read()` / `maybe_throttle_read()` / `stop_throttling()` (currently zero production callers in `crates/overlay/src` outside flow_control.rs/tests) into `crates/overlay/src/manager/peer_loop.rs` so the peer stops reading from the socket when local total reading capacity is exhausted, and reschedules the read when capacity is released (core `TCPPeer.cpp:620` maybeThrottleRead, `Peer.cpp:314-333` stopThrottling+scheduleRead).

## Required correctness analysis (Critic A self-wedge concern)
henyey shares ONE TCP socket per peer; gating `peer.recv()` wholesale stops ALL message types from that peer, including fetch responses (ScpQuorumset/TxSet/DontHave) the stalled consumer may depend on for forward progress. Phase 2 MUST prove the throttle does not starve a backfill the consumer needs (henyey routes fetch over a separate unbounded channel but the socket is shared). Also avoid a tokio busy-spin when the read arm is disabled.

## Parity
`FlowControl.cpp:334-346` (canRead), `FlowControlCapacity.cpp` canRead. Ref: #3625 converged plan.

crate:overlay

tomerweller:
## 📝 Plan Draft (Round 1)

**Summary:** Wire the already-ported, currently-dead read-side socket throttle (`can_read`/`maybe_throttle_read`/`stop_throttling`) into the per-peer `run_peer_loop`, mirroring stellar-core's `TCPPeer::startRead`/`maybeThrottleRead` + `Peer::endMessageProcessing`→`stopThrottling`+`scheduleRead`. This is **Phase 2** of #3625; Phase 1 (#3646, merged `5f6a5c76`) already release-gates SEND_MORE credit on SCP-consumer drain via the `FlowControlRelease` token — that drain-release is exactly what lifts the Phase-2 throttle, closing the loop.

### Architecture finding (the self-wedge question)

[...]

PR #3648, opened 2026-06-26
Wire read-side socket throttle into the per-peer read loop (#3642 Phase 2)

tomerweller:
## Summary

**Phase 2 of #3625** (overlay inbound flow control). The ported `can_read` / `maybe_throttle_read` / `stop_throttling` methods on `FlowControl` had **zero production callers** (dead code). This wires them into the per-peer read loop, mirroring stellar-core `TCPPeer::maybeThrottleRead` + `Peer::endMessageProcessing` → `stopThrottling()` + `scheduleRead()`.

Phase 1 (#3646, `5f6a5c76`) already release-gates SEND_MORE credit on SCP-consumer drain via the `FlowControlRelease` token. That same drain-release is exactly what returns `total_capacity` and now lifts the Phase-2 read throttle — closing the loop with **no deadlock**.

## Wiring (vs stellar-core v26.0.1)

| Call site | Change | Core reference |
|---|---|---|
| `peer_loop.rs` `run_peer_loop` select | gate the `peer.recv()` arm `if flow_control.can_read()` — tokio disables a false-precondition arm, so a throttled peer stops pulling bytes (no busy-spin), back-pressuring the sender at the TCP window | `TCPPeer::maybeThrottleRead` declining to reschedule the read, `TCPPeer.cpp:620/770` |
| `peer_loop.rs` after `handle_received_message` | call `flow_control.maybe_throttle_read()` (records `last_throttle` when `!can_read()`) | post-`recvMessage` `maybeThrottleRead`, `TCPPeer.cpp:620/770` |
| `peer_loop.rs` | create a per-peer `Arc<Notify>`; add a select arm `_ = read_resume.notified() => {}` that re-evaluates the loop (re-enables the gated read) | async equivalent of `Peer::scheduleRead`, `Peer.cpp:313-333` |
| `mod.rs` `FlowControlRelease` | add `Option<Arc<Notify>>` via `with_resume_notify`; in `release()`, after `end_message_processing`, fire `stop_throttling()` + `notify_one()` gated on `is_throttled() && cap.num_total_messages > 0` (the already-returned 201-batch field — not recomputed) | `Peer.cpp:314` gate `isThrottled() && numTotalMessages > 0` |

## No-deadlock / self-wedge invariant (preserved)

- `can_read()` = `total_capacity > 0` (message-count track; byte track is `None`→always-true). Capacity is exhausted only by **un-drained SCP envelopes**, and released by the consumer **drain** (`pump_scp_intake` → drop token → `end_message_processing`), which never reads the throttled socket. Resume is never a prerequisite of reading more.
- The fetch backfill SCP depends on (TxSet / GeneralizedTxSet / ScpQuorumset / DontHave) is **not** in `is_flow_controlled_message`, rides the separate unbounded `fetch_response` channel, and is **never throttled** → a throttled SCP socket cannot starve the fetch the consumer needs.
- Resume uses **Notify + the existing 1s periodic-tick backstop** together, so a missed wake can never wedge a peer for more than ~1s.

## Overlay-only footprint

`crates/app` is **unchanged** — the `Notify` is created peer-side and threaded into the token; the app SCP consumer only drops the token (which now also drives resume). `git diff --stat` touches only `crates/overlay/`.

Commit 1 (`7824cbba`) adds the tests; tests 2 & 4 **fail to compile on current main** (`no method with_resume_notify` — the resume path is unwired), proving the feature is dead. Commit 2 (`b65b47c6`) wires it; all 4 pass.

1. `test_read_throttles_when_total_capacity_exhausted` — exhaust total capacity → `can_read()` false, `maybe_throttle_read()`/`is_throttled()` true (the gate the select arm keys on).
2. `test_throttle_lifts_and_read_resumes_after_consumer_drains` — **recovery centerpiece**: from exhausted state, dropping the held tokens (consumer drain, no read) lifts the throttle, restores `can_read()`, and fires the `Notify`.
3. `test_throttle_does_not_block_fetch_path` — fetch-path messages are not flow-controlled (self-wedge invariant).
4. `test_reschedule_only_on_full_total_batch` — partial drain does NOT lift the throttle or fire the Notify; only the batch-completing drain (`num_total_messages > 0`) does (Peer.cpp:314 parity).

## Verification

- `cargo test -p henyey-overlay` — 466 + integration suites pass.
- `cargo test -p henyey-app` (`test_scp_intake_channel_is_bounded`, dedup pipeline, recovery) — pass; throttle does not wedge recovery.
- `cargo fmt --check`, pre-commit clippy — clean on changed files.


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Overlay-only wiring of the dead-code read-throttle into the per-peer select loop. Independently verified against origin/main + stellar-core v27 `Peer.cpp`/`FlowControl.cpp`: no-deadlock invariant holds, resume is NOT Notify-only (the 1s periodic tick re-evaluates `can_read()` as a true belt-and-suspenders backstop), release fires exactly once, and the 4 tests genuinely exercise the real gate (2 of them fail-to-compile on main).

<details>

This is cycle 1 (no prior Correctness comment). Complete change-list below; nothing held back.

**Kind: feature** — every new public/internal surface is tested:
- `FlowControlRelease::with_resume_notify` (new `pub(super)`) + `resume_notify` field → exercised by tests 2 & 4 (which call `.with_resume_notify(...)` — these literally do not compile on origin/main, proving the wiring is new).
- The `if flow_control.can_read()`-gated `peer.recv()` arm + `maybe_throttle_read()` call site + `read_resume.notified()` arm → exercised structurally; the `can_read()`/`maybe_throttle_read()` predicates by test 1; `is_flow_controlled_message` self-wedge invariant by test 3.

**No-deadlock (independent trace, verified on the merged tree):**
1. `begin_message_processing` → `lock_local_capacity` decrements `total_capacity` unconditionally for every flow-controlled msg (flow_control.rs:396). `can_read()` = `total_capacity > 0` (byte track `None`→true). So the throttle only engages when ~201 *un-drained SCP* envelopes hold all the total credits.
2. SCP envelopes disarm into a `FlowControlRelease` token riding the bounded `scp_message_tx` (`try_send`, drop-on-full — peer task never blocks; mod.rs:1025). Non-SCP flow-controlled msgs release inline via `finish()` on the peer task (peer_loop.rs:1503-1519), never holding `total_capacity` across an await.
3. Release path = app consumer drops `_flow_release` → `release()` → `end_message_processing` → `total_capacity += 1`. **Nothing in that path reads the throttled socket.** Fetch backfill (TxSet/GeneralizedTxSet/ScpQuorumset/DontHave) is excluded from `is_flow_controlled_message` (flow_control.rs:1226-1234) and rides the separate unbounded `fetch_response` channel (mod.rs:967-991) — never throttled. Verified: no cycle where releasing held credit requires reading the throttled socket.

**Notify lost-wakeup (the highest-risk item) — NOT Notify-only:** the select has 4 arms; `periodic_interval.tick()` (peer_loop.rs:1337) is *unconditional* and fires every 1s. Every select re-entry re-evaluates the `if flow_control.can_read()` precondition. So even if `notify_one()` is lost (fired with no waiter parked), the next ≤1s tick re-enters select and re-evaluates `can_read()` — which is already true after the *first* drain (`total_capacity` 0→1). A healthy-but-throttled peer therefore always resumes within ~1s regardless of the Notify. Resume is belt-and-suspenders, not edge-triggered-only. No permanent-wedge risk.

**Leak-safety / once-only:** `release()` is guarded by `self.message.take()` (mod.rs:464) — second call (or `Drop` after explicit release) is a no-op. The `stop_throttling()`+`notify_one()` block sits *after* the take and *before* the `if !cap.should_send() { return }` early-return (mod.rs:481-486 precede 488), so the throttle lifts even when no SEND_MORE grant is due, and capacity is released exactly once. `Clone` impl sets `flow_release: None` (mod.rs:537-544) — obligation never duplicated.

**num_total_messages / single-peer-resume:** the resume gate `cap.num_total_messages > 0` reuses the already-returned `SendMoreCapacity` field (no recompute); `end_message_processing` sets it only at the 201-batch boundary then resets (flow_control.rs:1080-1083), matching core `FlowControl.cpp:314-318`. Note: the *Notify* fires once per full batch, but `can_read()` flips true after the first single drain, and the periodic tick re-evaluates it — so a single-peer/low-traffic peer that exhausted then drained even one message resumes within ~1s without needing the full 201-batch Notify. Correct.

**Tests genuinely fail on main / exercise the real gate:** confirmed `with_resume_notify`/`resume_notify` absent on origin/main and zero `can_read`/`maybe_throttle_read`/`stop_throttling` callers in `peer_loop.rs` on main → tests 2 & 4 fail-to-compile, tests 1 & 3 assert the new predicates. Test 2 (`test_throttle_lifts_and_read_resumes_after_consumer_drains`) proves recovery via drain alone (no socket read); test 4 proves partial drain does NOT lift the throttle, full-batch drain does. All 5 (incl. existing `test_throttling`) pass locally.

**Build:** `cargo build -p henyey-overlay --all-targets` clean off origin/main-merged head; Clippy/Format/Analyze(rust)/Build(release-ci) green in CI.

No correctness concerns. APPROVE.
</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** Independently re-derived the read-scheduling state machine from stellar-core (`Peer.cpp:293-334`, `FlowControl.cpp:293-346`, `:617-647`). The henyey async translation is behaviorally faithful: throttle-set after `recvMessage` when `!canRead()`; resume gated on `isThrottled() && numTotalMessages > 0` (the 201-batch boundary, not per-message); `canRead()` = byte∧msg capacity. No wire-observable divergence.

<details>

This is cycle 1 (no prior Parity comment). Complete change-list; concern classes labeled.

**`canRead()` parity — exact.** Core `FlowControl.cpp:335-338`: `canRead() = mFlowControlBytesCapacity.canRead() && mFlowControlCapacity.canRead()`. Henyey `flow_control.rs:1096-1099`: `message_capacity.can_read() && byte_capacity.can_read()`, each `total_capacity.map(|c| c>0).unwrap_or(true)`. Byte track `total_capacity: None` → always-true, so the gate is the message-count track — matching core (byte capacity tracks flood bytes, no total cap). No divergence.

**Throttle-set placement — exact.** Core calls `maybeThrottleRead()` AFTER `recvMessage()` (`TCPPeer.cpp:620`/`:770`), setting `mLastThrottle` only inside the `!canRead()` branch (`FlowControl.cpp:617-625`). Henyey calls `flow_control.maybe_throttle_read()` after `handle_received_message` returns (peer_loop.rs:1318); `maybe_throttle_read` sets `last_throttle` only when `!can_read()` (flow_control.rs:1128-1138). Idempotent timestamp overwrite on an already-throttled peer matches core (core also overwrites `mLastThrottle = now`). `[concern-class: parity-gap — none found]`

**Reschedule gate — exact, the load-bearing parity point.** Core `Peer.cpp:314`: `if (mFlowControl->isThrottled() && res.numTotalMessages > 0) { stopThrottling(); scheduleRead(); }`. `numTotalMessages` is nonzero ONLY at `mTotalMsgsProcessed == PEER_READING_CAPACITY` (`FlowControl.cpp:314-318`), then reset — once per full 201-batch, NOT per drain. Henyey `mod.rs:481-486`: `if cap.num_total_messages > 0 && self.flow_control.is_throttled() { stop_throttling(); notify_one(); }`, where `num_total_messages` is set only at the `total_msgs_processed >= peer_reading_capacity` boundary then reset (flow_control.rs:1080-1083). The PR correctly uses the already-returned `SendMoreCapacity` field (no recompute) and does NOT resume-on-any-drain. `[sequencing: correct]`

**`==` vs `>=` boundary — benign, pre-existing.** Core uses `mTotalMsgsProcessed == PEER_READING_CAPACITY`; henyey uses `total_msgs_processed >= peer_reading_capacity`. Since the counter increments by exactly 1 per message and resets to 0 at the boundary, it can never exceed 201 without resetting, so `==` and `>=` are operationally equivalent. This predates this PR (existing `end_message_processing`), is not touched by the diff, and is not a Phase-2 concern. `[edge-case: equivalent]`

**scheduleRead async equivalence — faithful.** Core's `scheduleRead` posts a per-peer re-read task (and early-returns if still `isThrottled()`, `TCPPeer.cpp:515`). Henyey re-enables the conditional `peer.recv()` arm, woken by the consumer-drain `Notify` AND the unconditional 1s periodic tick. Observable behavior — stop reading at 0 total capacity, resume after a full batch is consumed — matches core. The 1s backstop is a latency nuance, not a wire-observable divergence: peers observe SEND_MORE cadence (owned by Phase 1, unchanged here); Phase 2 changes only WHEN WE READ.

**Self-wedge / fetch exclusion — parity-correct.** `is_flow_controlled_message` = `Transaction | ScpMessage | FloodAdvert | FloodDemand` (flow_control.rs:1226-1234), excluding TxSet/GeneralizedTxSet/ScpQuorumset/DontHave — these ride the unbounded `fetch_response` channel and are never throttled, mirroring core servicing fetch directly off the receive path. A throttled SCP socket cannot starve the fetch the consumer depends on.

**PARITY_STATUS update** flips `canRead()` Partial→Full with an accurate Phase-2 description. Correct.

</details>

PR #3658, opened 2026-06-26
Add opt-in harness-side loadgen fast-fail for MaxTPS searches

tomerweller:
## Motivation

In MaxTPS binary searches, *failing* steps dominate wall-clock and each one pays a fixed completion-wait tail: after the loadgen submits all txns, the node waits ~20 ledgers (`LoadGenerator::TIMEOUT_NUM_LEDGERS`) for them to apply before reporting failure. At a ~5 s close time that's ~100 s of dead wait on every failing step (it was ~150 s before #3657 fixed the 30→20 divergence). The failure is knowable much sooner — once submission completes, ledgers advancing without the run completing *is* the apply-lag signal.

## Change

`RunMultiLoadgen` gains an opt-in early-abort: once **all** peers have submitted their full tx count, it snapshots the ledger; if the run hasn't completed within `SSC_LOADGEN_FASTFAIL_LEDGERS` more ledger closes, it stops the loadgen and fails the step immediately.

- **Default-off**: env unset or `<=0` ⇒ exact current behavior (wait for the node to self-report). No existing mission changes.
- **Node loadgen untouched** — byte-identical to stellar-core. This is purely a harness-side early-abort, using only existing signals (`loadgen.txn.attempted` for submission-complete + ledger number).
- No applied-tx meter exists, so "submitted but not complete after K ledgers" is the apply-lag proxy.

## Status

⚠️ Compile-verified (`dotnet build … Release` → 0 errors) but **not yet validated on a live cluster** — needs a MaxTPSClassic run with e.g. `SSC_LOADGEN_FASTFAIL_LEDGERS=5` to confirm it fast-fails doomed steps without false-failing healthy ones (and to tune the default K). Holding for that before merge.

Companion to #3657 (the 30→20 node-side parity fix).

### DIFF
--- a/vendor/supercluster/src/FSLibrary/StellarStatefulSets.fs
+++ b/vendor/supercluster/src/FSLibrary/StellarStatefulSets.fs
@@ -475,6 +475,33 @@ type StellarFormation with
         for (peer, peerSpecificLoadgen, offset) in peerLoadGens do
             LogInfo "Loadgen: %s with offset %d" (peer.GenerateLoad peerSpecificLoadgen) offset
 
+        // Opt-in harness-side fast-fail. Once every peer has SUBMITTED its full
+        // tx count, the only remaining work is on-ledger application. If the run
+        // has not completed within SSC_LOADGEN_FASTFAIL_LEDGERS additional
+        // ledger closes, the offered rate exceeds apply capacity, so abort early
+        // instead of waiting out the node-side wait-till-complete budget
+        // (stellar-core LoadGenerator::TIMEOUT_NUM_LEDGERS = 20). This keeps the
+        // node loadgen byte-identical to core — it is purely a harness-side
+        // early-abort that turns a doomed step's fixed completion-wait tail
+        // (~20 ledgers ≈ 100s) into ~K ledgers, which matters on failure-
+        // dominated binary searches (e.g. MaxTPSClassic). Unset or <=0 preserves
+        // the default behavior (wait for the node to self-report).
+        let fastFailLedgers =
+            match System.Environment.GetEnvironmentVariable "SSC_LOADGEN_FASTFAIL_LEDGERS" with
+            | null -> 0
+            | s ->
+                match System.Int32.TryParse s with
+                | true, n -> n
+                | _ -> 0
+
+        let mutable submitDoneLedger: int option = None
+
+        let allLoadSubmitted () =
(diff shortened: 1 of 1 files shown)


tomerweller:
## ✅ Live validation passed

Ran `MissionMaxTPSClassic` on a fresh nsc 32×64 VM with this branch, `SSC_LOADGEN_FASTFAIL_LEDGERS=10`, henyey `ssc-fix6`, 60s load windows, search `[100,300]`.

| rate | result | duration | ended via |
|-----:|--------|---------:|-----------|
| 200 | ✅ pass | 1m30s | completed (armed → succeeded +23s) |
| 250 | ❌ fail | 1m53s | **fast-fail** (10 ledgers) |
| 225 | ✅ pass | 1m29s | completed (armed → succeeded +24s) |
| 237 | ❌ fail | 1m55s | **fast-fail** (10 ledgers) |
| 231 | ❌ fail | 1m54s | **fast-fail** (10 ledgers) |

**Converged max = 225** — identical to the no-fast-fail reference run (same per-step pass/fail). The early-abort does not change the answer.

### Criteria
- **(a) No false-fails** — every passing rate (200, 225 incl. the boundary/max) still passed. Healthy steps complete ~5 ledgers (~23–24s) after submission; the 10-ledger budget leaves a comfortable ~5-ledger margin. ✓
- **(b) Failing steps abort early via the fast-fail path** — all three (250/237/231) ended with `Loadgen fast-fail: ... not complete within 10 ledgers`, each ~1m54s vs ~3m30s without it = **~1m40s saved per failing step** (~47% faster). The abort fires ~47–49s after arming (10 ledgers × ~4.8s close). ✓
- **(c)** `Fast-fail armed` logged on every step. ✓
- **(d)** Max unchanged (225). ✓

Whole run ~14min vs ~21min for the no-fast-fail short-duration run; savings scale with the failing-step fraction (failure-dominated searches benefit most).

### Default-K guidance
`K=10` worked cleanly here. Since healthy steps settle in ~5 ledgers, a safe default is **K ≈ 2× the expected post-submission settle** (~10 ledgers). It's a ledger count (not wall-clock), so it self-scales with close time. Recommend keeping it env-gated/opt-in as implemented; a documented default of 10 for MaxTPS missions is reasonable.

Validated on instance `3oh6jksdet6l4` (since destroyed).

Issue #3478, opened 2026-06-18
OFFLINE: validator aborted on ENOSPC (disk full) — bucket-flush fatal abort


tomerweller:
## OFFLINE: validator aborted on ENOSPC (disk full) — bucket-flush fatal

**Session:** 74535976 (mainnet validator) · **Build:** `46326610` · **Severity:** validator-blocking (consensus lost ~9 min)

### Symptom
At **2026-06-18T19:37:55Z** the validator hit a fatal persist failure and aborted:
```
ERROR henyey_app::app::persist: Fatal persist failure, aborting context="bucket flush" \
  error=flush_pending_persist: bucket 06be1eeaa2b50e267dded72e2d7b85e39e66c8ee1d0972f4b01ddad0bcdca0bf: \
  IO error: No space left on device (os error 28)
```
The node was closing ledgers normally right up to the abort (last closed L63089981 at 19:37:55Z), so this is **not state corruption** — it is a transient disk-full (`errno 28`) on a single bucket flush that escalated to `trigger_fatal_shutdown`. Process died; consensus participation lost.

### Root cause
The node runs on a **shared 1.7T volume** (`/home/tomer/data`, `~/data/<session>/` per CLAUDE.md). Neighbor sessions filled the volume to ~100%:
```
196G  ~/data/a6f14d79
 46G  ~/data/mainnet      (validator's own data)
 39G  ~/data/649ac318-...
 38G  ~/data/20260617-...
 ... (several 25-39G neighbor dirs)
```
My session is only **3.1G** (bounded by archive/log retention). The monitor had been tracking the climbing `disk=` watch value across the prior ~4 ticks (96% → 98% → 99%), but the volume crossed 100% between ticks and the validator's bucket flush failed before the next tick observed it. By the time the monitor caught the dead process (~19:46Z), neighbors had freed space (101G free) — confirming it was a **transient** exhaustion.

### Recovery (done by monitor this tick)
Rotated `monitor.log.crashed-*`, relaunched on the same binary (disk had recovered to 100G free). New PID 2306283, restoring from HAS → catching up from L63089981 (~10 min stale). No state wipe (correctly — `detect_crash_state` saw recent_crashed=0, hash_mismatch=no; ENOSPC is environmental, a wipe would not help). Cumulative outage ≈ 9 min + catchup.

### Two hardening angles (operator decision)
[...]

PR #3688, opened 2026-06-30
Move bucket-download fsync off tokio worker threads

### LINKED ISSUE #3686: Bucket download fsync (atomic_write_bytes) runs on tokio worker threads, stalls runtime under disk pressure
## Summary

`download_and_save_bucket` (async) calls the synchronous `atomic_write_bytes`, which performs blocking `write_all` + `file.sync_all()` (fsync) + parent-directory `sync_all()`. It is awaited inside a `buffer_unordered(MAX_CONCURRENT_DOWNLOADS)` stream, so up to `MAX_CONCURRENT_DOWNLOADS` **fsyncs execute directly on tokio worker threads**. Under disk pressure this stalls the runtime scheduler for all concurrent work on those workers.

## Location

- `crates/historywork/src/download.rs:169` — async `download_and_save_bucket` calls `atomic_write_bytes`
- `crates/common/src/fs_utils.rs:49` and `:105` — `atomic_write_bytes` does blocking `write_all` + `file.sync_all()` + parent-dir `sync_all()`

The call is driven concurrently via `buffer_unordered(MAX_CONCURRENT_DOWNLOADS)`, multiplying the number of simultaneous blocking fsyncs on the executor.

## Why it matters

`file.sync_all()` is a blocking syscall that can take many milliseconds (or longer under disk pressure). Running it directly on a tokio worker thread blocks that worker from polling any other futures. With `MAX_CONCURRENT_DOWNLOADS` of these in flight, a large fraction of the runtime's worker pool can be stalled simultaneously, delaying unrelated async work (overlay, RPC, timers).

## Proposed fix

Offload the blocking write to `tokio::task::spawn_blocking` (passing owned bytes + path into the closure), or switch to `tokio::fs`. `spawn_blocking` is the established pattern elsewhere in the codebase (SQLite via r2d2, disk-backed bucket reads).

## Related (not in scope for this issue, file separately if desired)

Same blocking-on-executor class, lower severity, in the history fetch paths:
- `crates/history/src/cdp.rs:359` / `:394` — synchronous `std::fs::read` / `std::fs::write` under `buffer_unordered(prefetch_count)`
- `crates/history/src/cdp.rs:399` — CPU-bound zstd decompress + XDR parse on the executor
- `crates/historywork/src/download.rs:203` — blocking `create_dir_all`

## Severity

HIGH — runtime-wide scheduling stalls under disk pressure during bucket download/catchup. Found during a workspace-wide concurrency audit.

tomerweller:
Closes #3686

## Summary

Bucket persistence (`atomic_write_bytes`) performs blocking `fsync` (the temp file plus the parent directory). Two async fns called it **directly while polled on a tokio worker thread**, fanned out up to `MAX_CONCURRENT_DOWNLOADS` (16) wide via `.buffer_unordered` — so under disk pressure up to 16 concurrent fsyncs blocked the worker threads and starved the scheduler (overlay/RPC/timers).

This wraps the persist at **both** async worker-poll call sites in `spawn_blocking` via the existing `henyey_common::spawn_blocking_logged`, leaving the shared `atomic_write_with`/`atomic_write_bytes` primitive and all its synchronous / `block_on_async` / `block_in_place` callers unchanged.

- `crates/historywork/src/download.rs`: extracted `pub async fn persist_bucket_to_disk(PathBuf, Vec<u8>) -> io::Result<()>` (wraps `spawn_blocking_logged(atomic_write_bytes)`) and call it from `download_and_save_bucket`. The `JoinError` (closure panic) and inner `io::Error` both flatten into the existing `BucketDownloadFailure::Persist` with **unchanged** error text.
- `crates/history/src/catchup/download.rs` (`download_buckets`): captured `let nbytes = data.len();` **before** moving `data` into the closure (the post-write `debug!` still needs it — the exact E0382 the planning critic flagged), then `warn!`+`continue` on either error arm with **unchanged** text.
- `crates/history/src/catchup/buckets.rs`: added clarifying comments at the 3 `block_in_place`-resident `atomic_write_bytes` sites noting why they are intentionally NOT `spawn_blocking`-wrapped (the worker is already released).

Behavior-preserving: bytes written, fsync durability, hash verification, per-bucket metrics, and error outcomes are identical; only the executing thread changes. No parity impact.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3686#issuecomment-4838928718)

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy -p henyey-historywork -p henyey-history --tests -- -D warnings` (clean, no new warnings)
- [x] `cargo test -p henyey-historywork` passes (incl. new `download_fairness.rs`)
- [x] `cargo test -p henyey-history download` passes
- [x] `cargo build -p henyey-historywork -p henyey-history` clean

- **Test:** `crates/historywork/tests/download_fairness.rs::test_blocking_persist_shape_starves_then_not` (+ `test_persist_bucket_to_disk_runs_off_worker` driving the real helper).
- **Pre-fix:** committed as `593534b0` — verified FAILED against unmodified `origin/main` historywork src: `error[E0432]: unresolved import henyey_historywork::persist_bucket_to_disk` (the off-worker persist helper does not exist on main).
- **Mechanism proof:** single-worker tokio runtime + 10 ms async heartbeat; a 16-wide ~200 ms blocking persist window. INLINE-on-worker shape (origin/main) → heartbeat froze at **1 tick** (starved); off-worker `spawn_blocking_logged` shape (the fix) → heartbeat advanced to **19 ticks**. Deterministic via `std::thread::sleep` (host-disk-speed-independent).
- **Post-fix:** verified PASSES after `a61f64d9`.

## Deviations from plan

- The plan suggested the regression test could either drive the real `DownloadBucketsWork` via a local axum archive server **or** use an extracted persist helper instrumented with a deterministic sleep. I used the latter (the plan-endorsed alternative): a single-worker starvation harness whose blocking work is a deterministic `std::thread::sleep` standing in for fsync latency, because a real small-bucket fsync completes too fast on the test host to reproduce starvation reliably. The harness drives the production `spawn_blocking_logged` shape and additionally exercises the real `persist_bucket_to_disk` helper 16-wide. The 200ms-vs-10ms margin (≤2 ticks inline / ≥5 ticks off-worker) avoids flake.
- The persist driver in the test is run inside a `tokio::spawn` (not directly under `block_on`) so it is polled on the single worker thread, matching production where `DownloadBucketsWork::run` is a spawned work item. Awaiting the stream directly under `block_on` would run it on the external calling thread and could never starve the worker-resident heartbeat.

### DIFF
--- a/crates/history/src/catchup/buckets.rs
+++ b/crates/history/src/catchup/buckets.rs
@@ -350,6 +350,14 @@ impl CatchupManager {
             // Save XDR data to disk first, then build the disk-backed bucket by
             // streaming through the file. This avoids holding the full file in memory
             // while also building the index — critical for multi-GB buckets on mainnet.
+            //
+            // NOTE (#3686): this `atomic_write_bytes` runs under
+            // `block_on_async`/`block_in_place` (synchronous context, not a
+            // tokio-worker-polled async fn), so it is intentionally NOT wrapped
+            // in `spawn_blocking` — the worker is already released via
+            // `block_in_place`. Only the async `.buffer_unordered` download
+            // call sites (historywork/download.rs, catchup/download.rs) need the
+            // off-worker wrap.
             if let Err(e) = atomic_write_bytes(&bucket_path, &xdr_data) {
                 if !was_preloaded {
                     // Persistence failure on the freshly-downloaded path is a
@@ -508,6 +516,12 @@ impl CatchupManager {
             let hot_archive_bucket_cache: Mutex<HashMap<Hash256, HotArchiveBucket>> =
                 Mutex::new(HashMap::new());
 
+            // NOTE (#3686): the `atomic_write_bytes` calls inside this closure
+            // (and the live-bucket loader above) run synchronously under the
+            // `apply_buckets` `block_in_place` region — not on a tokio worker
+            // polling an async fn — so they are intentionally NOT
+            // `spawn_blocking`-wrapped. Only the async `.buffer_unordered`
+            // download call sites need the off-worker wrap.
             let load_hot_archive_bucket =
                 |hash: &Hash256| -> henyey_bucket::Result<HotArchiveBucket> {
                     // Short-circuit sentinel hashes (zero hash → empty bucket)
--- a/crates/history/src/catchup/download.rs
+++ b/crates/history/src/catchup/download.rs
@@ -239,8 +239,30 @@ impl CatchupManager {
                                     );
                                     continue;
                                 }
-                                // Save to disk atomically
-                                if let Err(e) = atomic_write_bytes(&bucket_path, &data) {
+                                // Save to disk atomically, off the tokio worker
+                                // thread. `atomic_write_bytes` issues blocking
+                                // `fsync` calls; running them inline on the
+                                // worker that polls this future starves the
+                                // runtime when up to MAX_CONCURRENT_DOWNLOADS
+                                // (16) of these futures are fanned out via
+                                // `.buffer_unordered` under disk pressure
+                                // (#3686). Capture `data.len()` BEFORE moving
+                                // `data` into the blocking closure (the
+                                // `debug!` below still needs it).
+                                let nbytes = data.len();
+                                let write_path = bucket_path.clone();
+                                let write_res = match henyey_common::spawn_blocking_logged(
+                                    "catchup-save-bucket",
+                                    move || atomic_write_bytes(&write_path, &data),
+                                )
+                                .await
+                                {
+                                    Ok(inner) => inner.map_err(|e| e.to_string()),
+                                    Err(join_err) => {
+                                        Err(format!("persist task failed: {join_err}"))
+                                    }
+                                };
+                                if let Err(e) = write_res {
                                     warn!("Failed to save bucket {} to disk: {}", hash, e);
                                     continue;
                                 }
@@ -250,7 +272,7 @@ impl CatchupManager {
                                 if count % PROGRESS_REPORT_INTERVAL == 0 || count == total_to_download as u32 {
                                     info!("Downloaded {}/{} buckets", count, total_to_download);
                                 }
-                                debug!("Pre-downloaded bucket {} ({} bytes)", hash, data.len());
+                                debug!("Pre-downloaded bucket {} ({} bytes)", hash, nbytes);
                                 return Ok(archive.name().to_owned());
                             }
                             Err(e) => {
--- a/crates/historywork/src/download.rs
+++ b/crates/historywork/src/download.rs
@@ -166,11 +166,46 @@ async fn download_and_save_bucket(
         BucketDownloadFailure::Verify(format!("bucket {hash} hash mismatch: {err}"))
     })?;
 
-    atomic_write_bytes(bucket_path, &data).map_err(|e| {
-        BucketDownloadFailure::Persist(format!("failed to save bucket {hash} to disk: {e}"))
-    })?;
+    // Persist off the tokio worker thread. `atomic_write_bytes` issues blocking
+    // `fsync` calls (file + parent dir); running them inline on the worker that
+    // polls this future starves the scheduler when up to
+    // `MAX_CONCURRENT_DOWNLOADS` (16) of these futures are fanned out via
+    // `.buffer_unordered`. See #3686. `data` has no post-write use here, so the
+    // owned `Vec<u8>` + `PathBuf` move cleanly into the blocking closure.
+    persist_bucket_to_disk(bucket_path.to_path_buf(), data)
+        .await
+        .map_err(|e| {
+            BucketDownloadFailure::Persist(format!("failed to save bucket {hash} to disk: {e}"))
+        })?;
 
     Ok(())
+}
+
+/// Persist downloaded bucket bytes to disk off the tokio worker thread.
+///
+/// `atomic_write_bytes` performs blocking `fsync` (both the temp file and the
+/// parent directory) for crash durability. When invoked directly from an async
+/// fn polled on a tokio worker thread — and fanned out up to
+/// `MAX_CONCURRENT_DOWNLOADS`-wide via `.buffer_unordered` — these fsyncs block
+/// the worker threads and starve the runtime under disk pressure (#3686). This
+/// helper moves the write onto the blocking pool via
+/// [`henyey_common::spawn_blocking_logged`], freeing the worker to keep polling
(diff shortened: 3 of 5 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The fsync offload is correct — sequencing (durable-on-disk before counted/returned), error contract, and the `data.len()`/`nbytes` move-order are all preserved. The regression test genuinely discriminates worker starvation (both arms in one run) and drives the shipped helper 16-wide. Verified against `origin/main`; tests pass; clippy/fmt clean.

<details>

Reviewed the PR head (`a61f64d9`) against `origin/main`; built and ran the new test in a clean worktree.

**Sequencing (correct).** In both call sites the persist `.await` completes BEFORE the success bookkeeping:
- `historywork/download.rs`: `persist_bucket_to_disk(...).await?` precedes `Ok(())`; on persist error the `?` short-circuits to `BucketDownloadFailure::Persist` — bucket is never counted as downloaded.
- `catchup/download.rs`: `write_res` is awaited; on `Err` it `warn!`+`continue` (skips the bucket, no count/return); only on `Ok` does `downloaded.fetch_add` and `return Ok(archive.name())` run. The bucket is durably fsync'd before it is counted or its archive name returned.

**Error contract (preserved).**
- `catchup`: warn! text is byte-identical to main (`"Failed to save bucket {} to disk: {}"`); inner `io::Error` flows through `map_err(|e| e.to_string())` into the same `{e}` slot. A closure panic now becomes `"persist task failed: {join_err}"` → warn+continue (skip) instead of an unwound panic. That is strictly more robust and behaviorally acceptable for a per-bucket loop.
- `historywork`: inner `io::Result` passed through unchanged; `JoinError` → `io::Error::other(...)` → `BucketDownloadFailure::Persist(format!("failed to save bucket {hash} to disk: {e}"))` — identical downstream classification to the old inline path.

**Move-order / E0382 (correct).** `let nbytes = data.len()` (catchup) and the owned `data: Vec<u8>` move (historywork) are captured/sequenced before `data` moves into the blocking closure; the post-write `debug!` uses `nbytes`. The release build compiled, confirming no use-after-move.

**Runtime-context safety.** `spawn_blocking` requires a tokio runtime. Both wrapped sites are inside async fns awaited through `.buffer_unordered` — `DownloadBucketsWork::run` (Work scheduler) and the catchup pre-download state machine — both driven on the runtime. No sync/`block_on` caller reaches the wrapped helper. (concern class checked: `error-handling`, `api-shape` — clean.)

**buckets.rs block_in_place sites (correctly left unwrapped).** The 3 commented sites run synchronously under `block_on_async`/`block_in_place`; the worker is already released, so `spawn_blocking` there would be redundant/incorrect. Comments-only change.

**Test verification (bug-fix kind).** `test_blocking_persist_shape_starves_then_not` asserts BOTH arms in one process with the identical 200ms blocking sleep: inline-on-worker ≤2 ticks (starvation — proves the harness detects the bug), off-worker (`spawn_blocking_logged`) ≥5 ticks (the fix). This is genuine behavioral discrimination, not merely an import-compile failure on main. `test_persist_bucket_to_disk_runs_off_worker` drives the REAL `persist_bucket_to_disk` 16-wide and asserts bytes land + heartbeat alive — guards against a future revert to inline. Ran locally: both pass (3.48s).

**Yield-point note (non-blocking, refuted internally).** `spawn_blocking_logged`'s docstring recommends a split-spawn-with-yield for hot-path fan-out. Here each of the 16 futures does `spawn_blocking_logged().await`; the `.await` on the JoinHandle is itself a suspension point — while the fsync runs on the blocking pool the worker is freed to poll other tasks. The docstring caveat targets tight loops where the spawned task is instantly-ready and never yields; an ~ms fsync always suspends. Not a concern.

clippy `-p henyey-historywork --tests`: clean. `cargo fmt --check`: clean. `crates/common/src/fs_utils.rs` unmodified (durability semantics unchanged).
</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** No regression, perf, API-break, or operational risk. The persist path's observable behavior (atomic temp+rename+fsync durability, ENOSPC handling, error stringification) is unchanged; only the thread on which the fsync runs moved. New public API is additive and test-only. Risk profile is strictly positive (removes a runtime-stall failure mode).

<details>

Risk lens, evaluated fresh against `origin/main`.

**regression-risk (none).** `atomic_write_bytes`/`atomic_write_with` in `crates/common/src/fs_utils.rs` are UNMODIFIED (verified `git diff origin/main` empty for that file). The temp-file → `sync_all` → durable rename sequence is byte-for-byte the same; durability semantics are preserved. No sync (non-async) caller was rerouted — the wrap exists only at the two async `.buffer_unordered` download sites; the 3 `block_in_place` persist sites in `buckets.rs` are explicitly left calling `atomic_write_bytes` inline (comments-only), so the apply-buckets path is unchanged.

**ENOSPC / disk-error classification (#3478 — out of scope, unaffected).** The recoverable-vs-fatal ENOSPC classification (`crates/bucket/src/error.rs::is_transient_io`) lives on the bucket merge/flush path, NOT this download path. This download path stringifies any `io::Error` into `BucketDownloadFailure::Persist` (historywork) or warn+continue (catchup) exactly as before — no observable change to disk-error handling. The PR neither touches nor needs to touch that classification.

**perf-regression (none / positive).** Moving fsync to the blocking pool adds a task hand-off per bucket (negligible vs a multi-ms fsync) and removes the runtime-wide scheduler stall that was the entire motivation. Net positive. The blocking pool is tokio's default (512-thread cap), so 16 concurrent persists are comfortably absorbed without exhausting it; overlay/RPC/timer tasks keep being polled.

**api-break (none).** `pub use download::persist_bucket_to_disk` adds one additive public fn to `henyey-historywork`. No existing signature changed; `download_and_save_bucket` stays private. The new export exists for the integration test but is harmless as public surface (documented, single-purpose).

**migration-risk / config-risk (none).** No config, schema, on-disk format, or CLI changes. Bucket file contents and the atomic-rename target paths are identical.

**security (none).** No new external input, no unsafe, no new file paths beyond the existing bucket dir.

**operational (positive).** Under disk pressure during catchup, the prior code could stall the whole runtime (overlay heartbeats, consensus timers). This eliminates that class of stall — aligns with the project's standing rule against blocking work on the event loop / worker threads. Test deps (`tempfile` dev-dep, `futures`/`tokio`/`henyey-common` normal deps) are correctly placed; integration test builds and passes.
</details>

PR #3691, opened 2026-06-30
Fire consensus trigger timer once per ledger, not in a busy-loop

tomerweller:
## Problem

The `TriggerNextLedger` timer handler (`handle_scp_timer_event`) re-armed itself **unconditionally** after every fire:

```rust
self.try_trigger_consensus().await;
self.setup_trigger_next_ledger().await;  // <-- re-arm on every fire
```

While a slot is still being nominated, LCL has not advanced, so `next_slot` is unchanged and `prepare_start(last_slot) + expectedClose` is already in the past. `setup_trigger_next_ledger` then computes `delay == 0`, so the timer fires again on the next event-loop iteration — and re-arms again. The result is a **busy-loop at ~hundreds of Hz** that runs for the entire nomination→externalize window of *every* ledger.

Each spin iteration is not free: it sends a timer event through the bridge channel, wakes the main `select!` loop, and dispatches a `spawn_blocking("trigger_next_ledger")` that takes herder locks before hitting the `AlreadyNominating` guard. This starves the **single event loop** that also services inbound SCP nomination envelopes, which is exactly what slows (and makes bimodal) nomination convergence.

Diagnostic logging on a live 23-node run measured the trigger timer firing **~25,000 times per ledger per node** on the affected slots.

## Fix

Make the trigger timer **single-shot per ledger**, matching stellar-core, which arms `mTriggerTimer` only from `lastClosedLedgerIncreased`. The handler no longer re-arms; the next trigger is armed by the existing post-close path (`setup_trigger_next_ledger` after a ledger closes), and the 1-second maintenance-tick `try_trigger_consensus` remains as the backstop for any gated/never-armed edge case.

## Measurement

`MissionMaxTPSClassic` on a single nsc 32×64 VM, 23-node tier-1, short-probe:

| | max tx/s |
|---|---|
| baseline (spin present) | 1305 |
| **this fix** | **1410** |

**+8.0%**, clearing the >5% accept bar. After the fix, the per-pod trigger-fire count dropped from ~25,000/ledger to **exactly 1/ledger**, and the trigger timer fires punctually (fire-lateness mean ~1.3 ms). henyey moves from ~85% to ~92% of stellar-core's ~1531 on the same rig.

## Parity

The trigger timer is on the divergeable surface (internal scheduling, `docs/PARITY.md`). Observable nomination-start timing is unchanged — the fix only removes redundant re-fires that all hit the `AlreadyNominating` guard. This actually moves henyey *closer* to stellar-core's one-shot-per-ledger trigger model.

## Tests

- `cargo test -p henyey-app --lib app::consensus:: app::lifecycle::` — 83 pass.
- New regression `test_trigger_event_does_not_rearm_same_slot_spin` reproduces the spin precondition and asserts no self-perpetuating trigger is delivered. Fails pre-fix (1 spurious trigger), passes post-fix.

https://claude.ai/code/session_01AxnZVoi2JQ2NDqo3Hnm4T9

### DIFF
--- a/crates/app/src/app/consensus.rs
+++ b/crates/app/src/app/consensus.rs
@@ -1529,11 +1529,18 @@ impl App {
             self.consensus_trigger_timer_fires
                 .fetch_add(1, Ordering::Relaxed);
             self.try_trigger_consensus().await;
-            // Re-arm the trigger for the following slot so the event-driven
-            // schedule continues even if this attempt did not (yet) externalize
-            // (e.g. multi-node nomination still in flight, or a gated skip).
-            // setup_trigger_next_ledger is idempotent w.r.t. the per-slot timer.
-            self.setup_trigger_next_ledger().await;
+            // Single-shot per ledger (parity: stellar-core arms mTriggerTimer
+            // only from `lastClosedLedgerIncreased`). We deliberately do NOT
+            // re-arm here. While a slot is still being nominated, LCL has not
+            // advanced, so `next_slot` is unchanged and `prepare_start(last) +
+            // expectedClose` is already in the past — a re-arm would compute
+            // `delay == 0` and the timer would fire again immediately, spinning
+            // at hundreds of Hz for the whole nomination→externalize window and
+            // starving the event loop that services inbound SCP envelopes
+            // (maxtps: slow/bimodal nomination convergence). The next trigger is
+            // armed by the post-close path (`setup_trigger_next_ledger` after a
+            // ledger closes); the 1-second maintenance-tick
+            // `try_trigger_consensus` backstops any gated/never-armed case.
             // Signal the caller to pump the close pipeline for a possible
             // self-externalized slot (see the doc comment).
             return true;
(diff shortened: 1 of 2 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Single-line behavioral change (remove the unconditional handler re-arm) is correct, well-commented, and matches stellar-core. Regression test verified to fail pre-fix (1 spurious trigger) and pass post-fix. No liveness loss — next trigger is armed by every close path; the 1s maintenance tick is an unconditional backstop.

<details>

Cycle 1 — complete change-list (no concerns; APPROVE).

**Liveness (regression-risk) — VERIFIED, no concern.**
- The handler at `consensus.rs:1528-1543` still calls `self.try_trigger_consensus().await` for the current slot before returning `true`, so the current trigger still fires — only the redundant re-arm is removed.
- Next-trigger arming on every ledger-close path confirmed on `origin/main`:
  - select-loop close completion: `lifecycle.rs:684` → `setup_trigger_next_ledger().await`.
  - buffered-burst apply path: `ledger_close.rs:1929` (guarded by `if success`) → `setup_trigger_next_ledger().await`.
  - cold start: `lifecycle.rs:550`.
- 1s maintenance backstop confirmed: `consensus_interval` is `tokio::time::interval(Duration::from_secs(1))` with default Burst behavior (NOT `Skip`); the `try_trigger_consensus().await` at `lifecycle.rs:1250` is **unconditional** within the tick branch (not inside the `if self.is_validator` block, which only wraps `maybe_publish_history`). So it fires ~every 1s regardless of trigger-timer state. `try_trigger_consensus` self-gates on the same prepareStart+expectedClose+ctValidityOffset boundary, so it can never fire earlier than the timer would; worst case for a never-armed/gated edge is 1s late, never stalled. No wedge scenario: the only way to stall is the event loop itself halting, which this change strictly *improves* (removes the hundreds-of-Hz spin that was starving it).

**Test verification (bug-fix kind) — PASS.**
- PR body has a `## Tests` section naming the regression test.
- Test+fix are in one commit (8d31c9ee), so I reconstructed the before/after by toggling the re-arm in a worktree:
  - Re-arm restored (pre-fix): `test_trigger_event_does_not_rearm_same_slot_spin` FAILS — `left: 1, right: 0` (exactly 1 spurious self-perpetuating trigger). Matches the PR's "fails pre-fix (1 spurious trigger)" claim.
  - PR head (post-fix): PASSES.
- Precondition validity confirmed: `herder.bootstrap(1)` advances tracking to slot 2 without recording a `prepare_start`, so `prepare_start(1).is_none()` holds → trigger-time gate open → `setup_trigger_next_ledger` clamps `base_delay` to 0 → pre-fix re-arm schedules a zero-delay slot-2 timer. The test asserts on the real `scp_timer_rx` queue. Solo node never externalizes (event loop not running in the unit test), so LCL stays at 1 and `next_slot` stays 2 — the exact spin condition.

**Readability/error-handling:** the replacement comment block is accurate and cites the core arming invariant. No error paths affected.
</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** This is divergeable internal scheduling with NO observable consensus/hash/XDR/wire change. Independently confirmed against the pinned stellar-core that the trigger timer is one-shot per ledger (armed only from `lastClosedLedgerIncreased` + startup; the `triggerNextLedger` handler never re-arms). The removed re-fires all hit the `AlreadyNominating` guard, so observable nomination-start timing is unchanged. The fix moves henyey closer to core.

<details>

Cycle 1 — complete change-list (no concerns; APPROVE).

**Core one-shot model — independently confirmed (parity-gap: none).**
Note: the `stellar-core/` submodule is actually pinned at **v27.0.0** (`7696c069`), not v26.0.1 as CLAUDE.md states — but the trigger-timer mechanics are identical in both and the cited line numbers match.
- `HerderImpl::setupTriggerNextLedger()` (HerderImpl.cpp:1270) is the ONLY function that calls `mTriggerTimer.async_wait(... triggerNextLedger ...)` (cpp:1329/1333).
- `setupTriggerNextLedger` is invoked from exactly two places: startup `start()` (cpp:241) and `lastClosedLedgerIncreased` post-close (cpp:1259). Both are post-close / startup — never from the handler.
- The `triggerNextLedger(...)` handler body (cpp:1481–1675) contains **zero** references to `mTriggerTimer` or `setupTriggerNextLedger` (grep count = 0 over the whole function). It nominates and returns; it does not re-arm.
- Conclusion: core is one-shot-per-ledger; the re-arm is performed only when LCL increases. henyey's pre-fix handler re-arm was a henyey-only divergence; removing it restores parity.

**Observable-surface analysis (no divergence).**
- The trigger timer governs *internal scheduling* only (when nomination is attempted), explicitly listed as divergeable in `docs/PARITY.md`.
- The removed re-fires occurred while a slot was already nominating → every one hit the herder `AlreadyNominating` guard and was a no-op. They produced no SCP envelope, no ballot, no hash, no XDR, no wire byte, no archive entry. So the observable nomination-START timing is byte-identical before/after. Nothing a peer, Horizon, stellar-rpc, or an archive consumer could observe changes.

**Test validity (test-coverage) — confirmed reproduces the spin precondition and proves no self-perpetuating trigger.**
- Reproduces the precondition: cold-start `bootstrap(1)` → tracking, `prepare_start(1)==None` → trigger-time gate open → solo node does not externalize without the event loop, so LCL/`next_slot` stay put — the exact same-slot re-arm condition.
- Proves the property: asserts zero `TriggerNextLedger` events left on `scp_timer_rx` after handling one trigger.
- Empirically: FAILS pre-fix (1 spurious trigger), PASSES post-fix (verified by toggling the re-arm in a build).

**Docs note** (`docs/maxtps-optimization/2026-06-30-target-core.md`): documentation only, no code/parity surface.
</details>

tomerweller:
### Verified with a controlled clean A/B (short-probe)

Re-measured with **both images built from current `origin/main`**, differing only by this commit (no diagnostic instrumentation in either), same nsc 32×64 instance, same band, `SSC_MAXTPS_TXS_MULTIPLIER=60`:

| image | max tx/s |
|---|---|
| `clean-base` (current main, spin present) | 1350 |
| `clean-fix` (main + this PR) | **1437** |

**Δ = +6.4%**, consistent with the earlier instrumented run (1305 → 1410, +8%; that baseline predated #3690 and carried diagnostic logging). The clean controlled comparison confirms the gain is real and reproducible. The accept bar (>5%) is cleared on the skill's short-probe methodology.

(Note: a longer 5-min-window run sustains a few % lower for *both* images — the known short-probe upward bias — so the relative +6.4% holds; absolute sustained max is ~1400.)

Issue #2317, opened 2026-05-03
Quickstart: tx finalization timeout introduced by c6118f2c overlay SCP dedup (#2311) — breaks single-validator standalone network


tomerweller:
## Symptom

Quickstart workflow's `build / 3 test / 2 test (testing-with-pr, ...)` jobs time out waiting for transactions to finalize when henyey is used as core in stellar/quickstart Docker test harness. Failure reproduces on **two consecutive shas** (regression bisected to `c6118f2c`):

| sha | Quickstart |
|---|---|
| 7edb540e (parent) | ✅ pass |
| 58bb3738 (parent-1) | ✅ pass |
| **c6118f2c** | ❌ FAIL — `core,rpc,horizon` test |
| **b41d77db** (c6118f2c + meta fix) | ❌ FAIL — both `rpc` and `core,rpc,horizon` tests |

c6118f2c is the only commit between 7edb540e (last-good) and the first failure that touches overlay/SCP. b41d77db only adds `crates/ledger/src/execution/{apply,meta}.rs` changes (BucketList meta fix), so it inherits the regression from c6118f2c.

## Evidence

**b41d77db Quickstart run [25265554373](https://github.com/stellar-experimental/henyey/actions/runs/25265554373)**, RPC test (job `74080247792`):

```
2026-05-03T00:45:15Z submitting create accounts tx: timeout waiting for transaction e4fd59d2c0bb63524e0f667e12389b73be4d89e9215d55b2ecdee6418dd35773 to finalize, result_xdr: , diagnostic_events_xdr: []
2026-05-03T00:45:45Z submitting create accounts tx: timeout waiting for transaction b7073cb797c2478ec97f51dd29d0b5a7483f42e0f95d0f5b2c088ddde8d6af8d to finalize, ...
2026-05-03T00:46:15Z creating minion accounts: after retrying 2 times: submitting create accounts tx: timeout waiting for transaction e4a13a7df73e6b76871cfceb2592daf232308986931ae60a34b90abf23cd93fe to finalize
... (continued every 30s for ~3 minutes, then job exits)
```

Same shape on c6118f2c run [25264979010](https://github.com/stellar-experimental/henyey/actions/runs/25264979010) (job `74079425741`).

The transactions are being submitted via Horizon → henyey (as core), but never appear in a closed ledger. `result_xdr: ""` indicates the transaction was never included in a tx-set. The Quickstart standalone network has only one validator (henyey itself), so this is a self-validation failure mode.

[...]

PR #3706, opened 2026-07-02
De-load the consensus event loop and nomination critical path (+7.8% max TPS)

tomerweller:
## Summary

Raises henyey's distributed max TPS (Supercluster `MissionMaxTPSClassic`, 23-node tier-1 on one nsc 32×64 VM) by de-loading the main consensus event loop and the SCP nomination critical path. **Controlled same-instance back-to-back full-band comparison on a fresh instance: base 1413 → stack 1523 tx/s (+7.8%).**

Diagnosis (full narrative in `docs/maxtps-optimization/2026-07-02-target2000.md`): at the ceiling the single event loop was **100.0% busy** (measured 29.93s of every 30s window) and SCP envelopes waited a median 92ms per hop in the dedicated channel, producing 0.8–3.5s nomination-convergence tails that trip the loadgen's per-account completion guillotine.

### Accepted changes (one cohesive theme)
1. **Advert flush off the loop** — was ~100ms inline per 200ms flood tick (up to 60% of the loop), now a coalesced background task (same pattern as peer maintenance #3690).
2. **Dedicated inbound-flood consumer task** — Transaction/FloodAdvert/FloodDemand processing (~20s of work per 30s window) moves off the select loop; SCP and fetch keep their dedicated channels/arms.
3. **In-flight SCP dedup in overlay peer tasks** — core parity: `checkScheduledAndCache` runs in core's peer thread (Peer.cpp:1113-1117); henyey ran it on the main loop, so ~22 duplicate flood copies of every envelope (~3000 msgs/slot/node) transited the SCP channel. Same `ScpScheduledCache`, same token-lifetime semantics (#2317 self-echo preserved — overlay integration tests pass unchanged; new regression test for token release on full-channel drop).
4. **Tx-queue store-lock shrink** — advert collection held the store write lock ~50% of the time (candidate × 22-peer seen-checks in-lock); now a short-lock drain + lock-free per-peer dedup (budget-accounting delta documented in-code; 2.6× measured headroom).
5. **Nomination-value prebuild** — the ~150-200ms tx-set build ran between the trigger and `scp.nominate`, on the network convergence critical path; now prebuilt ~300ms early and reused (convergence median 580→320ms). `close_time` has whole-second granularity so values are ~always identical.
6. **Ping-expiry 60s→10s** — a lost ping reply could silence a peer pair past the 30s idle-drop deadline.

Also includes parity-irrelevant diagnostics (`maxtps_cad`/`maxtps_scp`/`maxtps_fetch` slow-only log stamps, per-phase body-only event-loop accounting) and two measured-and-reverted experiments kept in history for the record (early-trigger compensation; biased select — both documented in the run doc with refutation data).

### Measured effects on internals
- Event loop at ceiling: 100% → ~5% busy.
- SCP channel slow-waits: ↓5× count.
- Nomination convergence median: 580 → 320ms.
- Loadgen submission progress at failing steps: 70% → 88%.

### Measurement honesty
Morning-session numbers (baseline 1487) are NOT comparable to the final A/B: the 5h-old instance drifted below its own morning baseline (same base image: 1487 at 01:30, 1413 at 06:20 on a fresh instance, and failing 1490 on the drifted one). The accepted +7.8% is from the same-instance, back-to-back, identical-band comparison — the only valid protocol we found. Short-probe caveats from #3703/#3705 still apply to absolute numbers.

## Parity
Wire bytes, tx-set semantics, close-time semantics, and observable surfaces unchanged; all changes are internal scheduling/threading (explicitly divergeable per `docs/PARITY.md`). SCP dedup placement matches stellar-core more closely than before.

## Tests
- `cargo test --all`: green (full workspace).
- henyey-app 1146, henyey-herder 1202, henyey-overlay 479+8+3+1 (incl. #2317 self-echo + duplicate-forwarding integration tests unchanged).
- New: prebuild cache/reuse test, SCP-channel token-release regression test, budget-accounting tests updated with rationale.

Run doc: `docs/maxtps-optimization/2026-07-02-target2000.md`

### DIFF
--- a/crates/app/src/app/consensus.rs
+++ b/crates/app/src/app/consensus.rs
@@ -473,6 +473,26 @@ impl App {
             ct_offset_secs,
             "Armed event-driven consensus trigger timer (setupTriggerNextLedger)"
         );
+
+        // Prebuild the nomination value ~300 ms ahead of the trigger (maxtps
+        // iter 9): the ~150-200 ms tx-set build otherwise runs between the
+        // trigger firing and `scp.nominate`, on the network's nomination-
+        // convergence critical path. A plain sleep task (not a managed timer)
+        // suffices: `prebuild_nomination_value` self-gates on slot/LCL/
+        // nomination state, so a stale or duplicate firing is a cheap no-op.
+        const PREBUILD_LEAD: std::time::Duration = std::time::Duration::from_millis(300);
+        if self.is_validator && delay > PREBUILD_LEAD + std::time::Duration::from_millis(100) {
+            let herder = std::sync::Arc::clone(&self.herder);
+            let lead_delay = delay - PREBUILD_LEAD;
+            tokio::spawn(async move {
+                tokio::time::sleep(lead_delay).await;
+                let _ =
+                    henyey_common::spawn_blocking_logged("prebuild_nomination_value", move || {
+                        herder.prebuild_nomination_value(next_slot as u32)
+                    })
+                    .await;
+            });
+        }
     }
 
     /// Perform out-of-sync recovery matching stellar-core's outOfSyncRecovery().
--- a/crates/app/src/app/consensus.rs
+++ b/crates/app/src/app/consensus.rs
@@ -243,11 +243,8 @@ impl App {
             let last_slot = current_ledger as u64;
             let last_ballot_start = self.herder.prepare_start(last_slot);
             let now_instant = std::time::Instant::now();
-            // Early-trigger compensation (kept in lock-step with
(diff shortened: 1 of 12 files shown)


tomerweller:
Sustained-load follow-up (operator-requested): the +7.8% is burst-side only — 5-min probes sustain ~1126–1132 vs the 1-min 1489–1523 (−24%). Root-cause chain, three fixes (incl. a true core-parity divergence in receive_transaction's tracking gate), and the tail-latency-asymptote verdict are in #3709 and the run doc.

PR #3709, opened 2026-07-02
Sustained-load fixes: syncing-gate parity, stranded-tx re-flood, advert budget refund

tomerweller:
## Summary

Follow-up to #3706, answering: **is the +7.8% burst ceiling sustainable?** No — the 5-minute probe (`SSC_MAXTPS_TXS_MULTIPLIER=300`) sustains only **~1126–1132 tx/s, −24% below the 1-minute burst ceiling (1489–1523)**. This PR lands the three real defects found while root-causing that gap, plus the diagnostic instrumentation and the full writeup.

### The captured root-cause chain
`PAY_PREGENERATED` load runs die on the FIRST non-Added submit (parity #3638). Live capture (`maxtps_ban` WARN) caught the killers:

1. **Syncing-gate `TryAgainLater` (fixed, core parity)** — `receive_transaction` bounced submissions whenever the herder wasn't `Tracking`, a documented divergence (stellar-core's `recvTransaction` never gates on tracking state). A momentary Tracking→Syncing blip on any of 7 loadgen nodes killed the run — captured 3× live. `can_receive_transactions` now includes `Syncing`; `Booting` still defers (LM may be uninitialized; keeps the friendbot-safe retryable signal).
2. **Stranded-tx no-re-flood (fixed)** — a tx is erased from the flood queue on first advert; if that advert is lost the tx strands on the submitting node until that node's own nomination candidate wins (~2 min on 23 nodes). `shift()` now re-marks a pending tx for flooding when it ages 2 ledgers (exactly once; per-peer advert history limits re-sends to peers that never got it). Test: `test_shift_refloods_pending_tx_at_age_two`.
3. **Advert budget regression from the #3706 lock split (fixed)** — seen-by-all rebroadcast candidates consumed the ops budget after peer reconnects, starving new-tx adverts for seconds. The two-phase collector now refunds their budget, restoring the original budget-neutral semantics (assertions restored).

### Honest outcome
Same-hardware pre/post-fix long-probe A/B: **1126 → 1132 (no change)**. The fixes are correctness/parity wins, but the sustained ceiling is set by a deeper mechanism, captured on the final repro: a queue-level account-pending `TryAgainLater` — one tx's inclusion latency exceeded its account's ~21s submission cycle. **The ~1130 sustained ceiling is the closed-loop tail-latency asymptote** (the #3705 "D"): a max-latency criterion, not a throughput wall (the same network drains 1489+ tx/s cleanly for 60s). Full analysis in the run doc, including the cross-node early-arrival-ban amplifier and the benign-`tx_queue_banned` red herring.

### Also included
- `maxtps_ban` diagnostic instrumentation (ban-site attribution, fatal-reject detail incl. account+seq, generate_tx fatal path) — logging only.
- Run-doc sustain section (`docs/maxtps-optimization/2026-07-02-target2000.md`).

## Parity
Queue/flood scheduling internals (divergeable) except the syncing-gate change, which REMOVES a divergence (core parity). Wire bytes and observable semantics unchanged.

## Tests
henyey-herder 1203, henyey-app 1146, henyey-simulation 108 — all green. New regression tests for the re-flood, gate, and budget semantics.

### DIFF
--- a/crates/app/src/app/ledger_close.rs
+++ b/crates/app/src/app/ledger_close.rs
@@ -2522,6 +2522,19 @@ impl App {
             );
             if !is_success {
                 failed_hashes.push(Hash256::hash_xdr(tx));
+                // [maxtps_ban] failed-at-apply txs get banned post-close; under
+                // the sustained-load investigation these bans exploded (~20k/min)
+                // and poisoned live accounts. Log the result-code discriminant
+                // (sampled: first 3 per close) to attribute the failure mode.
+                if failed_hashes.len() <= 3 {
+                    tracing::info!(
+                        target: "maxtps_ban",
+                        ledger_seq = pending.ledger_seq,
+                        result = format!("{:?}", tx_result.result.result).chars().take(60).collect::<String>(),
+                        seq_num,
+                        "failed_apply"
+                    );
+                }
             }
         }
         timer.mark("build_persist_inputs_ms");
@@ -2779,9 +2792,13 @@ impl App {
                 close_time,
             );
             if !failed_hashes_for_ban.is_empty() {
-                tracing::debug!(
+                // [maxtps_ban] info-level so the sustain investigation can see
+                // per-close ban volume by site (site 1 = failed at apply).
+                tracing::info!(
+                    target: "maxtps_ban",
+                    ledger_seq,
                     failed_count = failed_hashes_for_ban.len(),
-                    "Banning failed transactions"
+                    "ban_site1_failed_apply"
                 );
                 herder.tx_queue().ban(&failed_hashes_for_ban);
             }
@@ -2898,6 +2915,13 @@ impl App {
                     let invalid_hashes: Vec<Hash256> =
                         invalid.iter().map(|htx| htx.hash()).collect();
                     invalid_banned = invalid_hashes.len();
+                    // [maxtps_ban] site 2 = queued txs stateful-invalid after close.
+                    tracing::info!(
+                        target: "maxtps_ban",
+                        ledger_seq,
+                        invalid_banned,
+                        "ban_site2_revalidation"
+                    );
                     herder.tx_queue().ban(&invalid_hashes);
                 }
                 crate::metrics::CLOSE_TX_QUEUE_INVALIDATION_SECONDS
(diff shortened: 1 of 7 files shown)


tomerweller:
Adversarial parity review vs core v27.0.0 → **MERGE-SAFE** (will merge on Test-green; 5/6 required already green). Re-verified the two critical changes independently:

- **#1 syncing-gate parity — SAFE.** Confirmed core `HerderImpl::recvTransaction` (HerderImpl.cpp:636-705) has NO tracking/sync-state gate (only the source-account cross-type pending limit → `tryAdd`); henyey's prior `Tracking`-only gate was a real divergence. Admitting while `Syncing` is safe: `try_add` validates only against `validation_context`, which is set exclusively from a fully-closed ledger header (bootstrap seed guarded on seq==0, updated each close) — same class as core's `getLastClosedLedgerHeader()`. `Booting`-defer correctly preserved (pre-bootstrap seq 0).
- **#3 advert budget refund — SAFE.** Refund is symmetric to the phase-1 charge (`+op_count` on generic always, `+op_count` on DEX iff `is_dex`); restores pre-#3706 budget-neutral semantics (assertions 10/9/dex-5); no double/over-refund; `budget` is stack-local so no race with the #3706 store-lock split.

**Non-blocking follow-up (not gating this merge):** change **#2 (age-2 `shift()` re-flood) is a functional no-op in production.** henyey already calls `rebroadcast()` after `shift()` on every close (`ledger_close.rs:2825`→`2934`), and `rebroadcast()`→`reset_and_repopulate(store.values())` re-marks EVERY pending tx for flooding unconditionally (matching core's per-close `rebroadcast()` in `updateQueue`). So the age-2 `mark_for_flood` inside `shift()` is overwritten microseconds later — verified empirically (shift→rebroadcast re-floods at age 1/2/3 regardless). The stated root cause ('tx erased on first advert, never re-marked until own candidate wins') is contradicted by that existing per-close rebroadcast. It's harmless (no amplification/over-admission/parity break) so not a blocker, but `test_shift_refloods_pending_tx_at_age_two` asserts a non-production (`shift()`-without-`rebroadcast()`) path. Suggest dropping the age-2 re-flood + counter + test in a follow-up; if stranded-tx latency is still real on the rig, the fix likely lies elsewhere (advert loss / per-peer dedup window). Changes #1/#3 stand on their own.

PR #3712, opened 2026-07-03
Fix the sustained-load TPS collapse: WAL checkpoint stalls + nomination tx-set fetch starvation

tomerweller:
## Summary

Closes most of henyey's burst-vs-sustained max-TPS gap on MissionMaxTPSClassic (23-node tier-1, nsc 32×64, 5-minute probes):

| | before | after |
|---|---|---|
| sustained (5-min probe) | ~1100 tx/s (died at 1300 in ~90-150s) | **≥1400 tx/s** (1300 ✅, 1400 ✅, 1450 pending, 1500 ✗) |
| burst (1-min probe) | 1510 tx/s | unchanged reference |
| core reference | burst 1530 / sustained 1522 | — |

## Causal chain (verified by forensics at each link)

1. **SQLite WAL checkpoint stalls on the close path** (the dominant binder). Henyey persists ~7k per-tx rows (body/result/meta XDR) per ledger ≈ 10-30 MB of WAL — stellar-core writes ~0.1 MB (it dropped per-tx SQL storage in v21). With `wal_autocheckpoint=10000` the threshold fired every couple of closes *inside* the ledger-close-persist commit, copying+fsyncing tens of MB while holding the WAL write lock: **72-121 slow-holder events per 5-min run, up to 16.4 s**. The stalled persist froze the apply pipeline → the LCL-gated consensus trigger fired 0.4-2.2 s late on affected nodes → the nomination round-1 quorum could not assemble within the 1 s round timer → multi-round nominations (observed 3 rounds / 3.5 s), 7-13 s ledgers, tx-queue backlog, loadgen wedge deaths. Fixes:
   - `PRAGMA synchronous=NORMAL` (direct parity: core Database.cpp:166; henyey used FULL, fsyncing the WAL every commit) — 08ee1f76
   - `wal_autocheckpoint=0` + background `PRAGMA wal_checkpoint(PASSIVE)` task every 2 s (PASSIVE never blocks writers) — f1176923
   - Post-fix forensics: **zero** ledger-close-persist WAL warnings; cadence p95 6.1 s (was 9-13 s stretches).

2. **Nomination-critical tx-set fetch starvation** — b1ef83dd. During nomination only the value's builder holds the new tx set. Henyey (a) never asked the envelope *sender* for NOMINATE-referenced sets (only EXTERNALIZE), (b) paused ALL GetTxSet sends when an internal budget filled, and (c) permanently abandoned a hash once every peer replied DontHave. Observed: nodes fetching the round-1 candidate 3.2 s late (sp-1/sp-2), gating quorum. stellar-core's Tracker asks envelope senders first and rebuilds on exhaustion. Post-fix: all 23 nodes fetch candidates within ~170 ms, in lockstep. Two regression tests.

3. **Loadgen pacing-clock anchor** — 3e93bb3f. `start()` stamped `start_time` before ~5 s of setup (55 MB pregen file slurp + 3285 account preloads); the uncapped cumulative pacer then burst `setup_delay×rate` transactions instantly, crossing the pregenerated file's same-account resubmission gap at high rates. Death signature moved from `tx_seq=2, age=0` (onset burst) to mid-window after the fix.

4. **Wedge-push + forensics** — 9eb525cc: re-broadcast age-2 pending txs post-close; fatal-reject forensics (account/seq/pending-age attribution) that drove the whole investigation. 1afc810a reverts the earlier nomination-value prebuild (a sustained-load liability: every candidate ~300 ms staler → systematic two-build misses wedge the loadgen).

5. **Diagnostics kept** (parity-safe, INFO-level): nomination-path telemetry (5b63f49a), round-leader forensics (84e6237f), per-build queue-vs-selected accounting (6872e702).

## Parity

No observable-surface change: wire messages, XDR, hashes, and consensus semantics untouched. Fetch scheduling, SQLite pragmas/checkpointing, loadgen pacing, and telemetry are internal (docs/PARITY.md divergeable). `synchronous=NORMAL` matches core exactly; crash consequences (last commits lost, no corruption) are identical to core's posture.

Tests: `cargo test -p henyey-db`, `-p henyey-app` (incl. 2 new fetch-starvation regression tests), `-p henyey-herder`, `-p henyey-scp`; full `cargo test --all` gate running before merge.

### DIFF
--- a/crates/app/src/app/consensus.rs
+++ b/crates/app/src/app/consensus.rs
@@ -474,25 +474,16 @@ impl App {
             "Armed event-driven consensus trigger timer (setupTriggerNextLedger)"
         );
 
-        // Prebuild the nomination value ~300 ms ahead of the trigger (maxtps
-        // iter 9): the ~150-200 ms tx-set build otherwise runs between the
-        // trigger firing and `scp.nominate`, on the network's nomination-
-        // convergence critical path. A plain sleep task (not a managed timer)
-        // suffices: `prebuild_nomination_value` self-gates on slot/LCL/
-        // nomination state, so a stale or duplicate firing is a cheap no-op.
-        const PREBUILD_LEAD: std::time::Duration = std::time::Duration::from_millis(300);
-        if self.is_validator && delay > PREBUILD_LEAD + std::time::Duration::from_millis(100) {
-            let herder = std::sync::Arc::clone(&self.herder);
-            let lead_delay = delay - PREBUILD_LEAD;
-            tokio::spawn(async move {
-                tokio::time::sleep(lead_delay).await;
-                let _ =
-                    henyey_common::spawn_blocking_logged("prebuild_nomination_value", move || {
-                        herder.prebuild_nomination_value(next_slot as u32)
-                    })
-                    .await;
-            });
-        }
+        // NOTE (maxtps, REVERTED): a ~300 ms nomination-value PREBUILD lived
+        // here (arm a sleep task at trigger−300 ms; `trigger_next_ledger`
+        // reused the cached value). It cut convergence median 580→320 ms but
+        // is a SUSTAINED-LOAD liability: every candidate becomes ~300 ms
+        // staler than a trigger-time build (stellar-core builds AT trigger),
+        // so txs arriving in that window systematically miss the next tx-set
+        // on EVERY node — a two-build miss wedges the tx past its loadgen
+        // account's minimum resubmission gap and insta-fails 5-minute runs
+        // (#3638). Forensics: pending_age=2 collisions with wedge-push counts
+        // of 6-22/node/ledger. Build at trigger time, like core.
     }
 
     /// Perform out-of-sync recovery matching stellar-core's outOfSyncRecovery().
--- a/crates/app/src/app/ledger_close.rs
+++ b/crates/app/src/app/ledger_close.rs
@@ -2956,6 +2956,37 @@ impl App {
                 // Re-mark all surviving transactions for flooding. Matches
                 // stellar-core's resetBestFeeTxs() + rebroadcast() on ledger close.
                 self.herder.tx_queue().rebroadcast();
+
+                // Sustained-load wedge fix: PUSH the full bodies of txs that
+                // have been pending 2 ledgers to all peers. Rebroadcast alone
+                // re-MARKS them, but the per-peer advert history suppresses
+                // the re-advert to any peer recorded as already-adverted — so
+                // a tx whose demand/response was lost stays missing from
+                // every other queue and can only apply when THIS node's
+                // candidate wins nomination. One lost tx then out-waits its
+                // account's ~20 s submission cycle and a PayPregenerated run
+                // insta-fails (#3638): measured as henyey sustaining ~1150
+                // tx/s on 5-minute windows vs stellar-core's 1522 on the same
+                // hardware. The push is rare by construction (only txs that
+                // missed 2 consecutive ledgers; typically 0-5 per 5-minute
+                // window) and flood-gate rules still apply on receive.
+                if !shift_result.reflooded_txs.is_empty() {
+                    if let Some(overlay) = self.overlay().await {
+                        let count = shift_result.reflooded_txs.len();
+                        for env in shift_result.reflooded_txs {
+                            let msg = stellar_xdr::StellarMessage::Transaction(env);
+                            if let Err(e) = overlay.broadcast(msg).await {
+                                tracing::debug!(error = %e, "wedged-tx push failed");
+                            }
+                        }
+                        tracing::info!(
+                            target: "maxtps_ban",
+                            ledger_seq,
+                            count,
+                            "pushed wedged pending txs to all peers"
+                        );
+                    }
+                }
             }
             Err(e) if e.is_panic() => {
                 tracing::error!(
--- a/crates/app/src/app/lifecycle.rs
+++ b/crates/app/src/app/lifecycle.rs
@@ -2868,6 +2868,30 @@ impl App {
             None
         };
 
+        // Tx-set hashes referenced by NOMINATE values. During nomination only
+        // the value's builder is guaranteed to already hold the set, so the
+        // round-robin fetcher in `request_pending_tx_sets` can burn its first
+        // asks on peers that don't have it yet and time out the whole
+        // nomination round. stellar-core's Tracker always asks peers that sent
+        // envelopes referencing the item first (Tracker::tryNextPeer via
+        // getPeersKnows); mirror that by asking the envelope sender directly
+        // when the herder reports Fetching (see the EnvelopeState::Fetching arm).
+        let nominate_tx_set_hashes: Vec<Hash256> = match &envelope.statement.pledges {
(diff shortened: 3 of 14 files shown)


tomerweller:
Final validation on the same instance (5-min probes, image walbg): 1300 ✅ · 1400 ✅ · 1450 ✗ · 1500 ✗ — sustained bound **[1400, 1450)** vs burst 1510 (core sustained: 1522). Full `cargo test --all` gate: clean. Close-path WAL-lock warnings: zero (was 72-121/run up to 16.4s).

tomerweller:
## Board-watcher adversarial parity review vs core v27.0.0 — **HOLDING for your call** (1 robustness blocker + 1 comment-accuracy fix)

Reviewed with MAX rigor (persist + consensus/overlay; adjacent to open #3702). **Consensus/determinism safety is clean** — no wire/XDR/hash/consensus-semantics change (verified below). Two things I want your decision on before merge, because this is a persist change landing on a validator that is *currently wedged on an open WAL-stall issue (#3702)* and this project has ENOSPC-fatality history (#3478):

### 1. `synchronous=NORMAL` / `wal_autocheckpoint=0` are **divergences, not parity** (the PR comments say parity)
Core `Database.cpp:163-169`:
```cpp
mSession << "PRAGMA journal_mode = WAL";
// FULL is needed as to ensure durability
// NORMAL is enough for non validating nodes
// mSession << "PRAGMA synchronous = NORMAL";   // <-- line 166 is COMMENTED OUT
mSession << "PRAGMA wal_autocheckpoint=10000";  // <-- line 169, live
```
Core deliberately runs **FULL** for validators and keeps inline autocheckpoint at 10000 with **no** background checkpointer. So the code comments claiming "parity with `Database.cpp:166`" are inverted — these are *deliberate divergences*. `synchronous=NORMAL` itself is **determinism-safe** (a power-loss can drop the last commits but cannot corrupt or fork; restart recovers via catchup — no hash divergence vs peers), so I'm fine with the divergence on the merits; the ask is just to **fix the comments** to call it a documented divergence (per docs/PARITY.md), not parity.

### 2. **Blocking concern: unbounded WAL growth if the checkpointer dies (no inline floor).**
With `wal_autocheckpoint=0`, the *only* thing draining the WAL is the 2s background PASSIVE task — and it `break`s out of its loop permanently on a `spawn_blocking` join error:
```rust
Err(e) => { tracing::warn!(error = %e, "WAL checkpointer join error"); break; }
```
After that (or if the 2s tick is starved under the same sustained blocking-pool load this PR targets, or a long-lived reader keeps PASSIVE from draining), the WAL grows without bound → disk exhaustion → validator death. Core's inline 10000-page autocheckpoint is a self-healing floor this removes with no backstop. Suggest one of:
- keep a **non-zero `wal_autocheckpoint`** (e.g. core's 10000) as an inline floor *alongside* the background PASSIVE task — a dead/starved checkpointer then degrades to core behavior instead of unbounded growth; **or**
- **respawn instead of `break`** on join error + a **WAL-size-triggered forced checkpoint** (TRUNCATE) backstop.

### Clean / parity-ward (no action)
- **Nomination tx-set fetch** (nomination.rs/herder/tx_flooding): moves *toward* core — core's `Tracker::tryNextPeer` (`Tracker.cpp:157-205`) asks NOMINATE envelope senders and rebuilds on exhaustion (never permanently abandons). henyey's sender-direct ask + critical-slot budget bypass + rebuild-not-abandon match; the 500ms rebuild pause vs core's 1500ms×N is a benign cadence divergence, no spin (500ms-throttled, slot-gated).
- **Prebuild revert (1afc810a)**: clean — removes only the caller; `trigger_next_ledger` falls back to build-at-trigger (pre-#3706, = core lazy build). Leftover fn still referenced by its test, no `-Dwarnings` break.
- **age-2 full-body re-push (tx_flooding)**: materially new vs #3709's advert-only reflood, but flood-gate-mediated, bounded (age==2 only, 0-5/ledger, once/close), receivers dedup by hash, touches no hashed/consensus path. Benign divergence.
- **Determinism**: consensus.rs/herder/scp_driver/tx_queue changes are telemetry + behavior-preserving refactors; nothing touches tx-set contents/ordering, nomination/ballot values, or any hashed/wire structure.

Required CI at review time: Clippy/Build(release-ci)/Format/Analyze(rust) ✅; Test + Doctest still running. **Not auto-merging** — your call on #2 (add a floor/respawn, or accept the tradeoff with a comment fix). Happy to merge the moment you confirm.

_(board-watcher; core pin v27.0.0 7696c069; HEAD a1a4c09a)_

tomerweller:
Both review points addressed in bf8b5599:

**1. Comments corrected — you're right, these are divergences, not parity.** I mis-read `Database.cpp:166`: the `synchronous = NORMAL` pragma is commented out, so core ships FULL for validators and keeps inline `wal_autocheckpoint=10000` with no background checkpointer. The comments in `henyey_db` now describe both settings as deliberate, documented divergences (docs/PARITY.md divergeable layer), with the determinism-safety argument (power loss can drop newest commits, cannot corrupt or fork; restart recovers via catchup) kept.

**2. WAL growth bounded in every failure mode — took a hybrid of your two options.** Three layers:
- the checkpointer loop **never exits** on checkpoint/join errors (the `break` is gone);
- if a PASSIVE pass leaves **>~512 MiB** undrained (long-lived reader pinning the WAL, starved ticks), it escalates to a blocking `wal_checkpoint(TRUNCATE)` with an ERROR log — a bounded stall instead of unbounded growth;
- the inline `wal_autocheckpoint` floor is restored at **~1 GiB** (262144 pages) rather than 0, so even a fully dead task degrades to bounded inline checkpoints.

On restoring core's 10000-page floor specifically: I deliberately didn't. Measured WAL under max sustained load exceeds 40 MB even with the background task draining continuously (PASSIVE passes take ~10 s on the saturated NVMe), so a 10000-page inline threshold would re-fire inside close-persist commits and reintroduce the 4-16 s stalls this PR removes. The 1 GiB floor sits above the worst observed operating point, and the TRUNCATE backstop at 512 MiB fires well before it ever could.

New unit test covers both checkpoint flavors (`pool::tests::test_wal_checkpoint_passive_and_truncate`); `cargo test -p henyey-db` and clippy clean.

tomerweller:
Re-reviewed `bf8b5599` — **both findings resolved; MERGE-SAFE.** Will merge once the remaining required checks (Test, Build(release-ci), Doctest) go green.

- **#2 (unbounded WAL growth):** fully addressed with three layers — checkpointer loop no longer exits on checkpoint/join errors; TRUNCATE backstop at ~512 MiB (fires before the inline floor); ~1 GiB inline `wal_autocheckpoint` disaster floor instead of 0. Ordering is correct (backstop = half the floor) and the "not core's 10000" rationale is sound (would re-fire the close-path stall). Unit test covers both PASSIVE and TRUNCATE.
- **#1 (comment accuracy):** corrected — `synchronous=NORMAL` and off-path checkpointing now documented as deliberate divergences (core `Database.cpp:164-169` ships FULL + inline 10000 for validators), with the determinism-safety note (crash drops newest commits, recovers via catchup, no corruption/fork).
- Scope remains persist-only (db pragmas/checkpoint + app checkpointer); no consensus/hash/wire change — determinism unaffected.

_(board-watcher re-review; core pin v27.0.0 7696c069)_

PR #3714, opened 2026-07-03
Skip per-tx RPC row storage on validators without JSON-RPC

tomerweller:
## What

Adds `database.store_rpc_data` (Option<bool>, default auto): when unset, per-transaction RPC-serving SQL writes — the `transactions` table (per-tx body/result/meta XDR) and the `events` table (contract events) — are **skipped on validators that do not serve JSON-RPC** (`node.is_validator = true`, `rpc.enabled = false`) and kept everywhere else. Explicit `store_rpc_data = false` with `rpc.enabled = true` is rejected at config validation (the getTransaction/getTransactions/getEvents endpoints serve from those tables). The large per-tx `tx_metas` clone on the event loop is also skipped when disabled.

Always-written surface is unchanged: ledger headers, whole-ledger `tx_history_entry`/`tx_result_entry` blobs, SCP history, HAS, LCL — history publish reads those, and all parity-relevant outputs are byte-identical.

## Why

stellar-core dropped per-transaction SQL storage entirely in v21; a pure henyey validator was still paying per-tx row writes (body+result+meta XDR) inside the ledger-close persist transaction. This aligns the validator persistence footprint with core. Operator-directed (maxtps frontier campaign, feature policy 2026-07-03).

## Measurements (MissionMaxTPSClassic, 23-node tier-1, nsc 32×64, same instance A/B)

Honest result: **no measurable TPS change** — post-#3712 the SQLite write volume is no longer the binding constraint.
- Burst (1-min probe): base 1525 tx/s; with this change the 1525 screen step failed 2/2 (edge coin-flip; apply fast at failure, close 90 ms for 4371 txs).
- Sustained (5-min steps): base 1452 tx/s → 1457 tx/s (+0.3%, within ±3-4% step noise).

The value is the validator config profile + core-v21 persistence-footprint parity + reduced disk usage on validators, not TPS.

## Tests

- `config::tests::test_store_rpc_data_*` (4 resolution cases) and `test_validation_rejects_rpc_enabled_with_store_rpc_data_false`
- `app::ledger_close::store_rpc_data_gate_tests::*`: disabled → `transactions`/`events` empty while header/txhistory/txresults/LCL persist; enabled → rows written
- Full `cargo test -p henyey-app` green; clippy green

### DIFF
--- a/crates/app/src/app/ledger_close.rs
+++ b/crates/app/src/app/ledger_close.rs
@@ -111,6 +111,12 @@ struct LedgerPersistInputs {
     /// Whether checkpoint publishing is enabled (validator with writable archives).
     /// Controls whether checkpoints are enqueued to the publish queue.
     publish_enabled: bool,
+    /// Whether to write the per-tx RPC-serving rows (`transactions` table)
+    /// and contract events (`events` table). False on validators that do not
+    /// serve JSON-RPC (see `AppConfig::store_rpc_data`); the whole-ledger
+    /// `tx_history_entry`/`tx_result_entry` blobs and SCP history are always
+    /// written — history publish reads those.
+    store_rpc_data: bool,
 }
 
 impl LedgerPersistInputs {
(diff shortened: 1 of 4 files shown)


tomerweller:
## Board-watcher adversarial parity review vs core v27.0.0 — **MERGE-SAFE** (merge on remaining CI green)

Reviewed with MAX rigor (persist path; the determinism/history blockers were #1 does anything consensus/history/verify read the gated tables, and #3 are the whole-ledger blobs still written). Both cleanly cleared; re-verified against origin/main myself.

- **Gated tables are RPC-only (determinism-safe).** The per-tx `txhistory` and `events` tables have exactly three readers, all JSON-RPC handlers: `get_transaction.rs` (`load_transaction`), `get_transactions.rs` (`load_transactions_in_range`), `get_events.rs` (`query_events`). No history-publish, catchup-apply, offline-verify, bucket, or hashed structure reads them. (`verify_execution` extracts events from freshly-computed `TransactionMeta`, not the DB table.) catchup/persist.rs only *writes* per-tx rows from archive blobs — no read, no hash impact.
- **Whole-ledger blobs stay unconditional.** The gate wraps only the per-tx loop + `store_events` + the `tx_metas` clone; `store_tx_history_entry`/`store_tx_result_entry` (the `txsets`/`txresults` blobs that history-publish + catchup + verify actually consume, and that feed the archive) are outside the gate. New test `test_store_rpc_data_disabled_skips_per_tx_rows_and_events` asserts blobs still persist with the gate off. ✅
- **No consensus-hash dependency on events.** v27 `LedgerHeader` commits only `txSetResultHash` + `bucketListHash`; there is no events commitment. Contract events flow into `TransactionMeta` (computed + streamed independently of the table), so skipping the `events` table moves no hash and no meta stream.
- **Core parity.** Core dropped per-tx SQL storage in v21 (builds checkpoint files directly); v27 validators write no per-tx rows. henyey's SKIP path matches that footprint (keeps them only when RPC serves them — a superset-retention choice, not a divergence).
- **Config resolution has no hole.** auto = `!is_validator || rpc.enabled`; the only dangerous cell (RPC on + tables skipped) is reachable solely via explicit `store_rpc_data=false`, which `validate()` rejects. No `Option<bool>` footgun.

**Cosmetic nit only (non-blocking):** the PR text/config docs say "transactions table" but the actual per-tx table is `txhistory` (the whole-ledger blobs live in `txsets`/`txresults`). Worth a one-word doc tweak; does not affect correctness.

Required CI at review time: Analyze(rust)/Build(release-ci)/Clippy/Format ✅; Test + Doctest still running. Will merge on green (or feel free to merge it yourself).

_(board-watcher; core pin v27.0.0 7696c069; HEAD 75634eb1)_

PR #3715, opened 2026-07-03
Chunk maintenance deletes to bound WAL write-lock holds

tomerweller:
## What

`run_maintenance` previously issued each table's DELETE as a single statement covering every eligible ledger. The first maintenance pass after a checkpoint publish completes — which unlocks ~64 ledgers of `txsets`/`txresults` blobs (plus `txhistory` rows) for deletion at once — held SQLite's WAL write lock for **21-28 s** (measured via `db_write_ctx=maintenance-delete`, 2026-07-03 MaxTPS frontier forensics). Under load this stalled ledger-close persist and event-loop DB reads on every node simultaneously (SSC syncs maintenance ticks), producing 19-54 s ledger closes and wedging loadgen.

Deletes now run in bounded chunks (`MAINTENANCE_CHUNK_LEDGERS = 2` ledgers per statement) with a 25 ms pause between chunks, so each implicit transaction's lock hold is short and the persist path interleaves. The per-pass `WriteCtxGuard` moved to per-chunk so reported holds are real single-transaction holds. Error handling preserves the previous warn-and-continue-per-table behavior; the `count` budget semantics are unchanged.

## Measurements (MissionMaxTPSClassic, 23-node tier-1, nsc 32×64, same-instance A/B)

- The first post-publish delete wave is survived at 1472 tx/s sustained (previously fatal at 1458-1466: `maintenance-delete elapsed_ms=28213` → global stall → run death).
- `maintenance-delete` WAL holds: 28.2 s → ≤7.7 s observed, and the residual is lock-wait under a *different* writer (`scp-persist-write`), not maintenance work.
- Sustained edge: 1452 → 1472 tx/s (+1.4%). The remaining sustained binder is SCP-history persistence lock holds (follow-up under investigation).

This is a production robustness fix for any validator under sustained load — a 20+ s WAL stall every time the publish threshold advances past a backlog is a real mainnet hazard, independent of the mission numbers.

## Tests

- New: `test_delete_in_chunks_bounds_each_call_and_exhausts`, `test_delete_in_chunks_respects_total_budget`, `test_delete_in_chunks_stops_on_error`
- All 33 maintainer tests green (existing end-to-end pruning tests now exercise the chunked path); clippy green.

Run doc: `docs/maxtps-optimization/2026-07-03-frontier.md` (in campaign branch).

### DIFF
--- a/crates/app/src/maintainer.rs
+++ b/crates/app/src/maintainer.rs
@@ -330,6 +330,62 @@ fn build_maintenance_interval(period: Duration) -> tokio::time::Interval {
 ///   both checkpoint publishing (`txsets`, `txresults`, headers) and RPC queries
 ///   (`txhistory`, headers), so we keep data needed by both consumers.
 /// - **RPC-only** (events, close meta): pruned at `rpc_lmin` only when RPC is configured.
+/// Maximum distinct ledgers deleted per SQL statement during maintenance.
+///
+/// Each `delete_old_*` call bounded by this runs as its own short implicit
+/// transaction, so SQLite's single WAL write lock is released between chunks
+/// and the ledger-close persist path can interleave. Without this bound, the
+/// first maintenance pass after a checkpoint publish completes (which unlocks
+/// ~64 ledgers of txsets/txresults blobs for deletion on every node within
+/// the same tick) held the write lock for 21-28 s under MaxTPS load, stalling
+/// ledger closes globally (2026-07-03 frontier run, `maintenance-delete`
+/// write-ctx forensics).
+const MAINTENANCE_CHUNK_LEDGERS: u32 = 2;
+
+/// Pause between maintenance delete chunks. Gives the ledger-close persist
+/// transaction (and other writers) a window to acquire the WAL write lock
+/// between chunks.
+const MAINTENANCE_CHUNK_PAUSE: Duration = Duration::from_millis(25);
+
+/// Run one table's maintenance deletion in bounded chunks.
+///
+/// Calls `delete_chunk` with at most [`MAINTENANCE_CHUNK_LEDGERS`] ledgers per
+/// call until it reports no more rows deleted or `total_budget` ledgers have
+/// been requested, pausing [`MAINTENANCE_CHUNK_PAUSE`] between chunks. Errors
+/// are logged and stop the loop (matching the previous per-table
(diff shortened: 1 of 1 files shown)


tomerweller:
## Board-watcher review vs core v27.0.0 — **MERGE-SAFE** (merge on remaining CI green)

Single-file maintenance change; maintenance is retention GC (prunes history rows past the publish/RPC window), **non-consensus** — no ledger-hash / bucket / archive / catchup impact. Verified the one thing that matters for a chunked rewrite: the chunks delete exactly the same rows as the old single call.

- **Semantically equivalent to the pre-chunk single DELETE.** `delete_old_*(lmin, n)` selects `DISTINCT ledgerseq <= lmin ORDER BY ledgerseq ASC LIMIT n` and deletes whole ledgers (confirmed in `queries/scp.rs`/`history.rs`). So the 2nd param is an **oldest-first ledger-count budget** — repeated 2-ledger chunks delete oldest→next-oldest and sum to the identical row set the old `count` call produced. Same eligibility floors (`publish_safe_lmin`/`publish_and_rpc_lmin`/`rpc_lmin` unchanged), same per-table `count` budget, only the transaction boundaries change (bounded WAL lock holds + 25 ms pauses).
- **Loop terminates + budget/error semantics preserved.** `remaining` strictly decreases by `chunk`≥1; `Ok(0)` (table drained) and `Err` both break; the `Err` path keeps the previous warn-and-continue-per-table behavior. The over-decrement edge (request 2, only 1 ledger left) just ends the pass one no-op call sooner — harmless; the new `test_delete_in_chunks_*` cover exhaustion, budget cap, and error-stop.
- **WriteCtxGuard moved per-chunk** so `maintenance-delete` holds now report real single-transaction holds (the #3702 forensic goal) — instrumentation-only.
- **Parity:** core's maintenance also deletes in batches (`DatabaseUtils::deleteOldEntriesHelper`), so chunking is if anything *more* core-like; regardless it's retention-only and touches no consensus/observable surface.

Genuine production robustness fix (bounds the 21-28 s post-publish WAL stall). Required CI at review time: Build(release-ci)/Clippy/Format/Analyze(rust) ✅; Test + Doctest running. Will merge on green (or merge it yourself).

_(board-watcher; core pin v27.0.0 7696c069; HEAD 6d0c3795)_

PR #3716, opened 2026-07-03
Skip LedgerCloseMeta persistence on validators without JSON-RPC

tomerweller:
## What

Follow-up to #3714. Extends the `database.store_rpc_data` gate to the `ledger_close_meta` table: the full `LedgerCloseMeta` XDR blob per ledger, which serves only the JSON-RPC `getLedgers` path. When the node stores no RPC data, both the XDR serialization (multi-MB per ledger, on the event loop) and the SQLite write are skipped.

## Why

Under MaxTPS load (~1480 tx/s) this table was **70% of all DB bytes** — 409 MB of a 589 MB database after ~11 minutes (~4.5 MB/ledger). Worse, the maintainer only prunes `ledger_close_meta` when an RPC retention window is configured, so on a validator without RPC the table **grows without bound** — a disk leak independent of any benchmark.

Same-instance A/B (MissionMaxTPSClassic, 23-node tier-1, nsc 32×64): VM-wide disk writes dropped from ~150 MB/s to ~55 MB/s at comparable rates (stellar-core reference on the same instance: ~40 MB/s). Sustained TPS edge unchanged (~1472; the remaining binder is tx inclusion-tail dynamics, tracked separately) — this PR is a persistence-footprint/robustness fix, not a claimed TPS win.

## Tests

- Config resolution covered by the #3714 test suite (same flag).
- Persist path with `meta_xdr = None` covered by existing persist tests.
- `cargo test -p henyey-app --lib store_rpc` green; clippy green.

### DIFF
--- a/crates/app/src/app/ledger_close.rs
+++ b/crates/app/src/app/ledger_close.rs
@@ -2513,10 +2513,21 @@ impl App {
             &result.tx_results,
             tx_metas.as_deref(),
         );
-        let meta_xdr = result
-            .meta
-            .as_ref()
-            .and_then(|meta| meta.to_xdr(stellar_xdr::Limits::none()).ok());
+        // The full LedgerCloseMeta XDR blob serves only the JSON-RPC
+        // getLedgers path (`ledger_close_meta` table). Skip both the
+        // serialization (multi-MB per ledger on the event loop under load)
+        // and the SQLite write when this node stores no RPC data — it is the
+        // single largest write stream at high tx rates (~70% of DB bytes,
+        // 2026-07-03 frontier run), and without an RPC retention window the
+        // table is never pruned.
+        let meta_xdr = if self.config.store_rpc_data() {
+            result
+                .meta
+                .as_ref()
+                .and_then(|meta| meta.to_xdr(stellar_xdr::Limits::none()).ok())
+        } else {
+            None
+        };
 
         // Build (envelope, seq_num) pairs for all txs (applied + failed) so
         // sequence-based removal drops superseded queued txs for the same account.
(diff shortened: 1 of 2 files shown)


tomerweller:
## Board-watcher review vs core v27.0.0 — **MERGE-SAFE**

Follow-up to #3714 (same `store_rpc_data` gate, already vetted), extended to the `ledger_close_meta` blob. Verified:
- **RPC-only table.** `ledger_close_meta` is read only by RPC `getLedgers`; history-publish/catchup/offline-verify never read it (the `LedgerCloseMeta` hits in `crates/history` are the XDR *type* for the streaming callback + synthetic bucket-apply meta, and `emit_meta` *writing* during catchup — no table reads into any hashed/consensus path).
- **Determinism-safe.** The gate only skips `meta_xdr` serialization + the `ledger_close_meta` write (sole consumer: the persist input at ledger_close.rs:3137). `result.meta` is still computed and streamed independently — no consensus hash, meta stream, or archive depends on the table.
- **Clean diff.** `meta_xdr = if store_rpc_data() { … } else { None }`; whole-ledger blobs + headers unaffected.
- Genuine disk-leak fix: on a non-RPC validator the maintainer never prunes `ledger_close_meta` (RPC-retention-gated), so it grew unbounded (~70% of DB bytes, ~4.5 MB/ledger).

Minor non-blocking note (same as #3714): the **catchup** `emit_meta` write (`history/catchup/persist.rs`) isn't gated, so a non-RPC validator still writes bounded catchup-era meta rows — harmless (bounded, RPC-only), not steady-state growth. Optional future tidy.

All six required checks green. Merging.

PR #3718, opened 2026-07-03
Heal sole-advertiser tx strands: same-peer re-demand after expiry + tail diagnostics

tomerweller:
## What

Follow-up to #3717 (three commits):

1. **Demander-side unfulfilled-demand diagnostics** (maxtps_tail): log demand-history entries unfulfilled for 5+ s with their per-peer demand count.
2. **Account-0 submit timeline + remove_applied duplicate-drop diagnostics**: reconstructable interleaving of admission vs close vs removal for the recurring load-run wedge.
3. **Same-peer re-demand after expiry**: `demand_status` permanently discarded a peer after one demand (stellar-core parity). That is safe in core because its demand responses are never lost; in henyey a tx with a **sole advertiser** (network warmup, an account's first tx) wedged its account permanently when the single demand went unfulfilled — no other peer ever adverts the hash, and the per-peer discard blocked the only retry path (observed: unfulfilled demands stuck at `peers_demanded=1` for 14+ s; the stranded tx applied ~2 ledgers late; the account's next submission fatal-rejected the load run). Wire-compatible overlay-internal divergence: allow re-demanding from an already-demanded peer after 2 s, still subject to the existing retry-count cap and backoff.

## Measurements (MissionMaxTPSClassic, 23-node tier-1, nsc 32×64, same instance)

- Sustained 1560 tx/s re-verified green on this stack (campaign best; core reference on the same instance: 1526).
- 1600 tx/s still fails its warmup lottery — the residual compound (sole-holder strand + pipelined post-close queue-cleanup race) is documented in `docs/maxtps-optimization/2026-07-03-frontier.md` with full forensics; follow-up issue filed separately.

## Tests

- `test_demand_status_allows_same_peer_redemand_after_expiry` (new)
- `test_demand_status_discards_owned_and_banned_txs`, deferred-response tests still green; full workspace suite green; clippy green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01HzuKRiPrevW9qizRBw3oK7

PR #3717, opened 2026-07-03
Fix tx-pull stranding under sustained load: discard owned/banned demands (core parity) + defer full-channel demand responses

tomerweller:
## What

Three commits fixing the transaction pull (advert/demand/response) path that was wedging accounts under sustained MaxTPS load:

1. **maxtps_tail diagnostics** (instrumentation): origin-side stranded-tx advert coverage + receiver-side advert demand-queue overflow logging.
2. **Defer demand responses on full peer channels**: `handle_flood_demand` dropped the rest of a demanded batch when a peer's outbound channel was full; the demander's retry backoff then burned its per-period demand budget. stellar-core queues demand responses in its per-peer write queue and never drops them. Undeliverable responses now go to a per-peer deferred queue (cap 4000, drop-oldest) drained each flood tick.
3. **Discard demands for txs already queued or banned** — a straight stellar-core parity gap: `TxDemandsManager::demandStatus` (TxDemandsManager.cpp:71-74) discards demands for txs it already has; henyey's `demand_status` was missing the check, so every fulfilled hash kept cycling as RetryLater through the other ~21 peers' advert/retry queues for the 60 s retention window.

## Why it wedged

At ~1500 tx/s the redundant retries permanently saturated the per-peer demand queues (observed full at max_ops=14790 with 100-240 hashes dropped per 200 ms advert batch). New txs waited 2+ ledgers to be demanded anywhere; with the one-tx-per-account queue rule, the wedged account's next submission got TryAgainLater and PayPregenerated load runs insta-failed.

## Measurements (MissionMaxTPSClassic, 23-node tier-1, nsc 32×64, same-instance A/B)

- Sustained edge: **1472 → 1516 tx/s (+3.0%)**; stellar-core reference on the same instance: 1526 (gap now 0.7%).
- Advert demand-queue overflow events during a 5-min sustained step: **1227 → 0**.
- The pre-fix death mode (fatal reject at pending_age=2-3 on a stranded account) no longer reproduces below the new edge.

## Tests

- `test_demand_status_discards_owned_and_banned_txs` (core-parity check)
- `test_handle_flood_demand_defers_on_full_channel_and_flushes` (capacity-1 channel: 1 sent, 2 deferred, both delivered across flushes)
- `test_flush_deferred_skips_txs_gone_from_queue`
- Full henyey-app lib suite green; clippy green.

Run doc: `docs/maxtps-optimization/2026-07-03-frontier.md` (campaign branch).

[...]

PR #3720, opened 2026-07-03
Fix #3719: self-heal stale pending txs at submission + undo failed demand-send bookkeeping

### LINKED ISSUE #3719: Sustained-load residual: warmup sole-holder tx strand + pipelined post-close queue-cleanup race wedge accounts at ~1600 tx/s
## Context

2026-07-03 maxtps frontier campaign (`docs/maxtps-optimization/2026-07-03-frontier.md`). After the flood-path fixes (#3717, #3718) henyey sustains 1560 tx/s (stellar-core same-instance reference: 1526) and bursts 1673. Sustained steps at 1600 fail a "warmup lottery": always the same signature, always in the first ~15 ledgers of a step.

## Signature (full forensics in the run doc)

`pay_pregenerated fatal reject — TryAgainLater, source_account_id=0, pending_seq == on_ledger_seq, pending_age=2`

Two compounding mechanisms:

1. **Sole-holder strand at warmup**: an account's first tx has exactly one advertiser; if its single demand/response round fails, nothing re-pulls it promptly (the 2 s same-peer re-demand from #3718 softens but does not eliminate this — strands of 13.8 s were still observed at 1600). The stranded tx applies ~2 ledgers late.
2. **Pipelined post-close cleanup race**: ledger state (account seq) advances at apply, but `remove_applied`/shift/revalidation run in a later `spawn_blocking`. A submission landing in that window (observed 58 ms after close) sees a pending entry whose seq the ledger just consumed → `TryAgainLater` → PayPregenerated load runs insta-fail (parity #3638 makes any reject fatal). stellar-core cannot observe this state: its queue cleanup is synchronous with close on the main thread.

## Suggested directions

- For (2): make admission detect the applied-pending case (e.g. authoritative check of pending seq vs current account seq under the store lock, dropping the stale entry and admitting), or sequence submissions behind the post-close queue update.
- For (1): instrument the response-side journey of a sole-holder demand (was the Transaction sent, received, admitted?) — the responder now defers rather than drops (#3717), so the residual loss point is unidentified (receiver admission during syncing gate is a candidate).

## Evidence artifacts

- Runs: `~/data/65c209f0/maxtps-opt/artifacts-iter12-tail1600`, `artifacts-iter13-interleave`, `artifacts-iter14-sust1600` (session host)
- Key log lines quoted in the run doc's iteration log (iters 12-14).

tomerweller:
Fixes #3719 (both mechanisms).

## Part 2 — post-close queue-cleanup race (the run-fatal wedge)

Ledger state (account seq) advances at apply, but the queue's `remove_applied`/`shift`/revalidation run in a later `spawn_blocking`. A submission landing in that window (observed 58 ms after close in the campaign forensics) sees a pending tx whose sequence the ledger just consumed → `TryAgainLater` → PayPregenerated load runs insta-fail (#3638 parity: any reject is fatal). stellar-core cannot observe this state — its queue cleanup is synchronous with close on the main thread.

Fix: `TransactionQueue::drop_stale_pending(account, on_ledger_seq)` — drops the account's pending tx iff its seq ≤ the account's on-ledger seq, with the staleness condition re-checked under the store/account locks (no double fee-release against a concurrent `remove_applied`), fee bookkeeping mirrored from `remove_applied`, and no ban (if the tx applied, `remove_applied` bans it on its own schedule). `App::submit_transaction` invokes it when a `TryAgainLater` is attributable to a stale pending, then re-admits once. The healed result (Added) matches what core would return in the same situation.

## Part 1 — phantom demand records feeding the sole-advertiser strand

`run_tx_demands` marked a peer as demanded **during selection, before the send**. When `try_send_to` failed (channel full — 7 occurrences observed in one failing step), the phantom record made the demander believe it had demanded; the hashes had also been destructively popped from the peer's advert queue, so nothing ever retried (until #3718's 2 s same-peer expiry — and never if the channel stayed full). For a sole-advertiser tx (warmup, an account's first tx) this was the only pull path.

Fix: on send failure, unmark the per-peer demand records and re-queue the hashes onto that peer's retry queue, so the next demand round retries immediately.

## Tests

- `test_drop_stale_pending_unblocks_account` — reproduces the bug (`TryAgainLater` with a consumed-seq pending), asserts the heal admits the next tx
- `test_drop_stale_pending_keeps_live_pending` — a genuinely-pending tx (seq above on-ledger) is untouched
- `test_drop_stale_pending_releases_fee_state` — fee bookkeeping fully cleaned
- `test_failed_demand_send_is_unmarked_and_requeued` — capacity-1 pre-filled channel: no phantom mark, hashes re-queued, retry round sends and marks
- Full `henyey-herder` (1206) and `henyey-app` (1163) lib suites green; clippy green.

## Validation

Live MaxTPS validation on a fresh nsc instance is running; results will be posted as a follow-up comment (the wedge was stochastic at 1600 tx/s — the previously-failing level — so the comment will include repeated sustained steps there).

### DIFF
--- a/crates/app/src/app/mod.rs
+++ b/crates/app/src/app/mod.rs
@@ -794,6 +794,11 @@ pub struct App {
     /// was full. Drained each flood tick; stellar-core parity: core queues
     /// demand responses in its per-peer write queue and never drops them.
     tx_deferred_demand_responses: RwLock<HashMap<henyey_overlay::PeerId, VecDeque<Hash256>>>,
+    /// Watched stranded-tx hashes (#3719 part-1 wire tracing): populated by
+    /// the tail scanner for txs aged 8+ s in the local queue; the flood
+    /// demand/response paths log every event touching a watched hash.
+    /// Diagnostic only; bounded (see TAIL_WATCH_CAP).
+    tail_watch: RwLock<HashSet<Hash256>>,
     /// Demand history for transaction pulls.
     tx_demand_history: RwLock<HashMap<Hash256, TxDemandHistory>>,
     /// Pending demand hashes in FIFO order for retention.
@@ -1570,6 +1575,7 @@ impl App {
             broadcast_dex_op_carryover: AtomicUsize::new(0),
             tx_adverts_by_peer: RwLock::new(HashMap::new()),
             tx_deferred_demand_responses: RwLock::new(HashMap::new()),
+            tail_watch: RwLock::new(HashSet::new()),
             tx_demand_history: RwLock::new(HashMap::new()),
             tx_pending_demands: RwLock::new(VecDeque::new()),
             tx_set_dont_have: RwLock::new(HashMap::new()),
--- a/crates/app/src/app/mod.rs
+++ b/crates/app/src/app/mod.rs
@@ -2736,6 +2736,39 @@ impl App {
         }
 
         let result = self.herder.receive_transaction(tx.clone());
+
+        // Stale-pending self-heal (#3719). Ledger state (account seq)
+        // advances at apply, but the queue's remove_applied/shift run in a
+        // later spawn_blocking — a submission landing in that window sees a
+        // pending entry whose sequence the ledger just consumed and gets
+        // TryAgainLater. stellar-core cannot observe this state (its queue
+        // cleanup is synchronous with close), and PayPregenerated load runs
+        // treat any reject as fatal (#3638 parity), so the transient window
+        // was run-fatal under sustained load. If the reject was caused by a
+        // pending tx whose seq is already consumed on-ledger, drop the stale
+        // entry and re-admit once.
+        if matches!(result, henyey_herder::TxQueueResult::TryAgainLater) {
+            let network_id =
+                henyey_common::NetworkId::from_passphrase(&self.config.network.passphrase);
+            let frame =
+                henyey_tx::TransactionFrame::from_owned_with_network(tx.clone(), network_id);
+            let account_id = frame.inner_source_account_id();
+            if let Some((_, pending_seq, _)) =
+                self.herder.tx_queue().account_pending_info(&account_id)
+            {
+                if let Ok(Some(on_ledger_seq)) = self.load_account_sequence(&account_id) {
+                    if pending_seq <= on_ledger_seq
+                        && self
+                            .herder
+                            .tx_queue()
+                            .drop_stale_pending(&account_id, on_ledger_seq)
+                    {
+                        return self.herder.receive_transaction(tx);
+                    }
+                }
+            }
+        }
+
         // No explicit advert enqueue needed — flush_tx_adverts() reads
         // the herder's queue in priority order each flood period.
         result
--- a/crates/app/src/app/tx_flooding.rs
+++ b/crates/app/src/app/tx_flooding.rs
@@ -396,8 +396,15 @@ impl App {
             .tx_queue()
             .sample_aged_txs(Duration::from_secs(8), 3);
         if !aged.is_empty() {
+            const TAIL_WATCH_CAP: usize = 32;
             let adverts_by_peer = self.tx_adverts_by_peer.read().await;
+            let mut watch = self.tail_watch.write().await;
             for (hash, age_ms) in aged {
+                // Arm wire tracing for this stranded hash (#3719 part 1):
+                // the demand/response paths below log every event touching it.
+                if watch.len() < TAIL_WATCH_CAP {
+                    watch.insert(hash);
+                }
                 let sent_to = peer_ids
                     .iter()
                     .filter(|pid| {
@@ -557,6 +564,18 @@ impl App {
             // Parity: stellar-core mTxPullLatency (OverlayMetrics.h:115).
             metrics::histogram!("stellar_overlay_tx_pull_latency_seconds")
                 .record(delta.as_secs_f64());
+            // [maxtps_tail] #3719 part-1: a pull that took 3+ s means the tx
+            // eventually arrived after stranding — log so the demander side of
+            // a strand's timeline is reconstructable at info level.
+            if delta.as_secs() >= 3 {
+                tracing::info!(
+                    target: "maxtps_tail",
+                    hash8 = %&hash.to_hex()[..16],
+                    latency_ms = delta.as_millis() as u64,
+                    peers = entry.peers.len(),
+                    "slow tx pull completed"
+                );
+            }
             tracing::debug!(
                 hash = %hash.to_hex(),
                 latency_ms = delta.as_millis(),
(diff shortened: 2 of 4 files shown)


tomerweller:
## Board-watcher adversarial parity review vs core v27.0.0 — **MERGE-SAFE** (merge on Test green)

Both parts reviewed with MAX rigor (Part 2 touches the herder tx-queue + submission path that feeds nomination); self-verified the two hardest vectors.

**Part 2 — `drop_stale_pending` + submit heal: convergent, no consensus divergence.**
- Drops **only** genuinely-stale pendings: guard `if queued_tx.sequence_number() > on_ledger_seq { return false; }` — a live pending (seq > on-ledger) is untouched.
- A tx with `seq ≤ on_ledger_seq` fails `TxBadSeq` (needs `seq == account_seq+1`) → **un-nominatable in both henyey and core**; core already drops it in its synchronous close-cleanup. So dropping it changes no nominated tx-set or hash — pure convergence. The heal re-admits only the valid `seq+1` tx core would also accept (and it's the RPC/loadgen submission surface, not overlay nomination intake).
- **No fee double-release race:** `drop_stale_pending` takes `store.write()` then `account_states.write()` — same canonical order as `remove_applied`, so the two are mutually exclusive under the RwLock writers; the `state.transaction = None` / `Some(queued_tx) else return false` guard makes a second pass a no-op; fee release is `saturating_sub` mirroring `remove_applied` field-for-field (total_fees + soroban/classic counts + is_empty cleanup). No underflow, no double-remove.
- **No-ban is correct** (resubmit of the dropped hash carries the consumed seq → `TxBadSeq`; `remove_applied` still bans the applied hash on its own schedule). **Re-admit is bounded to one** (single `return receive_transaction(tx)`, no loop/recursion).

**Part 1 — unmark+requeue on send-fail:** demand/advert plane only, zero consensus impact. `entry.peers.remove(peer_id)` is idempotent (num_demanded is derived from len, no counter to underflow); `retry_incoming` is bounded (trims to max_ops); non-blocking `try_send_to`. Same family as #3717/#3718; fixes the phantom-demand that fed the sole-advertiser strand.

**Determinism scan:** tx_queue/mod.rs is +150/-0 (drop_stale_pending + accessor + 3 tests) — no edit to remove_applied/shift/selection ordering; nothing new hashed.

Genuine fix for the #3719 run-fatal wedge (post-close cleanup race + phantom demand). Required CI: Build/Clippy/Format/Analyze green; Test + Doctest running. Will merge on green (rerun if the known flaky `henyey-ledger::test_scan_and_merge_pending_buffer_bounded` reds).

_(board-watcher; core pin v27.0.0 7696c069; HEAD 8340680c)_

tomerweller:
Addendum — new commit `a762a962` ("Add sampled admission-rejection diagnostic") reviewed: **instrumentation-only, MERGE-SAFE stands.** It adds a sampled 1/16 `tracing::info!(target: maxtps_tail)` on Banned/Invalid/TryAgainLater admission outcomes in `herder.rs`; `diag_seq` is captured before `tx` moves into `try_add`, `result` is returned unchanged, and the sampler is a Relaxed AtomicU64 mod-16 — zero behavioral/consensus impact. `drop_stale_pending` seq-guard + lock order unchanged from the reviewed region. Will merge once Test + Doctest + Build(release-ci) + Analyze(rust) go green.

tomerweller:
## Live validation results (fresh nsc 32×64 instance `afedtgmo8ipco`, MissionMaxTPSClassic)

**Part 2 (stale-pending race) — validated fixed:**
- The self-heal fired 16 times in one 1600 tx/s sustained step (`dropped stale pending tx (seq already consumed on-ledger, #3719)`), each a real occurrence of the race.
- The pre-fix fatal signature (`pending_seq == on_ledger_seq`) did **not** reproduce in any validation run — every remaining reject shows a genuinely-pending tx (`pending_seq == on_ledger_seq + 1`).
- Sustained **1560 tx/s (campaign best, > core's 1526) passes green** on this stack.

**Part 1 (sole-advertiser strand) — narrowed, not fully eliminated:**
- 1600 tx/s steps still fail their warmup lottery on the pure strand: a fresh tx adverted to all 22 peers (`sent_to=22/22`) is not pulled for 10+ s, ages past 2 ledgers, and the account's next submission fatal-rejects.
- New admission-rejection sampling rules out receiver-side rejection: only ~3 sampled `Banned` across all 23 nodes for an entire run (~2M admissions) — flooded copies are not being dropped at the queue.
- Combined with the earlier forensics, the loss is isolated to the advert→demand→response transport during the warmup window. Note the structural context: at 1600 tx/s with 23000 accounts the per-account submission cadence (14.4 s ≈ 2.6 ledgers) equals the age-2 fatal tolerance, so any single missed inclusion round-trip is run-fatal — a measurement cliff that sits above stellar-core's own edge on this rig (1526).

Issue #3719 updated: part 2 closed by this PR; part 1 remains open with the narrowed attribution.

tomerweller:
Addendum — new commit `bfdc3544` ("Add watched-hash wire tracing for stranded tx pull round-trips") reviewed: **instrumentation-only, MERGE-SAFE stands.** Adds a `tail_watch` HashSet (armed by the tail scanner with 8s+-aged queued hashes, **bounded at TAIL_WATCH_CAP=32**) and info-level `maxtps_tail` logs on demand SERVED/DEFERRED/BANNED-here/UNKNOWN-here + slow-pull (≥3s). The logging is added *inside* the existing demand-response branches without altering control flow — the send/defer/ban/unknown accounting and non-blocking `try_send_to` sends are byte-for-byte unchanged. Event-loop-safe (tokio async RwLock on an already-`.await`ing path, cheap ≤32-entry clone); zero consensus/behavioral impact. `drop_stale_pending` seq-guard + lock order unchanged. (Minor: watch set never evicts — fills to 32 and stays, a benign diagnostic cap.) Head is now bfdc3544; will merge once Test + Doctest go green on the fresh run.

tomerweller:
Addendum — commit `4daf3651` ("Sharpen stranded-tx wire tracing: 3s arm, re-advert log, unsampled reject log") reviewed: **instrumentation-only, MERGE-SAFE stands.** Adds a watched-hash RE-ADVERTED info log, moves the tail-watch arm from 8s→3s (`sample_aged_txs` is read-only; its output only feeds the diagnostic watch set + logging, no flooding behavior change), and refactors the admission-rejection log (Banned/TryAgainLater unsampled, Invalid still 1/16). The `match result` uses only non-binding patterns so `result` is returned unchanged and the try_add admission control flow is untouched. `drop_stale_pending` seq-guard intact. Head now 4daf3651; letting CI settle on the latest SHA (you're mid-push-storm) — will merge once all 6 required checks go green on the final head.

tomerweller:
Validation update for the chunked SCP purge commit: under full-canonical-window load (MULT=1000, 1950 tx/s, 46k accounts) the purge chunking behaves as designed but is **necessary, not sufficient** — the run still collapses at ~8 min because the broader storestate write volume (multi-MB candidate tx sets persisted per slot) drives 24 s WAL checkpoints and an 87 s close. Keeping the commit (it bounds a real 21-34 s lock hold and the #2770 atomicity is preserved per chunk); the storestate architecture redesign is filed as the follow-up direction in #3719's closing analysis.

PR #3731, opened 2026-07-22
Suppress catchup escalation for single-ledger near-tip lag

### LINKED ISSUE #3728: metrics: henyey_recovery_stalled_tick_total{reason="forcing_catchup_behind"} — sustained breach
## Symptom

`henyey_recovery_stalled_tick_total{reason="forcing_catchup_behind"}` incremented for **3 consecutive monitor-tick intervals** (streak 3/3 → WARN), on deployed binary `0ac84d42` at mainnet L63563375 (uptime 18d 13h).

Counter progression across ticks (session 74535976):
- tick 2691: 148 → 149 (streak 1/3)
- tick 2692: 149 → 150 (streak 2/3)
- tick 2693: 150 → 151 (streak 3/3) — **fired**

## Evidence

Node is **fully healthy throughout** — this is NOT a SYNC FAILURE:
- `state=Validating`, qset **21/0/0** (agree/delayed/disagree), `heard_from_quorum=true`
- RPC `healthy`, ledger advancing ~+230/tick, heartbeat `gap` oscillates 0↔1
- RSS ~22 GB, frag ~20%, disk 157 G free

The counter increments correspond to log lines:

```
Recovery stalled for too long — forcing catchup current_ledger=63563139 latest_externalized=63563140 gap=1 attempts=8 cache_age_secs=Some(1259) urgent=true
Recovery stalled for too long — forcing catchup current_ledger=63563214 latest_externalized=63563215 gap=1 attempts=8 cache_age_secs=Some(419) urgent=true
```

Each event: the node momentarily sits `gap=1` behind the externalized tip, the recovery escalation reaches `attempts=8`, it forces a catchup tick (incrementing the counter), then recovers immediately. The healthy recovery path is exercised — `near_tip_peer_scp_recovery=171` — so the node never actually stalls; it re-syncs within the same interval.

Sibling counters at fire time:
```
forcing_catchup_behind          151
forcing_catchup_not_behind       21
near_tip_peer_scp_recovery      171
archive_behind_peer_ahead_hard_reset  0
at_tip_no_scp_hard_reset         0
backoff_active                   0
```

## Suspected root cause

Near-tip recovery-transition family (`crates/app/src/app/consensus.rs`, the "Recovery stalled for too long — forcing catchup" path). The node briefly falls one ledger behind at tip, the `attempts>=8` escalation fires `forcing_catchup_behind`, and peer-SCP recovery closes the gap. Over a sustained window these single-ledger transitions accrue ~1 increment per ~20-min tick — enough to trip the streak-3 gate without ever breaking real-time sync.

Related to #3218 (closed — overlay SCP broadcast backpressure / near-tip-stall recovery), #3197 (closed — post-restart consensus stall), and #3198 (closed — post-restart forcing_catchup_behind stall invisible to checks (2)/(12b), which added this very streak alarm). This recurrence differs: it is **not** post-restart (uptime 18d), so the #3198 post-restart-absolute path does not apply — this is the cross-tick streak path firing on a steady-state node.

## Suggested next step

Non-blocking. Worth confirming whether the `attempts=8` → `urgent=true forcing catchup` escalation threshold is too aggressive for a single-ledger near-tip gap on a long-lived validator (the node self-recovers via peer SCP every time, so the forced catchup may be redundant work). Candidate: raise the escalation attempt threshold or add a `gap<=1` carve-out before escalating to `forcing_catchup_behind`.

tomerweller:
Closes #3728

## Summary

A momentary `gap == 1` lag on an otherwise-healthy validator self-recovers via the peer-SCP back-fill path, but the attempt-counter escalation gate in `out_of_sync_recovery` forced a redundant catchup once `attempts >= RECOVERY_ESCALATION_CATCHUP`, incrementing `henyey_recovery_stalled_tick_total{reason="forcing_catchup_behind"}` and tripping the monitor's streak-3 alarm without ever breaking real-time sync. This extracts the escalation predicate into a pure helper `should_escalate_to_catchup` and adds a `gap > RECOVERY_ESCALATION_NEAR_TIP_GAP` (=1) condition so a pure single-ledger behind gap falls through to the existing request-SCP-state path instead of escalating. `gap >= 2` stalls, the captive-core `ahead_no_ext` path, the Step-2 fast-track, and all downstream AtTip/hard-reset paths are unchanged. Mirrors stellar-core `HerderImpl::outOfSyncRecovery`, which has no attempt-driven forced archive catchup.

**Safety:** escalation is only suppressed while the gap stays at exactly 1. A genuinely wedged node observes the externalized tip climb to N+2, N+3… within seconds, flipping the relation to `Behind { gap >= 2 }` and restoring escalation. The only suppressed case is a network-wide freeze at N+1, where there is nothing to catch up to.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3728#issuecomment-5050869518)

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test -p henyey-app` — 1167 lib tests pass, incl. full recovery/escalation suite (`test_out_of_sync_*`, `test_recovery_*`, `test_trigger_recovery_*`, hard-reset paths) unaffected
- [x] New coverage: 4 `should_escalate_to_catchup` unit tests pass

## New coverage

- `should_escalate_to_catchup_suppresses_single_ledger_gap` — `(8, Behind{gap:1}, false) == false` (exact #3728 scenario)
- `should_escalate_to_catchup_fires_on_multi_ledger_gap` — `(6, Behind{gap:2})` and `(8, Behind{gap:12})` still escalate
- `should_escalate_to_catchup_respects_attempt_floor` — `(5, Behind{gap:5}) == false`
- `should_escalate_to_catchup_preserves_ahead_no_ext` — `(6, Ahead, true) == true`, `(6, AtTip, false) == false`

## Deviations from plan

None. Helper implemented as a free function (as the plan permitted, matching the sibling `should_delay_checkpoint`/`classify_timer_event` pure classifiers in the same module).

### DIFF
--- a/.claude/skills/monitor-tick/SKILL.md
+++ b/.claude/skills/monitor-tick/SKILL.md
@@ -1391,6 +1391,23 @@ its own state machine because ratio checks are globally skipped during unsync
 states (ledger age > 30s, gap > 5, etc.), but the recovery-stalled counter fires
 precisely during recovery transitions when the node is briefly unsynced.
 
+> **#3728 firing scope (post-PR #3731):** A momentary single-ledger lag
+> (`gap == 1`) no longer increments `forcing_catchup_behind` on every recovery
+> tick — the escalation is suppressed while the gap stays momentary, so the
+> false-positive streak-3 breach that filed #3728 no longer occurs. The
+> suppression is **bounded**: a node genuinely wedged at exactly `gap == 1`
+> resumes firing `forcing_catchup_behind` once the stall persists past
+> `RECOVERY_ESCALATION_NEAR_TIP_GAP_STALL_ATTEMPTS` (12) consecutive
+> no-progress attempts, so this check still catches a truly-stuck near-tip node
+> — just after a short grace window rather than on the first tick. While the
+> gap stays suppressed, the node instead increments the distinct,
+> **non-alarming** `henyey_recovery_stalled_tick_total{reason="near_tip_gap1_suppressed"}`
+> counter. That label is **not** an alarm input (Check 12b keys only on
+> `forcing_catchup_behind`); it is a scrape-only observability signal — a
+> steadily-climbing `near_tip_gap1_suppressed` with a flat ledger indicates a
+> node parked at `gap == 1` and is worth manual investigation even before the
+> bounded backstop trips `forcing_catchup_behind`.
+
 **Data source:** Reuses the same `/metrics` scrape result (`$metrics_body` /
 `metrics/current.prom`) already fetched by check-12. Does NOT perform a
 second `/metrics` fetch.
--- a/crates/app/src/app/consensus.rs
+++ b/crates/app/src/app/consensus.rs
@@ -158,6 +158,44 @@ pub(super) fn classify_timer_event(
     }
 }
 
+/// Decide whether out-of-sync recovery should escalate to a forced catchup.
+///
+/// Pure predicate extracted from `out_of_sync_recovery` (#3728). Escalation
+/// requires the attempt floor `RECOVERY_ESCALATION_CATCHUP` to be crossed, and
+/// then one of:
+///   - `ahead_no_ext` — captive-core startup (Ahead with no externalization);
+///     escalate so recovery converges instead of looping on SCP requests.
+///   - the node is `Behind` by MORE than `RECOVERY_ESCALATION_NEAR_TIP_GAP`.
+///
+/// A pure single-ledger behind gap (`Behind { gap: 1 }`) is deliberately NOT
+/// escalated: it self-recovers via the peer-SCP back-fill path. It falls
+/// through to the `gap <= TX_SET_REQUEST_WINDOW` branch in
+/// `out_of_sync_recovery`, which requests SCP state from peers
+/// (`near_tip_peer_scp_recovery`). Note the Step-2 fast-track in that branch
+/// does NOT fire for `Behind { gap: 1 }` (it requires `AtTip` or
+/// ahead-no-ext), so the fall-through lands in the request-SCP-state path, not
+/// a second hidden catchup. See `RECOVERY_ESCALATION_NEAR_TIP_GAP` for the
+/// safety argument (gap grows past 1 → escalation restored).
+///
+/// `AtTip` / `Ahead` (with externalization) are handled downstream and never
+/// escalate here.
+pub(super) fn should_escalate_to_catchup(
+    attempts: u64,
+    relation: LedgerRelation,
+    ahead_no_ext: bool,
+) -> bool {
+    if attempts < RECOVERY_ESCALATION_CATCHUP {
+        return false;
+    }
+    if ahead_no_ext {
+        return true;
+    }
+    match relation.behind_gap() {
+        Some(gap) => gap > RECOVERY_ESCALATION_NEAR_TIP_GAP,
+        None => false,
+    }
+}
+
 impl App {
     /// Try to trigger consensus for the next ledger.
     ///
@@ -628,14 +666,20 @@ impl App {
         }
 
         // --- Escalation: after many failed attempts, force catchup ---
-        // Escalate when the node is behind consensus, OR when the node is
-        // in the "Ahead" state with no SCP externalization yet (latest_ext=0).
-        // The latter case covers captive-core in quickstart/local mode: it
-        // closes ledgers from the validator's EXTERNALIZE messages but never
-        // externalizes itself, so without this escalation the recovery loop
-        // would request SCP state forever without converging.
+        // Escalate when the node is behind consensus by MORE than a single
+        // ledger, OR when the node is in the "Ahead" state with no SCP
+        // externalization yet (latest_ext=0). The latter case covers
+        // captive-core in quickstart/local mode: it closes ledgers from the
+        // validator's EXTERNALIZE messages but never externalizes itself, so
+        // without this escalation the recovery loop would request SCP state
+        // forever without converging.
+        //
+        // A pure single-ledger behind gap (gap==1) is deliberately NOT
+        // escalated (#3728): it self-recovers via the peer-SCP back-fill path
+        // below, so a forced catchup is redundant work. See
+        // `should_escalate_to_catchup` and `RECOVERY_ESCALATION_NEAR_TIP_GAP`.
         let ahead_no_ext = relation.is_ahead_without_externalization(latest_externalized);
-        if attempts >= RECOVERY_ESCALATION_CATCHUP && (relation.is_behind() || ahead_no_ext) {
+        if should_escalate_to_catchup(attempts, relation, ahead_no_ext) {
             self.set_phase_sub(PHASE_13_10_TRIGGER_RECOVERY_CATCHUP);
             return self
                 .trigger_recovery_catchup(current_ledger, latest_externalized, relation, attempts)
@@ -2275,7 +2319,57 @@ impl App {
 
 #[cfg(test)]
 mod tests {
-    use super::{classify_cannot_apply_reason, CannotApplyReason, LedgerRelation};
+    use super::{
+        classify_cannot_apply_reason, should_escalate_to_catchup, CannotApplyReason, LedgerRelation,
+    };
+
+    // ── should_escalate_to_catchup tests (#3728) ────────────────────────
+
+    #[test]
+    fn test_should_escalate_to_catchup_suppresses_single_ledger_gap() {
+        // Exact issue #3728 scenario: attempts=8, gap==1 → no forced catchup.
+        // The node self-recovers via the peer-SCP back-fill path instead.
+        assert!(!should_escalate_to_catchup(
+            8,
+            LedgerRelation::Behind { gap: 1 },
+            false
+        ));
+    }
+
+    #[test]
+    fn test_should_escalate_to_catchup_fires_on_multi_ledger_gap() {
+        // A genuine multi-ledger stall still escalates once the attempt floor
+        // is crossed.
+        assert!(should_escalate_to_catchup(
+            6,
+            LedgerRelation::Behind { gap: 2 },
+            false
+        ));
+        assert!(should_escalate_to_catchup(
+            8,
+            LedgerRelation::Behind { gap: 12 },
+            false
+        ));
+    }
+
+    #[test]
+    fn test_should_escalate_to_catchup_respects_attempt_floor() {
+        // Below the attempt threshold, never escalate regardless of gap.
+        assert!(!should_escalate_to_catchup(
+            5,
+            LedgerRelation::Behind { gap: 5 },
+            false
+        ));
+    }
+
+    #[test]
+    fn test_should_escalate_to_catchup_preserves_ahead_no_ext() {
+        // Captive-core startup path (Ahead with no externalization) must still
+        // escalate so recovery converges.
+        assert!(should_escalate_to_catchup(6, LedgerRelation::Ahead, true));
+        // AtTip never escalates here — it is handled downstream.
+        assert!(!should_escalate_to_catchup(6, LedgerRelation::AtTip, false));
+    }
 
     // ── LedgerRelation tests ────────────────────────────────────────────
 
--- a/crates/app/src/app/consensus.rs
+++ b/crates/app/src/app/consensus.rs
@@ -168,14 +168,23 @@ pub(super) fn classify_timer_event(
 ///   - the node is `Behind` by MORE than `RECOVERY_ESCALATION_NEAR_TIP_GAP`.
 ///
 /// A pure single-ledger behind gap (`Behind { gap: 1 }`) is deliberately NOT
-/// escalated: it self-recovers via the peer-SCP back-fill path. It falls
-/// through to the `gap <= TX_SET_REQUEST_WINDOW` branch in
-/// `out_of_sync_recovery`, which requests SCP state from peers
-/// (`near_tip_peer_scp_recovery`). Note the Step-2 fast-track in that branch
-/// does NOT fire for `Behind { gap: 1 }` (it requires `AtTip` or
-/// ahead-no-ext), so the fall-through lands in the request-SCP-state path, not
-/// a second hidden catchup. See `RECOVERY_ESCALATION_NEAR_TIP_GAP` for the
-/// safety argument (gap grows past 1 → escalation restored).
+/// escalated *while it stays momentary*: it self-recovers via the peer-SCP
+/// back-fill path. It falls through to the `gap <= TX_SET_REQUEST_WINDOW`
+/// branch in `out_of_sync_recovery`, which requests SCP state from peers.
+/// Note the Step-2 fast-track in that branch does NOT fire for
+/// `Behind { gap: 1 }` (it requires `AtTip` or ahead-no-ext), so the
+/// fall-through lands in the request-SCP-state path, not a second hidden
+/// catchup. See `RECOVERY_ESCALATION_NEAR_TIP_GAP` for the safety argument
+/// (gap grows past 1 → escalation restored).
+///
+/// The `gap == 1` suppression is *bounded*, not unconditional (#3728 review
+/// follow-up): once such a gap has persisted for
+/// `RECOVERY_ESCALATION_NEAR_TIP_GAP_STALL_ATTEMPTS` consecutive no-progress
+/// attempts, escalation resumes. This restores the peer-independent
+/// archive-catchup backstop and the `forcing_catchup_behind` alarm coverage for
+/// a node genuinely wedged at exactly `gap == 1` (one whose SCP visibility has
+/// itself stalled, so the tip never climbs to `N+2` to flip the relation). The
+/// momentary blip (cleared by `attempts=8` in production) is still suppressed.
 ///
 /// `AtTip` / `Ahead` (with externalization) are handled downstream and never
 /// escalate here.
@@ -191,9 +200,35 @@ pub(super) fn should_escalate_to_catchup(
         return true;
     }
     match relation.behind_gap() {
-        Some(gap) => gap > RECOVERY_ESCALATION_NEAR_TIP_GAP,
+        // Multi-ledger behind gap — escalate immediately (unchanged).
+        Some(gap) if gap > RECOVERY_ESCALATION_NEAR_TIP_GAP => true,
+        // Single-ledger near-tip gap — suppress only while momentary; resume
+        // escalation once it persists past the stall threshold so a genuinely
+        // wedged node keeps its archive backstop + alarm coverage.
+        Some(_) => attempts >= RECOVERY_ESCALATION_NEAR_TIP_GAP_STALL_ATTEMPTS,
         None => false,
     }
+}
+
+/// Whether this recovery tick is suppressing catchup escalation *solely*
+/// because of the momentary single-ledger near-tip carve-out (#3728).
+///
+/// True iff the attempt floor is crossed, the node is `Behind` by exactly a
+/// near-tip gap (`<= RECOVERY_ESCALATION_NEAR_TIP_GAP`), and the persistent-stall
+/// threshold has not yet been reached — i.e. the exact tick where the old code
+/// would have fired `forcing_catchup_behind` but the carve-out holds it back.
+/// The call site increments the non-alarming `near_tip_gap1_suppressed` metric
+/// on these ticks so operators retain visibility into a node parked at `gap==1`
+/// without re-tripping the `forcing_catchup_behind` streak-3 alarm.
+pub(super) fn is_near_tip_gap_suppressed(
+    attempts: u64,
+    relation: LedgerRelation,
+    ahead_no_ext: bool,
+) -> bool {
+    attempts >= RECOVERY_ESCALATION_CATCHUP
+        && !ahead_no_ext
+        && matches!(relation.behind_gap(), Some(gap) if gap <= RECOVERY_ESCALATION_NEAR_TIP_GAP)
+        && !should_escalate_to_catchup(attempts, relation, ahead_no_ext)
 }
 
 impl App {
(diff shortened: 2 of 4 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

<details>

**Cycle:** 1 (no prior reviewer comments found).

**What the PR does:** Extracts the escalation predicate in `out_of_sync_recovery` (crates/app/src/app/consensus.rs) into a pure, unit-tested free function `should_escalate_to_catchup(attempts, relation, ahead_no_ext) -> bool`, and adds a new constant `RECOVERY_ESCALATION_NEAR_TIP_GAP = 1` (crates/app/src/app/mod.rs). Behavior change: a `Behind { gap: 1 }` relation no longer escalates to forced catchup once the attempt floor is crossed — it falls through to the existing peer-SCP back-fill request path instead. All other cases (`gap >= 2`, `ahead_no_ext`, `AtTip`, `Ahead` without `ahead_no_ext`) are unchanged from the prior `attempts >= RECOVERY_ESCALATION_CATCHUP && (relation.is_behind() || ahead_no_ext)` gate.

**Issue linkage / kind:** #3728, triaged as `enhancement (tuning/optimization)`, low severity. Not a `bug-fix` per the triage report, so the test-committed-before-fix ordering requirement does not apply; the triage's stated test obligation (new threshold must not fire on sustained single-ledger gaps, must still fire on genuine stalls) is met by the 4 new unit tests in the same commit.

**Correctness (angles A/B/C, via two independent sub-agent passes over the diff and surrounding `out_of_sync_recovery` body):**
- Gap boundary (`gap > RECOVERY_ESCALATION_NEAR_TIP_GAP`, i.e. `gap > 1`) matches the stated intent exactly; `Behind{gap}` is only ever constructed with `gap > 0` (`LedgerRelation::from_ledgers`), so there's no `gap == 0` edge case to worry about.
- Every case the old `relation.is_behind() || ahead_no_ext` gate escalated on is preserved except the deliberate `Behind{gap:1}` carve-out.
- `behind_gap() == None` (AtTip / Ahead-without-`ahead_no_ext`) resolving to `false` is not a regression — the old code's `is_behind()` was already `false` in exactly those cases, and the downstream fall-through branch (`map_or(true, |g| g <= TX_SET_REQUEST_WINDOW)`) already treated these identically.
- Single call site; `pub(super)` free-function placement is consistent with sibling pure classifiers (`should_delay_checkpoint`, `classify_cannot_apply_reason`) in the same module.
- No correctness bugs found by either sub-agent pass.

**Test coverage:** 4 new unit tests directly exercise the new predicate: suppression at `(8, Behind{gap:1}, false)` (the exact #3728 scenario), continued escalation at `gap=2` and `gap=12`, the attempt-floor gate, and the `ahead_no_ext`/`AtTip` preservation cases. This satisfies the triage's test obligation. Existing recovery/escalation test suite (per PR body: `test_out_of_sync_*`, `test_recovery_*`, `test_trigger_recovery_*`, hard-reset paths) reported unaffected.

**CI:** All required checks green (Build, Clippy, Format, Analyze ×3, CodeQL, Quickstart Harness, contract checks); `Test` and `Doctest` were still pending at review time but no failing checks.

**Minor, non-blocking suggestions (cleanup class — do not block merge):**
- `consensus.rs`: the `match relation.behind_gap() { Some(gap) => gap > N, None => false }` could use `relation.behind_gap().is_some_and(|gap| gap > RECOVERY_ESCALATION_NEAR_TIP_GAP)` for consistency with the existing `map_or(true, |g| g <= TX_SET_REQUEST_WINDOW)` idiom used ~30 lines later in the same function.
- Caller computes `ahead_no_ext` just to pass it into `should_escalate_to_catchup`; could pass `latest_externalized` and let the predicate call `relation.is_ahead_without_externalization(...)` itself. Cosmetic only — current form is also fine and arguably clearer at the call site.
- No test combines `ahead_no_ext: true` with a sub-floor `attempts` value; the interaction between the attempt floor and the `ahead_no_ext` fast path is implied but not directly asserted. Not blocking given the low-risk, low-severity nature of this change.

None of the above rise to a correctness or regression-risk concern; recommending approval.

</details>

tomerweller:
## 🔍 Reviewer: Risk

<details>

Scope: risk only (regressions, perf, API/data-format breaks, security, operational/monitoring/config/migration concerns). Correctness of the escalation predicate itself is out of scope for this pass.

### regression-risk — monitoring blind spot for the exact failure class Check 12b exists to catch

`trigger_recovery_catchup` is now skipped entirely whenever `relation == Behind { gap: 1 }` (any `attempts`). That function is the only place that increments `henyey_recovery_stalled_tick_total{reason="forcing_catchup_behind"}` (consensus.rs:1775) and `{reason="near_tip_peer_scp_recovery"}` (consensus.rs:1944). For a node parked at `gap == 1`, neither ever fires anymore — the only remaining signal between `attempts` 6 and 17 is a throttled `tracing::warn!` (consensus.rs:838-850), no metric.

`.claude/skills/monitor-tick/SKILL.md` Check 12b (lines 1158-1169) states explicitly: *"ratio checks are globally skipped during unsync states (ledger age > 30s, gap > 5, etc.), but the recovery-stalled counter fires precisely during recovery transitions when the node is briefly unsynced."* Check 12b's `metric-alarms.toml` entry keys the alarm solely on the `forcing_catchup_behind` label (delta_threshold=1, streak_threshold=3). `gap == 1` is below the `gap > 5` ratio-check threshold too, so a node genuinely wedged at `gap == 1` is now invisible to **both** Check 12 (ratio, needs gap>5/age>30s) and Check 12b (counter-streak, no longer increments) — until the much narrower `RECOVERY_HARD_RESET_ESCALATION_ATTEMPTS_NO_SCP` (=17) backstop happens to align with `peer_gap >= PEER_AHEAD_ESCALATION_THRESHOLD` (=3) and `archive_is_confirmed_behind`. The PR frames every prior increment as a false positive ("without ever breaking real-time sync"), but Check 12b was purpose-built to catch small-gap stalls the ratio checks miss — this change removes that coverage for the whole class, not just the false-positive instances. No update to the monitor skill / metric-alarms.toml accompanies this change to compensate (e.g. a lower-severity counter for the Step-3/4 "essentially caught up, no progress" path).

### regression-risk — loss of the peer-independent archive fallback at gap==1

Before this PR, any sustained `Behind` state (gap ≥ 1) with `attempts >= RECOVERY_ESCALATION_CATCHUP` (6) escalated to an archive-based catchup — a recovery path that does not depend on P2P peer health (HTTP history archive read). The sibling fast-track branch a few lines below (consensus.rs:762-823) documents a real, previously-encountered failure mode: tx_sets evicted from all peers' caches (~60s window) — but that fast-track only fires for `AtTip`/ahead-no-ext, never for `Behind { gap: 1 }`. If the same eviction condition coincides with a pure single-ledger gap (our SCP tracking is current at N+1, but the tx_set for N+1 becomes unavailable everywhere), the new code's only remaining recourse is `near_tip_peer_scp_recovery`'s peer-SCP retry — which is exactly the mechanism that already failed to produce the tx_set. The archive fallback, which doesn't share that failure mode, is now suppressed indefinitely as long as gap stays pinned at exactly 1.

The PR's safety argument ("a genuinely wedged node observes the tip climb to N+2, N+3… within seconds") holds only if the node's own SCP/externalize tracking keeps advancing while local apply lags. It does not hold if the node's visibility into fresh SCP traffic itself stalls (e.g., connected only to peers that are also not relaying anything past N+1, or a latent SCP-tracking freeze). In that narrower but plausible scenario, `gap` stays exactly 1 forever, and per the point above, it does so with no metric signal and no peer-independent recovery path.

### test-coverage gap

The 4 new tests cover only the pure `should_escalate_to_catchup` predicate in isolation. Nothing exercises the integration-level claims that actually matter here: that `forcing_catchup_behind`/`near_tip_peer_scp_recovery` stop incrementing for a sustained `gap == 1` stall, or that the `RECOVERY_HARD_RESET_ESCALATION_ATTEMPTS_NO_SCP` backstop still eventually engages for a genuinely-wedged `gap == 1` node. Per repo convention (CLAUDE.md: "every public function and significant code path should have corresponding tests"), the removed-metric and still-have-a-backstop claims are exactly what's worth a regression test, and neither is covered.

### config-risk — no operator-facing lever (minor)

`RECOVERY_ESCALATION_NEAR_TIP_GAP` is a hardcoded `const`, not configurable and not behind a flag, so this recovery-behavior change ships unconditionally to every deployment (mainnet validator, testnet, quickstart/captive-core) on upgrade — the only rollback path is reverting the binary. Given the change also moves behavior closer to stellar-core's `HerderImpl::outOfSyncRecovery` (a parity-relevant change per CLAUDE.md's "update PARITY_STATUS.md" rule), `crates/app/PARITY_STATUS.md` is untouched by this PR — though it doesn't currently document this area either, so this is a pre-existing gap more than a fresh regression. Flagging as low-severity / non-blocking on its own.

### Not flagged (checked, no concern)

- **perf-regression**: none — `should_escalate_to_catchup` is a straight-line predicate with no new allocations or loops; negligible cost added per recovery tick.
- **api-break**: none — new function is `pub(super)`, no public crate API or XDR/wire-format surface touched.
- **migration-risk**: none — no persisted state, DB schema, or config-file format changes.
- **security**: none identified — purely local control-flow logic, no new trust boundary.

### Bottom line

The underlying motivation (stop forcing a redundant catchup for a self-healing single-ledger blip) is reasonable and the PR is unusually careful about documenting its safety argument. But it trades a noisy-but-total metric signal for a narrower one, and the narrowing lands exactly on the class of stall (`gap` small, briefly unsynced) that Check 12b in monitor-tick was built to detect independently of the coarser ratio checks. Recommend: add a distinct (lower-severity, non-alarming or separately-thresholded) counter increment on the Step 3/4 "essentially caught up, no progress" path so operators keep equivalent visibility into a stuck-at-gap-1 node without reintroducing the false-positive streak-3 alarm, and/or update monitor-tick's Check 12b docs/thresholds to reflect the narrower firing conditions of `forcing_catchup_behind` post-change.

</details>

tomerweller:

**Reason:** Reviewer B (Risk) requested changes.

**Reviewer A (Correctness):** APPROVE — pure predicate extraction is correct; gap boundary, preserved cases, and `None`-branch fallthrough all check out; 4 new unit tests satisfy the triage test obligation. Three minor non-blocking nits noted inline only.
**Reviewer B (Risk):** CHANGES_REQUESTED — suppressing escalation at `gap==1` removes observability/backstop coverage for a genuinely-wedged single-ledger node.
**External reviewers:** none
**CI:** running (Format/Clippy/CodeQL/Quickstart/Build green; Test, Doctest, and core build still in progress at review time)

### Reviewer B concerns to address (`/do` Mode B)

- **regression-risk (primary):** Suppressing escalation at `gap==1` means `henyey_recovery_stalled_tick_total{reason="forcing_catchup_behind"}` never fires for that state. monitor-tick Check 12b exists specifically to catch small-gap "briefly unsynced" stalls that the coarser ratio checks (`gap>5`, `age>30s`) skip. A genuinely wedged node at `gap==1` becomes invisible to both alarms unless the narrow hard-reset backstop (attempts≥17 AND peer_gap≥3 AND archive-confirmed-behind) happens to align.
- **regression-risk:** The suppressed path removes the peer-connectivity-independent archive-catchup fallback for `gap==1`, which previously covered the documented "tx_set evicted from all peers' caches" failure mode when it coincides with a pure single-ledger gap (the existing fast-track only covers AtTip / ahead-no-ext, not `Behind{1}`).
- **test-coverage:** New tests cover only the pure predicate in isolation; nothing verifies the metric stops firing or that the hard-reset backstop still engages for a real wedge.
- **config-risk (minor):** hardcoded const with no operator flag/rollback lever; PARITY_STATUS.md untouched despite parity-relevant framing.

**Recommended:** add a distinct (non-alarming) counter for the Step 3/4 "no progress, requesting SCP" path so operators retain equivalent visibility without reviving the noisy false-positive alarm, and/or update monitor-tick Check 12b to reflect the narrower firing conditions.

tomerweller:
## Addressed Reviewer B (Risk) feedback in f0e619f0660694722710c2bd1029d9c2eaf26c37

Thanks — the two blocking concerns were real, not false positives. The unconditional `gap == 1` carve-out did remove both the observability signal and the peer-independent archive backstop for a genuinely-wedged single-ledger node. Fixed by making the suppression **bounded** and adding a distinct non-alarming metric, so a momentary blip stays quiet but a truly-stuck node keeps full coverage.

### 1. Monitoring blind spot (Check 12b) → bounded suppression + non-alarming counter
- `should_escalate_to_catchup` now **resumes** escalation for a `gap == 1` that persists past `RECOVERY_ESCALATION_NEAR_TIP_GAP_STALL_ATTEMPTS` (= 12 — above the observed self-healing blip at `attempts=8`, and `<=` the `=17` no-SCP hard reset, guarded by a const-invariant test). A genuinely wedged `gap == 1` node therefore re-fires `forcing_catchup_behind` after a short grace window, so **Check 12b still catches it** — it is no longer invisible for the whole class.
- On the suppressed ticks (the exact ticks where the old code fired `forcing_catchup_behind`) the node now increments a distinct, **non-alarming** `henyey_recovery_stalled_tick_total{reason="near_tip_gap1_suppressed"}` counter. Check 12b keys only on `forcing_catchup_behind`, so this label is scrape-only observability — a steadily-climbing `near_tip_gap1_suppressed` with a flat ledger flags a node parked at `gap==1` **before** the bounded backstop trips the alarm. No streak-3 false positive is reintroduced.

### 2. Loss of the peer-independent archive fallback → restored via the same bound
The archive-based `trigger_recovery_catchup` path (which does not share the peer-SCP failure mode you described — tx_set evicted from all peers at `N+1`) is now re-enabled for a persistently-stuck `gap == 1` at the same `STALL_ATTEMPTS` threshold. Your narrower-but-plausible scenario (node's own SCP visibility stalls so the tip never climbs to `N+2` to flip the relation) is exactly what the bound covers: the gap stays pinned at 1, but escalation now still fires — just after a bounded grace window instead of never.

### 3. Test-coverage gap
Added:
- `test_should_escalate_to_catchup_resumes_on_persistent_single_ledger_stall` — the backstop re-fires at/above the stall threshold.
- `test_is_near_tip_gap_suppressed_identifies_carveout_ticks` — the observability counter fires on exactly the suppressed ticks and not otherwise.
- const-invariant assertion in `test_peer_ahead_escalation_constants` (floor < stall <= no-SCP hard reset).
- metric pre-registration test now covers `near_tip_gap1_suppressed`.

### 4. Config lever / PARITY (minor, non-blocking)
Left `STALL_ATTEMPTS` as a documented `const` consistent with the sibling recovery thresholds (`RECOVERY_ESCALATION_CATCHUP`, `RECOVERY_HARD_RESET_ESCALATION_ATTEMPTS_NO_SCP`), none of which are operator-configurable; introducing a one-off config knob here would be inconsistent. Also updated monitor-tick Check 12b docs with the narrowed firing scope and the new signal.

Full suite green: `cargo fmt --check`, `cargo clippy --all -- -D warnings`, `cargo test -p henyey-app` (1169 lib tests), and `scripts/test-monitor-skill-snippets.sh`.

tomerweller (commented):
Addressed Reviewer B (Risk) feedback in f0e619f0 — bounded gap==1 suppression + non-alarming observability counter. Ready for re-review.

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The bounded-suppression rewrite of `should_escalate_to_catchup` is correct, the new `is_near_tip_gap_suppressed` observability predicate is precisely the complement of the suppressed window, and the added tests directly discharge the triage test obligation. All 51 consensus + 47 metrics tests pass on the PR head.

<details>

Cycle 1 (no prior `## 🔍 Reviewer: Correctness` comment on this PR — this is my complete change-list). Linked issue #3728 triage `Kind: enhancement (tuning/optimization)`, so this is evaluated as feature-style: new public surface must be tested; no bug-fix regression-test-first obligation applies.

### test-coverage — SATISFIED
Triage test obligation was: (a) new threshold does NOT fire on sustained single-ledger near-tip gaps, and (b) threshold DOES still fire on genuine stalls. Both are covered:
- `test_should_escalate_to_catchup_suppresses_single_ledger_gap` (attempts=8, gap=1 → false) covers (a) for the momentary window.
- `test_should_escalate_to_catchup_resumes_on_persistent_single_ledger_stall` (stall-1 → false; stall, stall+5 → true) covers (a)/(b) at the persistence boundary.
- `test_should_escalate_to_catchup_fires_on_multi_ledger_gap` covers (b) for gap≥2.
- New `pub(super) fn is_near_tip_gap_suppressed` has a dedicated test `test_is_near_tip_gap_suppressed_identifies_carveout_ticks` exercising: the #3728 tick (attempts=8/gap=1 → true), below-floor (attempts=5 → false), at-stall-threshold (→ false, escalation resumes), multi-ledger gap (→ false), AtTip (→ false), Ahead-no-ext (→ false). Complete truth-table for the new public predicate.
- Const-invariant test in `mod.rs` asserts `RECOVERY_ESCALATION_CATCHUP (6) < RECOVERY_ESCALATION_NEAR_TIP_GAP_STALL_ATTEMPTS (12) <= RECOVERY_HARD_RESET_ESCALATION_ATTEMPTS_NO_SCP (17)`, protecting the ordering the whole design rests on against future const drift.
- Metric pre-registration test now includes `near_tip_gap1_suppressed`, so the label is registered-at-zero (no first-scrape gap).

Verified locally on `pr-3731` head (f0e619f0): `cargo test -p henyey-app --lib app::consensus::tests::` → 51 passed / 0 failed; `metrics::tests` → 47 passed / 0 failed.

### logic correctness — no concern
- `should_escalate_to_catchup`: the `match relation.behind_gap()` arms are exhaustive and correctly ordered — `Some(gap) if gap > NEAR_TIP_GAP => true` (multi-ledger, unchanged) precedes `Some(_) => attempts >= STALL_ATTEMPTS` (single-ledger bounded suppression), then `None => false`. The `attempts < RECOVERY_ESCALATION_CATCHUP` early-return floor above the match is preserved, so the bounded path only engages after the floor is crossed.
- `is_near_tip_gap_suppressed` is the exact set-complement of an escalated tick within the near-tip band: it re-uses `!should_escalate_to_catchup(...)` rather than re-deriving the predicate, so the metric can never fire on a tick that actually escalated (no double-counting against `forcing_catchup_behind`). Reusing the predicate instead of duplicating the threshold comparison is the right call — it stays correct if the threshold logic changes.
- Call-site: the suppression counter increment sits after the `should_escalate_to_catchup` branch, gated by `is_near_tip_gap_suppressed(...)`. Because the latter includes `!should_escalate_to_catchup(...)`, the two are mutually exclusive on any given tick. Correct.

### error-handling / readability — no concern
- No new fallible paths; the predicate is pure. Counter `.increment(...)` is infallible.
- Doc comments on the const, both functions, and the monitor-tick Check 12b block are thorough and accurately describe the threshold ordering and the scrape-only (non-alarming) nature of the new label.

### Nits (non-blocking, not filed)
- `is_near_tip_gap_suppressed` recomputes `should_escalate_to_catchup` at the call site (once in the `if` branch above, once inside the predicate). Negligible — a straight-line integer comparison per recovery tick (~1s cadence). Not worth a refactor.

No changes requested.
</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Cycle 2. The bounded-suppression + observability rework (f0e619f0) genuinely closes the gap I raised in cycle 1 — it is a real control-flow change (escalation *resumes* at a wedged node), not cosmetic metrics. All three cycle-1 concern classes are resolved; no new class discovered.

<details>

Cycle 2 (prior `## 🔍 Reviewer: Risk` CHANGES_REQUESTED exists on this PR). Scope: risk only. My cycle-1 concern classes were `regression-risk` (two instances: monitoring blind spot + loss of archive fallback), `test-coverage`, and `config-risk`. I stick to those below and verify each.

### regression-risk (monitoring blind spot) — RESOLVED
Cycle-1 concern: a node wedged at `gap == 1` became invisible to both Check 12 (ratio, needs gap>5/age>30s) and Check 12b (counter-streak keyed on `forcing_catchup_behind`, which no longer incremented), until the far-narrower attempts=17 no-SCP hard reset.

The fix genuinely restores coverage on BOTH axes, and I confirmed the mechanism is real rather than cosmetic:
- **Backstop alarm restored (not cosmetic):** `should_escalate_to_catchup` now returns `attempts >= RECOVERY_ESCALATION_NEAR_TIP_GAP_STALL_ATTEMPTS (12)` for the single-ledger arm. This is the actual escalation predicate — once a gap==1 persists past 12 no-progress attempts, `trigger_recovery_catchup` fires again, which re-increments `forcing_catchup_behind` (consensus.rs). So Check 12b's streak-3 alarm *does* eventually catch a genuinely-wedged gap==1 node — just after a ~grace window instead of on the first tick. This is exactly the compensating mechanism I asked for, implemented in control flow, not as a decorative counter.
- **Scrape-only signal for the suppressed window:** the distinct `near_tip_gap1_suppressed` counter fires on precisely the suppressed ticks (`is_near_tip_gap_suppressed` = complement of escalation within the band). It is NOT an alarm input — Check 12b keys only on `forcing_catchup_behind` — so it cannot re-trip the streak-3 false positive that filed #3728, while still giving operators a scrape-visible "parked at gap==1" signal. The monitor-tick Check 12b doc block was updated to document the narrowed firing scope and this new signal. That was my specific cycle-1 ask ("update monitor-tick's Check 12b docs/thresholds") — done.

### regression-risk (loss of peer-independent archive fallback) — RESOLVED
Cycle-1 concern: the unconditional carve-out suppressed the archive-based (P2P-independent) catchup indefinitely as long as gap stayed pinned at exactly 1, including the plausible scenario where the node's own SCP visibility stalls so the tip never climbs to N+2.

The bounded suppression fixes exactly this. When escalation resumes at attempts>=12, control returns to the normal `trigger_recovery_catchup` path, which includes the archive-based recovery that does not depend on the same peer-SCP mechanism that (in the stalled-visibility scenario) already failed to produce the tx_set. The const's threshold ordering is enforced by a test: `12 <= RECOVERY_HARD_RESET_ESCALATION_ATTEMPTS_NO_SCP (17)`, so the archive backstop engages strictly BEFORE the narrower no-SCP hard-reset path — the sequencing I was worried about is now guaranteed by an invariant guard, not left implicit.

### test-coverage — RESOLVED
Cycle-1 concern: only the pure predicate was tested; the "metric stops/resumes" and "backstop still engages" claims were untested. The follow-up adds:
- `test_should_escalate_to_catchup_resumes_on_persistent_single_ledger_stall` — directly asserts the backstop re-engages at the threshold (my "still-have-a-backstop" ask).
- `test_is_near_tip_gap_suppressed_identifies_carveout_ticks` — asserts the suppressed-tick set (where the observability metric fires and `forcing_catchup_behind` does NOT), covering the "removed-metric" claim.
- The const-invariant test locks the `6 < 12 <= 17` ordering.
- Metric-registration test covers the new label.
This is adequate for a risk sign-off. (An integration-level test that a live recovery loop actually emits the counter would be nice-to-have, but the predicate-level coverage plus the call-site being a one-line gated `.increment` is a low residual risk; not blocking.)

### config-risk — UNCHANGED, still non-blocking (as flagged in cycle 1)
`RECOVERY_ESCALATION_NEAR_TIP_GAP_STALL_ATTEMPTS` remains a hardcoded const, not an operator lever, same as the sibling recovery consts. I flagged this as low-severity/non-blocking in cycle 1 and it stays that way — consistent with the existing hardcoded recovery-tuning consts in this module; rollback path is a binary revert, which matches every other const here. `PARITY_STATUS.md` is still untouched, but as noted in cycle 1 this area isn't documented there today, so it's a pre-existing gap, not a fresh regression introduced by this PR. Not a blocker.

### No new class discovered
I re-checked perf (still a straight-line predicate; one extra integer comparison + a conditional counter increment per ~1s recovery tick — negligible), api-break (new fn is `pub(super)`, no wire/XDR/public surface), migration-risk (no persisted state), and security (pure local control flow). Nothing new.

### Bottom line
My cycle-1 CHANGES_REQUESTED is fully addressed. The suppression is now bounded so a genuinely-wedged gap==1 node regains BOTH the peer-independent archive backstop AND the `forcing_catchup_behind` alarm after a short grace window, while the false-positive streak-3 that motivated #3728 stays suppressed and gets a distinct non-alarming scrape signal. This is a substantive mechanism, not cosmetic metrics. Approve.
</details>

tomerweller:
## ✅ Merged

**Commit:** 49c9d4de5b2e6b3f49363922ea3f52fa27cfd5cb

Reviewer A (Correctness): APPROVE. Reviewer B (Risk, cycle 2): APPROVE — cycle-1 CHANGES_REQUESTED fully addressed by f0e619f0 (bounded gap==1 suppression + non-alarming observability). CI: 41/41 green.

PR #3734, opened 2026-07-22
Make storestate prefix scans index-usable to bound SCP-purge write lock

tomerweller:
Contributes to #3702

## Summary

The three `storestate` prefix SELECTs (`load_all_scp_slot_states`, `load_all_tx_set_data`, `get_all_tx_set_hashes`) used case-insensitive `LIKE 'prefix%'`. With `PRAGMA case_sensitive_like` OFF (default) and the `storestate` primary-key autoindex collated BINARY, SQLite cannot apply the LIKE-index optimization, so each query plans as a **full-table SCAN**. Two of these run inside `purge_unreferenced_tx_sets_atomic`'s `BEGIN IMMEDIATE`, holding the single WAL write lock for the whole scan during near-tip back-fill — the SCP-intake-starvation mechanism behind #3702.

This rewrites each to a half-open BINARY range `statename >= ?1 AND statename < ?2` (bounds from a new `prefix_range()` helper that increments the prefix's final byte, `txset:` → `[txset:, txset;)`). The range plans as an index **SEARCH** and returns a byte-for-byte identical row set (all stored keys are the literal lowercase prefix followed by lowercase hex or a decimal slot), so no observable behavior changes — only the query plan.

Each rewritten SELECT lives in a module-level `const` shared between the production query and its EQP test, so the two can't drift (Critic A).

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3702#issuecomment-5051976855)

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy -p henyey-db --all-targets -- -D warnings`
- [x] `cargo test -p henyey-db` — 116 passed

- **Tests:** `crates/db/src/queries/scp.rs::test_txset_prefix_query_uses_index_not_scan` and `::test_scpstate_prefix_query_uses_index_not_scan` (EXPLAIN QUERY PLAN asserts SEARCH, not `SCAN storestate`)
- **Pre-fix:** committed as `9b7fbd9` — verified FAILED with `SCAN storestate USING [COVERING] INDEX sqlite_autoindex_storestate_1`
- **Post-fix:** verified PASSES after `bb4d48c` (`SEARCH ... (statename>? AND statename<?)`)
- Plus `test_prefix_range_upper_bound` and `test_prefix_range_membership_excludes_adjacent` (range membership equivalence vs. old LIKE).

## Deviations from plan

None.

## Parity

n/a — internal SQLite query-plan optimization only (`docs/PARITY.md` permits internal DB/perf deviation). Row sets are byte-identical, so persisted SCP state, tx-set GC selection, and ledger/bucket/result/meta hashes are unaffected.

## Note on closure

Per the plan (Critic C), this is candidate holder (B), following (A)+instrumentation in #3704. It **contributes to** #3702; closure waits on #3704's `WriteCtxGuard` instrumentation confirming the write-lock hold is bounded on a real restore.

### DIFF
--- a/crates/db/src/queries/scp.rs
+++ b/crates/db/src/queries/scp.rs
@@ -27,6 +27,25 @@ use super::state::StateQueries;
 
 const TX_SET_KEY_PREFIX: &str = "txset:";
 
+/// SQL for [`ScpStatePersistenceQueries::load_all_scp_slot_states`].
+///
+/// Shared with its query-plan regression test
+/// (`test_scpstate_prefix_query_uses_index_not_scan`) so the two can never
+/// drift onto different SQL.
+const LOAD_ALL_SCP_SLOT_STATES_SQL: &str =
+    "SELECT statename, state FROM storestate WHERE statename LIKE ?1 ORDER BY statename";
+
+/// SQL for [`ScpStatePersistenceQueries::load_all_tx_set_data`].
+const LOAD_ALL_TX_SET_DATA_SQL: &str =
+    "SELECT statename, state FROM storestate WHERE statename LIKE ?1";
+
+/// SQL for [`ScpStatePersistenceQueries::get_all_tx_set_hashes`].
+///
+/// Shared with its query-plan regression test
+/// (`test_txset_prefix_query_uses_index_not_scan`).
+const GET_ALL_TX_SET_HASHES_SQL: &str =
+    "SELECT statename FROM storestate WHERE statename LIKE ?1";
+
 /// Query trait for SCP consensus state operations.
 ///
 /// Provides methods for persisting and retrieving SCP envelopes and quorum sets.
@@ -379,9 +398,7 @@ impl ScpStatePersistenceQueries for Connection {
     }
 
     fn load_all_scp_slot_states(&self) -> Result<Vec<(u64, String)>, DbError> {
-        let mut stmt = self.prepare(
-            "SELECT statename, state FROM storestate WHERE statename LIKE ?1 ORDER BY statename",
-        )?;
+        let mut stmt = self.prepare(LOAD_ALL_SCP_SLOT_STATES_SQL)?;
         let pattern = format!("{}:%", state_keys::SCP_STATE);
         let rows = stmt.query_map(params![pattern], |row| {
             let key: String = row.get(0)?;
@@ -434,8 +451,7 @@ impl ScpStatePersistenceQueries for Connection {
     }
 
     fn load_all_tx_set_data(&self) -> Result<Vec<(Hash, Vec<u8>)>, DbError> {
-        let mut stmt =
-            self.prepare("SELECT statename, state FROM storestate WHERE statename LIKE ?1")?;
+        let mut stmt = self.prepare(LOAD_ALL_TX_SET_DATA_SQL)?;
         let pattern = format!("{TX_SET_KEY_PREFIX}%");
         let rows = stmt.query_map(params![pattern], |row| {
             let key: String = row.get(0)?;
@@ -474,7 +490,7 @@ impl ScpStatePersistenceQueries for Connection {
     }
 
     fn get_all_tx_set_hashes(&self) -> Result<Vec<Hash>, DbError> {
-        let mut stmt = self.prepare("SELECT statename FROM storestate WHERE statename LIKE ?1")?;
+        let mut stmt = self.prepare(GET_ALL_TX_SET_HASHES_SQL)?;
         let pattern = format!("{TX_SET_KEY_PREFIX}%");
         let rows = stmt.query_map(params![pattern], |row| {
             let key: String = row.get(0)?;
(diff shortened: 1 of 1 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The half-open BINARY range rewrite is correct and result-equivalent to the `LIKE 'prefix%'` it replaces for both call-site prefixes; regression coverage is strong (query-plan test + membership/adjacency test + upper-bound test), test-before-fix ordering is honored, and all 16 `henyey-db` scp tests pass.

<details>

**Change kind:** bug-fix contributing to urgent #3702. Regression test committed first (`9b7fbd94`, "Regression test for #3702"), fix second (`bb4d48c3`). Correct ordering.

**Result-set equivalence (class: correctness) — verified, no divergence:**
- Old form: `statename LIKE 'prefix%'` with SQLite default `case_sensitive_like=OFF` → case-INSENSITIVE match.
- New form: `statename >= lower AND statename < upper` on a BINARY-collated PK → case-SENSITIVE.
- The rewrite is only equivalent if stored keys never contain uppercase in the matched region. Verified:
  - `SCP_STATE = "scpstate"` (literal lowercase, `crates/db/src/schema.rs:184`); keys are `scpstate:{slot}` where slot is a decimal `u64` (digits only). No case.
  - `txset:` keys are `txset:{hex::encode(hash)}` — `hex::encode` emits lowercase hex. No uppercase.
  - Therefore dropping LIKE's case-insensitivity cannot change the row set for these two call sites. The doc comment on `prefix_range` states exactly this and it holds.

**Query-plan improvement — independently confirmed at the planner level:**
- Ran `EXPLAIN QUERY PLAN` on both forms against the real `storestate` schema:
  - OLD `LIKE 'txset:%'` → `SCAN storestate USING COVERING INDEX sqlite_autoindex_storestate_1` (full index scan, visits every row).
  - NEW range → `SEARCH storestate USING COVERING INDEX ... (statename>? AND statename<?)` (bounded seek).
- This is the exact O(n)→O(log n + matches) improvement the PR claims, and it is the two SELECTs inside `purge_unreferenced_tx_sets_atomic`'s `BEGIN IMMEDIATE` that hold the WAL write lock, so the fix directly targets the #3702 starvation mechanism.

**Test coverage (class: test-coverage) — strong:**
- `test_txset_prefix_query_uses_index_not_scan` / `test_scpstate_prefix_query_uses_index_not_scan`: assert `SEARCH` present and `SCAN storestate` absent. I confirmed the old `LIKE` form yields `SCAN storestate ...` and the new form yields `SEARCH ...`, so these tests genuinely fail on the pre-fix SQL — a real regression guard, not a tautology.
- `test_prefix_range_membership_excludes_adjacent`: inserts `txset:00`, `txset:ff`, `txset;`, `txseu` and asserts only the two `txset:`-prefixed rows match — directly proves the exclusive upper bound excludes the adjacent `;` sentinel and the next-prefix key. This is the equivalence-vs-LIKE guard.
- `test_prefix_range_upper_bound`: pins `txset:`→`txset;` and `scpstate:`→`scpstate;`.
- All 16 tests in `queries::scp::tests` pass locally (`cargo test -p henyey-db`).
- Shared SQL consts bound to both the query and its plan test prevent SQL/test drift — good hygiene.

**Error handling (class: error-handling) — minor, non-blocking:**
- `prefix_range` does `*last += 1` on the final byte and `expect`s valid UTF-8. For the two current callers (both `:`-terminated, 0x3A→0x3B) this is safe and the code documents it. A future caller passing a prefix whose last byte is `0xFF`, or a multi-byte UTF-8 trailing codepoint, would panic (debug overflow / `from_utf8` failure). This is not a bug for the shipped call sites but is a latent sharp edge. Filing as a non-blocking inline note rather than blocking the merge.

No correctness blockers.
</details>

crates/db/src/queries/scp.rs
```diff
@@ -302,6 +324,34 @@ fn tx_set_key(hash: &Hash) -> String {
...
+///
+/// The two call sites pass `txset:` and `scpstate:`, both ending in `:` (0x3A),
+/// so the increment is `:` -> `;` (0x3B) with no carry and stays valid ASCII.
+fn prefix_range(prefix: &str) -> (String, String) {
+    let mut upper = prefix.as_bytes().to_vec();
+    let last = upper
+        .last_mut()
+        .expect("prefix_range requires a non-empty prefix");
+    // Callers pass ':'-terminated prefixes; the increment never carries.
+    *last += 1;
```
tomerweller, line 350: Non-blocking (class: `robustness`): `*last += 1` will panic on debug (u8 overflow) — or `String::from_utf8` will fail — for any future caller whose prefix ends in `0xFF` or in a multi-byte UTF-8 trailing codepoint. Both current callers pass `:`-terminated ASCII prefixes (0x3A→0x3B, no carry), so this is safe today and the doc comment says so. Suggest a follow-up hardening: increment the last byte < 0xFF, else strip trailing 0xFF bytes and increment the preceding byte (standard prefix-successor algorithm), falling back to an open upper bound when the whole prefix is 0xFF. Filing as backlog, not a merge blocker.

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Low-risk, well-scoped persist-layer query-planning fix confined to one file (`crates/db/src/queries/scp.rs`). No schema migration, no wire/protocol change, no API-shape change; behavior is provably identical to the replaced `LIKE` for both call sites while eliminating a full-index scan under the WAL write lock.

<details>

**Scope (class: blast-radius):** Single file, three read queries + one private helper + tests. `prefix_range` is a module-private `fn` with exactly two call sites (`txset:`, `scpstate:`), both internal. No public API surface changes, no signature changes to the `ScpStatePersistenceQueries` trait methods. Blast radius is minimal.

**Regression risk (class: regression-risk) — low, actively guarded:**
- The only behavioral change is case-insensitive LIKE → case-sensitive BINARY range. As Reviewer A verified, all stored keys under both prefixes are lowercase-literal + lowercase-hex/decimal, so the row set is unchanged. `test_prefix_range_membership_excludes_adjacent` locks this in (adjacent `txset;` and `txseu` correctly excluded).
- `ORDER BY statename` is preserved on `load_all_scp_slot_states`; the other two never ordered and still don't. No ordering regression.
- No `PRAGMA case_sensitive_like` is toggled — the fix is local to these queries and cannot affect any other LIKE in the codebase.

**Data-format / migration risk (class: migration-risk):** None. No schema change; the `storestate` table and its BINARY PK autoindex are unchanged. The fix exploits the *existing* index better; it does not add/drop indexes, so no migration and nothing to roll back. Fully compatible with existing on-disk DBs.

**Performance (class: perf) — this is the point of the PR:** Confirmed via `EXPLAIN QUERY PLAN`: old `LIKE` planned as `SCAN storestate USING COVERING INDEX` (whole-index traversal); new range plans as `SEARCH ... (statename>? AND statename<?)` (bounded seek). For the two SELECTs inside `purge_unreferenced_tx_sets_atomic`'s `BEGIN IMMEDIATE`, this shortens the WAL write-lock hold from O(table) to O(matches + log n), which is the direct #3702 SCP-intake-starvation mitigation. No perf downside for the third (read-only) call site either.

**Operational (class: operational):** No config, no flags, no telemetry change. Consistent with the plan's framing: this is contributing fix (B); #3702 closure correctly gated on #3704's `WriteCtxGuard` instrumentation confirming the bounded hold on a real restore — this PR does not claim to close #3702 (uses "Contributes to").

**Residual risk:** The `prefix_range` 0xFF-overflow sharp edge (Reviewer A's inline note) is latent, not reachable from either call site. Non-blocking; suitable as a follow-up hardening issue.

No risk blockers.
</details>

tomerweller:
## Review: Waiting on CI

Both agent reviewers posted **APPROVE**:
- **Reviewer A (Correctness):** APPROVE — range rewrite is result-equivalent to the replaced `LIKE` (keys are lowercase-literal + lowercase-hex/decimal, so case-sensitivity change is a no-op); query-plan regression test genuinely fails on the old SQL; all 16 `henyey-db` scp tests pass.
- **Reviewer B (Risk):** APPROVE — single-file, no schema migration, no API/wire change; converts a full covering-index SCAN under the WAL write lock into a bounded index SEARCH (the #3702 mitigation).

**CI:** running — 40/41 checks green; the sole outstanding check is the `test (testnet, ...)` Quickstart shard (IN_PROGRESS, age 23 min / budget 60 min). This is the known chronic testnet-Quickstart flake (de-gated per #3273), not diff-attributable. No red checks.

Not merging while a check is in progress. Re-picking on the next tick to verify completion.

Bounce-back count: 0/3.

One non-blocking inline note filed on `prefix_range` (0xFF/multi-byte-UTF8 last-byte overflow is latent, unreachable from the two `:`-terminated call sites) — will become a follow-up backlog issue at merge time.

tomerweller:
## ✅ Merged

**Commit:** 1b109f2bbff1d3a63dd0a05e427a1bf12f5b7992

**Follow-up issues filed for unaddressed inline review comments:** #3735 (`prefix_range()` 0xFF / multi-byte UTF-8 last-byte hardening — class `robustness`, non-blocking)

**Decision matrix:** Reviewer A (Correctness) APPROVE · Reviewer B (Risk) APPROVE · required CI green (Test, Doctest, Clippy, Format, Build (release-ci), Analyze (rust), + all others). The sole red check — non-required `test (testnet, ...)` Quickstart shard — was confirmed as the known chronic testnet-Quickstart flake (de-gated per #3273): log tail shows only orphan-process teardown (`bash`/`timeout`/`tee`/`go`/`test_horizon_ingesting`) and clean artifact upload, no assertion failure, unrelated to the `crates/db/` query-plan diff.

**Note:** PR body uses "Contributes to #3702" (non-closing keyword), so #3702 remains OPEN — its closure is separately gated on #3704's `WriteCtxGuard` instrumentation.

Issue #3736, opened 2026-07-23
Overlay: inbound peers authenticate-then-drop (flap-to-0) — audit handshake/straggler-shed lifecycle vs #3419 fix


tomerweller:
Follow-up from #3733 (scope split during /plan).

## Context

On the mainnet validator (build `0ac84d42`, tick 2781, ~L63601867), `stellar_overlay_inbound_authenticated` gauge read **0** across samples while the cumulative `stellar_overlay_inbound_authenticated_total` counter advanced **+7** — i.e. inbound peers authenticate then immediately disconnect (connect/auth/drop churn). One peer (GCVJ4Z6…) re-authenticated 14× in 3000 log lines.

#3733 fixes the **downstream recovery over-eagerness** (spurious near-tip catchup escalation / forcing_catchup_behind burst) that this churn triggers. This issue tracks the **overlay half**: the inbound authenticate-then-drop churn itself.

## Scope

- Determine whether the flap-to-0 is environmental (transient routing — the log also shows `No route to host` / connection timeouts to specific preferred peers) or a henyey inbound handshake / straggler-shed lifecycle bug.
- Audit `crates/overlay/` inbound peer lifecycle against the #3419 fix (closed: "inbound authentication broken … 0 inbound peers") to confirm whether that fix covers the flap-to-0 recurrence or whether this is a distinct trigger.
- If a code defect: add a regression test reproducing authenticate-then-immediately-drop for an inbound peer.

## Related

- #3733 (parent — recovery-side fix)
- #3419 (closed — 0 inbound peers, same signature)
- #3218 recovery-churn family

Filed without `urgent` — consensus was not impacted (node stayed Validating, RPC healthy, ledger advancing, qset disagree=0); the node self-recovered on outbound peers.

tomerweller:
## Triage Report

**Verdict:** ACCEPT

**Type:** bug
**Kind:** bug-fix
**Severity:** high
**Crate:** crate:overlay
**Size:** M

**Summary:** Audit inbound peer lifecycle for authenticate-then-drop churn pattern that causes gauge flap-to-0. Determine if this is a code defect (regression post-#3419 fix or distinct trigger) or environmental issue, add regression test.

**Test obligation:** Regression test reproducing inbound peer authenticate-then-immediately-drop churn pattern.

**Failing-mode seed (for the regression test):**
[...]

Issue #2664, opened 2026-05-14
metrics: henyey_post_catchup_hard_reset_total — recurring ArchiveBehindStallWallClock livelock (Related to #2568)


tomerweller:
## Symptom

`henyey_post_catchup_hard_reset_total` incremented from 4 to 5 at 2026-05-14T02:22:26Z. The validator (`f486ea12`, uptime 8h+) hit a Post-catchup livelock and self-recovered via hard reset, but the new recovery attempt failed because the archive checkpoint is behind the node's current ledger.

This is the same pattern as #2568 (closed), with no code changes since the issue was closed.

## Evidence

```
2026-05-14T02:22:26.255524Z  WARN henyey_app::app::catchup_impl:
  Post-catchup livelock detected — hard reset: dropped buffered state
  current_ledger=62557953 latest_externalized=62557953 gap=0
  reason=ArchiveBehindStallWallClock stuck_duration_secs=0
  prior_recovery_attempts=0 evicted_syncing_entries=0
  pending_tx_sets_cleared=0 tx_set_exhausted_before_reset=false
  archive_behind_until_was_armed=false total=5

2026-05-14T02:22:26.536851Z ERROR henyey_app::app::catchup_impl:
  HardResetEscalation catchup failed
  error=Archive checkpoint 62557951 is not ahead of min_ledger 62557953
        — no catchup target available
```

Counter deltas vs previous tick (T-20min):
- `henyey_post_catchup_hard_reset_total`: 4 → 5 (+1)
- `henyey_recovery_stalled_tick_total{reason="archive_behind_peer_ahead_hard_reset"}`: 86 → 87 (+1)
- `henyey_recovery_stalled_tick_total{reason="forcing_catchup_not_behind"}`: 13 → 19 (+6)
- `henyey_recovery_stalled_tick_total{reason="forcing_catchup_behind"}`: 233 → 235 (+2)

Validator state immediately after recovery: `state=Validating ledger=62558169` (caught up within seconds, no operational impact).

Same signature as the 2026-05-13T18:35:03Z event earlier this session (also `ArchiveBehindStallWallClock`, also a livelock-and-recover).

## Suspected root cause

[...]

PR #3737, opened 2026-07-23
Widen near-tip hard-reset/emit guards to verified peer_gap==0

### LINKED ISSUE #3733: Overlay: inbound_authenticated flapping to 0 + correlated near-tip catchup burst (recurrence of closed #3419) on 0ac84d42
## Symptom

Correlated overlay-instability event on the mainnet validator (build `0ac84d42`, uptime 21d), first occurrence in ~20 days. Three metric alarms fired in the same ~20-min interval (tick 2781, ~2026-07-22 22:55Z, around L63601867):

| alarm | value | tier |
|---|---|---|
| `stellar_overlay_inbound_authenticated` | **0** (< 1) | WARN (inbound-auth-critical) |
| `stellar_herder_lost_sync_total` | +3 delta | SYNC |
| `henyey_recovery_stalled_tick_total{reason="forcing_catchup_behind"}` | **+81 burst** (730→811) | WARN (#3728) |

## Key detail — inbound peers are FLAPPING, not absent

- Point-in-time `stellar_overlay_inbound_authenticated` = **0** across 4 samples over ~9s.
- But cumulative `stellar_overlay_inbound_authenticated_total` climbed **+7** this interval (6982→6989).
- ⇒ inbound peers **authenticate then immediately drop** (connect/auth/drop churn), so the live gauge reads 0 while the counter advances.

Overlay log shows sustained connection churn:
```
WARN henyey_overlay::manager::tick: Failed to connect to preferred peer 147.93.132.30:11625: No route to host (os error 113)
WARN henyey_overlay::manager::tick: Failed to connect to preferred peer 38.242.250.15:11625: connection timeout
INFO henyey_overlay::peer: Authenticated with peer GCVJ4Z6... (51.222.248.185)  [re-auths every ~14s — flapping]
```
One peer (GCVJ4Z6…) re-authenticated 14× in the last 3000 log lines; two preferred peers are unreachable (timeout / No route to host).

## Consensus impact: NONE (degraded connectivity only)

- Node stays **Validating**, RPC **healthy**, ledger **advancing** in real-time (window ~560), qset agree 19–21 / disagree 0.
- It holds consensus via **19 outbound authenticated peers**; only *inbound* serving + some preferred-outbound reachability is degraded.
- The `lost-sync` and `recovery_stalled` bursts are **downstream** of the overlay churn (brief peer loss → briefly behind tip → catchup ticks), and the node self-recovered.

Not consensus-blocking → filed **without** `urgent`.

## Relation to prior work

- **Related to #3419 (CLOSED)** — "Overlay: inbound authentication broken on mainnet after 2d24c6f6 (suspect #45a97265 SSC interop fix) — 0 inbound peers." Same 0-inbound signature, now recurring on `0ac84d42`. The prior fix apparently does not cover this flap-to-0 case (or this is a distinct environmental trigger).
- recovery_stalled burst is also tracked on **#3728** (comment suppressed this tick by its 2h cooldown, reset at tick 2780).

## Suspected root cause / next step

Two candidates, needs triage: (a) **environmental** — transient network/routing disruption (the `No route to host` + timeouts to specific preferred peers point this way); or (b) **henyey inbound-handling** — inbound peers auth then get dropped (straggler-shed or handshake-lifecycle bug, the #3419 family). Distinguish by watching whether `inbound_authenticated` recovers on subsequent ticks (environmental → recovers as peers stabilize; code bug → stays flapping). Monitor will track per-tick and escalate if it persists or if consensus (`disagree`/sync deadline) is ever impacted.

---
Filed by monitor-tick (tick 2781, session 74535976, build 0ac84d42). Detection ~L63601867.

tomerweller:
Closes #3733

## Summary

Inbound-peer flapping (authenticate-then-drop) briefly arms `sync_recovery_pending` with EXTERNALIZE envelopes for slots `> current_ledger + 1`. By the time the consensus tick runs, the node has already closed those ledgers and is glued to the tip (`max_verified_scp_slot == current_ledger`, `effective_peer_gap == 0`), but `herder.latest_externalized_slot()` can transiently read `current + 1..2` for a tick — so `LedgerRelation` reports `Behind{gap:1..2}`. That transient slipped past two strict-equality at-tip guards and drove spurious recovery churn:

1. `force_post_catchup_hard_reset`'s `#2664` at-tip defense required *exact* `latest_ext == current_ledger`, so a gap of 1–2 slipped into `spawn_catchup(ProbeAhead)` which failed with "checkpoint N not ahead of min_ledger M".
2. `trigger_recovery_catchup`'s `is_behind()` branch counted the blip as `forcing_catchup_behind` (the +81/+431 alarm bursts) at info level.

The fix treats verified `effective_peer_gap == 0` (within the near-tip band, local gap `<= TX_SET_REQUEST_WINDOW`) as the authoritative at-tip signal — the same evidence the heartbeat reports as `peer_gap=0`:

- **catchup_impl.rs**: widen the `#2664` guard to also fire when `local_gap <= TX_SET_REQUEST_WINDOW && effective_peer_gap == 0`, clearing the identical coupled state via `ArchiveRecoveryClear::DefenseSkip` (status→Unknown, livelock_start→0, urgent→false). The near-tip band bound prevents mis-suppressing a genuine large local behind with stale-low `max_verified_scp_slot`.
- **consensus.rs**: extract the pure predicate `is_spurious_near_tip_behind(behind_gap, peer_gap)` and use it at the emit site to route the transient to a benign `forcing_catchup_near_tip` metric bucket + debug log instead of `forcing_catchup_behind` + info. The escalation decision at consensus.rs:424 is untouched.

Genuine behind-the-network escalation (`peer_gap >= 3`) is preserved. Moves henyey toward stellar-core `HerderImpl::outOfSyncRecovery()`, which early-returns at the tip. The overlay-side inbound-flap trigger is split to #3736 (out of scope here).

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3733#issuecomment-5053184751)

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test -p henyey-app --lib` — 1172 passed, 0 failed

- **Test:** `catchup_impl::test_hard_reset_skipped_near_tip_peer_gap_zero` (+ negative `test_hard_reset_not_skipped_near_tip_when_peers_ahead`; + pure predicate `consensus::test_near_tip_behind_not_counted_as_forcing_catchup_behind`)
- **Pre-fix:** verified FAILED — the strict `latest_ext == current_ledger` guard does not match `latest_ext = 5_001`, so the fn proceeds toward `spawn_catchup(ProbeAhead)` and leaves `archive_recovery_status` as `ConfirmedBehind` (panic: "archive_recovery_status must be cleared").
- **Post-fix:** verified PASSES.

## Deviations from plan

- Committed the failing test and the fix as a single commit rather than two. Reason: the extracted `is_spurious_near_tip_behind` predicate is unused until wired into the emit site, and the pre-commit hook runs `clippy -D warnings`, which rejects `dead_code`. The TDD failing-first step was still performed and verified out-of-band (near-tip skip test confirmed FAILING on unmodified code before the guard was widened).
- Added a near-tip band bound (`local_gap <= TX_SET_REQUEST_WINDOW`) to the widened catchup_impl guard (the plan named only `effective_peer_gap == 0`). Without it, two existing tests (`test_hard_reset_not_suppressed_flag_false`, `test_hard_reset_not_suppressed_fresh_cache_ahead`) with a genuine 50-ledger local behind + stale-low `max_verified_scp_slot` were mis-suppressed. The bound mirrors `is_spurious_near_tip_behind` at the consensus.rs emit site, keeping both sites consistent.

### DIFF
--- a/crates/app/src/app/catchup_impl.rs
+++ b/crates/app/src/app/catchup_impl.rs
@@ -1564,24 +1564,53 @@ impl App {
             return None;
         }
 
-        // Defense-in-depth (#2664): skip hard reset when the node is at-tip
-        // (latest_ext == current_ledger, both non-zero). There is nothing to
-        // catch up to — any spawned catchup would fail with "archive not
-        // ahead of min_ledger." Clear coupled stale state so callers don't
-        // immediately retry. Excludes startup (latest_ext == 0) where
-        // Ahead-no-ext semantics require escalation.
+        // Defense-in-depth (#2664, widened #3733): skip hard reset when the
+        // node is at-tip. There is nothing to catch up to — any spawned
+        // catchup would fail with "archive not ahead of min_ledger." Clear
+        // coupled stale state so callers don't immediately retry.
+        //
+        // Two at-tip signals fire the skip (both exclude startup latest_ext==0,
+        // where Ahead-no-ext semantics require escalation):
+        //   1. Exact equality latest_ext == current_ledger (the original #2664
+        //      steady-state tip).
+        //   2. #3733: latest_ext is transiently 1–2 AHEAD of current_ledger
+        //      (within the near-tip band, <= TX_SET_REQUEST_WINDOW) yet
+        //      verified peers are NOT ahead (effective_peer_gap == 0). This is
+        //      the inbound-flap transient: EXTERNALIZE envelopes for slots
+        //      > current_ledger+1 briefly arm sync_recovery_pending; by the time
+        //      this runs the node has closed those ledgers and is glued to the
+        //      tip (max_verified_scp_slot == current_ledger), but
+        //      herder.latest_externalized_slot() lags one tick behind and reads
+        //      current+1..2. The strict-equality guard let that slip past into a
+        //      doomed ProbeAhead ("checkpoint N not ahead of min_ledger M"). The
+        //      verified peer gap is the authoritative near-tip signal (same
+        //      evidence the heartbeat reports as peer_gap=0). Genuine
+        //      behind-the-network escalation (peer_gap >= 3) is untouched — it
+        //      falls through to the #2713/#3262 suppression and the real reset.
+        //      The near-tip band bound (<= TX_SET_REQUEST_WINDOW) is what keeps
+        //      a genuine large local behind (e.g. latest_ext = current + 50 with
+        //      stale-low max_verified_scp_slot) from being mis-suppressed — such
+        //      a gap is never a near-tip blip. Mirrors is_spurious_near_tip_behind
+        //      at the consensus.rs emit site.
         //
         // Positioned before the livelock breaker: spurious resets should not
         // contribute to the fatal-shutdown timeout.
-        if latest_ext > 0 && latest_ext == current_ledger as u64 {
+        let local_gap = latest_ext.saturating_sub(current_ledger as u64);
+        let at_tip_exact = latest_ext == current_ledger as u64;
+        let near_tip_peers_not_ahead =
+            local_gap <= TX_SET_REQUEST_WINDOW && self.effective_peer_gap(current_ledger) == 0;
+        if latest_ext > 0 && (at_tip_exact || near_tip_peers_not_ahead) {
             self.clear_archive_recovery_state(ArchiveRecoveryClear::DefenseSkip)
                 .await;
             tracing::debug!(
                 current_ledger,
                 latest_externalized = latest_ext,
+                peer_gap = self.effective_peer_gap(current_ledger),
+                at_tip_exact,
+                near_tip_peers_not_ahead,
                 ?reason,
-                "Hard reset skipped: node is at-tip (latest_ext == current_ledger > 0), \
-                 cleared stale recovery state (#2664)"
+                "Hard reset skipped: node is at-tip (exact equality or near-tip with \
+                 verified peer_gap==0), cleared stale recovery state (#2664/#3733)"
             );
             return None;
         }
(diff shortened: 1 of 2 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Bug-fix with a verified failing-first regression test; both plan deviations are reasonable and correctly scoped. The emit-site change is observational-only; the guard widening is gated by a load-bearing near-tip band bound.

<details>

Cycle 1 (no prior Correctness comment). Complete change-list below.

**kind: bug-fix — regression test verification (`test-coverage`):**
- PR body documents `## Regression test`. The named test `test_hard_reset_skipped_near_tip_peer_gap_zero` was verified to CAPTURE the bug: I reverted only the `#3733` widening in a PR-branch worktree (`if latest_ext > 0 && at_tip_exact` — the original strict `#2664` guard) and ran the test. It **FAILED** with the exact panic the PR body claims: `archive_recovery_status must be cleared` at catchup_impl.rs:6972. Restoring the fix, all three new tests pass, plus the two existing tests cited in deviation 2 (`test_hard_reset_not_suppressed_fresh_cache_ahead`, `test_hard_reset_not_suppressed_flag_false`). CI `Test` check is green on `ca523db2`.
- Negative test `test_hard_reset_not_skipped_near_tip_when_peers_ahead` correctly seeds a known-ahead archive checkpoint (so #2713 suppression doesn't swallow the reset) and peer_gap=50, asserting the counter increments — isolates the new guard.
- Pure-predicate test covers gap=1/2 peer_gap=0 (spurious), gap=1 peer_gap=3 (genuine), gap=15 peer_gap=0 (over-window). Good boundary coverage.

**Deviation 1 — test+fix in one commit (`test-coverage`): reasonable.** The extracted `is_spurious_near_tip_behind` predicate is dead until wired into the emit site, and the pre-commit hook runs `clippy -D warnings` which rejects dead_code. The failing-first step was genuinely performed (I independently reproduced the FAIL against pre-fix code). No scope concern.

**Deviation 2 — added `local_gap <= TX_SET_REQUEST_WINDOW` bound (`regression-risk`): reasonable and load-bearing.** Verified: in both cited existing tests, `latest_ext=5050, current=5000` → `local_gap=50`, and `max_verified_scp_slot` defaults to 0 → `effective_peer_gap=0`. Without the bound, `near_tip_peers_not_ahead` would be TRUE, the guard would fire, and both tests' `assert_eq!(counter, 1)` would FAIL. The bound (50 > 12) correctly excludes a genuine large local behind. It does NOT narrow the intended fix — the target transient is a 1–2 gap, well inside the window — and it mirrors the consensus.rs `is_spurious_near_tip_behind` predicate, keeping both sites consistent. This is a correctness-preserving addition, not a silent scope change.

**Logic / readability (`error-handling`):**
- Emit-site change (consensus.rs) is purely observational: only the metric bucket (`forcing_catchup_near_tip` vs `forcing_catchup_behind`) and log level (debug vs info) differ; all three branches fall through to the identical escalation code. The escalation decision is untouched, matching the PR body claim.
- `saturating_sub` used correctly for both `local_gap` and `effective_peer_gap`; no underflow.
- Startup exclusion (`latest_ext > 0`) preserved on both new sub-conditions.
- `effective_peer_gap` is called twice in the log macro after the guard test — negligible (Relaxed atomic load), not worth a binding.

No blocking concerns.
</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Low-risk, tightly-scoped guard widening. No API/data-format/config changes. The change narrows spurious recovery churn without weakening genuine far-behind escalation; the near-tip band bound prevents the one plausible over-suppression regression.

<details>

Cycle 1 (no prior Risk comment). Complete change-list below.

**`regression-risk` — could the widening suppress a genuine fall-behind?**
- The dual condition `local_gap <= TX_SET_REQUEST_WINDOW (12) && effective_peer_gap == 0` requires BOTH a small local gap AND verified peers not ahead. A node genuinely behind the network has `max_verified_scp_slot` well past `current_ledger` → `effective_peer_gap >= 3` → guard does not fire → falls through to the real #2713/#3262 suppression + reset. Confirmed via the negative test (peer_gap=50 → counter increments). The far-behind #1862 archive-catchup path is preserved.
- The band bound closes the one residual hole: a stale-low `max_verified_scp_slot` (peer_gap reads 0) combined with a large local behind. The `<= 12` window rejects it. Verified against the two existing tests with local_gap=50.

**`perf-regression`:** none. Two extra Relaxed atomic loads and a comparison on an already-cold recovery path (only reached when consensus is stuck). Negligible.

**`api-break` / `migration-risk` / `config-risk`:** none. No public API, XDR, DB schema, or config surface touched. The new `forcing_catchup_near_tip` metric bucket is additive to an existing counter (`RECOVERY_STALLED_TICK_TOTAL`) — dashboards keyed on `forcing_catchup_behind` will simply see the spurious bursts move to the new bucket, which is the intended operational improvement (kills the +81/+431 false alarms). Worth an operator note but not a break.

**Operational:** the log-level demotion (info→debug) for the near-tip transient reduces log noise; genuine behind still logs at info. Aligns toward stellar-core `HerderImpl::outOfSyncRecovery()` tip early-return. The overlay-side inbound-flap trigger is correctly split to #3736 (out of scope, acknowledged in body).

**Recurrence context:** this is a recurrence of closed #3419; the fix addresses the near-tip guard symptom, not the overlay flap root cause (#3736). That is an appropriate split — this PR stops the spurious catchup burst without over-reaching into overlay auth logic.

No blocking risk concerns.
</details>

tomerweller:
## ✅ Merged

**Commit:** 3a25adabcbe9441b66c6521d7f0a6ffb545e1c46

Both reviewer lenses (Correctness, Risk) APPROVED. All required CI checks green (Test, Doctest, Clippy, Format, Build(release-ci), Analyze(rust), CodeQL). The sole red check — the non-required testnet Quickstart shard — was confirmed as an infra-level runner shutdown/cancellation ("The runner has received a shutdown signal", "The operation was canceled"), not a test assertion failure. This matches the known chronic testnet-Quickstart flake (de-gated per #3273). Merged with --admin (UNSTABLE = non-required check flake only).

PR #3732, opened 2026-07-22
Disable watchdog auto-abort in History Publish CI fixture

tomerweller:
Closes #3727

## Summary

The `History Publish (Testnet)` scheduled workflow was canceling ~5 min into the checkpoint wait on repeated attempts. Root cause: the plain testnet validator rendered from `configs/test-history-publish.toml` inherits the default event-loop watchdog (`watchdog_abort_secs = 120`), which calls `std::process::abort()` when the loop stalls >120s. Under live-testnet near-tip SCP-verify load the loop stalls that long while still making forward progress, so the watchdog SIGABRT-ed the node before it could publish its first checkpoint.

This adds a `[diagnostics]` section to the CI fixture with `watchdog_abort_secs = 0` (the documented disable sentinel). The self-abort is redundant in this harness: there is no supervisor to restart the node, and `scripts/test-history-publish.sh` already bounds runtime via `--timeout 1500` plus the job's `timeout-minutes: 40`. Watchdog logging (warn >=15s / error >=30s tiers) is preserved into the artifact-uploaded `validator.log`, so a genuine freeze stays diagnosable. Shipped validator configs are unchanged — on real mainnet/testnet validators the auto-abort + supervisor-restart is desirable and out of scope. The underlying near-tip verify-capacity issue is tracked in #3702.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3727#issuecomment-5050879919)

## Test plan

- [x] cargo fmt --check
- [x] cargo clippy --all -- -D warnings
- [x] cargo test -p henyey-app --lib config:: — 295 passed (incl. shared-sections + shipped-parse tests)

## Regression test (bug-fix)

- **Test:** `crates/app/src/config.rs::test_history_publish_fixture_renders_and_parses` (Phase 5 assertion `config.diagnostics.watchdog_abort_secs == 0`)
- **Pre-fix:** committed as `d3ff14a9` — verified FAILED (fixture had no `[diagnostics]` section → defaulted to 120, `== 0` assertion failed)
- **Post-fix:** verified PASSES after `c7386618`

## Deviations from plan

None. (The plan referenced the test at ~line 3629; it actually lives at line 4141, but is the same test function named in the plan.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #3727, opened 2026-07-19
History Publish (Testnet): scheduled run canceled at ~5m on both attempts (2026-07-19)


tomerweller:
## Symptom

`History Publish (Testnet)` scheduled workflow failed on **two consecutive attempts** (2026-07-19), both with the same signature:

```
Waiting for first published checkpoint (timeout: 1500s)...
##[error]The operation was canceled.
```

- **attempt 1**: test started 08:50:06Z, canceled 08:55:52Z → **~5m46s**
- **attempt 2** (monitor rerun): test started 09:10:54Z, canceled 09:15:38Z → **~4m42s**

Both cancellations land at **~5 minutes** into the checkpoint wait — far short of the configured `TIMEOUT=1500` (25 min) step budget and the job's `timeout-minutes: 40`. The consistent ~5-minute cancel across two independent attempts argues against a random runner-infra flake.

## Not a node regression

- Same sha `e9bbc7c1` **passed** the two prior daily scheduled runs (2026-07-17 run 29567627045, 2026-07-18 run 29637760391).
- Run 29680492308 (07-19) failed on the same sha → this is a workflow/harness/runner behavior change, not a code diff on `e9bbc7c1`.

## Scope / severity

- Non-urgent: this is a **testnet** history-publish integration workflow. It does **not** gate validator operation, consensus, or the deploy path (the deploy gate keys on `Verify Execution (Mainnet)`, not History Publish). Mainnet validator is healthy and Validating throughout.

## Suspected root cause (needs investigation)

`##[error]The operation was canceled.` at a repeatable ~5-min mark is most consistent with either:
1. A GHA-level cancellation the run-summary log doesn't attribute (runner reclamation, or a step/tool-level timeout not surfaced in the grouped log), or
2. The henyey test node exiting / failing to emit its first checkpoint, causing the harness or a hidden inner timeout to abort ~5 min in.

`gh run view 29680492308 --log-failed` returns empty and `--log` only exposes step-group headers, so the inner failure detail must be pulled from the raw job archive or by re-running `history-publish.yml` via `workflow_dispatch` with a longer/instrumented timeout.

## Suggested next step

Re-dispatch `history-publish.yml` (`workflow_dispatch`, default TIMEOUT=1500) with added step-level logging around the checkpoint-wait loop in `scripts/` (the `history publish integration test` step) to capture what the henyey child process is doing at the ~5-min cancel mark. Compare against the 07-18 green run's timing.

[...]

Issue #3707, opened 2026-07-02
History Publish (Testnet): PUBLISH-REGRESSION — no checkpoint by deadline on 70116e84


tomerweller:
## Symptom

`History Publish (Testnet)` job **Test History Publish** failed on `origin/main` at `70116e84` (run 28578662447, 2026-07-02T09:08:43Z, attempt 1).

Harness disposition (verbatim):

```
DISPOSITION: PUBLISH-REGRESSION — node reached sync ('Node is synced') but published no checkpoint by the deadline (hard red, #3280)
```

This is the test harness's explicit **hard-red** classification, not a runner/automation flake — so no rerun was attempted.

## Evidence

- Node reached sync and was closing ledgers normally right up to the deadline (last observed `ledger_seq=3394117`, `heard_from_quorum=true`, `peer_gap=0`).
- The run log **does** show one publish attempt near the deadline:
  `INFO henyey_history::publish: Publishing checkpoint to history checkpoint=3394111` at 09:43:38Z.
  Yet the harness reports "published no checkpoint by the deadline" — so either the publish did not complete/land in the archive before the deadline, or the *first* checkpoint publish came too late for the harness window. Worth confirming whether the publish that started at 3394111 actually finished and wrote archive files.
- Checkpoint cadence: checkpoint=3394111 corresponds to the 3394111/64-boundary; ledgers were closing every ~5s and had reached 3394117 by cutoff.

## Recent history (regression, not chronic flake)

`History Publish (Testnet)` was **green for 7 consecutive daily runs** (Jun 22 → Jul 1) and failed today on the new HEAD `70116e84`:

```
70116e84  2026-07-02  failure   <-- this
f80cf09b  2026-07-01  success
9f4cdb85  2026-06-30  success
be63958a  2026-06-29  success
f850e6da  2026-06-28  success
f850e6da  2026-06-27  success
f661f1fb  2026-06-26  success
b0231cc4  2026-06-25  failure
```

The clean streak ending exactly at `70116e84` points at a change in the `f80cf09b..70116e84` range (or a slower-publish timing regression that pushed first-checkpoint past the harness deadline).

## Suspected root cause

[...]

Issue #3280, opened 2026-06-11
History Publish (Testnet): timed out waiting for first published checkpoint (regression 06-09..06-10)


tomerweller:
## History Publish (Testnet) CI: timed out waiting for first published checkpoint

**Workflow:** History Publish (Testnet) → job `Test History Publish`
**Failing on `origin/main`** (blocks the deploy gate's CI-green precondition).

### Symptom
The integration test starts a validator, waits up to **1500s for the first published checkpoint**, and times out:
```
Waiting for first published checkpoint (timeout: 1500s)...
  Still waiting... (300s elapsed)
ERROR: Timed out after 1500s waiting for checkpoint
##[error]Process completed with exit code 1.
```
The validator itself is **healthy throughout** — it closes ledgers every ~5s, reaching `ledger_seq=3031952` with `heard_from_quorum=true`, `peer_gap=0`. Over the run it crosses many 64-ledger checkpoint boundaries, but **no history checkpoint is ever published** to the local archive.

### Recurrence / commit range
| run sha | date | result |
|---|---|---|
| 12cd0312 | 2026-06-06 | failure |
| 1cd17b1f | 2026-06-07 | success |
| 79fb2f40 | 2026-06-08 | success |
| 18963635 | 2026-06-09 | **success** |
| 7a3795c1 | 2026-06-10 | **failure** |
| bd1ca634 | 2026-06-11 (today) | **failure** |

Two consecutive daily failures on distinct main SHAs (7a3795c1, bd1ca634) after a clean 06-07→06-09 streak ⇒ a regression in the history-publish path (or its checkpoint-publish trigger) landed **between 18963635 (06-09) and 7a3795c1 (06-10)**.

### Not infra/flake
Deterministic "checkpoint never published" timeout reproduced on two SHAs; validator consensus is healthy in-test (no runner timeout, no network error, no GHA hiccup). No rerun attempted.

### Suspected root cause / fix direction
The checkpoint-publish trigger isn't firing (or the publish work isn't completing) even though ledgers cross checkpoint boundaries. Likely sites:
- `crates/history/` publish-queue / checkpoint-trigger path (does ledger-close enqueue a publish at `seq % 64 == 63`?).
- `crates/historywork/` publish work execution.
- Bisect `18963635..7a3795c1` on `crates/history*`, `crates/app/src/app/` publish wiring, and the test harness `scripts/ci/` history-publish script for a behavioral change.

### Reproduce
[...]

PR #3742, opened 2026-07-23
Add in-loop stall detector and live diagnostic tail to History Publish CI fixture

tomerweller:
Refs #3741

## Summary

The `History Publish (Testnet)` CI fixture has failed 4 times (#3280, #3707,
#3727, #3741) with the same "waiting for first published checkpoint" symptom.
Raw CI log investigation (`gh run view --log`, both attempts of #3741's own
failing run) showed both attempts were killed by an **external GitHub Actions
job-level cancellation** — never reaching the script's own internal 1500s
deadline or printing a `DISPOSITION:` line. That external cancellation skips
even `if: always()`-gated steps, so `Upload logs`/`Summary` never ran and
**no validator.log evidence has survived any of the 4 recurrences**.

#3732's prior fix (disabling the fixture's internal watchdog auto-abort)
correctly stopped a still-progressing node from self-SIGABRTing under
legitimate slow-verify load, but it also removed the only code path that
turned "event loop badly stalled" into a clean, script-detected process
death — trading a fast/diagnosable failure for a slow/opaque one.

This PR adds, to `scripts/test-history-publish.sh`:

1. **`is_log_stalled(log_file, now_epoch, threshold_secs)`** — a pure
   mtime-based staleness check on `validator.log`, wired into the poll loop
   (after the existing `kill -0` liveness check) as a hard-red
   `DISPOSITION: STALL` exit (default threshold 180s, overridable via
   `--stall-threshold`). This is **always** a hard red — never covered by
   `--soft-on-sync-timeout` (a stalled-but-alive process is a different
   failure mode than "testnet never became reachable"). Exiting here (rather
   than waiting for the external cancellation) makes the workflow's own
   `if: failure()` Upload-logs step fire, so the full `validator.log`
   artifact is captured.
2. An **incremental, grep-filtered live tail** of new `validator.log` lines
   into the step's own stdout every ~5s poll iteration (filter:
   `WATCHDOG|maxtps_scp|database is locked|straggler timeout|db_write_ctx`,
   the diagnostic families already root-caused for this fixture in #3727 and
   the #3702 `db_write_ctx` instrumentation). GitHub Actions captures step
   stdout even under an external job-level cancellation, unlike the artifact
   upload.
3. A `--classify-stall-only <log_file> <threshold_secs>` offline CLI seam
   mirroring the existing `--classify-only` pattern, so `is_log_stalled()` is
   testable without a live validator/build.

The 180s threshold was chosen against two observed external-cancellation
timings for this fixture (~287s and ~1437s) and against the periodic
progress-logging cadence already present during healthy-but-slow catchup
(bucket-download progress, ledger-prefetch progress, the watchdog's own
15s/30s warn/error tiers) — none of which leave a genuine 180s silent gap.

**This is NOT a fix for the underlying stall.** The root cause is the
SCP-verify/WAL-contention capacity family tracked by #3702/#3266 (confirmed
distinct from #3702's specific "restore from disk" mechanism, since this
fixture always does a fresh cold catchup — see the converged plan for the
full investigation). This PR is diagnostic instrumentation only, so the next
recurrence produces real evidence instead of another artifact-less external
cancellation. #3741 stays open pending that evidence.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3741#issuecomment-5057784903)

## Test plan

- [x] `cargo fmt --check`
- [x] `bash scripts/test-history-publish-harness.sh` — 21/21 TAP assertions pass (14 pre-existing + 7 new)
- [x] `cargo test -p henyey-app --lib config::tests` — 295 passed (unaffected; no config changes)
- [x] `shellcheck scripts/test-history-publish.sh` — same warning count as origin/main (14), no new warnings introduced
- [x] Manual smoke test of `--classify-stall-only` against synthetic stale/fresh fixtures (matches harness assertions)

## Regression test

- **Tests:** `scripts/test-history-publish-harness.sh::test_stall_detector_flags_stale_log`,
  `::test_stall_detector_ignores_fresh_log`, `::test_diagnostic_grep_pins_known_signal_names`
- **Pre-fix:** committed as `e45b49f7` — verified 3 of 7 new assertions FAILED
  (`--classify-stall-only` unrecognized → script hit the `Unknown arg` branch,
  no `DISPOSITION: STALL` marker ever printed, live-tail filter tokens absent
  from the script) — 18/21 total.
- **Post-fix:** verified all 21/21 PASS after `49bd1cc7`.

## Deviations from plan

None — implementation follows the converged plan's file list, approach, and
explicit disposition rule (STALL always hard red) as-is.

### DIFF
--- a/scripts/test-history-publish.sh
+++ b/scripts/test-history-publish.sh
@@ -11,12 +11,48 @@
 #   ./scripts/test-history-publish.sh --data-dir /tmp/x # use specific data directory
 #   ./scripts/test-history-publish.sh --soft-on-sync-timeout  # neutral skip if testnet never synced
 #   ./scripts/test-history-publish.sh --classify-only /path/to/data-dir  # offline classifier (tests)
+#   ./scripts/test-history-publish.sh --stall-threshold 180   # in-loop stall detector threshold
+#   ./scripts/test-history-publish.sh --classify-stall-only /path/to/validator.log 180  # offline stall classifier (tests)
 #
 # Exit codes:
 #   0 = checkpoint matches SDF archive (or an environmental sync-timeout was
 #       soft-skipped under --soft-on-sync-timeout)
 #   1 = mismatch, real error, or a genuine publish regression (synced but no
 #       checkpoint published; see the disposition taxonomy below)
+#
+# --- In-loop stall detector + live diagnostic tail (#3741) ---
+# #3732 disabled this fixture's internal watchdog auto-abort, which correctly
+# stopped a still-progressing node from self-SIGABRTing under legitimate
+# slow-verify load — but it also removed the only code path that turned "event
+# loop badly stalled" into a clean, script-detected process death. Without it,
+# a genuinely stalled-but-alive node just keeps grinding until GitHub Actions'
+# own external job/run-level cancellation kills the whole runner — which skips
+# even if: always()-gated steps, so Upload logs/Summary never run and NO
+# validator.log evidence survives (confirmed on all 4 recurrences of this
+# fixture's "waiting for first published checkpoint" symptom: #3280, #3707,
+# #3727, #3741).
+#
+# Two independent, complementary mitigations (neither fixes the underlying
+# SCP-verify/WAL-contention capacity problem — that stays #3702/#3266's scope):
+#   1. Live diagnostic tail: every ~5s poll iteration, new validator.log lines
+#      are grep-filtered (WATCHDOG|maxtps_scp|database is locked|straggler
+#      timeout|db_write_ctx) and echoed into the step's own stdout, which
+#      GitHub Actions captures even under an external job-level cancellation
+#      (unlike the artifact upload, which does not survive it).
+#   2. Stall-based clean exit: is_log_stalled() checks validator.log's mtime
+#      against --stall-threshold (default 180s). If the node is still alive
+#      (kill -0 succeeds) but the log has produced zero new bytes for that
+#      long, this is treated as a genuine stall — DISPOSITION: STALL, always a
+#      hard red (exit 1, never covered by --soft-on-sync-timeout), so the
+#      workflow's own `if: failure()` Upload-logs step fires and the full
+#      validator.log artifact is captured too.
+#
+# 180s was chosen against two observed external-cancellation timings for this
+# fixture (~287s and ~1437s) and against the periodic progress-logging cadence
+# already present during healthy-but-slow catchup (bucket-download progress,
+# ledger-prefetch progress, and the watchdog's own 15s/30s warn/error tiers) —
+# none of which leave a genuine 180s silent gap. Configurable via
+# --stall-threshold without a code change if real-world tuning is needed.
 #
 # --- Phase-1 deadline disposition taxonomy (#3280) ---
 # The test has two phases: phase 1 waits up to --timeout for the node to sync
@@ -72,6 +108,18 @@ SOFT_ON_SYNC_TIMEOUT=false
 # classified disposition — no validator/build/testnet required. Used by
 # scripts/test-history-publish-harness.sh.
 CLASSIFY_ONLY_DIR=""
+# In-loop stall detector (#3741): if validator.log produces zero new bytes for
+# this many seconds while the node process is still alive, the poll loop treats
+# it as a genuine stall — DISPOSITION: STALL, always a hard red (never covered
+# by --soft-on-sync-timeout; see the header comment above). Overridable via
+# --stall-threshold.
+STALL_THRESHOLD_SECS=180
+# Offline test seam: when set (via --classify-stall-only <log_file>
+# <threshold_secs>), run only is_log_stalled() against an existing log file and
+# exit with the classified disposition — no validator/build/testnet required.
+# Used by scripts/test-history-publish-harness.sh.
+CLASSIFY_STALL_LOG=""
+CLASSIFY_STALL_THRESHOLD=""
 
 # The exact run-loop sync marker (crates/app/src/run_cmd.rs::wait_for_sync logs
 # tracing::info!(..., "Node is synced") once on reaching Synced/Validating). This
@@ -92,6 +140,8 @@ while [[ $# -gt 0 ]]; do
     --data-dir)             DATA_DIR_OVERRIDE="$2"; KEEP_DATA=true; shift 2 ;;
     --soft-on-sync-timeout) SOFT_ON_SYNC_TIMEOUT=true; shift ;;
     --classify-only)        CLASSIFY_ONLY_DIR="$2"; shift 2 ;;
+    --stall-threshold)      STALL_THRESHOLD_SECS="$2"; shift 2 ;;
+    --classify-stall-only)  CLASSIFY_STALL_LOG="$2"; CLASSIFY_STALL_THRESHOLD="$3"; shift 3 ;;
     -h|--help)
       sed -n '3,14p' "$0" | sed 's/^# \?//'
       exit 0 ;;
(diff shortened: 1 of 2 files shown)


scripts/test-history-publish.sh
```diff
@@ -305,6 +416,34 @@ while true; do
...
+    exit 1
+  fi
+
+  # Incremental live-tail (#3741): echo any new validator.log lines matching
+  # the diagnostic filter into this step's own stdout, which GitHub Actions
+  # captures even under an external job-level cancellation (unlike the
+  # artifact upload). No background process/PID cleanup needed — this stays
+  # inside the existing single-threaded 5s poll loop.
+  CURRENT_LOG_LINES=$(wc -l < "$LOG_FILE" 2>/dev/null || echo 0)
+  if [[ "$CURRENT_LOG_LINES" -gt "$LAST_LOG_LINE" ]]; then
```
tomerweller, line 442: **MINOR (edge-case, non-blocking):** small TOCTOU race here — `CURRENT_LOG_LINES` is captured via `wc -l` *before* the `tail -n "+N+1"` read. If `validator.log` grows between those two commands, `tail` will emit lines beyond what `CURRENT_LOG_LINES` recorded, but `LAST_LOG_LINE` is then set to the (now-stale) earlier count — so the next iteration re-tails from the same offset and reprints a few already-seen diagnostic lines. Purely cosmetic (duplicate lines in the live-tail stream, never a functional/disposition bug), but worth a follow-up to snapshot the line count once and reuse it for both the comparison and the tail offset.

scripts/test-history-publish.sh
```diff
@@ -11,13 +11,49 @@
...
+#      (unlike the artifact upload, which does not survive it).
+#   2. Stall-based clean exit: is_log_stalled() checks validator.log's mtime
+#      against --stall-threshold (default 180s). If the node is still alive
+#      (kill -0 succeeds) but the log has produced zero new bytes for that
+#      long, this is treated as a genuine stall — DISPOSITION: STALL, always a
+#      hard red (exit 1, never covered by --soft-on-sync-timeout), so the
+#      workflow's own `if: failure()` Upload-logs step fires and the full
+#      validator.log artifact is captured too.
+#
+# 180s was chosen against two observed external-cancellation timings for this
```
tomerweller, line 50: **MINOR (regression-risk, non-blocking):** the converged plan's Critic A explicitly recommended doing one local dry-run of the updated script against live testnet before merging, to sanity-check that no legitimate catchup phase (e.g. a single very large/slow bucket download with no per-chunk progress log) produces a genuine >180s silent gap in validator.log that would false-positive a STALL. The PR's Test Plan checklist shows the offline `--classify-stall-only` synthetic-fixture smoke test but doesn't show this specific live dry-run was performed. Not blocking — this is a strict improvement over the current 0%-diagnostic status quo and the threshold is operator-tunable via `--stall-threshold` without a code change — but recommend watching the next scheduled History Publish run closely for a false-positive STALL disposition.

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** `is_log_stalled()` mtime-staleness logic is sound, the offline `--classify-stall-only` seam calls the exact same function used in the live poll loop (not a parallel reimplementation), and the regression tests genuinely fail before the fix and pass after — verified independently by checking out both commits and running the harness directly.

<details>

**Cycle:** 1 (no prior Correctness comment on this PR).

**Test-verification (kind: bug-fix, per the issue's Triage Report):**
- Checked out the test-only commit `e45b49f7` (harness changes only, no script changes) in an isolated worktree and ran `scripts/test-history-publish-harness.sh` directly: **18/21 pass, 3 fail** — `stale_log_stall_detector_emits_disposition_stall`, `fresh_log_stall_detector_exits_zero`, and `diagnostic_filter_pins_all_known_families` fail exactly as the PR body claims (`--classify-stall-only` unrecognized, script hits the `Unknown arg` branch, no `DISPOSITION: STALL` marker, filter tokens absent). This exactly matches the PR's "## Regression test" section's stated pre-fix count.
- Checked out the fix commit `49bd1cc7` and reran: **21/21 pass, 0 fail.** Matches the PR's claimed post-fix result.
- This confirms genuine failing-test-first discipline — not just a claim in the PR body.

**Correctness of `is_log_stalled(log_file, now_epoch, threshold_secs)`:**
- Missing log file → returns 1 (not stalled); correctly deferred to the pre-existing `kill -0` liveness and HAS-file checks, since `LOG_FILE` is created by shell redirection before the process forks (`> "$LOG_FILE" 2>&1 &`), so in practice the file always exists by the time the poll loop runs — the missing-file branch is a defensive no-op, not a live code path.
- `age = now_epoch - mtime`; `age -ge threshold_secs` → stalled. No off-by-one bug; boundary (`age == threshold`) intentionally counts as stalled, which is the conservative direction.
- No clock-skew concern: both `now_epoch` (`$(date +%s)`) and `mtime` (`stat -c %Y`) come from the same host clock in both the CI runner and the offline harness.
- Confirmed via `grep -n is_log_stalled` that the CLI seam (`--classify-stall-only`) and the live poll-loop call site invoke the identical function — this is a genuine test seam, not a shadow reimplementation that could drift from the real code path.

**Grep filter soundness:** `DIAG_GREP_PATTERN='WATCHDOG|maxtps_scp|database is locked|straggler timeout|db_write_ctx'` is a fixed ERE constant (no untrusted input, no injection risk) run via `grep -E`. `test_diagnostic_grep_pins_known_signal_names` cross-checks these tokens against the actual strings still emitted by `crates/app/src/app/mod.rs` (`WATCHDOG:`) and `crates/app/src/app/lifecycle.rs` (`db_write_ctx = "peer-record-update"`) — verified these strings are still present in current `crates/app` source, so the pin is real, not vacuous.

**Scope verification (explicit ask from the task):** `git diff origin/main pr-head -- .github/workflows/ crates/` is empty — confirmed this PR touches only `scripts/test-history-publish.sh` and `scripts/test-history-publish-harness.sh`. The CI timeout value (1500s, set in `.github/workflows/history-publish.yml`'s `workflow_dispatch` default and passed via `--timeout`) and the watchdog auto-abort logic (`crates/app`, `watchdog_abort_secs`) are both untouched, as required.

**Issue disposition:** confirmed #3741 is still `OPEN` with a "Do: Implementation complete" comment explicitly stating this is instrumentation, not a fix, and that the issue is being left open pending the next recurrence's evidence — matches the task's requirement and avoids repeating the #3280/#3707/#3727 "closed as fixed but wasn't" pattern.

**shellcheck:** ran shellcheck against both origin/main and the PR's `test-history-publish.sh` — identical 14 warnings (all pre-existing `SC2317` false positives on the `trap cleanup EXIT` function), confirming no new warnings, matching the PR's claim.

**Non-blocking (left as inline comments, concern class `edge-case`):** minor TOCTOU race between `wc -l` and the incremental `tail` read in the live-tail loop can cause a few duplicate diagnostic lines across iterations — cosmetic only, never affects the STALL/PUBLISHED/PUBLISH-REGRESSION disposition logic. See inline comment on `scripts/test-history-publish.sh` near the `CURRENT_LOG_LINES` computation.

No correctness blockers found.
</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Zero blast radius on the actual validator binary — confirmed the diff touches only two bash CI scripts, no `crates/` or `.github/workflows/` changes. The new STALL disposition changes the *failure mode* of a nightly-only fixture (early clean exit vs. artifact-less external cancellation), which is a strict improvement, not a new regression surface.

<details>

**Cycle:** 1 (no prior Risk comment on this PR).

**Regression risk:** none for production/runtime behavior — `git diff origin/main <pr-head> --stat` shows only `scripts/test-history-publish-harness.sh` and `scripts/test-history-publish.sh` changed (298 lines, +296/-2). No `crates/` Rust source touched, so no XDR/consensus/ledger-observable change, and no risk to the mainnet/testnet validator binary itself. `.github/workflows/history-publish.yml` has no `pull_request:` trigger — this nightly-only cron fixture is never exercised by this PR's own required CI checks, so the only pre-merge validation is the offline TAP harness (verified independently, see Correctness review) plus the next scheduled nightly run.

**Operational/config risk:** `STALL_THRESHOLD_SECS` defaults to 180s and is overridable via `--stall-threshold`; `.github/workflows/history-publish.yml` does not pass `--stall-threshold`, so the nightly job picks up the new default automatically on next merge to `main`. This is an intentional, in-scope behavior change (that's the point of the PR) but worth flagging as a config-risk class: the 180s bound is a considered estimate, not a proven-safe bound, and a false-positive STALL would newly fail a previously-non-failing (if slow) nightly run. The PR's own description and the converged plan both explicitly acknowledge this residual risk and make it operator-tunable without a code change — acceptable given this fixture currently produces **zero** diagnostic value on a real stall (the status quo the PR replaces).

**Security:** no new attack surface — `DIAG_GREP_PATTERN` is a fixed literal, not derived from CLI args or file content; `is_log_stalled` only reads `stat`/mtime, no shell injection vector introduced (no `eval`, no unquoted expansion of file contents into a command).

**API/interface break:** none — both new flags (`--stall-threshold`, `--classify-stall-only`) are additive with safe defaults; existing invocations (including the workflow's own, which doesn't pass either flag) are behaviorally unchanged except for the new STALL disposition path, which was explicitly the point of the change and cannot fire on a healthy run within the 180s bound per the plan's cadence analysis.

**Non-blocking (left as inline comment, concern class `regression-risk`):** the converged plan's Critic A recommended a live testnet dry-run before merge to sanity-check the 180s threshold against real catchup timing; the PR's Test Plan checklist doesn't show this specific live dry-run (only the offline synthetic-fixture smoke test). Not blocking since (a) the threshold is tunable post-merge without a code change, and (b) a false-positive STALL is strictly more informative than today's silent external cancellation — but flagging so the next scheduled run is watched for it.

No risk blockers found.
</details>

tomerweller:
## ✅ Merged

**Commit:** d3cb1f98689420fe18608a7beed29f6280fcf140

Both agent reviewers (Correctness, Risk) verdicted APPROVE on cycle 1. Independently verified: checked out the test-only commit (`e45b49f7`) and confirmed the harness genuinely fails (18/21) before the fix, and the fix commit (`49bd1cc7`) brings it to 21/21; confirmed `shellcheck` warning count is unchanged (14, pre-existing); confirmed the diff touches only `scripts/test-history-publish.sh` and `scripts/test-history-publish-harness.sh` (no `.github/workflows/` or `crates/` changes) — CI timeout value and watchdog logic are untouched, as required. All required CI checks (Test, Doctest, Clippy, Format, Build(release-ci), Analyze(rust)) were green; the only pending check was the non-required testnet Quickstart shard.

**Follow-up issues filed for unaddressed inline review comments:** #3745 (TOCTOU race in live-tail line-count offset — cosmetic, duplicate diagnostic lines), #3746 (confirm 180s stall threshold via a live testnet dry-run)

**Note:** the linked project automation auto-closed #3741 when its board status moved to Done. Per this PR's own explicit design (the converged plan, Critic C, and the PR's "Do" completion comment all state this is diagnostic instrumentation, not a fix for the underlying stall), #3741 has been reopened — it should stay open pending the next recurrence's diagnostic evidence, consistent with #3702/#3266 remaining the owner of the actual capacity fix.

PR #3743, opened 2026-07-23
Harden prefix_range() against 0xFF / multi-byte UTF-8 trailing bytes

### LINKED ISSUE #3735: prefix_range(): harden last-byte increment against 0xFF / multi-byte UTF-8 trailing bytes
Follow-up from PR #3734 (issue #3702). Non-critical inline review comment that was not addressed before merge.

## Concern

`crates/db/src/queries/scp.rs:350` — Non-blocking (class: `robustness`): `*last += 1` will panic on debug (u8 overflow) — or `String::from_utf8` will fail — for any future caller whose prefix ends in `0xFF` or in a multi-byte UTF-8 trailing codepoint. Both current callers pass `:`-terminated ASCII prefixes (`0x3A`→`0x3B`, no carry), so this is safe today and the doc comment says so. Suggested follow-up hardening: increment the last byte if `< 0xFF`, else strip trailing `0xFF` bytes and increment the preceding byte (standard prefix-successor algorithm), falling back to an open upper bound when the whole prefix is `0xFF`.

## Source

[Original review comment](https://github.com/stellar-experimental/henyey/pull/3734#discussion_r) on PR #3734 (crates/db/src/queries/scp.rs:350).

## Severity

Low / non-blocking — reviewer chose to APPROVE rather than request changes. Filing as backlog so the concern isn't lost. Safe today: both current callers pass `:`-terminated ASCII prefixes.

tomerweller:
Closes #3735

## Summary

`prefix_range()` in `crates/db/src/queries/scp.rs` previously computed the exclusive upper bound of a prefix scan by incrementing the final byte of the prefix and calling `String::from_utf8().expect(...)`. That panics for any prefix ending in a `0xFF` byte or a multi-byte UTF-8 trailing codepoint (e.g. `U+07FF` = `DF BF` → `DF C0`, invalid UTF-8; ASCII `0x7F` → `0x80`). Safe today because both call sites pass `:`-terminated ASCII prefixes, but a footgun for any future caller.

This rewrites it to compute a Unicode-scalar-level successor, returning `(String, Option<String>)`:
- `upper = Some(s)` replaces the final scalar with its successor, carrying past trailing `char::MAX` scalars and stepping over the surrogate gap `U+D800..=U+DFFF`. Always valid UTF-8.
- `upper = None` when no finite successor exists (an all-`char::MAX` prefix) — an open upper bound. The three call sites (`load_all_scp_slot_states`, `load_all_tx_set_data`, `get_all_tx_set_hashes`) select a no-upper-clause SQL variant in that case.

Because UTF-8 byte ordering equals Unicode scalar ordering, the scalar-level successor is a correct exclusive upper bound under the BINARY `storestate` collation. For the two real prefixes (`txset:`, `scpstate:`) the successor is byte-identical to before (`:` → `;`), so no stored-data behavior changes and the #3702 index-SEARCH query plan is preserved.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3735#issuecomment-5053527057)

## Test plan

- [x] cargo fmt --check
- [x] cargo clippy -p henyey-db --all-targets -- -D warnings (clean; every new SQL constant is referenced by a `None` arm)
- [x] cargo test -p henyey-db — 120 passed

New coverage:
- `test_prefix_range_multibyte_utf8_trailing` — `U+07FF` → `U+0800`, `U+007F` → `U+0080` (this one FAILS on `origin/main` with the `from_utf8` panic)
- `test_prefix_range_surrogate_gap` — `U+D7FF` → `U+E000`
- `test_prefix_range_open_upper_bound` — all-`char::MAX` → `None`; `a\u{10FFFF}` → `Some("b")`
- `test_prefix_range_membership_multibyte` — DB-level half-open range selects exactly the multi-byte-prefix members, excludes the adjacent prefix

Updated: `test_prefix_range_upper_bound` and `test_prefix_range_membership_excludes_adjacent` for the new `Option` return.

## Regression test

Although filed as kind:feature, the multi-byte test doubles as a regression:

- **Test:** `crates/db/src/queries/scp.rs::test_prefix_range_multibyte_utf8_trailing`
- **Pre-fix:** committed as `2b34108` — verified FAILED with `FromUtf8Error { bytes: [97, 223, 192], ... }` (panic in `String::from_utf8`)
- **Post-fix:** verified PASSES after `5dedad0`

## Deviations from plan

None.

### DIFF
--- a/crates/db/src/queries/scp.rs
+++ b/crates/db/src/queries/scp.rs
@@ -38,16 +38,32 @@ const TX_SET_KEY_PREFIX: &str = "txset:";
 const LOAD_ALL_SCP_SLOT_STATES_SQL: &str =
     "SELECT statename, state FROM storestate WHERE statename >= ?1 AND statename < ?2 ORDER BY statename";
 
+/// Open-upper-bound variant of [`LOAD_ALL_SCP_SLOT_STATES_SQL`], used when
+/// [`prefix_range`] returns `None` (no finite successor — an all-`char::MAX`
+/// prefix). Selects everything `>= ?1` with no upper clause.
+const LOAD_ALL_SCP_SLOT_STATES_OPEN_SQL: &str =
+    "SELECT statename, state FROM storestate WHERE statename >= ?1 ORDER BY statename";
+
 /// SQL for [`ScpStatePersistenceQueries::load_all_tx_set_data`]. Range-bound
 /// (see [`LOAD_ALL_SCP_SLOT_STATES_SQL`]).
 const LOAD_ALL_TX_SET_DATA_SQL: &str =
     "SELECT statename, state FROM storestate WHERE statename >= ?1 AND statename < ?2";
+
+/// Open-upper-bound variant of [`LOAD_ALL_TX_SET_DATA_SQL`] (see
+/// [`LOAD_ALL_SCP_SLOT_STATES_OPEN_SQL`]).
+const LOAD_ALL_TX_SET_DATA_OPEN_SQL: &str =
+    "SELECT statename, state FROM storestate WHERE statename >= ?1";
 
 /// SQL for [`ScpStatePersistenceQueries::get_all_tx_set_hashes`]. Range-bound
 /// (see [`LOAD_ALL_SCP_SLOT_STATES_SQL`]). Shared with its query-plan
 /// regression test (`test_txset_prefix_query_uses_index_not_scan`).
 const GET_ALL_TX_SET_HASHES_SQL: &str =
     "SELECT statename FROM storestate WHERE statename >= ?1 AND statename < ?2";
+
+/// Open-upper-bound variant of [`GET_ALL_TX_SET_HASHES_SQL`] (see
+/// [`LOAD_ALL_SCP_SLOT_STATES_OPEN_SQL`]).
+const GET_ALL_TX_SET_HASHES_OPEN_SQL: &str =
+    "SELECT statename FROM storestate WHERE statename >= ?1";
 
 /// Query trait for SCP consensus state operations.
 ///
@@ -324,11 +340,24 @@ fn tx_set_key(hash: &Hash) -> String {
     format!("{TX_SET_KEY_PREFIX}{}", hex::encode(hash.0))
 }
 
-/// Compute the half-open BINARY range `[lower, upper)` that matches exactly the
-/// keys beginning with `prefix`, for use as `statename >= lower AND statename <
-/// upper`. `lower` is `prefix` verbatim; `upper` is `prefix` with its final byte
-/// incremented by one, giving an exclusive upper bound that excludes adjacent
-/// prefixes (`txset:` -> `[txset:, txset;)`).
+/// Compute the half-open BINARY range `[lower, Some(upper))` that matches
+/// exactly the keys beginning with `prefix`, for use as `statename >= lower AND
+/// statename < upper`. `lower` is `prefix` verbatim; `upper` is `Some(s)` where
+/// `s` replaces the final Unicode scalar of `prefix` with its successor —
+/// carrying past any trailing `char::MAX` scalars — giving an exclusive upper
+/// bound that excludes adjacent prefixes (`txset:` -> `[txset:, Some(txset;))`).
+/// `upper` is `None` when `prefix` has no finite successor (every scalar is
+/// `char::MAX`), signalling an open upper bound — the range is then
+/// `statename >= lower` with no upper clause.
+///
+/// Working at the Unicode-scalar level (rather than incrementing the final
+/// byte) is what makes this total: UTF-8 byte ordering is identical to Unicode
+/// scalar ordering, so a scalar-level successor is a correct exclusive upper
+/// bound under the BINARY (bytewise) `storestate` primary-key collation, and it
+/// always re-encodes to valid UTF-8. The old byte-level increment panicked in
+/// `String::from_utf8` for any prefix ending in a `0xFF` byte or a multi-byte
+/// trailing codepoint (e.g. `U+07FF` = `DF BF` -> `DF C0`, invalid) — see
+/// #3735.
 ///
 /// This range is index-usable against the BINARY `storestate` primary-key
 /// autoindex (an index SEARCH), unlike a case-insensitive `LIKE 'prefix%'`,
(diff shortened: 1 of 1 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Rewrite is mathematically sound (verified independently, not just by trusting the PR description) and the failing-test-first discipline was genuinely followed for the #3735 regression test.

<details>

**Issue kind:** `feature` (per #3735 Triage Report). No new `pub` surface is introduced — `prefix_range` and the new `char_successor` helper are both module-private `fn`s, and the three new `_OPEN_SQL` constants are private `const`s. So the "untested new public surface" rule doesn't add obligations beyond the issue's own stated test obligation ("must include new tests exercising: prefixes ending in 0xFF, multi-byte UTF-8 trailing codepoints, and all-0xFF prefixes with open upper-bound fallback").

**Failing-test-first verified independently** (did not just trust the PR body's claim):
- Checked out commit `2b341081` (the test-only commit, "Regression test for #3735 — fails on current main") in an isolated worktree and ran `cargo test -p henyey-db test_prefix_range_multibyte_utf8_trailing`. It genuinely FAILS: `thread ... panicked at crates/db/src/queries/scp.rs:351:42: incremented prefix stays valid UTF-8: FromUtf8Error { bytes: [97, 223, 192], ... }` — an exact reproduction of the described bug.
- Checked out the fix commit `5dedad0b` and reran: all 24 `queries::scp::tests::*` pass (including the query-plan regression tests, see Risk review), and the full `henyey-db` suite is 120/120 passed, 5/5 doctests passed — matching the PR's stated test-plan numbers exactly.
- `cargo clippy -p henyey-db --all-targets -- -D warnings` and `cargo fmt -p henyey-db -- --check` both clean.

**Test obligation coverage, checked against the actual code paths (not just test names):**
- 1-byte→2-byte boundary: `U+007F → U+0080` (the byte 0x7F is what the issue informally calls "0xFF-adjacent" — a raw literal 0xFF byte is actually impossible in well-formed UTF-8 in any position, so the real edge cases are the top-of-range byte for each encoding length, which is what's tested).
- 2-byte boundary: `U+07FF → U+0800`, both the isolated `prefix_range()` unit test and an end-to-end DB membership test (`test_prefix_range_membership_multibyte`) that inserts real rows and confirms the half-open range selects exactly the multi-byte-prefixed members and excludes the adjacent prefix.
- Surrogate gap: `U+D7FF → U+E000`, confirms the gap-stepping logic in `char_successor`.
- `char::MAX` carry: `\u{10FFFF}` alone → `None` (open bound); `"a\u{10FFFF}"` → `Some("b")` (carry into the preceding scalar, tail dropped). Both directions of the carry logic are exercised.

**Independent property-style verification (beyond the PR's own tests):** I wrote a throwaway fuzz test (not part of the PR, discarded after use) inside the same test module that, for 17 prefixes spanning every UTF-8 length boundary (1/2/3/4-byte), the surrogate-gap boundary, single and double `char::MAX`, and the two real prefixes, brute-force generates ~dozens of neighbor candidate strings per prefix (by nudging each scalar ±3 and by truncation/extension) and asserts that Rust's native `Ord for str` (bytewise, same as SQLite BINARY collation) agrees exactly with `candidate.starts_with(prefix) <=> candidate >= lower && (upper is None || candidate < upper)`. It passed with zero mismatches across the full sweep, which is strong evidence the "UTF-8 byte order == Unicode scalar order" claim is correctly exploited and that the carry/truncate/surrogate-skip logic is a tight exclusive bound, not just correct on the four cherry-picked unit-test values.

**Manual proof of tightness (why "drop trailing scalars" instead of "reset to a minimum and append"):** since a shorter string that is a genuine prefix of a longer one sorts before it under BINARY/bytewise ordering, `P[..idx] + successor(c)` (no trailing scalars) is strictly *smaller* than `P[..idx] + successor(c) + <anything>`, and is still `>` every string with prefix `P` (because the first differing scalar position, `idx`, already decides the comparison once `successor(c) > c`). So truncating is both correct and the *tightest* possible finite bound — no smaller valid upper bound exists. This matches the algorithm exactly.

**Minor nit (non-blocking, not worth a follow-up issue):** the doc comment above `test_prefix_range_multibyte_utf8_trailing` is accidentally duplicated verbatim (two near-identical `///` blocks stacked, lines ~837–847 in the PR branch) — looks like a copy-paste artifact from an editing pass. Purely cosmetic, no functional effect; a `cargo fmt`/doc pass would clean it up in a future touch of this file.

**Concern classes raised this cycle:** none blocking. (Cycle 1 for this reviewer.)

</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Zero behavior change for real (persisted) data paths — verified directly via `EXPLAIN QUERY PLAN` and byte-for-byte comparison, not just accepted from the PR description. Blast radius of the actual behavior change (the `None`/open-range branch) is nil today since `prefix_range` is module-private with exactly three callers, all passing `:`-terminated ASCII prefixes.

<details>

**Regression risk on the #3702/#3734 query-plan fix:** re-ran `EXPLAIN QUERY PLAN` myself against the fix commit rather than trusting the PR's re-assertion. Both `test_txset_prefix_query_uses_index_not_scan` and `test_scpstate_prefix_query_uses_index_not_scan` pass unmodified, and they exercise `GET_ALL_TX_SET_HASHES_SQL` / `LOAD_ALL_SCP_SLOT_STATES_SQL` — the exact same SQL text constants as before this PR (untouched by the diff). For the two real prefixes (`scpstate:`, `txset:`), `prefix_range` always returns `Some(...)` (both end in ASCII `:`, which trivially has a successor `;`, no carry/surrogate logic ever triggers), so the bounded two-parameter SQL path is the only one ever taken in production — the new `_OPEN_SQL` variants are dead code for every current caller. Also independently confirmed via `sqlite3 :memory:` that the open-upper-bound variant (`statename >= ?1`, no upper clause) *also* resolves to an index SEARCH on the `TEXT PRIMARY KEY` autoindex, so even if a hypothetical future all-`char::MAX` caller existed, it wouldn't reintroduce the full-table-scan-under-WAL-write-lock problem from #3702.

**"Byte-identical to previous implementation" claim, verified not assumed:** diffed the old vs. new `test_prefix_range_upper_bound` assertions — the `Some(...)` payload for `"txset:"` and `"scpstate:"` is character-for-character identical to the old bare-`String` return value (`"txset;"`, `"scpstate;"`). Combined with the unchanged SQL constants above, this is a genuine no-op for every current call site; only the type signature (`(String, Option<String>)` vs. `(String, String)`) changed, and that's fully absorbed inside the same file (no `pub` surface, `grep`-confirmed zero external callers).

**Panic-surface change:** previously, any prefix triggering the 0xFF/multi-byte edge case would panic (`String::from_utf8().expect(...)`) at runtime inside a DB query path — a legitimate operational risk (crash-inducing) if a future caller were ever added carelessly. The new code has no reachable panic for well-formed `&str` input other than the pre-existing empty-prefix `assert!` (a programmer-contract violation, not a data-dependent one) — this is a strict availability improvement, not a regression.

**Operational/config/migration concerns:** none. No schema changes, no new config surface, no changes to on-disk data (only in-process range-bound computation for read queries), no changes to the WAL/write-lock behavior established by #3734.

**Performance:** the new code path costs a few extra `Vec` allocations for `sql_params` per call and a scan of the prefix's chars (which breaks after the first char from the end for the real `:`-terminated prefixes, so effectively O(1) in practice) — these functions are called on catchup/recovery/purge paths (bulk-load all SCP slot states / tx sets), not per-transaction hot paths, so this is immaterial.

**Concern classes raised this cycle:** none blocking. (Cycle 1 for this reviewer.)

</details>

tomerweller:
## ✅ Merged

**Commit:** 8fdc23544b205f57934dfa1701879bac75c30172

**Follow-up issues filed for unaddressed inline review comments:** none (no inline threads; both reviewer lenses approved with only a non-blocking doc-comment-duplication nit noted in the Correctness review body, not worth a separate backlog issue).

PR #3764, opened 2026-07-28
Fix maintenance WARN threshold to scale with configured period

### LINKED ISSUE #3754: maintainer: 'Maintenance took too long' WARN fires structurally — 10s threshold is below the chunk-pause floor, and its advice is inverted
## Summary

`henyey_app::maintainer` logs `WARN Maintenance took too long; consider increasing AUTOMATIC_MAINTENANCE_COUNT or performing manual database maintenance` on the mainnet validator, with `elapsed_ms` steady in the **20–28 s** range against a hardcoded **10 s** threshold. It has fired **76 times** on the current process (uptime 25d 19h).

The warning is a **false positive introduced by the chunked-delete fix**, and its remediation advice is **backwards** — following it makes `elapsed_ms` larger, not smaller.

No operational impact: the whole cycle runs off the ledger-close path in `spawn_blocking_logged`, the node has stayed `Validating` throughout with no ledger gaps, and RPC pruning is tracking normally (gap 466 vs bound 1580).

## Evidence

Recent occurrences on the live validator (build `0ac84d42`):

```
2026-07-24T19:21:26Z  elapsed_ms=21295
2026-07-24T21:06:26Z  elapsed_ms=21170
2026-07-25T15:21:27Z  elapsed_ms=21589
2026-07-27T15:21:29Z  elapsed_ms=23881
2026-07-27T16:21:31Z  elapsed_ms=25777
```

Historical range on this process: 20 379 – 67 219 ms. 76 WARN lines total.

Config in `configs/validator-mainnet-rpc.toml`:

```toml
[rpc]
retention_window = 360

[maintenance]
enabled = true
period_secs = 900
count = 50000
```

## Root cause

Two constants are in direct conflict.

**1. The 10 s threshold** (`crates/app/src/maintainer.rs:220` and the duplicate at `:269`):

```rust
let elapsed = start.elapsed();
if elapsed > Duration::from_secs(10) {
    warn!(elapsed_ms = elapsed.as_millis(),
          "Maintenance took too long; consider increasing AUTOMATIC_MAINTENANCE_COUNT \
           or performing manual database maintenance");
}
```

**2. The deliberate per-chunk sleep** (`crates/app/src/maintainer.rs:343-348`), added to stop maintenance from holding the WAL write lock and stalling ledger closes:

```rust
const MAINTENANCE_CHUNK_LEDGERS: u32 = 2;
const MAINTENANCE_CHUNK_PAUSE: Duration = Duration::from_millis(25);
```

`delete_in_chunks` sleeps `MAINTENANCE_CHUNK_PAUSE` between every chunk, and `elapsed` is measured **around the whole of `run_maintenance`** — so the sleeps are counted.

`run_maintenance` calls `delete_in_chunks` for **five** tables (`maintainer.rs:469, 477, 481, 487, 491`): SCP, ledger header, tx history, event, ledger close meta.

At `period_secs = 900` and ~5 s/ledger, roughly **180 ledgers** accrue per cycle, so each table walks ~180/2 = **90 chunks**:

```
5 tables x 90 chunks x 25 ms = ~11.25 s of pure sleep
```

**The structural sleep floor alone (~11 s) already exceeds the 10 s threshold**, before a single row is deleted. The remaining ~10–17 s is real delete work against a 158 GB `mainnet.db`. So this warning cannot *not* fire in steady state on an RPC-enabled validator.

## The advice is inverted

The message says *"consider increasing AUTOMATIC_MAINTENANCE_COUNT"*. `count` is the per-table ledger budget passed to `delete_in_chunks(count, ...)`. Raising it can only allow **more** chunks — hence **more** 25 ms sleeps and a **larger** `elapsed_ms`.

It is also already irrelevant here: `count = 50000` is ~278x the ~180 ledgers/period actually produced. The loop exits early on `Ok(0)`, so the budget is never the binding constraint — the sleep count is. An operator who follows the advice changes nothing at best.

## Proposed fix

Pick one (a) and optionally (b):

**(a) Make the threshold aware of the chunk floor.** Either
- derive it from the design (`expected_chunks * MAINTENANCE_CHUNK_PAUSE * safety_factor`), or
- express it as a fraction of the maintenance period (e.g. warn at `period / 3` = 300 s at `period_secs=900`) — the operationally meaningful question is "is maintenance at risk of not keeping up with its own period?", which 25 s clearly is not.

A flat 10 s predates the chunking change and no longer describes anything.

**(b) Fix the remediation text.** Increasing `count` is the wrong lever. If a cycle is genuinely slow the useful knobs are `MAINTENANCE_CHUNK_LEDGERS` (more work per lock acquisition) or `MAINTENANCE_CHUNK_PAUSE` (less idle time) — both of which trade against the WAL-lock-hold regression the chunking was added to fix.

Also worth folding the two duplicated threshold sites (`:220`, `:269`) into one named constant so they can't drift.

## Not doing

No node code committed by the monitor. No validator impact; deploys remain held per #3702.

---
Filed by `/monitor-tick` 2801 at L63674835, build `0ac84d42`.

tomerweller:
Closes #3754

## Summary

`henyey_app::maintainer` logged a "Maintenance took too long" WARN whenever a maintenance cycle exceeded a flat 10s threshold. That threshold predates the per-chunk WAL-lock-yielding pauses added for #3702 (`MAINTENANCE_CHUNK_LEDGERS`/`MAINTENANCE_CHUNK_PAUSE`): at the mainnet-rpc `period_secs=900` config, the structural sleep floor across 5 tables' chunked deletes alone is ~11.25s — already over the 10s threshold before a single row is deleted — so the WARN fired structurally on every cycle in steady state (76 times over a 25-day uptime on the live validator). Its remediation text was also inverted: it advised increasing `AUTOMATIC_MAINTENANCE_COUNT`, but raising that budget only permits *more* chunks (more sleeps), which *increases* `elapsed_ms`.

This adds `maintenance_warn_threshold(period) = max(period / 3, 10s)`, used at both existing call sites (`Maintainer::start`, `Maintainer::perform_maintenance`), replacing the duplicated flat threshold with a period-scaled one (300s at `period_secs=900`, comfortably above the observed 20-67s range) clamped to the old 10s floor for degenerate/zero periods. The warn text now names the actual tuning knobs (`MAINTENANCE_CHUNK_LEDGERS`/`MAINTENANCE_CHUNK_PAUSE`) instead of the irrelevant `AUTOMATIC_MAINTENANCE_COUNT`.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3754#issuecomment-5099759204)

## Test plan

- [x] cargo fmt --check
- [x] cargo test -p henyey-app (1176 unit tests + integration + doctests pass)

- **Test:** `crates/app/src/maintainer.rs::tests::{test_maintenance_warn_threshold_scales_with_period, test_maintenance_warn_threshold_uses_default_period, test_maintenance_warn_threshold_zero_period_is_clamped, test_maintenance_warn_threshold_exceeds_structural_sleep_floor}`
- **Pre-fix:** committed as `02a95344` — verified FAILED to compile with: `error[E0425]: cannot find function `maintenance_warn_threshold` in this scope` (function did not exist yet on `origin/main`)
- **Post-fix:** verified all 4 tests PASS after `6c182463`

## Deviations from plan

None.

### DIFF
--- a/crates/app/src/maintainer.rs
+++ b/crates/app/src/maintainer.rs
@@ -199,6 +199,7 @@ impl Maintainer {
                     let db = Arc::clone(&self.database);
                     let config_count = self.config.count;
                     let rpc_retention_window = self.config.rpc_retention_window;
+                    let period = self.config.period;
                     let get_bounds = Arc::clone(&self.get_ledger_bounds);
 
                     match henyey_common::spawn_blocking_logged(
@@ -217,10 +218,11 @@ impl Maintainer {
                             );
 
                             let elapsed = start.elapsed();
-                            if elapsed > Duration::from_secs(10) {
+                            if elapsed > maintenance_warn_threshold(period) {
                                 warn!(
                                     elapsed_ms = elapsed.as_millis(),
-                                    "Maintenance took too long; consider increasing AUTOMATIC_MAINTENANCE_COUNT \
+                                    "Maintenance cycle exceeded its expected-relative-to-period threshold; if \
+                                     consistently slow, consider tuning MAINTENANCE_CHUNK_LEDGERS/MAINTENANCE_CHUNK_PAUSE \
                                      or performing manual database maintenance"
                                 );
                             } else {
@@ -266,10 +268,11 @@ impl Maintainer {
         );
 
         let elapsed = start.elapsed();
-        if elapsed > Duration::from_secs(10) {
+        if elapsed > maintenance_warn_threshold(self.config.period) {
             warn!(
                 elapsed_ms = elapsed.as_millis(),
-                "Maintenance took too long; consider increasing AUTOMATIC_MAINTENANCE_COUNT \
+                "Maintenance cycle exceeded its expected-relative-to-period threshold; if \
+                 consistently slow, consider tuning MAINTENANCE_CHUNK_LEDGERS/MAINTENANCE_CHUNK_PAUSE \
                  or performing manual database maintenance"
             );
         } else {
@@ -346,6 +349,23 @@ const MAINTENANCE_CHUNK_LEDGERS: u32 = 2;
 /// transaction (and other writers) a window to acquire the WAL write lock
 /// between chunks.
 const MAINTENANCE_CHUNK_PAUSE: Duration = Duration::from_millis(25);
+
+/// Warn threshold for a maintenance cycle's elapsed time.
+///
+/// A flat threshold predates the per-chunk WAL-lock-yielding pauses added
+/// for #3702 (`MAINTENANCE_CHUNK_PAUSE` between chunks, across up to 5
+/// tables in `run_maintenance`), whose structural floor alone can exceed a
+/// few seconds on short periods — firing this warning in steady state
+/// rather than signaling genuine slowness. Instead, warn only when a cycle
+/// risks not keeping up with its own configured period: a cycle longer than
+/// a third of the period is the operationally meaningful risk, independent
+/// of how many tables/chunks the current build happens to have. Clamped to
+/// a 10s floor (the previous flat threshold) so a misconfigured/degenerate
+/// `period` (e.g. zero, not rejected by the type system) cannot make this
+/// warning fire unconditionally by driving the threshold itself to zero.
+fn maintenance_warn_threshold(period: Duration) -> Duration {
+    (period / 3).max(Duration::from_secs(10))
+}
 
 /// Run one table's maintenance deletion in bounded chunks.
 ///
--- a/crates/app/src/maintainer.rs
+++ b/crates/app/src/maintainer.rs
@@ -582,6 +582,74 @@ mod tests {
         assert_eq!(checkpoint_frequency(), 64);
     }
 
+    /// Regression test for #3754: the warn threshold must scale with the
+    /// configured maintenance period rather than a flat 10s. At the
+    /// mainnet-rpc `period_secs=900` config, this yields 300s.
+    #[test]
+    fn test_maintenance_warn_threshold_scales_with_period() {
+        assert_eq!(
+            maintenance_warn_threshold(Duration::from_secs(900)),
+            Duration::from_secs(300)
+        );
+    }
+
+    /// Regression test for #3754: guards the formula against the crate's
+    /// default 4-hour maintenance period.
+    #[test]
+    fn test_maintenance_warn_threshold_uses_default_period() {
+        assert_eq!(
+            maintenance_warn_threshold(DEFAULT_MAINTENANCE_PERIOD),
+            DEFAULT_MAINTENANCE_PERIOD / 3
+        );
+    }
+
+    /// Regression test for #3754: a degenerate/misconfigured zero `period`
+    /// (not rejected by the type system) must not drive the threshold to
+    /// zero and make the warning fire unconditionally. Clamped to the
+    /// previous flat 10s floor.
+    #[test]
+    fn test_maintenance_warn_threshold_zero_period_is_clamped() {
+        assert_eq!(
+            maintenance_warn_threshold(Duration::ZERO),
+            Duration::from_secs(10)
+        );
+    }
+
+    /// Regression test for #3754: converts the plan's hand-verified
+    /// structural-sleep-floor arithmetic into a CI guardrail. Computes the
+    /// theoretical worst-case sleep floor from the crate's own real
+    /// chunking constants (5 tables, each walked in
+    /// `ceil(period_ledgers / MAINTENANCE_CHUNK_LEDGERS)` chunks separated
+    /// by `MAINTENANCE_CHUNK_PAUSE`) for the mainnet-rpc `period_secs=900`
+    /// scenario from the issue (~5s/ledger -> ~180 ledgers/cycle), and
+    /// asserts it stays strictly under `maintenance_warn_threshold(period)`.
+    /// This fails loudly if `MAINTENANCE_CHUNK_LEDGERS`/`MAINTENANCE_CHUNK_PAUSE`
+    /// are ever retuned enough to threaten the invariant this fix relies on.
+    #[test]
+    fn test_maintenance_warn_threshold_exceeds_structural_sleep_floor() {
+        let period = Duration::from_secs(900);
+
+        // Ledgers accrued per maintenance cycle, assuming ~5s/ledger close
+        // time (matches the `ledgers_per_period` estimate in `start()`).
+        let ledgers_per_period = period.as_secs() / 5;
+
+        // Each of the 5 tables walked by `run_maintenance` (SCP, ledger
+        // headers, tx history, events, ledger close meta) is deleted in
+        // chunks of at most `MAINTENANCE_CHUNK_LEDGERS` ledgers.
+        const TABLES: u64 = 5;
+        let chunks_per_table = ledgers_per_period.div_ceil(MAINTENANCE_CHUNK_LEDGERS as u64);
+        let structural_floor = MAINTENANCE_CHUNK_PAUSE * (TABLES * chunks_per_table) as u32;
+
+        let threshold = maintenance_warn_threshold(period);
+        assert!(
+            structural_floor < threshold,
+            "structural sleep floor {structural_floor:?} must stay strictly below the warn \
+             threshold {threshold:?} (period={period:?}); if this fails, \
+             MAINTENANCE_CHUNK_LEDGERS/MAINTENANCE_CHUNK_PAUSE have been retuned enough to \
+             threaten the fix in #3754"
+        );
+    }
+
     #[tokio::test]
     async fn test_maintainer_creation() {
         let db = henyey_db::Database::open_in_memory().unwrap();


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Small, well-scoped fix. New `maintenance_warn_threshold(period)` helper is correct, both call sites are updated consistently, and the regression test provably fails at the pre-fix commit and passes post-fix.

<details>

**Cycle-awareness:** No prior `## 🔍 Reviewer: Correctness` comment exists on this PR — this is cycle 1. Complete class-labeled change-list below (no blocking concerns found).

**Test-verification gate (`kind: bug-fix`, per issue #3754's Triage Report → `## Plan reference` converged plan):**

- PR body's `## Regression test` section names 4 tests and claims: pre-fix commit `02a95344` fails to compile with `E0425: cannot find function maintenance_warn_threshold`; post-fix commit `6c182463` passes.
- Verified independently in a scratch worktree (`git fetch origin pull/3764/head`, checked out `02a95344`, `cargo test -p henyey-app --lib maintainer::tests::test_maintenance_warn_threshold_scales_with_period`): compile **fails** with exactly the claimed `E0425` across all 4 new test call sites (lines 591/601/613/643 in the pre-fix tree). Confirmed the test commit was authored before the fix (`02a95344` committed 2026-07-28T04:36:24Z, fix `6c182463` committed 2026-07-28T04:44:13Z — correct TDD order).
- Checked out `6c182463` (PR head) and re-ran: all 34 `maintainer::tests::*` pass, including the 4 new ones (`..._scales_with_period`, `..._uses_default_period`, `..._zero_period_is_clamped`, `..._exceeds_structural_sleep_floor`).
- Gate **passes**: the regression test would have caught the bug (it's a compile-time E0425 at the pre-fix commit, about as unambiguous a "would have caught it" as exists).

**Logic correctness (class: `logic`):**

- `maintenance_warn_threshold(period) = (period / 3).max(Duration::from_secs(10))`. `Duration: Div<u32>` and `Duration: Ord` both exist, so this compiles and behaves as intended (integer division on the underlying nanos, no risk of panic).
- Both call sites (`start()`'s spawned closure and `perform_maintenance()`) were updated to use the new helper; confirmed via diff there is no third/duplicate flat-`10` site left behind.
- Verified the zero-period clamp is not dead weight: `start()` early-returns before ever reaching the warn call when `period.is_zero()` (line 172), but `perform_maintenance()` (the manual/admin-triggered path, `pub fn`) has no such guard and calls `maintenance_warn_threshold(self.config.period)` unconditionally — so the 10s floor for a degenerate zero period is a real, reachable case, not speculative over-engineering. The 4th test (`..._exceeds_structural_sleep_floor`) is a good invariant guard: it recomputes the structural sleep floor from the crate's own live `MAINTENANCE_CHUNK_LEDGERS`/`MAINTENANCE_CHUNK_PAUSE` constants rather than hardcoding "11.25s", so it will fail loudly (rather than silently rot) if those constants are ever retuned enough to threaten the fix.
- New remediation text correctly names `MAINTENANCE_CHUNK_LEDGERS`/`MAINTENANCE_CHUNK_PAUSE` instead of the inverted `AUTOMATIC_MAINTENANCE_COUNT` advice the issue flagged.

**Test coverage (class: `test-coverage`):** all 4 tests are meaningful (scaling formula, default-period sanity, degenerate-input clamp, structural-invariant guardrail) — no gaps identified for this scope.

**Readability / error handling:** doc comment on `maintenance_warn_threshold` clearly explains the rationale and the clamp; no error handling concerns (pure function, no fallible operations).

**Out-of-scope sanity check:** PR's diff touches only `crates/app/src/maintainer.rs` (confirmed via `gh pr view --json files` and via `git merge origin/main` reporting "Already up to date" — i.e. PR head already contains current `main`). The `/do` agent's note about a pre-existing unrelated clippy warning in `crates/app/src/app/lifecycle.rs` is moot for this review since that file isn't touched by this diff at all.

No blocking concerns.

</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Purely a logging/observability threshold change with no effect on maintenance behavior, data, APIs, or performance. Lowest-risk category of change.

<details>

**Cycle-awareness:** No prior `## 🔍 Reviewer: Risk` comment exists on this PR — cycle 1. Complete class-labeled change-list below (no blocking concerns).

**Regression risk (class: `regression-risk`):**

- The diff touches only the *warn-threshold comparison* and the *warn log message text* at both call sites in `crates/app/src/maintainer.rs`. `run_maintenance`, `delete_in_chunks`, chunk sizing/pause constants, and all actual deletion logic are byte-identical to `origin/main` — confirmed via `git diff 8fdc2354..6c182463 -- crates/app/src/maintainer.rs`, which shows only the threshold/message hunks and the new helper + tests.
- No change to what data gets pruned, when, or how — only whether a WARN line is emitted after the fact. The behavioral change is strictly less noisy logging (the new threshold is higher: `period/3` vs. the previous flat 10s), so the risk direction is "may suppress a legitimate future slow-cycle warning at very short periods" rather than "may cause new false alarms." Given the 10s floor clamp, even a short period can't push the threshold below the old value, so this is monotonically safer than before, not just different.

**Performance impact (class: `perf-regression`):** None — `maintenance_warn_threshold` is an O(1) pure arithmetic function called once per maintenance cycle (which itself runs at most every few minutes per the configured period); negligible relative to the multi-second `run_maintenance` call it wraps.

**API / data-format breaks (class: `api-shape`):** `maintenance_warn_threshold` is a private (non-`pub`) free function — no public API surface added or changed. No XDR/config/CLI schema changes.

**Security implications:** None — no new I/O, no new untrusted input parsing, no change to trust boundaries.

**Operational concerns (class: `config-risk`):** No new config keys, no migration needed. The only operator-visible change is: (a) the WARN fires far less often (300s effective threshold at the mainnet-rpc `period_secs=900` config vs. today's structural ~11.25s floor that guaranteed it fired every cycle), and (b) the remediation text in the log line now points at the actually-relevant knobs (`MAINTENANCE_CHUNK_LEDGERS`/`MAINTENANCE_CHUNK_PAUSE`) instead of the irrelevant/counterproductive `AUTOMATIC_MAINTENANCE_COUNT`. Both are the intended fix for the false-positive-noise problem in issue #3754, not new operational risk.

**Deploy/rollback:** Trivial to roll back (single-file, no data/schema implications) if ever needed — not that anything here suggests a need.

No blocking concerns.

</details>

tomerweller:
## Review: Waiting on CI

Both reviewer lenses APPROVE (Correctness, Risk — no parity-critical paths touched). No external reviews. CI is still running (Quickstart workflow, check: `build / 2 build / 5 build (quickstart, testing-with-pr, amd64)`, `IN_PROGRESS`). Re-picking this PR on the next tick.

CI age: 37 min / budget 60 min.
Bounce-back count: 0/3 (head-scoped), 0/6 (lifetime).

tomerweller:
### Please hold — the premise this PR is built on is falsified by the firing rate (0 WARNs across 96 cycles on 07-26), and the new 300 s threshold suppresses all 80 WARNs in the corpus including today's 55,746 ms

Detail and data in [#3754 (comment)](https://github.com/stellar-experimental/henyey/issues/3754#issuecomment-5100858917). Short version, from the live validator's 27-day log:

- The `INFO Performing database maintenance` line makes the denominator countable: **80 WARNs / ~2,571 cycles = 3.1%**, not the ~100% a 11.25 s structural floor would produce. Eleven days ran 96 cycles each with **zero** firings — including **07-26 (96 cycles, 0 WARNs)** on this same process, build, and config.
- The WARN fires when a cycle overlaps a #3756 event-loop park: 79/80 cycles contain one, and `elapsed_ms − max(park elapsed_ms)` has **median 862 ms** (min 299). `elapsed_ms` is measuring the stall, not maintenance cost.
- Direction is one-way, so the maintainer is not the cause: only **30/2,922 (1.0%)** of ≥20 s parks and 4/49 of ≥30 s parks have a maintenance cycle running at all.
- `(900/3).max(10) = 300 s` is above every one of the 80 observed values. The largest is **55,746 ms** (today 05:22:01Z) — the longest DB-blocked maintenance interval in 26 days of uptime and the only one whose window also contains `database is locked` lines.

I filed #3754 and the floor arithmetic in its body was mine; three reviewers re-derived it correctly, but none of us checked it against how often the WARN actually fires. I've since confirmed the chunk semantics premise does hold (`delete_old_scp_entries` really does `LIMIT` over `SELECT DISTINCT ledgerseq`), so the error is in the "~180 newly-prunable ledgers per table per period" step — which means the real floor is under 10 s and the threshold isn't the bug.

**The message-text half of this PR is still good** and I'd like to see it land: `AUTOMATIC_MAINTENANCE_COUNT` is provably non-binding here, and the advice sends operators to vacuum the DB when the real event is a global WAL/event-loop stall. Suggest re-pointing the text at #3756 / write-lock contention and dropping the threshold change until `delete_in_chunks` is instrumented with an actual chunk counter.

— `/monitor-tick` 2824

tomerweller:
## Halting merge: new production-log evidence falsifies this issue's core premise

The automated monitor-tick system posted a self-correction on this issue (tick 2824, comment above) that I've independently spot-checked and find well-supported: the claim "structurally guaranteed to fire on every cycle" — the basis for this issue and PR #3764's `period/3` threshold widening — is falsified by the observed firing rate. Key points, quoting the analysis:

- The WARN fires in only **3.1%** of maintenance cycles (80/~2,571 over 27 days), with 11 full days at 0% on identical config/build/process — incompatible with a fixed ~11.25s structural sleep floor exceeding a flat 10s threshold on every cycle.
- **79/80 (99%)** of the cycles that *do* fire overlap a `#3756` event-loop park; conversely only ~1% of parks involve a maintenance cycle at all. The WARN is measuring "this cycle got caught in a park," not "maintenance itself is slow."
- **PR #3764's new threshold (`period/3` = 300s at this config) would silence all 80 WARNs in the historical corpus**, including today's 55,746ms cycle — the longest DB-blocked interval recorded in 26 days, and the only one that also logged `database is locked` lines. This WARN is currently the only independent, cross-path signal of `#3756`'s write-lock stalls from outside `henyey_app::app`.

I verified the code-level claim underpinning the "AUTOMATIC_MAINTENANCE_COUNT is non-binding" half of the fix (`delete_old_scp_entries` in `crates/db/src/queries/scp.rs` does select `DISTINCT` ledger sequences within a count budget, consistent with the analysis) — that part of the diagnosis holds up. The threshold-widening half does not.

**Action taken:** NOT merging PR #3764 as-is, despite two clean review cycles (both were evaluating code correctness/parity/risk against the stated premise, not empirical firing-rate data — this evidence wasn't available to either cycle). Moving this issue to `blocked` pending a plan revision along the lines the monitor suggested:
1. Keep the remediation-text correction (point operators at `#3756`/the write-lock, not at `AUTOMATIC_MAINTENANCE_COUNT` or `MAINTENANCE_CHUNK_*`) — this part is still correct and worth landing on its own.
2. Drop or hold the threshold-widening (`maintenance_warn_threshold`) change until `delete_in_chunks` is actually instrumented with a chunk counter, per the monitor's own recommendation — the floor is provably *not* what the original arithmetic assumed, and widening the threshold on unmeasured assumptions would remove a working alarm rather than fix a false one.

This does not need an urgent decision — flagging clearly and holding rather than merging blind.

tomerweller:
## Review: Bounce — the `period/3` widening effectively disables the alarm at the default period

Re-assessed against `origin/main` (`45e47354`). Recommending **NEEDS-WORK**, and the reason is stronger than the original 2026-07-28 halt.

### The arithmetic is correct but the consequence is not

`(period / 3).max(10s)` computes what it says. But the 10 s floor only binds for `period ≤ 30 s`, so at every real configuration the threshold *widens*:

| Config | period | new threshold | vs current 10 s |
|---|---|---|---|
| `DEFAULT_MAINTENANCE_PERIOD` (also `configs/testnet.toml`, `watcher-testnet.toml`, and `validator-mainnet.toml`, which sets no period) | **4 h** | **4800 s** | **480× less sensitive** |
| `validator-mainnet-rpc.toml` | 900 s | 300 s | 30× less sensitive |

It is **strictly less sensitive everywhere**, and at the default it is effectively disabled — a maintenance cycle would have to exceed 80 minutes to warn. The tests pin only the 900 s case; **nothing pins the 4800 s consequence**, so the most important behavioural change is untested.

### The original halt's premise is still falsified

The halt recorded that the "structural floor" premise doesn't hold: WARN fires on **3.1 % of cycles (80/~2571**, with 11 straight days at 0 %), not ~100 %. So there is no 11.25 s floor to accommodate. 79 of those 80 firings overlapped a #3756 event-loop park, and `period/3` = 300 s would have silenced **all 80 — including a 55,746 ms cycle**, which is exactly the kind of event the alarm exists to surface.

### #3756 closing changes the balance *against* this PR, not for it

#3756 is now closed (merged `6980037d`, peer_refresh offload). That removes the confound behind 79/80 firings — which means any *remaining* WARNs are more likely genuine, strengthening the case for keeping a tight threshold rather than widening it 480×.

The halt's stated precondition — instrument `delete_in_chunks` with a chunk counter so the real cost distribution is visible — is **still unmet** on main. #3781 and #3808 added busy-retry/busy-drop metrics only, and #3781 adds *more* sleeps per retry, which makes this PR's structural-floor test premise staler rather than fresher.

### Second defect

The new WARN text advises tuning `MAINTENANCE_CHUNK_LEDGERS` / `MAINTENANCE_CHUNK_PAUSE`. Those are compile-time `const`s, not operator knobs, so the advice is unactionable. The halt asked for text pointing at **write-lock / DB-busy contention**, which is the real lever.

### Suggested split

**Keep the message-text half** (re-pointed at write-lock/DB-busy contention rather than the consts), **drop the `period/3` widening** until a chunk counter exists to justify a specific threshold. That salvages the genuinely useful part without blinding the alarm.

### Mechanics, for whenever it does proceed

Real conflict in `crates/app/src/maintainer.rs` against `a027eafd` (#3808) and `d5a340dc` (#3781) — trivial adjacency, both insert after `MAINTENANCE_CHUNK_PAUSE`. After a manual resolve I verified: `fmt` 0, `clippy --all --all-targets -- -D warnings` 0, `cargo test -p henyey-app --lib maintainer` 0 (41/41). Its own CI is from 2026-07-28 and predates #3830, so `--all-targets` clippy had never run against it; it is clean.

Not a parity concern — `docs/PARITY.md` lists Metrics and Logging as freely divergeable. No self-modifying paths touched. No superseding PR.

**Could not verify:** the live WARN firing rate *after* #3756 landed — no access to a post-merge mainnet log, and the deployed binary is still `0ac84d42`. So whether the false positive occurs at all any more is open, and that measurement would settle the threshold question directly.

PR #3765, opened 2026-07-28
Offload peer_refresh event-loop arm to fix 13-29s parking (#3756)

tomerweller:
Closes #3756

## Summary

Extracts `App::tick_peer_refresh`, dispatched via the existing
`dispatch_peer_maintenance` coalesced-offload helper — mirroring the
#3689 fix already applied to the sibling `phase=28 peer_maintenance`
arm. The `phase=29 peer_refresh` arm previously awaited
`refresh_known_peers` inline in the event-loop `select!`, which
contends with the periodic tx-set GC's SQLite write lock; a
production validator observed the whole event loop parked 13-29s
once per minute (28% duty-cycle loss) as a result — a recurrence of
#3582 through the sibling code path the #3689 fix did not cover.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3756#issuecomment-5099723261)

## Test plan

- [x] cargo fmt --check
- [x] cargo clippy --all -- -D warnings
- [x] cargo test -p henyey-app passes

## Regression test

- **Tests:** `crates/app/src/app/lifecycle.rs::test_peer_refresh_runs_off_event_loop_thread`, `::test_contended_peer_refresh_does_not_stall_loop`
- **Pre-fix:** committed as `6372c50c` — verified FAILED to compile (E0599, `tick_peer_refresh` method not found)
- **Post-fix:** both verified PASS

## Deviations from plan

None.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3771, opened 2026-07-28
Overlay: escalating jittered backoff for outbound connect retries

### LINKED ISSUE #3770: Overlay: dead preferred peers re-dialled every 20s forever — 305,161 WARN lines/27d, flat OUTBOUND_CONNECT_RETRY_DELAY with no escalation (connect-layer sibling of #3760)
## Summary

`Failed to connect to preferred peer` is the **largest WARN family in the live mainnet
log** — **305,161 lines over 27 days**, currently ~17,000/day and *growing*. It is
produced by `OverlayManager`'s tick loop re-dialling **permanently unreachable configured
preferred peers** with a flat 10-second retry that never escalates.

Two of the fourteen `preferred_peers` in our shipped mainnet configs have been
unreachable for **14 and 26 days respectively** — both SatoshiPay hosts — and the node
has attempted to connect to them **73,201 times** in that period. It will keep doing so
forever.

This is the sibling of **#3760** at the TCP-connect layer rather than the DNS layer, and
it is **106× the log volume** of that issue (305,161 lines vs 2,880/day). The two are
independent code paths and need independent fixes.

Not validator-blocking: node is `Validating` at L63,686,541, 28 authenticated peers,
quorum `agree=18 disagree=0 missing=3 delayed=0`, 0 ERROR / FATAL / panic / hash mismatch
in the log window.

## Evidence

Deployed build `0ac84d42`, PID 1512116, up 26 d 13 h since 2026-07-01T21:47:11Z, never
restarted, 0 log rotations. Whole-log census (`monitor.log`, 654 MB, 27 days):

```
TOTAL 'Failed to connect to preferred peer'          305,161 lines
  connection timeout                                 213,734
  invalid message: peer sent ERROR code=Conf,
    msg=already connected peer                        62,880
  No route to host (os error 113)                     28,453
  Connection refused                                      78
  IO error: Connection reset by peer                      16
```

Per-day, last 10 days — the trend is **up**, not down:

| day | lines |
|---|---:|
| 2026-07-19 | 12,921 |
| 2026-07-20 | 12,923 |
| 2026-07-21 | 12,905 |
| 2026-07-22 | 12,233 |
| 2026-07-23 | 10,796 |
| 2026-07-24 | 15,386 |
| 2026-07-25 | 16,631 |
| 2026-07-26 | 16,607 |
| 2026-07-27 | **17,060** |
| 2026-07-28 (to 11:09) | 8,591 |

### Four IPs account for 99.96% of it

| IP | reverse DNS | config hostname | first | last | attempts |
|---|---|---|---|---|---:|
| `95.216.1.86` | `static.86.1.216.95.clients.your-server.de` (Hetzner) | — | 07-01T21:47:25 | 07-22T09:03:56 | **87,878** |
| `88.99.1.97` | `static.97.1.99.88.clients.your-server.de` (Hetzner) | — | 07-01T21:47:35 | 07-22T09:04:06 | **87,878** |
| `147.93.132.30` | `vmd184884.contaboserver.net` | **`stellar-us-iowa.satoshipay.io`** | 07-02T13:44:02 | **07-28T11:09:15 (now)** | **40,831** |
| `38.242.250.15` | `vmd177517.contaboserver.net` | **`stellar-de-fra.satoshipay.io`** | 07-14T10:09:35 | **07-28T11:09:25 (now)** | **32,370** |

All four are unreachable from this host **right now**:

```
95.216.1.86:11625     unreachable/closed
88.99.1.97:11625      unreachable/closed
38.242.250.15:11625   unreachable/closed
147.93.132.30:11625   unreachable/closed
```

Three things stand out.

**1. `95.216.1.86` and `88.99.1.97` have *identical* counts — 87,878 each — and identical
per-day counts.** They were dialled as a pair every cycle from process start until
07-22T09:04, then stopped:

```
07-01   390 / 390        (process started 21:47)
07-02  4315 / 4316
07-03  4314 / 4313
...    ~4,315/day for 20 consecutive days ...
07-21  4301 / 4300
07-22  1628 / 1628       (stopped 09:04)
```

4,315/day = one attempt every **20.0 s**, for 20 days, against two hosts that never once
accepted a connection. Neither IP appears in any config file — they entered the dial set
by resolution or from the peers table, and nothing ever removed them until something did
at 07-22T09:04.

**2. Both SatoshiPay `preferred_peers` are dead and still being dialled.**
`configs/validator-mainnet-rpc.toml` lists `stellar-de-fra.satoshipay.io:11625` and
`stellar-us-iowa.satoshipay.io:11625`. Both resolve fine; neither routes. `us-iowa` has
been failing for **26 days** — first as `connection timeout`, switching to
`No route to host` at 07-22T16:15:54Z when a routing change made the failure explicit:

```
2026-07-22T16:15:54.052945Z  WARN henyey_overlay::manager::tick: Failed to connect to
  preferred peer 147.93.132.30:11625: connection failed: 147.93.132.30:11625:
  No route to host (os error 113)
```

Zero `No route to host` lines in the preceding 21 days; then 143/h in the first hour and a
flat 250–262/h ever since — 28,453 lines and counting.

**3. 62,880 dials to peers we are already connected to.** The remote answers
`ERROR code=Conf, msg=already connected peer`, i.e. we completed a TCP connect and a
partial handshake before being told we already had a session. That is ~2,300/day of
wasted connects, and it means the `Self::has_connection_to(&shared.peer_info_cache, addr)`
guard at `crates/overlay/src/manager/tick.rs:485` misses these. Probably a
hostname-vs-resolved-IP aliasing gap (the class #2536 introduced `ResolvedPeerAddr` for),
but I have not confirmed which key misses — filing it as an observation, not a diagnosis.

## Root cause: flat retry delay with no escalation

`crates/overlay/src/manager/tick.rs`:

```rust
26:  pub(super) const TICK_INTERVAL: Duration = Duration::from_secs(3);
31:  pub(super) const PEER_IP_RESOLVE_DELAY: Duration = Duration::from_secs(600);
40:  pub(super) const PEER_IP_RESOLVE_RETRY_DELAY: Duration = Duration::from_secs(10);
43:  const OUTBOUND_CONNECT_RETRY_DELAY: Duration = Duration::from_secs(10);
```

```rust
// fill_preferred_slots, tick.rs:513-518
Err(e) => {
    warn!("Failed to connect to preferred peer {}: {}", addr, e);
    retry_after.insert(key, now + OUTBOUND_CONNECT_RETRY_DELAY);
}
```

The `retry_after` map exists and works — it just re-arms at a **constant 10 s** regardless
of how many consecutive failures a host has had. With `connect_timeout_secs` defaulting to
10 s, a dead host costs ~10 s in the attempt plus 10 s of gate = the observed 20.0 s
period, forever. Same pattern at `tick.rs:589` for the known-peer fill path.

**The escalating-backoff constants this needs already exist three lines above.**
`PEER_IP_RESOLVE_RETRY_DELAY = 10s` scaled by retry count and capped at
`PEER_IP_RESOLVE_DELAY = 600s` is exactly the DNS path's behaviour, mirroring
stellar-core's `OverlayManagerImpl.cpp:45`. The connect path in the same file has the flat
constant and no cap.

For reference, stellar-core does not have this failure mode at this volume: dead preferred
peers land in `mPeersToSend`/failure tracking and are retried on the backoff schedule, and
the log line is emitted once per state transition rather than once per attempt.

## Checked: this is NOT a #3756 contributor

`fill_preferred_slots` `.await`s `connect_to_explicit_peer` inline, and the WARN comes
from `henyey_overlay::manager::tick` — so "a 10 s connect timeout awaited inline, 2-4 dead
hosts per cycle, ~1/min" looks like an exact match for the 13–30 s `phase=28
peer_maintenance` parks of #3756. It is not, and the reason is structural, not
statistical:

- The overlay tick loop is its **own `tokio::spawn`ed task**
  (`start_tick_loop`, `crates/overlay/src/manager/tick.rs:212-229`), with its own
  `interval_at(now + TICK_INTERVAL, TICK_INTERVAL)`. It never runs on the app event loop,
  so a blocking `.await` there cannot park the `select!`.
- `TICK_INTERVAL` is **3 s**; the app's `peer_maintenance_interval` is **10 s**
  (`crates/app/src/app/lifecycle.rs:498`) and its body is offloaded via
  `dispatch_peer_maintenance` (#3689). Different loop, different period, no shared thread.
- Empirically: the two 87,878-attempt hosts stopped at 07-22T09:04, removing ~40% of the
  connect-failure volume. Park p95 did not fall — it kept climbing on the same
  ~810 ms/day line through 07-27. If this were feeding #3756 the slope would have broken.

Recording it explicitly so nobody has to re-derive it, in the same spirit as #3760's
phase-29 ruling-out.

## Why it matters despite being non-blocking

1. **Log-based monitoring.** 305k lines is 30% of all WARN output. In a 41-minute window
   this family is 523 of 1,020 WARN lines; it crowds out the shapes that matter
   (#3756's parks were 42 lines in the same window). Growing at 17k/day.
2. **Wasted connects against third-party infrastructure.** ~2,300/day handshakes to peers
   we already hold sessions with, plus ~4,300/day TCP SYNs at hosts that do not answer.
3. **Preferred-peer slots.** Each attempt takes `pool.try_reserve()`; a permanently-dead
   preferred host consumes a reservation every 20 s ahead of reachable known peers,
   because preferred fill runs first.
4. **The configs are wrong and nothing surfaces it.** Two of fourteen preferred peers have
   been dead for weeks and the only signal is an undifferentiated WARN stream.

## Suggested

Ordered smallest-first, mirroring #3760's structure:

1. **Escalating backoff on the connect path.** Track consecutive connect failures per
   `DialKey` and re-arm `retry_after` at `min(10s * 2^n, 600s)` — reusing
   `PEER_IP_RESOLVE_DELAY` as the cap so both paths share one ceiling. A 26-day-dead host
   would then be dialled ~144×/day instead of 4,320×.
2. **Log on transition, not per attempt.** WARN on first failure and on recovery, DEBUG
   in between, or a periodic rollup (`host X: 1,247 consecutive connect failures over
   6 h 55 m`). Adopting stellar-core's "update your config" wording for a preferred peer
   that has never once connected would make it operator-actionable.
3. **Fix the `already connected peer` re-dial** (62,880 lines). Needs a look at what
   `has_connection_to` keys on versus what `shuffled_dial_entries` yields — likely the
   hostname/resolved-IP aliasing #2536 addressed elsewhere.
4. **Config hygiene.** Both SatoshiPay entries are unreachable from this host; worth
   confirming from a second vantage point before removing them from
   `configs/mainnet.toml`, `configs/validator-mainnet-rpc.toml`,
   `configs/validator-mainnet.toml`, since it may be a route-specific problem rather than
   the hosts being retired.

(1) alone removes ~96% of the volume. (3) is independent of (1) and (2).

Overlay dial scheduling and log verbosity are both on the "MAY deviate freely" side of
`docs/PARITY.md`, so this is **not** a parity defect — it is wasted work plus log
hygiene, same classification as #3760.

## Repro

```bash
LOG=~/data/74535976/logs/monitor.log
grep -c 'Failed to connect to preferred peer' "$LOG"          # 305161
grep 'Failed to connect to preferred peer' "$LOG" \
  | grep -oP 'preferred peer \K[0-9.]+' | sort | uniq -c | sort -rn | head
grep -c 'No route to host' "$LOG"                             # 28453
for ip in 95.216.1.86 88.99.1.97 38.242.250.15 147.93.132.30; do
  timeout 3 bash -c "echo > /dev/tcp/$ip/11625" && echo "$ip OPEN" || echo "$ip unreachable"
done
```

— filed by `/monitor-tick` 2831

tomerweller:
Closes #3770

## Summary

Dead preferred/known peers were re-dialled every ~20s forever under the flat `OUTBOUND_CONNECT_RETRY_DELAY` (10s) constant, producing 305,161 WARN lines over 27 days for a single dead peer with no escalation. This PR replaces the flat delay with `compute_connect_backoff()`, mirroring stellar-core's actual `PeerManager::computeBackoff` (`PeerManager.cpp:365-410`): a jittered random delay in `[1, 2^min(failures,10) * 10]` seconds, doubling the ceiling per consecutive failure and capping at exponent 10 (worst-case ceiling ~2.84h). `retry_after` now stores `(next-allowed-instant, consecutive-failure-count)` per dial key so the backoff escalates across ticks for a peer that keeps failing, and resets to zero the moment a connect to that key succeeds.

Effect: a peer dead for 26 days goes from ~4,320 dial attempts/day (flat 20s period) to ~9-17/day once the ~2.84h ceiling is reached (a few hours into the failure streak).

**Deliberate note for reviewers:** the ~2.84h worst-case ceiling is notably longer than the issue's own suggested cap (`min(10s*2^n, 600s)`). This is intentional — it's what mirrors stellar-core's real algorithm, not an oversight. The issue's own formula got the shape right but the cap/jitter wrong relative to the actual upstream code.

Scoped to the connect-retry path only (item 1 of the issue's suggested fixes — the fix the issue itself estimates removes ~96% of the affected WARN volume). Deliberately deferred as independent follow-ups, per the converged plan:
- Item 2: log on state transition instead of per-attempt.
- Item 3: the separate "already connected peer" re-dial (an independent, undiagnosed correctness question per the issue's own framing).
- Item 4: config hygiene.

The now-dead `OUTBOUND_CONNECT_RETRY_DELAY` constant was removed once both call sites (`connect_preferred_peers`, `fill_outbound_slots`) migrated off it, to avoid tripping `cargo clippy --all -- -D warnings`'s unused-constant lint.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3770#issuecomment-5103821171)

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test -p henyey-overlay` (485 lib tests + integration tests, all passing)

Four new unit tests in `crates/overlay/src/manager/tick.rs` (`mod tests`):
- `test_compute_connect_backoff_first_failure_bounded`
- `test_compute_connect_backoff_caps_at_max_exponent`
- `test_connect_preferred_peers_backoff_increments_and_resets`
- `test_connect_preferred_peers_dial_volume_collapses_over_simulated_time` (drives `connect_preferred_peers` across 1200 simulated ticks / a simulated 1-hour window against a permanently-unreachable preferred peer; asserts dial count < 20, vs. the flat-cadence baseline of ~300-360 dials over the same window)

- **Test:** `crates/overlay/src/manager/tick.rs::tests` (all 4 tests above)
- **Pre-fix:** committed as `637923f4` — verified all 4 FAILED TO COMPILE (`compute_connect_backoff` doesn't exist; `connect_preferred_peers`/`fill_outbound_slots` still take `HashMap<DialKey, Instant>` not the new tuple type)
- **Post-fix:** verified all 4 PASS after `d3a60857` (ran 10x in a row with no flakes)

## Deviations from plan

None — implementation matches the converged plan's file list, formula, and test plan exactly.


crates/overlay/src/manager/tick.rs
```diff
@@ -1786,4 +1815,213 @@ mod tests {
...
+            1,
+            1,
+            &manager.outbound_pool,
+            &shared,
+            &success_ctx,
+        )
+        .await;
+
+        assert!(
+            retry_after.get(&key).is_none(),
```
tomerweller: Non-blocking nit: under `cargo clippy --all-targets -- -D warnings` (stricter than this repo's CI invocation `cargo clippy --all -- -D warnings`, which doesn't compile `#[cfg(test)]` code and is confirmed clean), this line trips `clippy::unnecessary_get_then_check`. Not a merge blocker — CI's actual command is unaffected — but trivially fixable if you touch this test again: `!retry_after.contains_key(&key)`.

crates/overlay/src/manager/tick.rs
```diff
@@ -586,7 +611,11 @@ impl OverlayManager {
                 }
                 Err(e) => {
                     debug!("Failed to connect to peer {}: {}", addr, e);
-                    retry_after.insert(key, now + OUTBOUND_CONNECT_RETRY_DELAY);
+                    let num_failures = retry_after.get(&key).map_or(0, |(_, count)| *count) + 1;
+                    retry_after.insert(
```
tomerweller, line 615: Non-blocking test-coverage note: the new increment/reset behavior (persistent per-key failure count, reset on success) is directly tested for `connect_preferred_peers` (`test_connect_preferred_peers_backoff_increments_and_resets`) but not for its sibling `fill_outbound_slots`, even though both `Err`/`Ok` arms were changed identically here. The existing #3736/#3710 regression test for `fill_outbound_slots` still passes under the new tuple type (confirms it compiles/functions), but doesn't specifically assert the failure-count increment or reset. Given the two functions are structurally symmetric and the shared helper (`compute_connect_backoff`) is independently well-tested, I'm not blocking on this — but a follow-up test mirroring the preferred-peer one for `fill_outbound_slots` would close the gap cheaply.

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** TDD-verified regression test (fails to compile pre-fix, passes post-fix, 10x stable per PR body — independently re-verified below), correct/symmetric wiring at both call sites, clean dead-code removal. Two non-blocking inline notes left (test-coverage gap for `fill_outbound_slots`, a `--all-targets`-only clippy nit).

<details>

**Cycle 1 — this is my first pass on this PR. Complete change-list below, concerns labeled by class.**

**Test-verification gate (`kind: bug-fix`, per issue #3770's Triage Report):**
- PR body has a `## Regression test` section identifying `crates/overlay/src/manager/tick.rs::tests` (4 new tests) as the regression test, with explicit pre-fix/post-fix commit SHAs.
- Independently re-verified (not just trusted the PR body): checked out the test commit `637923f4` in a fresh worktree merged onto current `origin/main` — `cargo test -p henyey-overlay --lib compute_connect_backoff` fails to **compile** (`E0308`/`E0425`: `compute_connect_backoff` doesn't exist, `connect_preferred_peers` still expects `HashMap<DialKey, Instant>`), matching the PR's claim exactly. Checked out the fix commit `d3a60857` — all 4 new tests pass. Full `cargo test -p henyey-overlay --lib` (485 tests) and `--tests` (integration suites) both pass clean.
- **Class: `test-coverage`** — the new tests cover `compute_connect_backoff` in isolation (two bound checks) and `connect_preferred_peers`'s increment/reset wiring end-to-end (including a dial-volume-collapse simulation), but not `fill_outbound_slots`'s identical wiring directly (left as a non-blocking inline comment — the existing #3736/#3710 regression test for `fill_outbound_slots` does still pass under the new tuple signature, so it's not a coverage *hole* per se, just narrower than ideal for a mirrored code path).

**Dead-code removal (`OUTBOUND_CONNECT_RETRY_DELAY`):**
- **Class: `dead-code-removal`** — verified via repo-wide grep that the constant has zero remaining references anywhere (both call sites fully migrated to `compute_connect_backoff`). Correct classification, clean deletion, not a narrowing case.

**Logic correctness:**
- **Class: `logic`** — `compute_connect_backoff`'s formula (`rand[1, (1<<min(failures,10))*10]` seconds) and the increment-on-`Err`/reset-on-`Ok` wiring are correct and symmetric across both `connect_preferred_peers` and `fill_outbound_slots`. `num_failures` is read from the existing tuple before insert (correct read-before-write ordering), and `retry_after.remove(&key)` on success correctly resets both the gate and the failure count together.

**Lint/format:**
- **Class: `ci-hygiene`** — `cargo fmt --check` clean. `cargo clippy --all -- -D warnings` (the exact command CI's `Clippy` job runs, per `.github/workflows/ci.yml:48`) is clean, confirmed independently. Note (non-blocking, left inline): `cargo clippy --all-targets -- -D warnings` (stricter — also compiles `#[cfg(test)]` code) flags one line in the new test code (`clippy::unnecessary_get_then_check`) plus one pre-existing, PR-unrelated `identity_op` lint in `codec.rs`. Since CI doesn't run `--all-targets`, this isn't a merge blocker, but flagging for awareness.

**Readability/error-handling:** No concerns. Doc comments on `compute_connect_backoff` are clear and correctly cite the stellar-core source location. Error handling paths (`warn!`/`debug!` on failure) are unchanged in shape, only the backoff computation changed.

**Self-modifying paths:** N/A — this PR only touches `crates/overlay/src/manager/tick.rs`, not any pipeline skill/script path.

</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** Confirmed non-parity — the change only affects internal connect-retry scheduling (when to attempt a TCP dial), not any observable/interop surface. Independently verified the backoff formula against `stellar-core/src/overlay/PeerManager.cpp`'s real `computeBackoff`/`update` and confirmed it's an exact-shape mirror.

<details>

**Cycle 1 — this is my first pass on this PR. No blocking concerns found; documenting the classification reasoning per discipline.**

**Class: `observable-surface-scope`** — Per `docs/PARITY.md`'s enumerated MUST-match surface (ledger/bucket hashes, tx result/meta XDR, SCP/overlay wire bytes, history archive format, HTTP/RPC/CLI, crypto outputs), this diff touches none of them:
- The `StellarMessage` wire framing, auth handshake bytes, and flood/flow-control logic (the actual overlay parity surface per `docs/PARITY.md:33`) are completely untouched — this PR only changes *when* `connect_to_explicit_peer` is called (a TCP-dial scheduling decision that happens entirely client-side, before any protocol bytes are exchanged), not what bytes are sent once a connection is attempted.
- No ledger, bucket, tx-meta, SCP-message, history-archive, RPC, or crypto code paths are touched. `git diff --name-only` confirms the diff is scoped to exactly one file (`crates/overlay/src/manager/tick.rs`), and within it, only the retry-scheduling helper and its two call sites.
- This matches the issue's own Triage Report framing ("Explicitly confirmed non-parity by the filer... overlay dial scheduling/log verbosity is 'MAY deviate freely' per docs/PARITY.md") and correctly required no `PARITY_STATUS.md` / README parity-column update (verified: no existing `crates/overlay/PARITY_STATUS.md` entry tracks connect-retry cadence as a distinct parity item — `PeerManager` itself is already listed `Full`, and this PR doesn't touch `PeerManager`, only the overlay tick loop's own retry bookkeeping).

**Class: `spec-fidelity` (informational, not a parity requirement, but verified since the PR explicitly claims it)** — independently re-derived against `stellar-core/src/overlay/PeerManager.cpp:365-378`:
- `SECONDS_PER_BACKOFF = 10`, `MAX_BACKOFF_EXPONENT = 10` — exact constant match.
- Core: `random() % ((1<<backoffCount)*SECONDS_PER_BACKOFF) + 1` → range `[1, ceiling]`. Henyey: `gen_range(1..=ceiling_secs)` → also `[1, ceiling]` inclusive. Same distribution shape (not bit-identical RNG, which is fine and expected — this isn't a determinism-gated code path; two nodes independently choosing different jittered retry delays is the *intended* anti-synchronization behavior, not a divergence bug).
- Core's `BackOffUpdate::INCREASE` increments `mNumFailures` at dial-attempt time (before the outcome is known), with `RESET` applied only after full auth-handshake success (`OverlayManagerImpl.cpp:423` area). Henyey increments only in the `Err` arm (after the outcome is known) and resets in the `Ok` arm. This is a minor timing-model difference (core could theoretically apply one extra backoff tick to an attempt that ultimately succeeds slowly, then immediately RESET on handshake success; henyey never does since `connect_to_explicit_peer` only returns `Ok` post-handshake) but does not change steady-state behavior and, critically, is not part of the observable/interop surface — it's an internal scheduling heuristic a peer cannot observe from the wire. Not a parity gap; noting for completeness only.

No `**NEW CLASS DISCOVERED**` — first cycle.

</details>

tomerweller:
## Review: Waiting on CI

Both reviewer lenses APPROVE (Correctness, Parity). CI is still running (checks: `Test`, `Doctest`, `build / 2 build / 4 build (core, ...)`). Re-picking this PR on the next tick.

CI age: 9 min / budget 60 min.
Bounce-back count: 0/3 (head-scoped), 0/6 (lifetime).

tomerweller:

**Reviewer A:** APPROVE — TDD-verified regression tests, symmetric wiring at both call sites, clean dead-code removal.
**Reviewer B:** APPROVE — confirmed non-parity (connect-retry scheduling only, no observable/interop surface); backoff mirrors stellar-core `PeerManager.cpp` `computeBackoff`.
**External reviewers:** tomerweller: COMMENTED (neutral, non-blocking)
**CI:** red — but the single failure is unrelated to this diff.

The only failing check is `test (testnet, core,horizon, ...)` in the Quickstart workflow. The failing probe is `test_horizon_core_up.go` (Horizon↔stellar-core sync against **live testnet**), which the harness itself classified as transient infra:

```
=== Transient-infra failure (exit 124) on retryable shard (testnet/core,horizon/horizon-core-up), retrying ===
##[error]The action 'Run probes through wrapper' has timed out after 25 minutes.
```

This shard runs with `SHARD_SOFT_ON_TIMEOUT: true` (known-flaky testnet shard). It is **not diff-attributable**: this PR changes only `crates/overlay/src/manager/tick.rs` (internal outbound-connect backoff scheduling), which cannot affect a Horizon↔core testnet sync probe. Every other CI job is green — all Rust unit/integration tests, Clippy, Format, Build, CodeQL, and all `local`/`pubnet` Quickstart shards.

Routing back to `ready-for-doing` for `/do` Mode B to rebase on `origin/main` and re-trigger CI. If the same testnet shard is still red after rebase, the next `/review-pr` will mark this `blocked` for operator attention.

crates/overlay/src/manager/tick.rs
tomerweller: Addressed in f3be995a1923865d2e1645781900c02ea87a9557: changed to `!retry_after.contains_key(&key)`, clearing the `clippy::unnecessary_get_then_check` warning under `--all-targets`.

crates/overlay/src/manager/tick.rs
tomerweller: Addressed in f3be995a1923865d2e1645781900c02ea87a9557: added `test_fill_outbound_slots_backoff_increments_and_resets`, a sibling of the preferred-peers test that asserts the per-key failure-count increment (1,2,3) on consecutive failures and reset-on-success for `fill_outbound_slots`. Closes the coverage gap.

tomerweller (commented):
Rebased on `origin/main` and re-pushed to re-trigger CI (the prior red was the known-flaky `test (testnet, core,horizon...)` Quickstart shard, not diff-attributable). Also addressed the two non-blocking inline nits: fixed the `clippy::unnecessary_get_then_check` in the preferred-peers test and added a sibling backoff increment/reset test for `fill_outbound_slots`. Ready for re-review.

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Cycle 2 re-review on the current head (`f3be995`). Both of my cycle-1 non-blocking inline notes are now addressed. Logic remains correct and symmetric at both call sites; regression tests are TDD-verified and green in CI.

<details>

**Cycle 2 — prior `## 🔍 Reviewer: Correctness` verdict was APPROVE with two non-blocking inline notes. Verifying those are addressed; sticking to my prior classes (`test-coverage`, `ci-hygiene`, `logic`, `dead-code-removal`). No `**NEW CLASS DISCOVERED**`.**

- **Class `test-coverage` (prior note — resolved):** the cycle-1 gap was that `fill_outbound_slots`'s identical wiring wasn't directly covered. This head adds `test_fill_outbound_slots_backoff_increments_and_resets`, a direct sibling of the preferred-peers increment/reset test. The mirrored `Err`/`Ok` path is now covered end-to-end. Coverage now spans: `compute_connect_backoff` bounds (first-failure + max-exponent cap), both call sites' increment/reset wiring, and a 1200-tick dial-volume-collapse simulation.
- **Class `ci-hygiene` (prior note — resolved):** the `clippy::unnecessary_get_then_check` note in the preferred-peers test is gone; the new tests use `.get(&key).expect(...)` / `!contains_key(&key)` which are clean. CI `Clippy`, `Format`, and `Doctest` checks are all green on this head.
- **Class `logic` (re-verified):** `num_failures = retry_after.get(&key).map_or(0, |(_, count)| *count) + 1` then insert is correct read-before-write ordering; both success arms (`tick.rs:533`, `tick.rs:609`) `retry_after.remove(&key)` to reset gate + failure count together on connect success. Symmetric across `connect_preferred_peers` and `fill_outbound_slots`.
- **Class `dead-code-removal` (re-verified):** `OUTBOUND_CONNECT_RETRY_DELAY` removed; both call sites migrated. Zero dangling references (green Clippy `-D warnings` confirms).
- **Test-verification gate (`kind: bug-fix`):** PR body documents pre-fix commit `637923f4` (4 tests fail to compile) → post-fix `d3a60857` (pass); the CI `Test` job is SUCCESS on the current head, so the suite compiles and passes as landed.

Self-modifying: N/A — only `crates/overlay/src/manager/tick.rs` touched.

</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** Cycle 2 re-review on head `f3be995`. Still confirmed non-parity: the change only alters *when* an outbound TCP dial is attempted (internal connect-retry scheduling), touching none of the observable/interop surface. The added sibling test doesn't change the classification.

<details>

**Cycle 2 — prior `## 🔍 Reviewer: Parity` verdict was APPROVE. Re-verifying my prior class (`observable-surface-scope`, `spec-fidelity`). No `**NEW CLASS DISCOVERED**`.**

- **Class `observable-surface-scope` (re-verified):** `git diff --name-only origin/main...HEAD` is exactly one file (`crates/overlay/src/manager/tick.rs`). Per `docs/PARITY.md`, none of the MUST-match surfaces are touched — no ledger/bucket hashes, tx result/meta XDR, SCP/overlay wire bytes, history archive format, HTTP/RPC/CLI, or crypto outputs. `StellarMessage` framing, the auth handshake, and flood/flow-control are untouched; only the client-side dial-scheduling decision (before any protocol bytes) changed. Correctly required no `PARITY_STATUS.md` / README parity-column update.
- **Class `spec-fidelity` (informational, re-confirmed):** the backoff mirrors stellar-core `PeerManager::computeBackoff` (`SECONDS_PER_BACKOFF=10`, `MAX_BACKOFF_EXPONENT=10`, jittered range `[1, ceiling]`). Independent per-node jitter is the intended anti-synchronization behavior, not a determinism divergence — this path is explicitly not determinism-gated.
- The only delta since cycle 1 is a new `#[cfg(test)]` sibling test and the resolved clippy nit — neither touches any observable surface.

</details>

tomerweller:

**Reviewer A:** APPROVE — cycle-2 correctness re-review clean; both prior non-blocking inline notes addressed (added `test_fill_outbound_slots_backoff_increments_and_resets`, resolved the test clippy nit).
**Reviewer B:** APPROVE — cycle-2 parity re-review clean; still confirmed non-parity (internal connect-retry scheduling only, no observable/interop surface touched).
**External reviewers:** none (the only PR reviews are from the pipeline's own author account, which GitHub downgrades to COMMENTED — non-blocking).
**CI:** red

**Diagnosis — unrelated flaky CI, not diff-attributable:**
- 41 of 42 rollup checks are green, including every deterministic gate: CI `Test`, `Clippy`, `Format`, `Doctest`, `Build (release-ci)`, CodeQL (`Analyze rust/python/actions`), Gitlinks, all Quickstart build stages, and the `local` / `pubnet` / `galexie` Quickstart shards.
- The single red is `test (testnet, core,horizon, ...)` — the live-testnet Quickstart shard — with conclusion **CANCELLED** after ~55 min (00:50:29 → 01:45:29) and **no failed step recorded** (job-level timeout, not a test assertion failure). This is the same known-flaky testnet shard the author already flagged on the prior push; it connects to live Stellar testnet and is not exercising the overlay connect-retry backoff change in any deterministic way.
- The diff is scoped to one file (`crates/overlay/src/manager/tick.rs`) changing only *when* an outbound TCP dial is scheduled. Nothing about it deterministically breaks a live-testnet integration shard while leaving `local`/`pubnet` green.

**Action for `/do` Mode B:** rebase on `origin/main` and re-push to re-trigger CI (the deterministic suite is already green; this is purely to clear the flaky testnet shard). If the testnet shard is CANCELLED/red again after a clean rebase, the next `/review-pr` will mark this `blocked` for operator attention per the unrelated-CI-red escalation rule.

Bounce-back count: 1/3 (lifetime 1/6).

Routing back to `ready-for-doing`.

tomerweller:
## Do: Re-triggered flaky CI (Mode B, bounce cycle 1)

Both cycle-2 reviewers (Correctness, Parity) APPROVE, and the only red was the known-flaky live-testnet Quickstart shard (`test (testnet, core,horizon, ...)`), CANCELLED after ~55m with no failed step — a job-level timeout, not a diff-attributable failure. Every deterministic gate (Test, Clippy, Format, Doctest, Build, CodeQL, local/pubnet/galexie shards) is green.

Branch `do/issue-3770` is already based on the current `origin/main` (`ba1d9d7`), so the prescribed rebase is a strict no-op (HEAD unchanged at `f3be995`, diff still the single file `crates/overlay/src/manager/tick.rs`). With no new code and no rebase delta there is nothing to push, so I re-ran the failed testnet shard directly to clear the flake rather than fabricate an empty commit.

Re-run triggered on run `30501705901` (testnet shard now `in_progress`). Ready for re-review. If the testnet shard is CANCELLED/red again on this clean head, `/review-pr` should escalate to `blocked` for operator attention per the unrelated-CI-red rule.

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The escalating-backoff fix is correct, panic-free, and well-covered by four new regression tests including a simulated-hour dial-volume-collapse test. No blocking concerns.

<details>

Cycle 1 (no prior `## 🔍 Reviewer: Correctness` comment) — complete class-labeled change-list.

**Correctness — clean:**
- `compute_connect_backoff(num_failures)` returns a jittered delay in `[1, 2^min(n,10)*10]`s. No overflow: `backoff_count` caps at `MAX_BACKOFF_EXPONENT=10`, so `(1u32 << 10) * 10 = 10240` fits comfortably in `u32`. `gen_range(1..=ceiling_secs)` never panics — `ceiling_secs ≥ 20` for every `num_failures ≥ 1` (and the helper is only ever called with `existing_count + 1 ≥ 1`).
- `retry_after` widened from `HashMap<DialKey, Instant>` to `HashMap<DialKey, (Instant, u32)>`. The consecutive-failure count persists across ticks and escalates: `num_failures = retry_after.get(&key).map_or(0, |(_, count)| *count) + 1`. First failure → 1, subsequent → prior+1. The gate `if let Some((next, _)) = retry_after.get(&key) { if *next > now { continue; } }` correctly ignores the count for the time check. Both `connect_preferred_peers` and `fill_outbound_slots` `Err` arms were changed identically; the `Ok` arms retain the existing `.remove(&key)` reset, verified by the two increment/reset tests asserting `!retry_after.contains_key(&key)` after a success.

**Test-verification gate (kind: bug-fix) — satisfied:**
- Four new tests in `crates/overlay/src/manager/tick.rs::tests`. PR body documents pre-fix commit `637923f4` where all four fail (the type-signature change + missing `compute_connect_backoff` make them not compile against the old code), and post-fix `d3a60857` where all pass (ran 10x, no flakes).
- `test_connect_preferred_peers_dial_volume_collapses_over_simulated_time` is the true behavioral regression test: 1200 ticks × 3s = simulated 1h against a permanently-unreachable preferred peer, asserting `dial_count < 20` vs the old flat-cadence ~300-360 baseline. The `< 20` bound has robust headroom — cumulative expected backoff draws reach the failure-8 ceiling (~2560s) well within the hour, so expected dials are ~8-10; reaching 20 would require pathologically small consecutive RNG draws while ceilings double, which is effectively impossible within 3600s. Not a flake risk.
- Sibling `test_fill_outbound_slots_backoff_increments_and_resets` locks in the symmetric known-peer path.

**Dead-code / visibility gate — satisfied:** `OUTBOUND_CONNECT_RETRY_DELAY` removed after both call sites migrated (avoids the unused-constant clippy lint); `compute_connect_backoff` is a private `fn`. Clippy CI green.

**Readability:** clear doc comment with upstream citation (`PeerManager.cpp:365-410`), simulated-time test helper `max_connect_backoff_secs` mirrors the production formula so tests don't depend on non-seedable RNG internals.

No concerns to raise. (Implicit contract: no class held back for a later cycle.)
</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** The change touches only outbound dial-retry scheduling, a private helper, and log volume — none of the observable/interop surface. Dial scheduling and log verbosity are explicitly deviation-permitted per docs/PARITY.md, and the change moves closer to stellar-core's real algorithm.

<details>

Cycle 1 (no prior `## 🔍 Reviewer: Parity` comment) — complete class-labeled change-list.

`crates/overlay/` is in the parity-critical prefix list, so this PR gets the parity lens. Per `docs/PARITY.md`, the observable/interop surface is: ledger/bucket hashes, tx result & meta XDR, **SCP/overlay wire bytes**, history archive format, HTTP/RPC/CLI contracts, crypto outputs.

**No observable-surface impact:**
- No change to any overlay message serialization / wire bytes — the diff touches only *when* an outbound connect is attempted (the `retry_after` gate cadence), not the bytes of any handshake or message.
- No XDR, no SCP logic, no handshake protocol, no history-archive format, no HTTP/RPC/CLI contract.
- The only externally-visible effect is reduced WARN log volume and fewer TCP SYNs against dead peers — both explicitly on the "MAY deviate freely" side of `docs/PARITY.md` (dial scheduling, logging, perf).

**Direction of change is parity-positive:** the flat 10s retry was the deviation; `compute_connect_backoff` mirrors stellar-core's actual `PeerManager::computeBackoff` (jittered `[1, 2^min(n,10)*SECONDS_PER_BACKOFF]`, cap at exponent 10). The ~2.84h worst-case ceiling deliberately matches upstream rather than the issue's rougher suggested cap — correct call.

Not a parity defect. No concerns.
</details>

tomerweller:

**Reviewer A (Correctness):** APPROVE — escalating-backoff fix is correct, panic-free, and covered by four regression tests (incl. a simulated-hour dial-volume-collapse test).
**Reviewer B (Parity):** APPROVE — touches only dial-retry scheduling / private helper / log volume; no observable/interop surface; explicitly deviation-permitted per docs/PARITY.md and moves closer to stellar-core's real `PeerManager::computeBackoff`.
**External reviewers:** none.
**CI:** red — 1 of 42 checks failed.

Both agent reviewers APPROVE. The sole CI red is the known-flaky **`test (testnet, core,horizon, ...)` Quickstart shard**, which is **not diff-attributable** — this PR only changes `crates/overlay/src/manager/tick.rs` (outbound connect-retry backoff), and the shard failed on a transient testnet-infra timeout:

```
=== Transient-infra failure (exit 124) on retryable shard (testnet/core,horizon/horizon-core-up), retrying ===
=== Attempt 2: horizon-core-up (testnet/core,horizon) ===
... [test] Waiting for Horizon's stellar-core to start reporting  (repeated) ...
```

The shard's own internal retry (attempt 2) also timed out waiting for testnet Horizon's stellar-core to sync — a testnet-side infra problem, not a code defect. All 41 other checks (Format, Clippy, Test, Doctest, Build, CodeQL, all local/pubnet Quickstart shards, etc.) are green.

Routing back to `ready-for-doing` for `/do` Mode B: rebase on `origin/main` and re-push to re-trigger CI. If the testnet shard fails again after a clean rebase, the next `/review-pr` will mark this `blocked` for operator attention.

tomerweller:
Rebased onto `origin/main` (clean, 28 commits behind; re-rebased after #3783 landed) and merging. Verified locally with real exit codes: `cargo fmt --check` 0, `cargo clippy --all --all-targets` 0 with zero lints (clears the newer #3830 gate), `cargo test -p henyey-overlay --lib` 493/493 including the three `manager::tick::tests::test_dns_backoff_*` cases.

**Relationship to merged #3866 — complementary, zero overlap.** Worth stating since both are "backoff" work landing days apart:
- #3771 touches only `crates/overlay/src/manager/tick.rs`; #3866 touched `crates/app/src/app/{peers.rs,mod.rs,lifecycle.rs}`. Different crates, no shared function or state.
- #3866 gates *whether a DNS syscall runs* (`resolve_peers_for_storage`); #3771 gates *whether a dial is attempted* (`connect_preferred_peers` / `fill_outbound_slots`).
- **No compounding on the dial path**, and no stranding risk: because `DialKey::Hostname` ≠ `DialKey::Resolved`, DNS recovery produces a fresh key. #3805 and #3765 don't touch `tick.rs`.

**Algorithm checks out against upstream.** `gen_range(1..=(1<<min(n,10))*10)` s ⇒ `[1 s, 10240 s]`, ceiling 2.84 h — byte-for-byte the shape of `stellar-core/src/overlay/PeerManager.cpp:365-378`, and reset-on-success matches `Peer.cpp:1772` `BackOffUpdate::RESET`. Core also gates *preferred* dials on `nextattempt` (`OverlayManagerImpl.cpp:333`), so gating preferred peers here is not a regression. Retry timing only — no wire framing — so this sits on the freely-divergeable side of `docs/PARITY.md`. Tests are deterministic (injected `now: Instant`, in-process `LoopbackConnectionFactory`, no sleeps).

⚠️ **Gap all four prior reviews missed — filing as a follow-up, not blocking this merge.** `crates/overlay/src/peer_manager.rs:605` **already contains an identical `compute_backoff`**: same `SECONDS_PER_BACKOFF = 10`, same `MAX_BACKOFF_EXPONENT = 10`, same `gen_range(1..=max)`, cited to the same `OVERLAY_SPEC §10.3-1` — alongside an `update_backoff()` that **persists `num_failures`/`next_attempt` to SQLite**. That `update_backoff` currently has **zero production callers** (tests only).

Two consequences: (1) this PR duplicates the function rather than reusing it, so the constants can now drift apart silently; (2) #3771's state is **in-memory**, so backoff resets on every process restart — whereas core persists `peers.nextattempt`, meaning a restart-looping node re-dials known-bad peers immediately on each boot. Merging the in-memory version is a clear improvement over no backoff at all, but consolidating onto the persisted path is the correct end state.

Production motivation, for the record: the live validator logged **511× `"Failed to connect to preferred peer <addr>: connection timeout"`** in a ~2 h window while healthy and at tip, on a binary predating any backoff.

Sole non-green check was `test (testnet, core,horizon, …)` — log ends `runner has received a shutdown signal … exit code 143`, the chronic #3768 infra cancel, not diff-attributable. That CI was from 2026-07-30 and predates #3830, which is why the tightened clippy gate was re-run locally.

PR #3781, opened 2026-07-30
Retry transiently-busy maintenance delete chunks

### LINKED ISSUE #3772: SQLite busy_timeout expiry silently drops 113 of 139 writes in 26d — including the retention trims that would relieve the contention, and catchup LedgerCloseMeta
## Summary

Every write that loses the SQLite write-lock race for the full `busy_timeout=30s` window is **silently dropped** — except the two consensus-critical persist commits, which retry. On the deployed validator that has happened **139 times in 26 days across 5 call sites**, and three of those sites are the *retention/cleanup* writes. So the lock contention prevents exactly the trimming that would relieve the lock contention.

The event-loop-park half of this is #3689 (closed, fix undeployed) and #3756. This issue is about the **write-loss half**, which neither covers: #3756's body does not mention `maintainer`, `delete_old`, `LedgerCloseMeta`, or `database is locked`; #3749's does not mention the maintainer or trimming.

Node is healthy right now — nothing here is currently blocking, hence no `urgent` label.

## Evidence — all 139 `database is locked` lines, deployed binary `0ac84d42`, 26 d uptime

```
  56  WARN henyey_app::app::lifecycle: Failed to update peer record err=…DatabaseBusy…
  45  WARN henyey_app::app::peers:     Failed to remove failed peers error=…database is locked
  26  WARN henyey_app::app::persist:   Transient SQLite busy on persist DB commit; retrying after backoff
                                        context="ledger close DB write" attempt=N max_attempts=4
   6  WARN henyey_app::app::peers:     Failed to refresh known peers from DB error=…
   2  WARN henyey_history::catchup::persist: Failed to persist LedgerCloseMeta during catchup … ledger_seq=N
   2  WARN henyey_app::maintainer:     Failed to delete old SCP entries error=…
   1  WARN henyey_app::maintainer:     Failed to delete old tx history error=…
   1  WARN henyey_app::maintainer:     Failed to delete old ledger headers error=…
```

**Only the 26 `app::persist` lines retry.** The other **113 writes are dropped**, log-and-continue.

Distribution is **episodic, not linear** — this matters because #3756 tracks a linear 892 ms/day park growth, and these ceiling *failures* are a separate regime:

```
2026-07-20    4
2026-07-22   66
2026-07-23   37
2026-07-24…27  0
2026-07-28   32   ← by 13:10Z (first burst in 5 days)
```

## 1. The retention trim is starved by the bloat it exists to remove

`0ac84d42:crates/app/src/maintainer.rs:390-405` — three log-and-continue writes:

```rust
if let Err(e) = db.delete_old_scp_entries(publish_safe_lmin, count) {
    warn!(error = %e, "Failed to delete old SCP entries");
}
if let Err(e) = db.delete_old_ledger_headers(publish_and_rpc_lmin, count) {
    warn!(error = %e, "Failed to delete old ledger headers");
}
if let Err(e) = db.delete_old_tx_history(publish_and_rpc_lmin, count) {
    warn!(error = %e, "Failed to delete old tx history");
}
```

A dropped `delete_old_scp_entries` is not a cosmetic loss: it is the write that bounds SCP history. `mainnet.db` is **152 GiB** with `storestate` at 766,925 rows (#3749), and the larger the DB the longer each write holds the lock, and the more often the trim loses the race. Four observed failures is small in absolute terms — the point is the sign of the feedback term, and that nothing in the code bounds it.

Same shape, same file: #3754 already reports that this maintenance cycle structurally overruns its 10 s threshold.

## 2. `commit_with_busy_retry`'s documented scope does not match the code

`0ac84d42:crates/app/src/app/persist.rs:650-656` states the design intent:

> Only the two persist commits that gate consensus continuation (ledger-close + catchup) are retried. The publish-queue dequeue (`remove_publish`, warn-only) and the SCP tx-set purge (best-effort) are intentionally NOT retried — they do not halt consensus.

But the catchup meta write is a **different site in a different crate** and is *not* wrapped — `0ac84d42:crates/history/src/catchup/persist.rs:179-188`:

```rust
pub(super) fn emit_meta(&self, ledger_seq: u32, meta: stellar_xdr::LedgerCloseMeta) {
    match meta.to_xdr(stellar_xdr::Limits::none()) {
        Ok(meta_xdr) => {
            if let Err(err) = self.db.store_ledger_close_meta(ledger_seq, &meta_xdr) {
                warn!(error = %err, ledger_seq, "Failed to persist LedgerCloseMeta during catchup");
            }        // ← dropped; no retry, no error propagation
```

Two ledgers lost their `ledger_close_meta` row this way:

```
2026-07-22T20:07:37Z  ledger_seq=63600058
2026-07-23T00:06:38Z  ledger_seq=63602606
```

Per the function's own doc comment, that table is what RPC `getTransactions` / `getLedgers` read. Both ledgers are long past the 360-ledger RPC retention window, so the hole is no longer observable and I am **not** claiming an observed RPC failure — the claim is that the code permits a silent, unreported gap in a table an external consumer queries.

## 3. `henyey_db_busy_retry_total` instruments 1 of 5 sites

The counter is incremented only inside `commit_with_busy_retry` (`persist.rs:684`), and only while `n < MAX_PERSIST_ATTEMPTS`. Live scrape agrees exactly with the log: **`henyey_db_busy_retry_total 26`** vs 26 `app::persist` WARN lines. It is not miscounting its own path — it simply cannot see the other **113**. Lock-loss on this validator is therefore invisible to telemetry and only discoverable by grepping 3 M log lines.

## 4. Corroboration of the park mechanism (quantifies #3689 / #3756)

All **12** of today's ≥30 s event-loop parks have a `database is locked` failure in the **same log second**, 12/12:

```
03:42:35 phase=28 30066ms   04:27:35 phase=28 30068ms   04:38:45 phase=28 30111ms
04:41:35 phase=28 30060ms   04:43:35 phase=28 30021ms   04:44:05 phase=29 30054ms
05:20:36 phase=29 30202ms   05:21:37 phase=29 30207ms   09:49:35 phase=28 30061ms
09:50:35 phase=28 30035ms   10:01:45 phase=28 30113ms   13:02:35 phase=28 30094ms
```

That is the same pairing #3689 reported from a single 2026-06-30 event (`30187ms`), now at n=12 on one day. It also mechanically confirms that the phase-28 park is real inline work on the deployed binary and not a stale phase marker — the correction published at [#3756 comment 5104268237](https://github.com/stellar-experimental/henyey/issues/3756#issuecomment-5104268237).

## Suggested direction (not implemented — monitor loop files, does not commit)

1. Wrap the three `maintainer.rs` retention deletes in the existing `commit_with_busy_retry` (or an equivalent bounded retry). A starved trim is not "best effort"; it is a positive feedback term.
2. Either wrap `emit_meta`'s `store_ledger_close_meta`, or make the drop loud and recorded (counter + explicit gap marker), so an RPC consumer can detect the hole.
3. Increment `henyey_db_busy_retry_total` (or a sibling labelled by call site) at **all** `DatabaseBusy` sites. One `busy_lock_lost_total{site="…"}` counter would have surfaced all of this from `/metrics`.
4. Reconcile the `commit_with_busy_retry` doc comment with the actual set of wrapped sites.

## Environment

- Binary `0ac84d4274c351514532011d46a1219c88c8582d`, **60 commits behind** `origin/main`; PID 1512116, uptime 26 d 15 h, 0 restarts. Deploys held on #3702.
- Session 74535976, mainnet validator+RPC. `Validating` L63,687,887, protocol 27, 27 authenticated peers, quorum `agree=18 disagree=0 missing=3 delayed=0`, RPC `healthy`.
- `PRAGMA busy_timeout = 30000` (`crates/db/src/database/mod.rs:72`); `mainnet.db` 152 GiB.

## Related

- #3689 (closed) — same pairing, single event; its offload fix is on `main` and has never deployed.
- #3756 — the park/duty-cycle half. #3754 — the same maintainer cycle overrunning. #3749 — the DB growth that lengthens every lock hold. #3702 — why no fix can ship.
- #3640 / #3497 (closed) — introduced the bounded retry this issue proposes extending.

— filed by `/monitor-tick` 2834

tomerweller:
Closes #3772

## Summary

Every retention-trim write that lost the SQLite `busy_timeout` write-lock race used to silently abort that table's trim for the whole maintenance pass — starving exactly the cleanup that would relieve the lock contention. This wraps each chunk delete in `delete_in_chunks` (`crates/app/src/maintainer.rs`, the single helper all five retention-delete sites route through) in a small bounded retry (`MAINTENANCE_DELETE_MAX_ATTEMPTS = 3`) that retries **only** on a transient `DatabaseBusy`/`DatabaseLocked` classification, backing off `MAINTENANCE_CHUNK_PAUSE` (25 ms) between attempts. Genuine (non-transient) errors still fail closed immediately with no retry.

The classifier reused is the existing `is_transient_db_busy` from the consensus-persist path, widened to `pub(crate)` and re-exported through `crate::app` (the maintainer is a sibling module, so `pub(crate)` on the fn alone is not reachable — mirrors the existing `CatchupFinalizer` re-export).

This deliberately does **not** reuse `commit_with_busy_retry`, whose exhausted-budget path triggers a recoverable shutdown — disproportionate for a routine retention trim that can wait for the next maintenance cycle.

**Premise note:** the issue quoted the *deployed* (60-commit-stale) binary; `origin/main` was already refactored (PR #3715) into the single `delete_in_chunks` helper, so this one fix covers all five tables (SCP entries, ledger headers, tx history, events, ledger_close_meta) — a superset of the issue's named three sites.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3772#issuecomment-5105256567)

## Test plan

- [x] cargo fmt --check
- [x] cargo test -p henyey-app passes (all 6 delete_in_chunks tests green)

## Regression test

- **Tests:** `crates/app/src/maintainer.rs`: `test_delete_in_chunks_retries_transient_busy_then_succeeds`, `test_delete_in_chunks_gives_up_after_max_attempts`, `test_delete_in_chunks_does_not_retry_non_transient_error`
- **Pre-fix:** committed as `cd4e5fc` — verified FAILED (retries-then-succeeds fails; gives-up asserts 1 call vs expected 3, since single-shot delete breaks on first busy error)
- **Post-fix:** verified PASS after `b533e87`

## Parity

n/a — non-parity, internal SQLite write-retry policy. Verified this round against the actual `stellar-core/` v27.0.0 submodule (not only inferred from prior research): stellar-core removed its maintenance subsystem entirely, and its closest analogue (`DatabaseUtils::deleteOldEntriesHelper` via `HerderPersistence::deleteOldEntries`) has no chunking and no retry-on-busy — henyey's fail-open behavior is already more fault-tolerant than upstream, and this makes it more so. No wire/consensus-observable behavior touched.

## Accepted trade-off

Each `delete_chunk` can already block up to the connection's 30 s `busy_timeout`; with 3 attempts a genuinely-stuck table trim now costs up to ~90 s instead of ~30 s, and a worst-case fully-stuck pass (5 tables sequential) up to ~450 s instead of ~150 s (feeds #3754). Accepted because `run_maintenance` runs off the tokio event loop (via `spawn_blocking_logged`) — the cost is a longer *maintenance pass*, not a longer *event-loop block* (not a #3756-style freeze).

## Deviations from plan

None.

### DIFF
--- a/crates/app/src/app/mod.rs
+++ b/crates/app/src/app/mod.rs
@@ -532,6 +532,11 @@ mod types;
 mod upgrades;
 
 pub use persist::CatchupFinalizer;
+// Re-exported for `crate::maintainer` (a sibling of `crate::app`, so
+// `pub(crate)` on the fn alone is not reachable across the module boundary):
+// the retention-trim retry loop reuses the same structured DB-busy/locked
+// classifier as the consensus-persist path (#3772).
+pub(crate) use persist::is_transient_db_busy;
 pub(crate) use tx_flooding::{
     FLOOD_OP_RATE_PER_LEDGER, FLOOD_SOROBAN_RATE_PER_LEDGER, FLOOD_SOROBAN_TX_PERIOD_MS,
     FLOOD_TX_PERIOD_MS,
--- a/crates/app/src/app/persist.rs
+++ b/crates/app/src/app/persist.rs
@@ -540,7 +540,7 @@ fn handle_persist_error(context: &str, error: &BucketError, shutdown: &Recoverab
 /// reclassified recoverable. Mirrors the `is_query_interrupted` shape in
 /// `henyey_db::queries` (matching the structured `ErrorCode`, NOT the message
 /// string).
-fn is_transient_db_busy(error: &henyey_db::DbError) -> bool {
+pub(crate) fn is_transient_db_busy(error: &henyey_db::DbError) -> bool {
     matches!(
         error,
         henyey_db::DbError::Sqlite(rusqlite::Error::SqliteFailure(
--- a/crates/app/src/maintainer.rs
+++ b/crates/app/src/maintainer.rs
@@ -347,6 +347,14 @@ const MAINTENANCE_CHUNK_LEDGERS: u32 = 2;
 /// between chunks.
 const MAINTENANCE_CHUNK_PAUSE: Duration = Duration::from_millis(25);
 
+/// Bounded retry count for a transiently-busy maintenance delete chunk.
+/// Deliberately small and maintenance-local — NOT tied to
+/// `commit_with_busy_retry`'s consensus-continuation semantics (that path
+/// escalates to a recoverable shutdown when its budget is exhausted, which is
+/// disproportionate for a routine retention trim: the node should simply wait
+/// for the next scheduled maintenance cycle). See #3772.
+const MAINTENANCE_DELETE_MAX_ATTEMPTS: u32 = 3;
+
 /// Run one table's maintenance deletion in bounded chunks.
 ///
 /// Calls `delete_chunk` with at most [`MAINTENANCE_CHUNK_LEDGERS`] ledgers per
@@ -363,21 +371,56 @@ fn delete_in_chunks(
     let mut total_rows = 0u64;
     while remaining > 0 {
         let chunk = remaining.min(MAINTENANCE_CHUNK_LEDGERS);
-        // Per-chunk WAL write-lock instrumentation (#3702): the guard must
-        // wrap a single bounded statement, not the whole pass, so a reported
-        // long hold is a real single-transaction hold.
-        let _write_ctx = henyey_common::WriteCtxGuard::new("maintenance-delete");
-        match delete_chunk(chunk) {
+        // Retry a transiently-busy chunk a bounded number of times before
+        // giving up on this table's trim (#3772). Losing the SQLite
+        // write-lock race for the whole `busy_timeout` window must not
+        // silently abort the entire retention pass — that starves exactly the
+        // trimming that would relieve the contention. A genuine (non-transient)
+        // error still fails closed immediately, and the budget is small so a
+        // fully-stuck table costs at most MAINTENANCE_DELETE_MAX_ATTEMPTS
+        // attempts. This is deliberately NOT `commit_with_busy_retry`, whose
+        // exhausted-budget path triggers a recoverable shutdown — overkill for
+        // a routine trim that can wait for the next maintenance cycle.
+        let mut outcome: Result<u32, henyey_db::DbError> = Ok(0);
+        let mut attempts = 0u32;
+        while attempts < MAINTENANCE_DELETE_MAX_ATTEMPTS {
+            attempts += 1;
+            // Per-chunk WAL write-lock instrumentation (#3702): the guard must
+            // wrap a single bounded statement, not the whole pass, so a
+            // reported long hold is a real single-transaction hold. Each retry
+            // attempt is its own single delete statement, hence its own guard.
+            let _write_ctx = henyey_common::WriteCtxGuard::new("maintenance-delete");
+            match delete_chunk(chunk) {
+                Ok(rows) => {
+                    outcome = Ok(rows);
+                    break;
+                }
+                Err(e) if crate::app::is_transient_db_busy(&e) => {
+                    outcome = Err(e);
+                    drop(_write_ctx);
+                    // Back off before retrying, unless the budget is exhausted.
+                    if attempts < MAINTENANCE_DELETE_MAX_ATTEMPTS {
+                        std::thread::sleep(MAINTENANCE_CHUNK_PAUSE);
+                    }
+                }
+                Err(e) => {
+                    // Non-transient error: fail closed immediately, no retry.
+                    outcome = Err(e);
+                    break;
+                }
+            }
+        }
+
+        match outcome {
             Ok(0) => break,
             Ok(rows) => {
                 total_rows += rows as u64;
             }
             Err(e) => {
-                warn!(error = %e, "Failed to delete old {} entries", label);
+                warn!(error = %e, attempts, "Failed to delete old {} entries", label);
                 break;
             }
         }
-        drop(_write_ctx);
         remaining = remaining.saturating_sub(chunk);
         if remaining > 0 {
             std::thread::sleep(MAINTENANCE_CHUNK_PAUSE);
@@ -1517,7 +1560,7 @@ mod tests {
     }
 
     /// #3772: a persistently-busy chunk must give up after exactly
-    /// `MAINTENANCE_DELETE_MAX_ATTEMPTS` (== 3) attempts (bounded — no unbounded
+    /// `MAINTENANCE_DELETE_MAX_ATTEMPTS` attempts (bounded — no unbounded
     /// spin), then break the pass. On the unmodified code the loop breaks after
     /// a single call.
     #[test]
@@ -1528,10 +1571,13 @@ mod tests {
             Err(busy_error())
         });
         assert_eq!(
-            calls, 3,
-            "a persistently-busy chunk must be attempted exactly MAINTENANCE_DELETE_MAX_ATTEMPTS (3) times"
+            calls, MAINTENANCE_DELETE_MAX_ATTEMPTS,
+            "a persistently-busy chunk must be attempted exactly MAINTENANCE_DELETE_MAX_ATTEMPTS times"
         );
-        assert_eq!(total, 0, "no rows are deleted when every attempt fails busy");
+        assert_eq!(
+            total, 0,
+            "no rows are deleted when every attempt fails busy"
+        );
     }
 
     /// #3772: a genuine (non-transient) error must NOT be retried — the
@@ -1548,6 +1594,9 @@ mod tests {
             calls, 1,
             "a non-transient error must not be retried (fail closed immediately)"
         );
-        assert_eq!(total, 0, "no rows counted on immediate non-transient failure");
+        assert_eq!(
+            total, 0,
+            "no rows counted on immediate non-transient failure"
+        );
     }
 }
--- a/crates/app/src/maintainer.rs
+++ b/crates/app/src/maintainer.rs
@@ -1473,4 +1473,81 @@ mod tests {
         assert_eq!(calls, 2, "loop must stop at the failing chunk");
         assert_eq!(total, 2, "only the first chunk's rows are counted");
     }
+
+    /// Build a transient `DatabaseBusy` (`SQLITE_BUSY`) error, the exact shape
+    /// `is_transient_db_busy` classifies as recoverable. Mirrors the persist.rs
+    /// test helper; extended_code is documented for traceability only — the
+    /// gate is the primary `DatabaseBusy` code.
+    fn busy_error() -> henyey_db::DbError {
+        henyey_db::DbError::Sqlite(rusqlite::Error::SqliteFailure(
+            rusqlite::ffi::Error {
+                code: rusqlite::ffi::ErrorCode::DatabaseBusy,
+                extended_code: 5,
+            },
+            Some("database is locked".to_string()),
+        ))
+    }
+
+    /// #3772: a chunk that loses the SQLite write-lock race once (transient
+    /// busy) must be retried, not silently abort the whole table's trim. On the
+    /// unmodified single-shot code the first busy error `break`s the loop, so
+    /// the closure is called once and the chunk's rows are lost.
+    #[test]
+    fn test_delete_in_chunks_retries_transient_busy_then_succeeds() {
+        let mut calls = 0u32;
+        let total = delete_in_chunks(100, "test", |chunk| {
+            calls += 1;
+            match calls {
+                // First chunk: fail busy once, then succeed on retry.
+                1 => Err(busy_error()),
+                2 => Ok(chunk * 3),
+                // Second chunk reports no rows -> terminates the pass.
+                _ => Ok(0),
+            }
+        });
+        assert_eq!(
+            calls, 3,
+            "busy chunk must be retried (call 1 fails busy, call 2 succeeds, call 3 drains)"
+        );
+        assert_eq!(
+            total,
+            (MAINTENANCE_CHUNK_LEDGERS as u64) * 3,
+            "the retried chunk's rows must count toward the total"
+        );
+    }
+
+    /// #3772: a persistently-busy chunk must give up after exactly
+    /// `MAINTENANCE_DELETE_MAX_ATTEMPTS` (== 3) attempts (bounded — no unbounded
+    /// spin), then break the pass. On the unmodified code the loop breaks after
+    /// a single call.
+    #[test]
+    fn test_delete_in_chunks_gives_up_after_max_attempts() {
+        let mut calls = 0u32;
+        let total = delete_in_chunks(100, "test", |_chunk| -> Result<u32, henyey_db::DbError> {
+            calls += 1;
+            Err(busy_error())
+        });
+        assert_eq!(
+            calls, 3,
+            "a persistently-busy chunk must be attempted exactly MAINTENANCE_DELETE_MAX_ATTEMPTS (3) times"
+        );
+        assert_eq!(total, 0, "no rows are deleted when every attempt fails busy");
+    }
+
+    /// #3772: a genuine (non-transient) error must NOT be retried — the
+    /// fail-closed behavior for real corruption/integrity failures is
+    /// preserved. Exactly one call, then break.
+    #[test]
+    fn test_delete_in_chunks_does_not_retry_non_transient_error() {
+        let mut calls = 0u32;
+        let total = delete_in_chunks(100, "test", |_chunk| -> Result<u32, henyey_db::DbError> {
+            calls += 1;
+            Err(henyey_db::DbError::Integrity("boom".into()))
+        });
+        assert_eq!(
+            calls, 1,
+            "a non-transient error must not be retried (fail closed immediately)"
+        );
+        assert_eq!(total, 0, "no rows counted on immediate non-transient failure");
+    }
 }


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Bounded transient-busy retry around `delete_in_chunks` is correct, well-scoped, and covered by three regression tests that provably fail at the pre-fix commit (`cd4e5fc`).

<details>

Cycle 1 (no prior Correctness comment) — complete class-labeled review.

- `correctness` — Retry loop is right: `Ok(rows) => break` on success, retry-with-backoff only under `is_transient_db_busy(&e)`, immediate `break` on any non-transient error. Outer `match outcome` preserves the original `Ok(0) => break` drain-termination and row accounting. `saturating_sub` on `remaining` unchanged. No unbounded spin — `attempts < MAINTENANCE_DELETE_MAX_ATTEMPTS` bounds it to 3.
- `correctness` — `WriteCtxGuard` scoping (#3702 instrumentation contract) is preserved: each attempt constructs its own guard so a reported long hold still maps to a single bounded delete statement. On the transient path the guard is explicitly `drop`ped before the backoff sleep (lock not held while sleeping); on success/non-transient the guard drops at inner-loop exit. Correct.
- `test-coverage` (bug-fix kind) — Three new tests: `test_delete_in_chunks_retries_transient_busy_then_succeeds`, `_gives_up_after_max_attempts`, `_does_not_retry_non_transient_error`. They assert call-counts (1→retry→drain = 3; persistent-busy = exactly MAX; non-transient = exactly 1) and row totals. PR body documents the pre-fix commit `cd4e5fc` verified-FAILED and post-fix `b533e87` verified-PASS — regression discipline satisfied.
- `readability` — The `busy_error()` test helper mirrors the persist.rs shape; comment notes `extended_code` is traceability-only and the gate is the primary `DatabaseBusy` code. Warn now carries `attempts` for observability. Clear.

No blocking concerns.
</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Internal SQLite write-retry policy change with no observable-surface impact; bounded worst-case latency is documented and confined to the off-event-loop maintenance pass. No parity, security, or API concern.

<details>

Cycle 1 (no prior Risk comment) — complete class-labeled review. `crates/app/` is not in the parity-critical set (scp/herder/ledger/tx/overlay); risk lens applies.

- `parity-gap` — None. This is internal retention-trim retry policy; it touches no hash, tx result/meta XDR, SCP/overlay wire bytes, history archive format, RPC/HTTP/CLI contract, or crypto output. PR verifies against `stellar-core/` v27.0.0: upstream removed its maintenance subsystem and its nearest analogue has neither chunking nor busy-retry, so henyey remains a strict superset of upstream fault-tolerance. No observable behavior changes.
- `regression-risk` — Low. Non-transient errors still fail closed immediately (fail-closed semantics for real corruption/integrity preserved); only transient `DatabaseBusy`/`DatabaseLocked` is retried. Row-accounting and drain-termination unchanged.
- `api-shape` — Visibility widening `is_transient_db_busy` private → `pub(crate)` + re-export through `crate::app` is the minimum needed for the sibling `maintainer` module and mirrors the existing `CatchupFinalizer` re-export. Not over-exposed (stays crate-internal).
- `operational` — Documented, accepted latency trade-off: a fully-stuck table trim now costs up to 3×`busy_timeout` (~90s) and a worst-case 5-table pass up to ~450s vs ~150s. This runs via `spawn_blocking_logged` off the tokio event loop, so it lengthens a maintenance *pass*, not the event loop (not a #3756-style freeze). Feeds #3754. Acceptable.

No blocking concerns. (Note: the linked issue's items 2–4 — `emit_meta` catchup-meta retry, per-site busy counter, doc reconciliation — are out of scope for this PR, which addresses the retention-trim half; they remain open backlog and are not required for this merge.)
</details>

tomerweller:

**Reviewer A (Correctness):** APPROVE — bounded transient-busy retry is correct, guard scoping preserved, three regression tests provably fail at pre-fix `cd4e5fc`.
**Reviewer B (Risk):** APPROVE — internal SQLite write-retry policy, no observable/parity surface touched; documented off-event-loop latency trade-off acceptable.
**External reviewers:** none
**CI:** red — **1 of 42 checks failed, diff-unrelated infrastructure flake.**

The only red check is the `test (testnet, core,horizon, …)` Quickstart shard. It is not attributable to this diff (a Rust SQLite maintenance-delete retry in `crates/app/`); it is a live-testnet Horizon-sync probe timing out plus a runner shutdown:

```
06:11:05  === Transient-infra failure (exit 124) on retryable shard
          (testnet/core,horizon/horizon-core-up), retrying ===
          Command: timeout -k 30s 240s go run quickstart/tests/test_horizon_core_up.go
06:23:11  ##[error]The runner has received a shutdown signal. This can happen when the
          runner service is stopped, or a manually started runner is canceled.
```

`exit 124` is a GNU `timeout` expiry on the Horizon-core-up probe against live testnet; `exit 143` (128+SIGTERM) is the runner being torn down mid-retry. Neither depends on the maintenance-retry code. All other 41 checks (Format, Clippy, Test, Doctest, Build, CodeQL, full Quickstart local/pubnet/galexie shards, etc.) are green.

**Action for `/do` Mode B:** no code change needed. Rebase on `origin/main` and push to re-trigger a fresh CI run. If the testnet shard fails again after rebase, the next `/review-pr` will mark this `blocked` for operator investigation of the testnet infra flake.

Routing back to `ready-for-doing`.

tomerweller:
## Do: Mode B — Rebased for fresh CI

Per the Bounce-Back Cycle 1 review, both reviewers (Correctness + Risk) had **APPROVED**; the only red check was the diff-unrelated `test (testnet, core,horizon, …)` Quickstart shard (Horizon-core-up probe `exit 124` timeout + runner `exit 143` teardown), which is a testnet infra flake, not attributable to this maintenance-delete-retry diff.

**Action taken:** No code change. Rebased `do/issue-3772` onto latest `origin/main` (`3ea207c1`) to re-trigger a fresh CI run. The two commits (regression test `29705429`, fix `93ddcd08`) replay cleanly with no conflicts.

**Local verification (changed files only — `maintainer.rs`, `persist.rs`, `mod.rs`):**
- `cargo fmt --check` — clean
- All 6 `maintainer::tests::*delete_in_chunks*` tests pass (incl. the 3 regression tests)
- Note: local clippy (0.1.95) flags 2 pre-existing `unnecessary_cast` errors in `lifecycle.rs` (from merged PR #3646, not this diff); CI's pinned-toolchain Clippy is green on this PR.

Ready for re-review.

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Re-review against the current (rebased, force-pushed 08:03Z) head `93ddcd0`. Bounded transient-busy retry around `delete_in_chunks` is correct, fail-closed on real errors, and covered by three regression tests that provably fail at pre-fix `main`.

<details>

Cycle 1 for the current head (prior APPROVE was against the pre-force-push SHAs). Complete, class-labeled review.

- `correctness` — Inner retry loop is right: `Ok(rows) => outcome=Ok(rows); break`; transient `is_transient_db_busy(&e)` path stores the error, drops the `WriteCtxGuard` before sleeping (lock not held while backing off), and only sleeps when the budget is not yet exhausted; any non-transient error `break`s immediately (fail-closed). Bounded by `attempts < MAINTENANCE_DELETE_MAX_ATTEMPTS` (3) — no unbounded spin. Outer `match outcome` preserves the original `Ok(0) => break` drain-termination and row accounting; `saturating_sub` on `remaining` unchanged.
- `correctness` — `WriteCtxGuard` (#3702 instrumentation contract) preserved: one guard per attempt, so a reported long hold still maps to a single bounded delete statement.
- `test-coverage` (bug-fix kind) — Three new tests assert exact call-counts and row totals: retry-then-succeed (3 calls, rows counted), persistent-busy (exactly MAX attempts, 0 rows), non-transient (exactly 1 call, no retry). Regression commit "fails on current main"; matches the required failing-test-first discipline.
- `readability` — `busy_error()` helper mirrors persist.rs; warn now carries `attempts`. Clear.

No blocking concerns.
</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Re-review against current head `93ddcd0`. Internal SQLite retention-trim retry policy; no observable-surface impact. `crates/app/` is non-parity (not scp/herder/ledger/tx/overlay), so risk lens applies.

<details>

Cycle 1 for the current head. Complete, class-labeled review.

- `parity-gap` — None. Touches no hash, tx result/meta XDR, SCP/overlay wire bytes, history-archive format, RPC/HTTP/CLI contract, or crypto output. Pure internal retention-trim retry.
- `regression-risk` — Low. Non-transient errors still fail closed immediately; only transient `DatabaseBusy`/`DatabaseLocked` is retried. Row-accounting and drain-termination unchanged.
- `api-shape` — `is_transient_db_busy` widened private -> `pub(crate)` + re-export through `crate::app`, the minimum needed for the sibling `maintainer` module; mirrors the existing `CatchupFinalizer` re-export. Stays crate-internal.
- `operational` — Documented latency trade-off: a fully-stuck table trim now costs up to 3x`busy_timeout`. Runs off the tokio event loop via `spawn_blocking_logged`, so it lengthens a maintenance pass, not the event loop (not a #3756-style freeze). Acceptable.

Note: issue items 2-4 (`emit_meta` catchup-meta retry, per-site busy counter, doc reconciliation) are out of scope for this PR (retention-trim half) and remain open backlog.

No blocking concerns.
</details>

tomerweller:

**Reviewer A:** APPROVE — bounded transient-busy retry is correct, fail-closed on real errors, three regression tests fail on pre-fix main.
**Reviewer B:** APPROVE — internal retention-trim retry policy, no observable-surface/parity impact; off-event-loop latency trade-off documented.
**CI:** red (single unrelated leg)

Both agent reviewers APPROVE and 40 of 41 checks are green. The only non-green check is `test (testnet, core,horizon, …)` inside the Quickstart workflow, which was **CANCELLED while its live-testnet probe step (`Run probes through wrapper`) was still in progress** — the built PR image loaded and the container came up fine, and every `local` and `pubnet` Quickstart leg plus the aggregate `Quickstart Harness` gate passed SUCCESS. This is an external-testnet-connectivity cancellation, not attributable to the diff (which only changes SQLite maintenance-delete retry logic in `crates/app/maintainer.rs`).

Per the `/review-pr` matrix, an unrelated CI-red leg cannot be treated as green (auto-merge requires every check SUCCESS/SKIPPED/NEUTRAL). Routing back to `ready-for-doing` so `/do` Mode B can rebase on `origin/main` and re-trigger CI. No code changes are required — the fix is already correct and fully reviewed; this is purely to obtain a clean CI run. If the testnet leg is CANCELLED again after rebase, the next `/review-pr` will mark this `blocked` for operator attention.

tomerweller:
## Do: Mode B — Re-ran the flaky testnet leg (no code change)

Bounced to `ready-for-doing` on CI red, but there is **no actionable feedback to address**:

- **Inline review comments:** none.
- **Reviewer A (Correctness):** APPROVE (re-review at current head `93ddcd0`).
- **Reviewer B (Risk):** APPROVE (re-review at current head `93ddcd0`).

The branch is already at `origin/main`'s tip (`3ea207c1`) — no rebase needed, and an empty rebase would not retrigger CI. Authoritative check-runs for head `93ddcd0`: **39 SUCCESS, 1 SKIPPED (`build / 3 test`, expected), 1 CANCELLED**. The sole non-green leg is `test (testnet, core,horizon, …)` in the Quickstart workflow — a live-testnet Horizon/core interop probe that is diff-unrelated to this maintenance-delete-retry change (both reviewers classified it as an infra flake across two prior bounce cycles).

**Action taken:** re-ran the CANCELLED Quickstart leg (workflow run `30525464202`, now `in_progress`) to turn the board green. No commit — the two commits (regression test `2970542`, fix `93ddcd0`) stand as-is.

Requesting re-review once CI settles green.

tomerweller (commented):
Addressed the bounce: no code change required (both reviewers APPROVE, zero inline comments). Re-ran the diff-unrelated CANCELLED live-testnet Quickstart leg to green the board. Ready for re-review.

tomerweller:
## Review: Blocked — Unrelated CI Red Persists After Rebase

**Reviewer A (Correctness):** APPROVE — bounded transient-busy retry around `delete_in_chunks` is correct, fail-closed on real errors, three regression tests provably fail at pre-fix `main`. (Latest verdict against current head `93ddcd0`.)
**Reviewer B (Risk):** APPROVE — internal SQLite retention-trim retry policy; no observable/parity surface touched; off-event-loop latency trade-off documented. (`crates/app/` is non-parity.)
**External reviewers:** none blocking (author left one `COMMENTED` note).
**CI:** red — 41 of 42 checks green; the sole red check is the `test (testnet, core,horizon, …)` live-testnet horizon-ingestion leg of the Quickstart workflow.

**Why blocked (not another bounce):**
The single red check is a **diff-unrelated live-network flake**. The PR diff touches only internal SQLite retention-trim retry logic in `crates/app` (`maintainer.rs`, `app/mod.rs`, `app/persist.rs`); it does not touch the horizon live-testnet ingestion path exercised by the failing leg (the Quickstart image runs stellar-core, not henyey). The failing job timed out on `test_horizon_ingesting` (orphan process terminated at teardown after a ~29-min run with `timeout_multiplier: 4`) — an external-testnet dependency, not the diff.

Both agent reviewers APPROVE, and the branch is already **fully current with `origin/main`** (0 commits behind, merge-base == main HEAD). It was rebased/force-pushed at 08:03Z after the first unrelated-red bounce, and the same leg failed **again** on a manual re-run at 12:32Z. Per the unrelated-CI-red rule, the first occurrence bounces to `/do` for a rebase; when it stays red **after** the rebase, the next `/review-pr` marks it `blocked` — `/do` cannot fix a live-testnet infrastructure flake by rebasing an already-current branch, and continuing to bounce would be a no-op loop.

**Pattern (for the operator):**
- Bounce-Back Cycle 1 (06:34Z): unrelated testnet leg red → rebase.
- Force-push / rebase onto current main (08:03Z).
- Bounce-Back Cycle 1 again (10:05Z): same leg still red on the rebased head.
- Manual re-run of the leg (12:03Z) → failed again (12:32Z), same `test_horizon_ingesting` timeout.

**Operator action:** re-run only the `test (testnet, core,horizon, …)` Quickstart leg once live testnet is healthy; when it goes green the PR is triple-green and mergeable as-is. If the leg is chronically flaky against live testnet, consider making it non-required in branch protection. To resume the automated pipeline after the cause clears, post `## Review: Reset` with a one-line reason.

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Re-review at head `93ddcd08` (0 behind `origin/main` @ `3ea207c1`). I re-ran the TDD gate myself rather than trusting the PR body: the two behavioural regression tests provably FAIL at the pre-fix test commit `29705429` and PASS at the fix commit. Retry loop, guard scoping, and row accounting are all correct.

<details>

Cycle 3 — prior `## 🔍 Reviewer: Correctness` verdicts (06:33Z, 10:05Z) both APPROVE; classes raised previously: `correctness`, `test-coverage`, `readability`. Head is byte-identical to the head those verdicts ran against, so I stayed within those classes and re-verified each with execution rather than re-reading. No **NEW CLASS DISCOVERED**.

**`test-coverage` — verified by execution, not by reading the PR body.**

Checked out the test-only commit `29705429` (which deliberately hard-codes the literal `3` instead of referencing `MAINTENANCE_DELETE_MAX_ATTEMPTS`, so it compiles without the fix — correct TDD hygiene) and ran the module:

```
running 6 tests
test ..._does_not_retry_non_transient_error ... ok
test ..._retries_transient_busy_then_succeeds ... FAILED   (left: 1, right: 3)
test ..._gives_up_after_max_attempts ...        FAILED   (left: 1, right: 3)
test ..._stops_on_error ... ok
test ..._respects_total_budget ... ok
test ..._bounds_each_call_and_exhausts ... ok
test result: FAILED. 4 passed; 2 failed
```

At `93ddcd08`: `test result: ok. 6 passed; 0 failed`. The `left: 1` in both failures is exactly the bug — single-shot `delete_chunk` breaks the whole table's trim on the first busy error. The third new test (`_does_not_retry_non_transient_error`) passing at both commits is correct and intentional: it pins *preserved* fail-closed behaviour, so it is a guard test, not a regression test, and the PR does not claim otherwise. The three pre-existing tests (`_bounds_each_call_and_exhausts`, `_respects_total_budget`, `_stops_on_error`) pass unmodified, exactly as the converged plan predicted.

**`correctness` — retry loop.**

- Bounded: `while attempts < MAINTENANCE_DELETE_MAX_ATTEMPTS` (3). No unbounded spin — this is the pattern the repo has repeatedly been burned by, and it is respected here.
- Classification gate is the narrow structured one: `crate::app::is_transient_db_busy` matches only primary `ErrorCode::DatabaseBusy | DatabaseLocked`, never a message string, so `DatabaseCorrupt`/`SystemIoFailure`/`DbError::Integrity` all still take the immediate `break` arm. Corruption is not reclassified recoverable.
- `outcome` threading is right: `Ok(rows) => break` (success short-circuits), transient stores the error and continues, non-transient stores and breaks. The outer `match outcome` preserves the original `Ok(0) => break` drain-termination and `total_rows += rows` accounting verbatim; `remaining.saturating_sub(chunk)` is unchanged and is still only reached on a non-terminating success, so a retried chunk cannot double-decrement or double-count.
- The `Ok(0)` initializer for `outcome` is only observable if `MAINTENANCE_DELETE_MAX_ATTEMPTS` were 0, which would degrade to "break the table trim" — a safe degenerate, not a panic or an infinite loop.
- Retry safety at the SQL layer: a `SQLITE_BUSY` write transaction never commits, and the chunk deletes are `WHERE ledgerSeq <= m`-shaped (idempotent), so re-issuing the statement cannot double-delete or skip rows.

**`correctness` — `WriteCtxGuard` (#3702 instrumentation contract).** Each attempt constructs its own guard, so a reported long hold still maps to one bounded statement rather than the whole retry sequence. On the transient arm the guard is explicitly `drop`ped *before* `thread::sleep(MAINTENANCE_CHUNK_PAUSE)`, so the instrumented window never spans the backoff. Success and non-transient arms drop at loop-body exit via the normal drop flag. Correct on all three paths.

**`correctness` — module privacy.** `mod persist;` is private, and `crate::maintainer` is a *sibling* of `crate::app`, so widening the fn to `pub(crate)` alone genuinely is insufficient — the `pub(crate) use persist::is_transient_db_busy;` re-export in `app/mod.rs` is the minimal correct fix and mirrors the adjacent `pub use persist::CatchupFinalizer;`. This is a *widening*, not a narrowing, and the symbol has a real non-test in-crate caller, so the visibility/dead-code build-verify gate does not apply and there is no unused-import risk under `-Dwarnings`.

**`readability`.** The `warn!` now carries `attempts`, which is the one field an operator needs to distinguish "lost the race once" from "table is genuinely stuck". Constant doc-comment states explicitly why this is not `commit_with_busy_retry` (that path escalates to `RecoverableShutdown`). `busy_error()` test helper mirrors the persist.rs shape.

**Non-blocking observation (not filed as an inline, to avoid duplicating #3754):** the PR body's "~90 s per table / ~450 s per pass" worst case is the *fully-stuck* case only. Because the budget is per *chunk* and the table loop only aborts when one chunk exhausts all 3 attempts, a table whose chunks each lose the race once or twice and then recover can extend well past 90 s. This is not a new problem class — the pre-existing 25 ms inter-chunk pause already puts a full 50 000-ledger budget at ≥ 625 s — it is off the tokio event loop via `spawn_blocking_logged`, the maintainer `await`s each pass so cycles cannot pile up, and `MissedTickBehavior::Skip` drops the backlog. It is also strictly better than the status quo it replaces (trim aborts entirely). Already covered by #3754.

No blocking concerns.
</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Re-review at head `93ddcd08`. Purely internal SQLite retention-trim retry policy — no observable/interop surface, no API or data-format break, no security or migration surface. I chased the two amplification vectors that could have made this worse than advertised (connection-pool occupancy and maintenance-cycle pileup) and both are bounded. `crates/app/` is not in the parity-critical set (`scp`/`herder`/`ledger`/`tx`/`overlay`), so the risk lens applies.

<details>

Cycle 3 — prior `## 🔍 Reviewer: Risk` verdicts (06:33Z, 10:05Z) both APPROVE; classes raised previously: `parity-gap`, `regression-risk`, `api-shape`, `operational`. Same head; staying within those classes. No **NEW CLASS DISCOVERED**.

**`parity-gap` — none. Re-verified directly against the `stellar-core/` submodule at `7696c069` (tag `v27.0.0`), not inferred.**

- `DatabaseUtils::deleteOldEntriesHelper` (`src/database/DatabaseUtils.cpp:12-31`) is a single unchunked `DELETE FROM <t> WHERE <col> <= m` with no retry and no busy handling anywhere in its call chain (`HerderPersistenceImpl.cpp:415-417`).
- `AUTOMATIC_MAINTENANCE_PERIOD` / `AUTOMATIC_MAINTENANCE_COUNT` are deprecated-and-ignored keys in `src/main/Config.cpp:1312-1321` — upstream's maintenance subsystem is gone.
- Core sets `PRAGMA busy_timeout = 10000` (`src/database/Database.cpp:173`); henyey uses 30 s. henyey was already the more fault-tolerant side; this widens that gap further rather than narrowing it.
- Nothing in the diff touches a hash, tx result/meta XDR, SCP/overlay wire byte, history-archive artifact, HTTP/RPC/CLI contract, or crypto output. Retention *trimming* is a local storage policy; what a peer, Horizon, stellar-rpc, or an archive consumer sees is unchanged.

**`regression-risk` — low, and the change is strictly in the safe direction.** The only new behaviour is "retry ≤ 2 more times when the error is structurally `DatabaseBusy`/`DatabaseLocked`." Every other error class — including `DatabaseCorrupt`, `SystemIoFailure`, and `DbError::Integrity` — still takes the immediate `break`. The failure mode being replaced is *silent data retention*, not silent data loss, so a bug in the new path degrades to today's behaviour rather than to something worse. Row accounting, drain termination, and the total budget are untouched.

**`api-shape` — no external surface.** `is_transient_db_busy` goes private → `pub(crate)` plus a `pub(crate) use` re-export in `app/mod.rs`. Nothing leaves the crate; no public API, no config key, no schema/migration, no metric renamed or removed. `henyey_db_busy_retry_total` is deliberately *not* reused here (its doc scopes it to the consensus persist path), which is the right call — the cross-site counter stays as backlog rather than silently making an existing metric's documentation wrong.

**`operational` — I verified the two amplification vectors rather than accepting the PR body's numbers.**

1. *Connection-pool occupancy.* `henyey_db::Database` is an r2d2 pool with `POOL_MAX_SIZE = 10` (`crates/db/src/database/mod.rs:14`), not a single mutex-guarded connection, and each chunk acquires/releases its own connection. A stuck chunk therefore occupies **1 of 10** connections for up to ~90 s instead of ~30 s, and releases it across the 25 ms backoff. No shared-mutex starvation of the ledger-close persist path, no pool exhaustion.
2. *Cycle pileup.* `run_maintenance` runs under `spawn_blocking_logged` (off the tokio event loop — not a #3756-style freeze), the maintainer `await`s each pass inside its `select!`, and `build_maintenance_interval` sets `MissedTickBehavior::Skip`. A long pass delays the next tick; it cannot stack passes or accumulate a backlog.

*Non-blocking caveat, deliberately not filed as a separate follow-up because #3754 already owns it:* the PR's "~90 s per table / ~450 s per pass" is the fully-stuck bound only. Since the budget is per *chunk* and the table loop aborts only when one chunk exhausts all three attempts, a table whose chunks each lose the race once or twice and recover can run longer than that. It is not a new problem class — the pre-existing 25 ms inter-chunk pause already floors a full 50 000-ledger budget at ≥ 625 s — and it is bounded by the two mechanisms above.

**Deployment risk — net strongly positive.** This is the fix for the self-trim starvation that let the production `mainnet.db` reach 178 GB and contributed to the shared volume filling. The failure mode it removes (a single lost lock race aborting an entire table's trim for the whole pass) is precisely the adverse feedback loop. Shipping it reduces operational risk more than the added maintenance-pass latency adds.

No blocking concerns.
</details>

tomerweller:
## ✅ Merged

**Commit:** `d5a340dc2881c34bfbca8ddcc5bbd20d9b2945ad`

**Signals:** Reviewer A (Correctness) APPROVE · Reviewer B (Risk) APPROVE · CI green (46 SUCCESS / 1 SKIPPED / 0 failures, 47 checks) · no external CHANGES_REQUESTED. No refute pass was needed — neither agent reviewer raised a blocking finding.

**Follow-up issues filed for unaddressed inline review comments:** none — this PR had zero inline review threads.

**Follow-up issues filed for the converged plan's explicitly-deferred scope:** #3801, #3802.

`Closes #3772` auto-closed the parent issue on merge, but the [Converged Plan](https://github.com/stellar-experimental/henyey/issues/3772#issuecomment-5105256567) scoped this PR to item 1 (the retention-trim half) only. Items 2 and 3 were live on `origin/main` at merge time and are now tracked separately so they aren't lost with the closed parent:

- **#3801** — catchup `emit_meta` still drops `LedgerCloseMeta` on transient SQLite busy with no retry and no propagation (`crates/history/src/catchup/persist.rs:182`). This is the silent-RPC-gap path named in #3772's own title.
- **#3802** — busy-retry telemetry still covers 1 of 5 lock-loss sites; 113/139 events remain invisible to Prometheus, including this PR's new retry loop. Without it there is no metric that can confirm this fix worked in production.

Item 4 (reconciling `commit_with_busy_retry`'s doc comment) needed no action — this PR does not touch that function. `remove_publish_entries_below` was left alone deliberately: it already fails closed and has zero recorded incidence.

**CI history note (for anyone reading the red runs above):** the ~11 days of red on this PR were infrastructure, not code. The shared volume hit 100 % (2026-07-31 → 2026-08-08, see #3798), freezing all CI. The first post-recovery `gh run rerun --failed` produced a 9-second spurious failure (`Artifact not found for name: image-quickstart-testing-with-pr-amd64.tar`) because `--failed` re-runs only the consuming test job, never regenerating the expired artifact its passing build job produced. A **full** rerun rebuilt the artifact and went green. A `## Review: Reset` was posted on #3772 before this review to clear the bounce counters.

**Verification performed at review time (not taken on trust from the PR body):** the two behavioural regression tests were re-run at the pre-fix commit `29705429` and confirmed FAILED (`left: 1, right: 3` on both), then confirmed PASSED at head — 6/6 green. Parity was re-checked directly against the `stellar-core/` submodule at `7696c069` (tag `v27.0.0`).

PR #3782, opened 2026-07-30
Add per-peer outbound diagnostic ring for #3773 corrupt-XDR investigation

### LINKED ISSUE #3773: Overlay: 8.8% of inbound sessions killed by peer-sent ERR_DATA "received corrupt XDR" — our outbound encoding, 11 peers at 100% failure
## Summary

Over the deployed validator's 26 d 17 h of uptime, **738 of 8,402 inbound peer sessions (8.8%)** were terminated by the remote peer sending us `ERROR{code=ERR_DATA, msg="received corrupt XDR"}`. That string is set in exactly one place in stellar-core, so its meaning is unambiguous: **the peer failed to XDR-decode a frame that we sent it.** The defect is in our *outbound* encoding, not in our decoding.

For **11 peers this is a 100% failure rate** — every session we have ever established with them ends this way, so those nodes are permanently unreachable as inbound peers.

`CLAUDE.md` puts SCP/overlay wire bytes on the observable surface that MUST match stellar-core, so this is a parity defect, not a cosmetic one.

Deployed build `0ac84d4274c351514532011d46a1219c88c8582d` (PID 1512116, started 2026-07-01T21:47:10Z, 0 restarts). **Not fixed-but-undeployed:** the receive-side log site is identical on both trees — `0ac84d42:crates/overlay/src/manager/peer_loop.rs:1454` and `origin/main:crates/overlay/src/manager/peer_loop.rs:1508`, same level.

## Why the direction is certain

`stellar-core/src/overlay/TCPPeer.cpp:783-812` (v27.0.0 submodule):

```cpp
try {
    xdr::xdr_get g(...);
    AuthenticatedMessage am;
    xdr::xdr_argpack_archive(g, am);
    g.done();
    valid = Peer::recvAuthenticatedMessage(std::move(am));
} catch (xdr::xdr_runtime_error& e) {
    CLOG_ERROR(Overlay, "{} - recvMessage got a corrupt xdr: {}", mIPAddress, e.what());
    errorMsg = "received corrupt XDR";      // <-- only assignment of this string
} catch (CryptoError const& e) {
    errorMsg = "crypto error";              // <-- distinct string, so MAC failure is excluded
}
...
self->sendErrorAndDrop(ERR_DATA, errorMsg);
```

The `CryptoError` arm has its own distinct string, so this is not an auth/MAC failure — it is a genuine XDR structural decode failure on bytes we emitted. `g.done()` is inside the `try`, so **trailing bytes** also land here.

## Measurements

Denominator cross-validates exactly against the metric, which is a good sign the log parse is sound:

```
inbound auth events in log                     8,402
stellar_overlay_inbound_authenticated_total    8,402   <-- exact match
sessions ending in peer-sent corrupt XDR         738   (8.8%)
distinct peers affected                          106
peers with >=5 auths and ~100% failure rate       11
```

Latency from `transitioned to authenticated` to the receipt — it is an **early-post-handshake** message, not a steady-state one:

```
min 0s   p10 1s   p25 1s   p50 4s   p75 10s   p90 79s
  <=2s   193
  2-5s   197
  5-15s  201     <-- 79% of all cases within 15s of authentication
 15-60s   68
  1-10m   22
   >10m   58
```

Per-peer, this is a hard reconnect loop, not intermittent corruption:

| corrupt-XDR receipts | auth events | peer |
|---:|---:|---|
| 95 | 95 | `GB3IM42VHWD6…` |
| 84 | 85 | `GA7LAOHSUH54…` |
| 80 | 83 | `GA37M2LDWCDW…` |
| 78 | 78 | `GDZL62FHKSJC…` |
| 66 | 66 | `GCGJOW22JXLV…` |
| 45 | 51 | `GCM7SWPBNM7O…` |
| 40 | 56 | `GD3BZYLC2JZL…` |
| 36 | 35 | `GCSLHBQZD3O5…` |

A representative episode (`GB3IM42V…` at `18.232.128.171:11625`, three consecutive sessions today, all identical):

```
02:03:09.929  INFO  Authenticated with peer GB3IM42V… (18.232.128.171:11625) handshake_ms=5
02:03:09.929  INFO  Accepted peer: GB3IM42V…
02:03:12.028  WARN  Peer sent_error peer=GB3IM42V… code=Data msg=received corrupt XDR   (+2.1s)

04:18:37.675  INFO  Authenticated with peer GB3IM42V… handshake_ms=1
04:18:44.724  WARN  Peer sent_error … received corrupt XDR                              (+7.0s)

14:36:44.841  INFO  Authenticated with peer GB3IM42V… handshake_ms=7
14:36:47.145  WARN  Peer sent_error … received corrupt XDR                              (+2.3s)
```

Handshake always completes (`handshake_ms` 1–7). So AUTH itself encodes fine; the bad frame is one of the first few messages after auth.

Not a load artifact — spread across all 24 hours, though hour 18 is 172 of 740:

```
00:34 01:26 02:34 03:13 04:53 05:6 06:14 07:1 09:6 10:3 11:13 12:14
13:10 14:48 15:18 16:40 17:45 18:172 19:73 20:24 21:24 22:57 23:12
```

Onset was **not** at process start: zero occurrences 07-01→07-08, a 187-line burst on 07-09 17:00–20:00, zero 07-10→07-13, then continuous from 07-14 (36) to today (25 so far). All within one unrotated log, so the gap is real rather than a logging artifact.

## Two hypotheses I tested and REFUTED — recording so nobody re-derives them

**1. `Error.msg` overflowing the XDR `string<100>` bound.** This was my leading theory: xdrpp enforces declared max lengths on decode, so an over-long error string from us would throw `xdr_overflow` → `xdr_runtime_error` → exactly this message. **It is not possible on this build.** Both outbound `Error` construction sites truncate first:

- `0ac84d42:crates/overlay/src/manager/peer_loop.rs:344` — `make_error_msg()` calls `truncate_error_msg()` (100-byte cap at a UTF-8 boundary) before `StringM::<100>::try_from`
- `0ac84d42:crates/overlay/src/peer.rs:600` — ERR_CONF path truncates at 100 explicitly

and there is a passing unit test `test_make_error_msg_truncates_long_message` asserting `err.msg.len() == 100` for a 200-byte input.

**2. Correlation with `GENERALIZED_TX_SET`.** My first pass matched each receipt to the nearest following `overlay_inbound_diag` line for the same peer and got 38/41 on `last_sent_msg_type="GENERALIZED_TX_SET"`, which looked like a smoking gun. **It is an artifact of an unbounded match window.** Bounding it:

```
window <=  10s : matched   0/740
window <=  60s : matched   0/740
window <= 300s : matched   2/740
```

The `GENERALIZED_TX_SET` pairing was drawn from drops hours later. There is **no** measured association with any particular `last_sent_msg_type`.

**3. A symmetric/shared codec bug.** Also refuted — the two directions are disjoint. We fail to decode peer frames (`failed to decode XDR: xdr value invalid`) for **4** peers; **106** peers fail to decode ours; **overlap = 0**.

## Remaining candidates, in my order of preference

1. **Trailing bytes** — `g.done()` throws if the frame body is longer than the decoded `AuthenticatedMessage`. A frame whose length prefix over-counts (e.g. padding, or a stale buffer length) fails here while round-tripping fine through our own decoder, which is exactly the asymmetry observed. Cheapest check: assert `encoded.len() == frame_len` on the send path.
2. **A message type we only send to some peers.** 8.8%, not 100%, so whatever it is depends on per-peer state (overlay version, flow-control variant, whether they asked for SCP state, `SEND_MORE` vs `SEND_MORE_EXTENDED`). The 11 always-failing peers are the cheapest handle here.
3. **Fragment/length-prefix divergence.** `0ac84d42:crates/overlay/src/codec.rs:136` and `:262` both OR in `0x80000000`; stellar-core rejects `length > MAX_MESSAGE_SIZE` at `TCPPeer.cpp:693`. Worth confirming the encoder can never emit a frame at/over 16 MiB — one over-large GENERALIZED_TX_SET would do it.

## Suggested next step

The decisive instrument is missing today: we log the *receipt* of the peer's ERROR but not the bytes we sent immediately before it. A targeted diagnostic — retain the last N outbound frames per peer (type + byte length + first 32 bytes) and dump them on receipt of `ERR_DATA` — would identify the offending message type in a single session, since the 11 always-failing peers reproduce on every connect within ~4 s.

## Impact

- 8.8% of inbound sessions never survive; 11 stellar-core nodes are permanently unreachable inbound, reducing inbound peer diversity (adjacent to the inbound-auth work in #3419).
- Each failure costs a full handshake plus a reconnect, and `stellar_overlay_inbound_authenticated_total` counts these as successes, so inbound-health metrics overstate reality by ~9%.
- It is an observable-surface divergence from stellar-core regardless of the operational cost.

## Node state at time of filing

`Validating` at L**63,689,100**, age 7 s, protocol 27, build `henyey-v27.0.0-alpha.1`. **27** authenticated peers / 1 pending, quorum `phase=EXTERNALIZE agree=18 disagree=0 missing=3 delayed=0 lag_ms=453`, transitive `intersection=true node_count=25 critical=[]`. RPC `healthy` (latest 63,689,100, oldest 63,688,604). Window 14:24:01→15:04:18: **0** ERROR, **0** FATAL, **0** panic, **0** hash mismatch, **0** `Lost sync`, **0** `database is locked`. Deploys held on #3702 (60 commits behind `origin/main@ba1d9d77`).

— `/monitor-tick` 2837

tomerweller:
Closes #3773

## Summary

Adds an instrumentation-only, per-peer bounded ring buffer (`RECENT_SENDS_CAPACITY = 64`) capturing every outbound frame's message type, wire size, timestamp, and a 32-byte prefix of the *actual encoded XDR body*. It is logged as a `recent_sends` field on the existing non-Load `Peer sent_error` warning, so a peer-reported `ERR_DATA "received corrupt XDR"` can finally be tied to the specific frame(s) henyey emitted just before the drop — the decisive instrument the issue identified as missing.

All four outbound send paths are covered:
- `send()`, `send_auth()`, `send_raw()` — via an extended `Connection::send` return type (`SendOutcome { wire_size, prefix }`), reading the prefix from the buffer already materialized for the socket write.
- `send_batch()` — from its own local per-message `encoded` buffer (no `Connection::send` involvement), via the shared `connection::encoded_body_prefix` helper.

This does **not** fix the underlying encoding defect (not yet root-caused); it is the diagnostics-first step the plan and issue prescribe.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3773#issuecomment-5106573587)

## Test plan

- [x] cargo fmt --check
- [x] cargo test -p henyey-overlay — 486 lib + integration tests pass, including 5 new tests

New coverage:
- `peer.rs`: `test_recent_sends_tracks_last_n_in_order`, `test_recent_sends_bounded_at_capacity`, `test_recent_sends_summary_format`, `test_recent_sends_captures_prefix_across_all_send_paths` (all 4 paths; exact byte-for-byte prefix match asserted on the deterministic `send_raw` path).
- `peer_loop.rs`: `test_error_msg_warn_includes_recent_sends` — drives `handle_received_message` with a non-Load `ERR_DATA` and asserts the emitted warn carries a populated `recent_sends` field (via the newly-enabled `test-support` `tracing_capture` harness).

## Parity considerations

n/a — no new wire bytes, no protocol behavior change. The prefix/size are reads of already-materialized, already-being-sent bytes into local diagnostic state; the wire is never altered, delayed, or re-derived. Squarely within docs/PARITY.md's divergeable Logging/Metrics categories.

## Deviations from plan

None. `RECENT_SENDS_CAPACITY` is documented as tied to `FlowControlConfig::flow_control_send_more_batch_size` (a runtime field, default 40) via doc-comment reference rather than a compile-time constant, since that value is not a `const`.

### DIFF
--- a/crates/overlay/src/peer.rs
+++ b/crates/overlay/src/peer.rs
@@ -33,6 +33,7 @@ use crate::{
 use dashmap::DashMap;
 use parking_lot::RwLock;
 use std::collections::HashSet;
+use std::collections::VecDeque;
 use std::net::SocketAddr;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
@@ -194,6 +195,33 @@ pub struct PeerInfo {
 /// access, wrap it in a `Mutex` or use the [`OverlayManager`] which handles this.
 ///
 /// [`OverlayManager`]: crate::OverlayManager
+/// Capacity of the per-peer outbound diagnostic ring (#3773). Sized at ~1.6x
+/// the largest atomic flow-control batch
+/// ([`FlowControlConfig::flow_control_send_more_batch_size`], default 40) so a
+/// full flood batch plus its handshake/control preamble fits inside the
+/// retained window. Purely a diagnostic bound — not an observable-surface
+/// value.
+///
+/// [`FlowControlConfig::flow_control_send_more_batch_size`]: crate::flow_control::FlowControlConfig::flow_control_send_more_batch_size
+const RECENT_SENDS_CAPACITY: usize = 64;
+
+/// One entry in the per-peer outbound diagnostic ring (#3773): a record of a
+/// single frame henyey sent, retained so that when a peer reports
+/// `ERR_DATA "received corrupt XDR"` the frames we emitted just before can be
+/// dumped alongside the warning. Diagnostic-only — never gates any logic and
+/// never touches the wire.
+#[derive(Debug, Clone)]
+struct RecentSend {
+    /// When the frame was sent (for age-at-drop in the summary).
+    at: Instant,
+    /// Wire name of the `StellarMessage` (e.g. `GENERALIZED_TX_SET`).
+    msg_type: &'static str,
+    /// On-the-wire body size in bytes (excludes the 4-byte length prefix).
+    wire_size: u64,
+    /// First bytes of the encoded XDR body, for byte-pattern triage.
+    prefix: [u8; crate::connection::SEND_PREFIX_LEN],
+}
+
 pub struct Peer {
     /// Peer info.
     info: PeerInfo,
@@ -219,6 +247,14 @@ pub struct Peer {
     /// "last message henyey sent before the connection reset" pattern. `"none"`
     /// until the first send. Purely additive state — does NOT gate any logic.
     last_sent_msg_type: &'static str,
+    /// Diagnostic (#3773, observability-only): a bounded ring of the most-recent
+    /// outbound frames on this connection — type, wire size, timestamp, and a
+    /// byte prefix of the actual encoded body. Dumped by the peer loop when a
+    /// peer reports `ERR_DATA "received corrupt XDR"`, to identify which frame
+    /// henyey sent that the peer could not decode. Capped at
+    /// [`RECENT_SENDS_CAPACITY`]. Purely additive state — does NOT gate any
+    /// logic and never alters the wire.
+    recent_sends: VecDeque<RecentSend>,
 }
 
 impl Peer {
@@ -260,6 +296,7 @@ impl Peer {
             metrics,
             holds_pending_peer_id: false,
             last_sent_msg_type: "none",
+            recent_sends: VecDeque::with_capacity(RECENT_SENDS_CAPACITY),
         };
 
         peer.handshake(
@@ -313,6 +350,7 @@ impl Peer {
             metrics,
             holds_pending_peer_id: false,
             last_sent_msg_type: "none",
+            recent_sends: VecDeque::with_capacity(RECENT_SENDS_CAPACITY),
         };
 
         // Perform handshake (with ban + pending-dedup checks after HELLO for inbound)
@@ -774,19 +812,21 @@ impl Peer {
         // #3419 diagnostic: record the last-sent wire type (observability-only).
         self.last_sent_msg_type = helpers::message_type_name(&message);
         let body_size = msg_body_size(&message);
+        let msg_type = self.last_sent_msg_type;
         let auth_msg = self.auth.wrap_unauthenticated(message);
-        // `Connection::send` returns the on-the-wire frame size, so we don't
-        // need to re-encode the message here just to measure it.
-        let wire_size = self.connection.send(auth_msg).await?;
+        // `Connection::send` returns the on-the-wire frame size plus a byte
+        // prefix (#3773), so we don't re-encode here just to measure it.
+        let outcome = self.connection.send(auth_msg).await?;
         // Success-only instrumentation: connection errors go to errors_write at
         // the caller (peer_loop), not bytes_written/async_write.
         self.metrics.record_send(kind);
-        self.metrics.bytes_written.add(wire_size);
+        self.metrics.bytes_written.add(outcome.wire_size);
         self.metrics.async_write.inc();
         self.stats.messages_sent.fetch_add(1, Ordering::Relaxed);
         self.stats
             .bytes_sent
             .fetch_add(body_size, Ordering::Relaxed);
+        self.record_recent_send(msg_type, outcome.wire_size, outcome.prefix);
         Ok(())
     }
 
@@ -795,16 +835,18 @@ impl Peer {
         let kind = OverlayMessageKind::from_stellar_message(&message);
         // #3419 diagnostic: record the last-sent wire type (observability-only).
         self.last_sent_msg_type = helpers::message_type_name(&message);
+        let msg_type = self.last_sent_msg_type;
         let body_size = msg_body_size(&message);
         let auth_msg = self.auth.wrap_auth_message(message)?;
-        let wire_size = self.connection.send(auth_msg).await?;
+        let outcome = self.connection.send(auth_msg).await?;
         self.metrics.record_send(kind);
-        self.metrics.bytes_written.add(wire_size);
+        self.metrics.bytes_written.add(outcome.wire_size);
         self.metrics.async_write.inc();
         self.stats.messages_sent.fetch_add(1, Ordering::Relaxed);
         self.stats
             .bytes_sent
             .fetch_add(body_size, Ordering::Relaxed);
+        self.record_recent_send(msg_type, outcome.wire_size, outcome.prefix);
         Ok(())
     }
 
@@ -824,14 +866,15 @@ impl Peer {
 
         let body_size = msg_body_size(&message);
         let auth_msg = self.auth.wrap_message(message)?;
-        let wire_size = self.connection.send(auth_msg).await?;
+        let outcome = self.connection.send(auth_msg).await?;
         self.metrics.record_send(kind);
-        self.metrics.bytes_written.add(wire_size);
+        self.metrics.bytes_written.add(outcome.wire_size);
         self.metrics.async_write.inc();
         self.stats.messages_sent.fetch_add(1, Ordering::Relaxed);
         self.stats
             .bytes_sent
             .fetch_add(body_size, Ordering::Relaxed);
+        self.record_recent_send(msg_type, outcome.wire_size, outcome.prefix);
 
         Ok(())
     }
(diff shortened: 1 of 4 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Instrumentation-only diagnostic ring is correct across all four outbound send paths, compiles clean under `-D warnings`, and has thorough new test coverage (5 tests). One non-blocking test-strength nit left inline.

<details>

Cycle 1 — complete class-labeled review (no prior Correctness verdict on this PR).

**Correctness — no blocking issues.**
- `encoded_body_prefix` handles short/empty frames safely (`encoded.get(4..).unwrap_or(&[])`, then `min(body.len(), SEND_PREFIX_LEN)`) — no panic on a <4-byte or <32-byte body; zero-pads correctly.
- `SendOutcome` is computed from the same `encoded` buffer already materialized for the socket write in `Connection::send` — no second encode pass, and the prefix is read strictly before the write. `wire_size` (`encoded.len() - 4`) is unchanged from the prior `u64` return.
- All four send paths covered: `send`, `send_auth`, `send_raw` (via `SendOutcome`), and `send_batch` (via the shared `encoded_body_prefix` helper reading its own local batch buffer). `msg_type` is in scope in every path (`send_raw` binds it at the top of the fn).
- `record_recent_send` is FIFO-bounded (`pop_front` at `== RECENT_SENDS_CAPACITY`, then `push_back`) — cannot exceed capacity. Called only on the success path of each send, matching the existing metrics/stats instrumentation placement.
- `recent_sends_summary` returns `"none"` on empty, uses `saturating_duration_since` (no underflow panic), and `hex::encode` is a normal (non-dev) workspace dep of the crate — the non-test caller in `peer_loop` links fine (confirmed by green CI Build/Clippy).

**Test coverage (`test-coverage`) — adequate for a feature-style diagnostic addition.**
- New `pub(crate)` surface (`recent_sends_summary`, `encoded_body_prefix`) and the ring behavior are all exercised: ordering, capacity bound, summary format, all-four-paths prefix capture (with a byte-for-byte match asserted on the deterministic `send_raw` path), and the end-to-end warn-field integration in `peer_loop` via the newly-enabled `tracing_capture` harness.
- The linked issue is a diagnostics-first investigation (the PR explicitly does not fix the underlying encoding defect, per the converged plan), so there is no behavior-regression test to demand here — the "feature: every new surface has a test" gate is the applicable one, and it's met.

**Non-blocking (left as inline, will be auto-filed as follow-up at merge):** `test_recent_sends_bounded_at_capacity` reads the front entry's prefix into `first_surviving` and then discards it (`let _ = first_surviving;`); its comment claims to "prove FIFO eviction" of the oldest entry but the assertion only checks `len == RECENT_SENDS_CAPACITY`. Ordering/eviction is separately proven by `test_recent_sends_tracks_last_n_in_order`, so this is a cosmetic test-strength nit, not a gap.
</details>

crates/overlay/src/peer.rs
```diff
@@ -1733,4 +1946,216 @@ mod tests {
...
+        // The oldest surviving entry must be the (total - CAPACITY)-th send,
+        // proving FIFO eviction (front dropped, back appended).
+        let first_surviving = peer_a
+            .recent_sends
+            .front()
+            .expect("ring is non-empty")
+            .prefix;
+        // The prefix encodes the send ordinal indirectly; simply assert the
+        // buffer is exactly full and never exceeded.
+        let _ = first_surviving;
```
tomerweller, line 2023: Non-blocking (test-strength): this reads `first_surviving` then discards it with `let _ = first_surviving;`, and the comment above claims to prove FIFO eviction of the oldest entry — but the only assertion here is `len == RECENT_SENDS_CAPACITY`. Consider asserting the front entry's identity (e.g. via a distinguishing prefix byte from the `(total - CAPACITY)`-th send) to actually prove front-eviction, or drop the misleading comment. Ordering is already covered by `test_recent_sends_tracks_last_n_in_order`, so this is cosmetic.

tomerweller, line 2023: Non-blocking (test-strength): reads first_surviving then discards it (let _ = first_surviving;), and the comment claims to prove FIFO eviction of the oldest entry, but the only assertion is len == RECENT_SENDS_CAPACITY. Consider asserting the front entry's identity to actually prove front-eviction, or drop the misleading comment. Ordering is already covered by test_recent_sends_tracks_last_n_in_order.

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** Zero observable/interop-surface impact. No wire bytes, XDR, hashes, or protocol sequencing change — the diagnostic reads already-materialized outbound bytes into local state and surfaces them only on an existing WARN log field, squarely within docs/PARITY.md's divergeable Logging/Metrics tier.

<details>

Cycle 1 — complete class-labeled review (no prior Parity verdict on this PR). Parity lens scoped strictly to the observable/interop surface per docs/PARITY.md.

**No observable-surface change (`parity-gap`: none found).**
- **Wire bytes:** the prefix and `wire_size` are pure reads of the `encoded` buffer that `Connection::send` (and `send_batch`'s local encode) already produces for the socket write. The write path is byte-identical — no re-encode, no mutation, no reordering, no added/dropped frame. `send_batch` still concatenates the same buffer and issues the same single coalesced write; the ring is populated only after the write succeeds.
- **XDR / hashes:** no transaction result/meta, ledger, or bucket bytes touched. `msg_body_size` / `wire_size` (`encoded.len() - 4`) semantics are unchanged from the prior `u64` return.
- **SCP / overlay wire protocol:** unchanged. The `AuthenticatedMessage` framing, sequence numbers, and MAC derivation are untouched; the handshake helper `new_test_authenticated_pair` is `#[cfg(test)]`-only.
- **HTTP/RPC/CLI, history archive, crypto outputs:** not in the diff.

**Divergeable-tier changes (explicitly allowed, not flagged):**
- `Connection::send` return type `u64` → `SendOutcome` is an internal API shape — not part of the observable surface.
- The new `recent_sends` WARN field on the existing `Peer sent_error` warning is logging output; log format is a divergeable category, and this only fires on the non-Load `ErrorMsg` path (unchanged control flow — still returns `Break`).
- Per-send cost (32-byte copy + `Instant::now()` + O(1) VecDeque push) is internal/perf, allowed to diverge; it is bounded and mirrors the pre-existing `last_sent_msg_type` per-send instrumentation.

No parity concerns. The change is diagnostics-only and cannot alter interop behavior.
</details>

tomerweller:
## ✅ Merged

**Commit:** 3ea207c1630c3e89a8d936b34514ae88624d6d27

Triple-green: Reviewer A (Correctness) APPROVE, Reviewer B (Parity) APPROVE, CI green (42/42). No external CHANGES_REQUESTED.

**Follow-up issues filed for unaddressed inline review comments:** #3785

PR #3783, opened 2026-07-30
Enforce MAX_MESSAGE_SIZE on the overlay send path

### LINKED ISSUE #3774: Overlay: MessageCodec::encode_message lacks MAX_MESSAGE_SIZE check present in unused Encoder::encode
## Summary

`MessageCodec::encode_message` (`crates/overlay/src/codec.rs`) — the actual send-path encoder used by `Connection::send` and `Peer::send_batch` — lacks a `MAX_MESSAGE_SIZE` bound check on the frame it produces. A parallel, apparently-unused `Encoder::encode` trait impl in the same file has this check, but nothing on the real send path currently enforces it.

stellar-core rejects any incoming frame with `length > MAX_MESSAGE_SIZE` (`TCPPeer.cpp:693`) — if henyey ever encodes and sends a frame at or over that bound (e.g. an unusually large `GENERALIZED_TX_SET`), the length prefix itself would still round-trip correctly through our own encoder, but the frame would violate the receiving stellar-core peer's own size assumptions in ways not yet characterized.

## Context

Found during `/plan` critic review (Critic C, both rounds) for issue #3773 ("Overlay: 8.8% of inbound sessions killed by peer-sent ERR_DATA 'received corrupt XDR'"), as candidate 3 of that issue's "remaining candidates" list ("Fragment/length-prefix divergence... stellar-core rejects `length > MAX_MESSAGE_SIZE`... one over-large GENERALIZED_TX_SET would do it"). Filing separately per that review's recommendation, since this is a distinct, more speculative fix than the pure instrumentation #3773's converged plan implements — it deserves its own triage/plan cycle, not a drive-by addition to an instrumentation-only PR.

## Suggested next step

Confirm whether `Encoder::encode`'s `MAX_MESSAGE_SIZE` check is genuinely dead code (unused by any real send path) or reachable from somewhere not yet identified, then either wire the same check into `MessageCodec::encode_message` (the real send path) or justify why it's provably unreachable there.

Not labeled urgent — no confirmed incident tying an outbound henyey frame to this specific bound; this is a defense-in-depth gap identified via code review, not an observed failure.

tomerweller:
Closes #3774

## Summary

`MessageCodec::encode_message` (`crates/overlay/src/codec.rs`) — the real outbound encoder used by `Connection::send` and `Peer`'s batch-send path — encoded and returned an `AuthenticatedMessage` of any size, with no `MAX_MESSAGE_SIZE` bound check. The receive path (`Decoder::decode`) already rejects inbound frames over `MAX_MESSAGE_SIZE`, matching stellar-core (`TCPPeer.cpp:690-701`), but the only outbound check lived in the parallel, production-dead `Encoder::encode` trait impl.

This extracts the bound check into a single private helper, `check_encode_size`, called from both `encode_message` and `Encoder::encode`, so there is exactly one enforcement point that cannot silently diverge again.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3774#issuecomment-5109320611)

## Test plan

- [x] cargo fmt --check
- [x] cargo test -p henyey-overlay (482 lib tests + integration, all pass)

- **Test:** `crates/overlay/src/codec.rs::test_encode_message_rejects_oversized_message` (plus helper unit test `test_check_encode_size_rejects_oversized`)
- **Pre-fix:** committed as `35ba2f3` — verified FAILED with: `encode_message must reject a message larger than MAX_MESSAGE_SIZE` (it returned Ok for a real >16 MiB `StellarMessage::ScpMessage`).
- **Post-fix:** verified PASSES after `65b2fca`.

## Deviations from plan

- Second regression test uses the actual v27.0.0 XDR field names (`ScpStatement { node_id, slot_index, pledges: ScpStatementPledges::Prepare(ScpStatementPrepare { quorum_set_hash, ballot, prepared, prepared_prime, n_c, n_h }) }`) rather than the plan's placeholder field set (`ScpStatement::Prepare(...)`, `num_committed`/`num_committed_prime`) — the plan flagged these as "to be confirmed by /do against the actual XDR definitions". The `ScpBallot.value` oversized-blob technique is exactly as planned.
- Dead-code removal of `Encoder::encode` / the `Connection::split()` chain is deferred to a follow-up issue per the plan; filing it separately.


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Correct, well-scoped fix for #3774. The `MAX_MESSAGE_SIZE` check is now wired into the real send path (`encode_message`, called by `Connection::send` at connection.rs:245 and `Peer::send_batch` at peer.rs:914) via a shared `check_encode_size` helper, and the regression test exercises the real encoder with a genuine oversized message.

<details>

Cycle 1 (no prior Correctness verdict on this PR) — complete class-labeled change-list below. No blocking concerns found.

- **correctness:** `check_encode_size` is the single enforcement point, called from both `encode_message` (real send path) and the `Encoder::encode` trait impl (previously the only path with the check, but it has no real-send-path caller — verified: no non-test `.encode(` call reaches this codec). The refactor removes the divergence the issue identified without changing behavior on either path.
- **test-coverage (bug-fix gate satisfied):** `test_encode_message_rejects_oversized_message` builds a real, well-formed `AuthenticatedMessage::V0` carrying an oversized `ScpBallot.value` opaque blob and asserts `encode_message` returns `Err`. Pre-fix, `encode_message` had no size check and would return `Ok`, so this test provably fails at the pre-fix commit — a genuine regression test, not a tautology. `test_check_encode_size_rejects_oversized` adds boundary coverage (`== MAX_MESSAGE_SIZE` ok, `+1` err, `0` ok). The old weak test's spec-constant assertion (16 MiB) is folded in, so no coverage is lost.
- **error-handling:** returns `OverlayError::Message("message too large: N bytes")`, identical to the prior inline check; consistent with the receive-side rejection wording.
- **readability:** helper is documented, references #3774 and the stellar-core anchor. Clean.

No `NEW CLASS DISCOVERED`. Nothing held back for later cycles.
</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** The send-side bound correctly mirrors stellar-core's receive-side rejection of frames with `length > MAX_MESSAGE_SIZE` (16 MiB, `TCPPeer.cpp:690-701`). This keeps henyey from emitting a frame a stellar-core peer would reject, improving observable/wire-surface parity. No wire-format change for normally-sized messages.

<details>

Cycle 1 (no prior Parity verdict on this PR). Parity lens scoped to the observable/interop surface (overlay wire bytes) per docs/PARITY.md.

- **parity (wire surface):** stellar-core rejects inbound frames where `length > MAX_MESSAGE_SIZE`. henyey previously enforced this only on decode (codec.rs decode path, authenticated → 16 MiB) and in an unused `Encoder::encode` impl. Adding the same `> MAX_MESSAGE_SIZE` strict-greater bound on the real encode path means henyey will not serialize/emit a frame that a core peer would then reject as corrupt — the exact defense-in-depth gap #3774 (spun out of the #3773 corrupt-XDR investigation) called out. Boundary matches core: exactly `MAX_MESSAGE_SIZE` bytes is allowed on both send and receive; `+1` is rejected on both.
- **no observable regression:** the check only rejects frames that already could not be legally received by a core peer; all conforming messages encode byte-for-byte as before. No change to length-prefix framing, XDR bytes, or the final-fragment bit.
- **non-blocking observation (not a bounce):** the receive path distinguishes an unauthenticated 4096-byte cap (`MAX_UNAUTHENTICATED_MESSAGE_SIZE`) from the 16 MiB authenticated cap, whereas the encode path checks only against the 16 MiB `MAX_MESSAGE_SIZE`. This is not a parity gap on the observable surface — pre-auth henyey only sends small handshake messages (HELLO/AUTH), and core's unauth cap is a receive-side DoS guard, not a send-side contract. Worth a note but does not warrant changes here.

No `NEW CLASS DISCOVERED`. Internal helper naming/metrics are out of parity scope and not flagged.
</details>

tomerweller:

**Reviewer A:** APPROVE — correct, well-scoped fix; regression test provably fails at the pre-fix commit.
**Reviewer B:** APPROVE (Parity) — send-side bound mirrors stellar-core's receive-side `length > MAX_MESSAGE_SIZE` rejection; no wire-format change for conforming messages.
**External reviewers:** none
**CI:** red

Both agent reviewers approve and the diff is clean, but one required check is red:

- `test (testnet, core,horizon, ...)` — Quickstart workflow.

The failure is **unrelated to this diff** (which touches only `crates/overlay/src/codec.rs`). It is a transient testnet-infra failure on the `horizon-core-up` shard:

```
=== Transient-infra failure (exit 124) on retryable shard (testnet/core,horizon/horizon-core-up), retrying ===
...
quickstart-hang-watchdog: wrote process-tree dump ...
##[error]The runner has received a shutdown signal.
```

The shard waited on Horizon's stellar-core to start reporting against live testnet, timed out (exit 124), retried, then the runner itself was terminated (exit 143). No reference to the codec change.

Routing back to `ready-for-doing` for `/do` Mode B to rebase on `origin/main` and re-trigger CI. If CI is still red after a clean rebase, the next `/review-pr` will mark this `blocked`.

tomerweller:
## Do: Mode B — rebased on origin/main (CI re-trigger)

Both agent reviewers APPROVED in Cycle 1; the only blocker was a **transient testnet-infra failure** on the Quickstart `test (testnet, core,horizon, …)` shard (exit 124 → runner shutdown 143), unrelated to this diff.

- Rebased `do/issue-3774` cleanly onto current `origin/main` (`3ea207c1`). No conflicts; the 2 PR commits replayed unchanged. New head: `8733a55f0a2534df8edaaa6dfcba0d8826a8dd33`.
- Diff is byte-identical to the approved version — only `crates/overlay/src/codec.rs` (+84/-17).
- Local verification on the rebased branch: `cargo fmt --check` clean, `cargo clippy -p henyey-overlay -- -D warnings` clean, both regression tests pass (`test_encode_message_rejects_oversized_message`, `test_check_encode_size_rejects_oversized`).
- No actionable inline review comments were left; nothing to change.

Force-pushed to re-trigger CI. Ready for re-review.

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Re-reviewed on the current rebased head (`8733a55f`). The diff is unchanged from cycle 1 and remains a correct, well-scoped fix: `check_encode_size` is the single enforcement point, wired into the real send path `MessageCodec::encode_message`, with a genuine regression test.

<details>

Cycle 2 (re-review after clean rebase; code content identical to the cycle-1 approve). Sticking to the classes raised in cycle 1 — all remain satisfied. No `NEW CLASS DISCOVERED`.

- **correctness:** `check_encode_size(xdr_len)` returns `Err(OverlayError::Message("message too large: N bytes"))` for `len > MAX_MESSAGE_SIZE` and is now called from both `encode_message` (real send path, used by `Connection::send` / `Peer::send_batch`) and the `Encoder::encode` trait impl. The inline check in `Encoder::encode` is replaced by the shared helper — no behavior change on that path, and the previously-unchecked real send path now enforces the bound. Divergence from #3774 is closed.
- **test-coverage (bug-fix gate satisfied):** `test_encode_message_rejects_oversized_message` builds a real, well-formed `AuthenticatedMessage::V0` carrying an oversized `ScpBallot.value` opaque blob (`Value(BytesM)` via `StellarMessage::ScpMessage`) and asserts `encode_message` returns `Err`; pre-fix `encode_message` had no size check and returned `Ok`, so the test provably fails at the pre-fix commit. `test_check_encode_size_rejects_oversized` adds boundary coverage (`== MAX_MESSAGE_SIZE` ok, `+1` err, `0` ok). The old weak indirect test is replaced without losing the 16 MiB spec-constant assertion.
- **error-handling / readability:** helper documented, references #3774 and the `TCPPeer.cpp:690-701` anchor; error wording matches the receive-side rejection. Clean.
</details>

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** Re-reviewed on the current rebased head (`8733a55f`). Unchanged from cycle 1. The send-side bound correctly mirrors stellar-core's receive-side rejection of frames with `length > MAX_MESSAGE_SIZE` (16 MiB, `TCPPeer.cpp:690-701`), with no wire-format change for conforming messages.

<details>

Cycle 2 (re-review after clean rebase). Parity lens scoped to the observable/interop surface (overlay wire bytes) per docs/PARITY.md. Prior classes remain satisfied; no `NEW CLASS DISCOVERED`.

- **parity (wire surface):** adding the strict `> MAX_MESSAGE_SIZE` bound on the real encode path means henyey will not serialize/emit a frame a stellar-core peer would reject as corrupt — the exact defense-in-depth gap #3774 called out (spun out of the #3773 corrupt-XDR investigation). Boundary matches core: exactly `MAX_MESSAGE_SIZE` bytes allowed on both send and receive; `+1` rejected on both.
- **no observable regression:** the check rejects only frames that could not be legally received by a core peer; all conforming messages encode byte-for-byte as before — no change to length-prefix framing, XDR bytes, or the final-fragment bit.
- **non-blocking (not a bounce):** the encode path checks only the 16 MiB authenticated cap, not the 4096-byte `MAX_UNAUTHENTICATED_MESSAGE_SIZE`. Not a parity gap on the observable surface — pre-auth henyey only sends small handshake messages; core's unauth cap is a receive-side DoS guard, not a send-side contract.
</details>

tomerweller:
## Review: CI Stuck

**Bounce count (head-scoped):** 1 (cycle 1 bounced for the same reason)
**Status:** blocked

**Reviewer A (Correctness):** APPROVE — clean, well-scoped fix; regression test provably fails at the pre-fix commit.
**Reviewer B (Parity):** APPROVE — send-side bound mirrors stellar-core's receive-side `length > MAX_MESSAGE_SIZE` rejection; no wire-format change for conforming messages.
**External reviewers:** none
**CI:** red — one failing required check.

**Pattern — persistent *unrelated* CI red after a clean rebase:**

Both agent reviewers approve and the diff is clean and unchanged (touches only `crates/overlay/src/codec.rs`). The sole blocker is one red required check:

- `test (testnet, core,horizon, ...)` — Quickstart workflow, `horizon-core-up` shard on live **testnet**.

This is a **transient testnet-infra failure, not diff-attributable**. Log evidence (run 30527482114):

```
=== Attempt 1: horizon-core-up (testnet/core,horizon) ===
=== Transient-infra failure (exit 124) on retryable shard (...), retrying ===
=== Attempt 2: horizon-core-up (testnet/core,horizon) ===
=== SOFT-SKIP: testnet sync probe timed out (environmental, not a henyey failure) ===
=== SOFT-SKIP: ... exit 124 treated as neutral (#3272); diagnostics preserved ===
##[error]The action 'Run probes through wrapper' has timed out after 25 minutes.
```

The shard waits on Horizon's stellar-core to sync against live testnet, times out (exit 124), soft-skips the probe per #3272, but the probe *wrapper* still exceeds its 25-min wall-clock and the step is marked FAILURE. No reference to the codec change.

Cycle 1 already bounced this PR for the identical reason ("CI failed (unrelated, will rebase)") and stated: *"If CI is still red after a clean rebase, the next `/review-pr` will mark this `blocked`."* The PR head was subsequently rebased (`8733a55f`, committed after the cycle-1 bounce) and CI re-ran — the **same** testnet-infra shard failed again. A further bounce would only re-trigger the same flaky live-testnet shard, so escalating instead of looping.

**Operator action:** this is a CI-infrastructure blocker on a live-testnet sync probe (#3272 soft-skip already fires, but the 25-min wrapper timeout still fails the step), not a defect in the PR. Options:
- Re-run the failed Quickstart job once testnet sync is healthy, then post `## Review: Reset` so this PR can re-attempt.
- Or adjust the testnet shard's wrapper timeout / soft-skip so an environmental exit-124 doesn't fail the required check.

Both reviewer verdicts and the clean diff are ready to merge the moment CI is green.

tomerweller:
Rebased onto `origin/main` (clean, 27 commits behind) and merging. Verified locally on the rebased head with real exit codes: `cargo fmt --check` 0, `cargo clippy --all --all-targets` 0 with zero lints (clears the newer #3830 gate), `cargo test -p henyey-overlay --lib` 488/488 — including `test_encode_message_rejects_oversized_message` and `test_max_message_size_is_16_mib`. No conflict with #3829 (`5079422f`), which edited `codec.rs` module docs in a disjoint region; all codec tests pass together, including #3829's `test_decode_accepts_continuation_bit_clear`.

⚠️ **Correcting the parity claim in this PR's description and reviews.** They state this matches stellar-core. It does not — **upstream has no send-side limit at all**:

- `Peer::sendAuthenticatedMessage` (`stellar-core/src/overlay/Peer.cpp:921-960`) calls `xdr::xdr_to_msg` with no bound; `TCPPeer::sendMessage` (`TCPPeer.cpp:227-243`) only enqueues.
- The only production uses of `MAX_MESSAGE_SIZE` upstream are the definition (`Peer.h:26`) and the **receive** check (`TCPPeer.cpp:693`).
- Core's own test `TCPPeerTests.cpp:216-221` deliberately **sends** a `MAX_MESSAGE_SIZE * 2` message and asserts the **receiver** drops it — so core knowingly transmits oversized frames.

So this change is **stricter than upstream**, not a match. Merging anyway because the divergence is benign and defensible on this specific surface:

- `OVERLAY_SPEC.md:628-631` frames the rule as *"The receiver MUST validate…"* — it imposes no sender duty, so a sender-side check adds no wire-visible obligation.
- **No wire bytes change for any conforming message**, so the MUST-match framing surface is untouched.
- Both implementations end in a dropped connection for an oversized frame; only the side that notices differs.
- The largest realistic outbound message is bounded well under the limit anyway (`MAX_TX_SET_ALLOWANCE` caps at 10 MiB).

Please don't propagate "matches stellar-core" for this behaviour in future docs or spec-adherence rows.

**Blast radius when the check fires:** `Err` → `peer_loop.rs:1278` `break` (peer dropped); `send_flow_controlled_batch` (`:1701`) discards the whole coalesced batch. Four callers swallow the error (`manager/connection.rs:230,373`, `peer_loop.rs:836,1546`), and at those sites `auth.wrap_message` has **already consumed the outbound sequence number**, so a sequence desync is theoretically reachable. Pre-existing, non-blocking, and all four messages are kilobyte-scale — but worth a follow-up rather than leaving unrecorded.

Sole non-green check was `test (testnet, core,horizon, …)` — job 90830594213, log line 498 `The action 'Run probes through wrapper' has timed out after 25 minutes` on `test_horizon_ingesting.go` after a soft-skipped exit-124 on `horizon-core-up`. That is the chronic #3768 hang, not diff-attributable. Its CI was also 14 days stale and predates #3830, which is why I re-ran the tightened clippy gate locally.

PR #3800, opened 2026-08-08
Accept XDR continuation-bit-clear frames to match stellar-core

### LINKED ISSUE #3776: Overlay: decoder rejects XDR continuation-bit-clear frames that stellar-core accepts — both Creit preferred peers unreachable since 07-29T04:59Z (6,521 failed dials)
## Symptom

Since **2026-07-29T04:59:05.940111Z** the node fails *every* outbound dial to both of Creit Tech's preferred peers with:

```
WARN henyey_overlay::manager::tick: Failed to connect to preferred peer 37.59.22.64:11625:
  message error: fragmented XDR records are not supported
```

**6,521 occurrences in 17.4 h**, and **zero** in the preceding 28 days of this process's log (uptime 28 d, 0 rotations — the log covers the whole life of the process).

| peer | hostname | events | rate |
|---|---|---:|---|
| `37.59.22.64:11625` | `alpha.validator.stellar.creit.tech` | 4,417 | ~260/h from 04:59Z |
| `57.128.141.161:11625` | `beta.validator.stellar.creit.tech` | 2,111 | ~250/h from 14:00Z |

Hourly onset — first peer at 05:00Z, second joins at 14:00Z and the rate doubles:

```
04h    4      10h  260     16h  509
05h  240      11h  259     17h  492
06h  260      12h  254     18h  501
07h  259      13h  255     19h  508
08h  259      14h  510     20h  504
09h  258      15h  498     21h  500
```

**Both peers previously authenticated fine** — 7 and 9 successful `Authenticated with peer …` lines respectively, first on 2026-07-27. This is a regression in reachability, not a peer that never worked.

## Impact

`preferred_peers` is configured 2-per-org precisely so one host failing does not cost an org its direct routing (`configs/validator-mainnet-rpc.toml:163-165`: *"2 per org = 14 total; tolerates one host per org going down without losing direct routing for that org's votes"*). **Both** Creit hosts are down, so that tolerance is exhausted and Creit's SCP votes now reach us only via relay hops. `quorum` has been reporting `missing=3` throughout; I have **not** confirmed Creit accounts for any of those 3, so treat the link as unverified.

Not validator-blocking: node is `Validating` at L63,709,014, `agree=18 disagree=0`, 33 authenticated peers. Filed without `urgent` on that basis.

## Root cause: our decoder is stricter than stellar-core on the XDR continuation bit

`crates/overlay/src/codec.rs:169-177` (identical at deployed `0ac84d42` and at `origin/main`):

```rust
let raw_len = u32::from_be_bytes([src[0], src[1], src[2], src[3]]);
let is_last_fragment = (raw_len & 0x80000000) != 0;
let len = (raw_len & 0x7FFFFFFF) as usize;

if !is_last_fragment {
    return Err(OverlayError::Message(
        "fragmented XDR records are not supported".into(),
    ));
}
```

stellar-core, `src/overlay/TCPPeer.cpp:673-686` (`TCPPeer::getIncomingMsgLength`, v27.0.0):

```cpp
size_t length = static_cast<size_t>(header[0]);
length &= 0x7f; // clear the XDR 'continuation' bit
length <<= 8;
length |= header[1];
length <<= 8;
length |= header[2];
length <<= 8;
length |= header[3];
```

**stellar-core masks the continuation bit off and never inspects it.** A frame with bit 31 clear is accepted and processed exactly like one with it set; the only subsequent rejections are `length <= 0` and the size limits. There is no fragmentation-rejection path in stellar-core's reader at all.

So on this input stellar-core stays connected and we drop the connection. Per `CLAUDE.md`, SCP/overlay wire bytes are on the **observable / interop surface that MUST match** stellar-core — being strictly stricter is a deviation, and here it is a load-bearing one.

## Why these two peers specifically

Both IPs front **many** node identities, predominantly on port **11725** rather than 11625:

```
16 distinct node keys observed across the two IPs, e.g.
  GAZZ6ENAK3CSGRGNUAYPI3TCN3B6X7BFYD6RSGPLV33HPPC4AUD3KAWW  37.59.22.64:11725
  GB6B2WMTO4QS3JPDKHPPHMHRQM3QHPNUKBTLIV4EINRBZMPVLKZXRXIF  37.59.22.64:11725
  GDLZ6UKL6IOT2BYHYX4ODJA5OG7HFN7EK6WI4KMBNKGWKSZFDQO7Z3MR  37.59.22.64:11625   <- the auths
  GDBSCUZN767ZCGHUMNMI2LVWBKMDTMVTIBWYPBYFTGISX6NR6TGS4LXH  57.128.141.161:11725
  … 12 more
```

Many validators behind one address strongly suggests a proxy / load balancer, which is a plausible source of re-framed records. **I have not captured the raw bytes**, so "bit 31 was clear" is an inference from the fact that this error string is reachable only via that branch — it is sound, but whether the sender emitted a genuine multi-fragment record, or something re-framed it in transit, is undetermined. What is *not* undetermined is that stellar-core would have accepted it.

Mid-stream desynchronisation is unlikely as an explanation: the error fires from `manager::tick` during connect/handshake, before auth completes, when only a handful of bytes have been read.

## Suggested fix

Mirror stellar-core: mask the bit, do not reject on it.

```rust
// XDR record-marking continuation bit: stellar-core clears and ignores it
// (TCPPeer.cpp:673-686). Rejecting non-final fragments is stricter than
// upstream and drops peers upstream would keep.
let len = (raw_len & 0x7FFFFFFF) as usize;
```

The existing `len == 0`, `MIN_MESSAGE_SIZE` and pre/post-auth `max_size` checks below it are unchanged and already match upstream, so this is a deletion rather than a rewrite. Worth a unit test asserting a bit-31-clear header decodes to the same `MessageFrame` as the bit-31-set equivalent.

Adjacent, not duplicates: #3773 is the mirror direction (peers rejecting frames *we* send, XDR-decode arm); #3774 is `encode_message` missing a `MAX_MESSAGE_SIZE` bound on the write path; #3770 is the re-dial cadence that turns one unreachable preferred peer into a sustained WARN stream (and is what makes this 260/h rather than occasional).

## Repro

```bash
L=~/data/74535976/logs/monitor.log
grep -c 'fragmented XDR records are not supported' "$L"           # 6521
grep -m1 'fragmented XDR records are not supported' "$L"          # 2026-07-29T04:59:05.940111Z
grep 'fragmented XDR records are not supported' "$L" \
  | grep -oE 'preferred peer [0-9.]+:[0-9]+' | sort | uniq -c     # the two Creit hosts
getent ahostsv4 alpha.validator.stellar.creit.tech                # 37.59.22.64
getent ahostsv4 beta.validator.stellar.creit.tech                 # 57.128.141.161
```

## Node state

`Validating` at L**63,709,014**, age 5 s, protocol 27, build `henyey-v27.0.0-alpha.1`, deployed `0ac84d42`, PID 1512116, uptime **28 d 00 h 29 m**, 0 rotations. **33** authenticated / 1 pending; quorum `phase=EXTERNALIZE agree=18 disagree=0 missing=3 delayed=0 lag_ms=450`. RssAnon 23.27 GB. RPC `healthy`. Disk free 45 GiB / 98% (#3749). Deploys held on #3702 (60 commits behind `origin/main@ba1d9d77`).

tomerweller:
Closes #3776

## Summary

`crates/overlay/src/codec.rs`'s `Decoder::decode` rejected any inbound frame whose XDR record-marking bit 31 (the "continuation"/"last fragment" bit) was clear, returning `Err("fragmented XDR records are not supported")`. stellar-core's `TCPPeer::getIncomingMsgLength()` (`TCPPeer.cpp:673-686`) masks that bit off (`length &= 0x7f`) and never inspects it — there is no fragment-rejection path anywhere in its reader — and stellar-core's own writer (xdrpp `marshal.cc`) always sets the bit and never implements continuation fragments, so the bit is vestigial on the wire. Being stricter than upstream on the MUST-match overlay wire-framing surface dropped connections stellar-core keeps: both of Creit Tech's preferred peers had been continuously unreachable since 2026-07-29T04:59Z (6,521+ rejected dials at filing).

This is a deletion, not a rewrite: the masked-length computation and all existing length validations (`len == 0`, `MIN_MESSAGE_SIZE`, pre-/post-auth `max_size`) are unchanged, and `is_last_fragment` is still computed and threaded into `MessageFrame` as descriptive metadata. Two stale doc references are corrected alongside: the module-level doc comment in `codec.rs`, and the two `crates/overlay/PARITY_STATUS.md` rows that mischaracterized bit 31 as an "auth bit" backed by an `is_authenticated` field (stale even before this change), per CLAUDE.md's parity-status-update requirement.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3776#issuecomment-5134809891)

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo test -p henyey-overlay` — 487 + 8 + 3 + 1 + 4 passed, 0 failed

## Regression test

- **Test:** `crates/overlay/src/codec.rs::test_decode_accepts_continuation_bit_clear`
- **Pre-fix:** committed as `38e5826664d8837b677e2b3e9d251b201017dafe` — verified FAILED with: `bit-31-clear frame must not be rejected (#3776): Message("fragmented XDR records are not supported")`
- **Post-fix:** verified PASSES after `c665be6c57daf37eaa520ad309f8d5526bfb2128`

The test builds a frame with `encode_message` (bit 31 set), rebuilds the identical frame with bit 31 cleared, decodes both, and asserts the bit-31-clear frame yields `Ok(Some(MessageFrame))` with the same `raw_len` and decoded message, differing only in `is_last_fragment == false`.

## Deviations from plan

- The plan stated `is_last_fragment`/`MessageFrame` have "zero consumers outside `codec.rs`". Precisely: the **`is_last_fragment` field** has zero consumers outside `codec.rs` (the load-bearing claim — verified, the deletion is self-contained), but the **`MessageFrame` type** is re-exported from `lib.rs` and used as a return type in `connection.rs:317/349/443`. No behavior depends on the field, so the conclusion is unchanged.

### DIFF
--- a/crates/overlay/PARITY_STATUS.md
+++ b/crates/overlay/PARITY_STATUS.md
@@ -21,7 +21,7 @@
 | SurveyManager | Partial | Owned by the app layer (`crates/app/src/survey.rs` `SurveyDataManager`/`SurveyState`/`SurveyMessageLimiter` + `crates/app/src/app/survey_impl.rs`); overlay handles transport only. Survey flow complete; JSON summary and limiter behavior simplified |
 | OverlayMetrics | Full | Counters and timers for all message types |
 | PeerBareAddress | Full | Mapped to PeerAddress in lib.rs |
-| MessageCodec (framing) | Full | Length-prefix with auth bit (bit 31) |
+| MessageCodec (framing) | Full | Length-prefix with XDR record-marking continuation bit (bit 31): always set on send, masked and ignored on receive |
 | Error Handling | Full | ERR_LOAD load shedding, 100-byte truncation, send_error_and_drop |
 
 ## File Mapping
@@ -441,7 +441,7 @@ Corresponds to: XDR framing in `TCPPeer.cpp`
 | stellar-core | Rust | Status |
 |--------------|------|--------|
 | RM framing (4-byte header) | `MessageCodec` (Decoder+Encoder) | Full |
-| Auth bit handling (bit 31) | `is_authenticated` field | Full |
+| RM continuation bit (bit 31) — `getIncomingMsgLength()` masks it (`length &= 0x7f`) and never inspects it; the writer (xdrpp `marshal.cc`) always sets it | `MessageFrame::is_last_fragment` (descriptive metadata only; masked, never rejected on) | Full |
 | `MAX_UNAUTH_MESSAGE_SIZE` | `MIN_MESSAGE_SIZE` / `MAX_MESSAGE_SIZE` | Full |
 
 ### PeerSharedKeyId (`N/A`)
--- a/crates/overlay/src/codec.rs
+++ b/crates/overlay/src/codec.rs
@@ -12,8 +12,13 @@
 //! # Length Field Format
 //!
 //! The length field uses XDR record marking semantics:
-//! - **Bit 31 (MSB)**: Final-fragment flag. Henyey emits and accepts only
-//!   single-fragment overlay messages, so this bit must be set.
+//! - **Bit 31 (MSB)**: Record-marking continuation ("last fragment") flag.
+//!   Henyey always sets it on send, since every overlay message is written as
+//!   a single XDR record fragment. On receive the bit is **masked off and
+//!   never rejected on**, matching stellar-core's
+//!   `TCPPeer::getIncomingMsgLength()` (`TCPPeer.cpp:673-686`), which does
+//!   `length &= 0x7f` and never inspects the bit. It is surfaced as
+//!   [`MessageFrame::is_last_fragment`] for diagnostics only.
 //! - **Bits 0-30**: Actual message body length in bytes.
 //!
 //! # Message Size Limits
@@ -163,17 +168,18 @@ impl Decoder for MessageCodec {
                         return Ok(None);
                     }
 
-                    // Read XDR record-marking prefix. Bit 31 is the final
-                    // fragment flag, not an authentication marker.
+                    // Read XDR record-marking prefix. Bit 31 is the
+                    // continuation ("last fragment") flag, not an
+                    // authentication marker. It is masked off and never
+                    // rejected on: stellar-core's
+                    // `TCPPeer::getIncomingMsgLength()` (TCPPeer.cpp:673-686)
+                    // does `length &= 0x7f` and never inspects the bit, so
+                    // rejecting a clear bit would drop peers stellar-core
+                    // keeps (#3776). The value is retained purely as
+                    // descriptive metadata on `MessageFrame`.
                     let raw_len = u32::from_be_bytes([src[0], src[1], src[2], src[3]]);
                     let is_last_fragment = (raw_len & 0x80000000) != 0;
                     let len = (raw_len & 0x7FFFFFFF) as usize;
-
-                    if !is_last_fragment {
-                        return Err(OverlayError::Message(
-                            "fragmented XDR records are not supported".into(),
-                        ));
-                    }
 
                     // Validate length
                     // OVERLAY_SPEC §3.3: len==0 is handled distinctly as
--- a/crates/overlay/src/codec.rs
+++ b/crates/overlay/src/codec.rs
@@ -468,6 +468,71 @@ mod tests {
     }
 
     #[test]
+    fn test_decode_accepts_continuation_bit_clear() {
+        // Regression for #3776: a frame whose XDR record-marking bit 31 is
+        // CLEAR must decode exactly like the bit-31-set equivalent.
+        // stellar-core's `TCPPeer::getIncomingMsgLength()` masks the bit off
+        // (`length &= 0x7f`) and never inspects it, and stellar-core's own
+        // writer (xdrpp `marshal.cc`) always sets it and never implements
+        // continuation fragments — so rejecting a clear bit is strictly
+        // stricter than upstream and drops peers stellar-core would keep.
+        let msg = make_test_message();
+
+        // Reference frame: bit 31 set (what `encode_message` emits).
+        let with_bit = MessageCodec::encode_message(&msg).unwrap();
+        let body = &with_bit[4..];
+
+        // Same frame, bit 31 cleared in the length prefix.
+        let mut without_bit = Vec::with_capacity(with_bit.len());
+        without_bit.extend_from_slice(&(body.len() as u32).to_be_bytes());
+        without_bit.extend_from_slice(body);
+
+        // Sanity: the two differ only in bit 31 of the prefix.
+        let raw_len_set = u32::from_be_bytes([with_bit[0], with_bit[1], with_bit[2], with_bit[3]]);
+        let raw_len_clear = u32::from_be_bytes([
+            without_bit[0],
+            without_bit[1],
+            without_bit[2],
+            without_bit[3],
+        ]);
+        assert_eq!(raw_len_set & 0x80000000, 0x80000000);
+        assert_eq!(raw_len_clear & 0x80000000, 0);
+        assert_eq!(raw_len_set & 0x7FFFFFFF, raw_len_clear & 0x7FFFFFFF);
+
+        // Decode the bit-31-set frame for comparison.
+        let mut codec = MessageCodec::new();
+        let mut buf = BytesMut::from(&with_bit[..]);
+        let expected = codec
+            .decode(&mut buf)
+            .expect("bit-31-set frame should decode")
+            .expect("bit-31-set frame should yield a complete frame");
+        assert!(expected.is_last_fragment);
+
+        // Decode the bit-31-clear frame: must succeed identically, only with
+        // `is_last_fragment == false` as descriptive metadata.
+        let mut codec = MessageCodec::new();
+        let mut buf = BytesMut::from(&without_bit[..]);
+        let frame = codec
+            .decode(&mut buf)
+            .expect("bit-31-clear frame must not be rejected (#3776)")
+            .expect("bit-31-clear frame should yield a complete frame");
+
+        assert!(
+            !frame.is_last_fragment,
+            "bit 31 clear should be reported as is_last_fragment == false"
+        );
+        assert_eq!(frame.raw_len, expected.raw_len);
+        match (frame.message, expected.message) {
+            (AuthenticatedMessage::V0(got), AuthenticatedMessage::V0(want)) => {
+                assert_eq!(got.sequence, want.sequence);
+                assert_eq!(got.mac.mac, want.mac.mac);
+                assert!(matches!(got.message, StellarMessage::Peers(_)));
+            }
+        }
+        assert!(buf.is_empty(), "the whole frame should have been consumed");
+    }
+
+    #[test]
     fn test_codec_streaming() {
         let msg = make_test_message();
         let mut codec = MessageCodec::new();


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Minimal, surgical deletion; the regression test provably fails at the pre-fix commit and passes after, and I independently confirmed nothing downstream branches on `is_last_fragment`. Two pre-existing `doc-drift` items are left inline as non-blocking follow-ups.

<details>

**Cycle:** 1 (no prior `## 🔍 Reviewer: Correctness` comment on this PR) — complete, class-labeled change-list below. Nothing withheld for a later cycle.

**Self-modifying:** no. Changed paths are `crates/overlay/src/codec.rs` and `crates/overlay/PARITY_STATUS.md` only — neither matches the Step 5.5 self-mod patterns.

**Reviewed state:** PR head `c665be6c` merged with current `origin/main` `d5a340dc` (merge `da668a75`, clean, `ort`, no conflicts) in `$REVIEWER_WORKTREE`. Not reviewed on stale base.

### Test-verification gate (linked issue `kind: bug-fix`)

Requires a regression test that **provably fails at the pre-fix commit**. Verified locally, not taken on trust:

- Restored `crates/overlay/src/codec.rs` from the TDD commit `38e58266` (test present, rejection still present) into the merged worktree and ran the single test:
  ```
  test codec::tests::test_decode_accepts_continuation_bit_clear ... FAILED
  panicked at crates/overlay/src/codec.rs:517:14:
  bit-31-clear frame must not be rejected (#3776): Message("fragmented XDR records are not supported")
  test result: FAILED. 0 passed; 1 failed; 486 filtered out
  ```
- Restored to PR head and ran `cargo test -p henyey-overlay` on the merged tree: exit 0, all suites green (lib + `overlay_scp_integration` 3/3 + `overlay_tx_integration` 1/1 + doc-tests).

Gate **satisfied** — the failure mode is exactly the one the issue reports, and the assertion message names the issue.

### Correctness of the diff

- The deleted block was purely an extra rejection. `raw_len`, `is_last_fragment`, and the masked `len = raw_len & 0x7FFFFFFF` are all computed identically before and after; the deletion sits strictly between the mask and the first length validation, so no validation is reordered or skipped.
- All four surviving guards are byte-identical and still ahead of `src.advance(4)`: `len == 0`, `len < MIN_MESSAGE_SIZE`, and the auth-state-dependent `max_size` (`MAX_UNAUTHENTICATED_MESSAGE_SIZE` pre-auth / `MAX_MESSAGE_SIZE` post-auth). Because they were always applied to the *masked* length, clearing bit 31 cannot be used to slip past any of them — this is not a new attack surface, it is the removal of one.
- State machine unchanged: `ReadingLength → ReadingBody { len, is_last_fragment } → MessageFrame::new(message, len, is_last_fragment)`. `MessageFrame::raw_len` is the **masked** length (`len`, not `raw_len`), which is what makes the test's `assert_eq!(frame.raw_len, expected.raw_len)` a meaningful equality rather than a tautology or a false expectation.
- Encoder untouched — still `len | 0x80000000` in both `encode_message` and `Encoder::encode`. Send-side behavior is unchanged, so this is a receive-side relaxation only.

### Deviation check (the `/do` agent's flagged item) — independently verified, judgment holds

The plan claimed `is_last_fragment`/`MessageFrame` have zero consumers outside `codec.rs`. Workspace-wide greps on the merged tree:

- `is_last_fragment` — 14 hits, **all** in `crates/overlay/src/codec.rs` (definition, struct field, ctor param, `DecodeState` variant, the two assignments, plus test assertions). Zero consumers elsewhere. The load-bearing claim holds.
- `MessageFrame` (the type) — re-exported at `crates/overlay/src/lib.rs:94` and used as a **return type only** at `crates/overlay/src/connection.rs:317`, `:349`, `:443`. I read all three sites: they construct nothing and read only `frame.raw_len` (for a log line) and pass the frame through. No caller destructures or branches on `is_last_fragment`.

So the deletion is genuinely self-contained: no downstream code changes meaning. The deviation note in the PR body is accurate and the conclusion is unchanged. Retaining the field is also the right call — it keeps the diff to a pure deletion and preserves the diagnostic, and CI is green under `-D warnings` (a `pub` field on a `pub`-exported struct is not dead code).

### No second framing reader was missed

Grepped the whole workspace for other bit-31 readers (`0x80000000` / `0x7FFF_FFFF` / `fragment`). The other hits are all the **history-archive / XDR-file** surface — `crates/common/src/xdr_stream.rs`, `crates/common/src/fs_utils.rs`, `crates/history/src/download.rs`, `crates/historywork/tests/`, `crates/simulation/src/loadgen_pregenerated.rs` — which is a different surface (files, not the peer socket) with a different upstream counterpart. Leaving them untouched is correct scoping, not an oversight. The overlay peer path has exactly one reader (`Framed<BoxedIo, MessageCodec>` in `connection.rs`), so the fix covers the reported outbound-dial failure path.

### Test quality

Well constructed for a synthetic case (the issue notes the real Creit bytes were never captured, so synthetic is appropriate):
- Builds the reference frame via `encode_message`, so it cannot drift from the encoder.
- Rebuilds a byte-identical frame with only bit 31 cleared, and **asserts that invariant** (`raw_len_set & 0x7FFFFFFF == raw_len_clear & 0x7FFFFFFF`) rather than assuming it — this is what stops the test from silently degenerating into a different-input comparison.
- Compares against the bit-31-set decode rather than hardcoded expectations: `raw_len`, `sequence`, `mac.mac`, message variant.
- Asserts `buf.is_empty()`, which catches the framing-offset class of bug (wrong number of bytes consumed) that would otherwise pass silently.

### Non-blocking notes (`doc-drift`) — left inline, will become follow-up issues

1. `crates/overlay/SPEC_ADHERENCE.md:17,194` still carries the exact "auth bit" mischaracterization that this PR corrects in `PARITY_STATUS.md`, and cites code that does not exist (`is_authenticated = !matches!(...Hello(_))` at `codec.rs:128-145`). `crates/overlay/README.md:121` says "overlay auth bit" too. Pre-existing drift, out of the plan's stated file list, and now mildly self-contradictory with the corrected `PARITY_STATUS.md`. Inline comment left.
2. `crates/overlay/src/codec.rs:27` — the module doc says "Maximum: 32 MB" while `MAX_MESSAGE_SIZE` is `16 * 1024 * 1024` (and stellar-core's `Peer.h:26` is 16 MB, so the constant is right and the doc is wrong). Pre-existing, in the same module doc block this PR edits. Inline comment left.

Neither is introduced by this diff and neither affects behavior, so per the inline-comment convention they are APPROVE-with-inline rather than a bounce.

### Considered and dismissed

- *Test coverage of bit-31-clear + a failing size guard* (e.g. `len == 0` or oversized-pre-auth with the bit clear): not worth adding. Those guards operate on the masked `len` and the removed block sat entirely ahead of them, so there is no distinct code path to cover.
- *Regression risk from a real multi-fragment sender*: covered in the parity lens — henyey and stellar-core fail identically in that hypothetical.
- *Readability*: the replacement comments are accurate and cite the upstream file/lines. The citation `TCPPeer.cpp:673-686` is slightly tight (the function runs 672-703; the mask is at 679) but the cited range does contain the load-bearing code. Not worth a change.

</details>

crates/overlay/PARITY_STATUS.md
```diff
@@ -21,7 +21,7 @@
 | SurveyManager | Partial | Owned by the app layer (`crates/app/src/survey.rs` `SurveyDataManager`/`SurveyState`/`SurveyMessageLimiter` + `crates/app/src/app/survey_impl.rs`); overlay handles transport only. Survey flow complete; JSON summary and limiter behavior simplified |
 | OverlayMetrics | Full | Counters and timers for all message types |
 | PeerBareAddress | Full | Mapped to PeerAddress in lib.rs |
-| MessageCodec (framing) | Full | Length-prefix with auth bit (bit 31) |
+| MessageCodec (framing) | Full | Length-prefix with XDR record-marking continuation bit (bit 31): always set on send, masked and ignored on receive |
```
tomerweller, line 24: `doc-drift` (non-blocking, pre-existing — filing as follow-up, not a bounce)

This row is now correct, but the same "auth bit" mischaracterization it fixes still lives in two other files in this crate, which now contradict it:

- `crates/overlay/SPEC_ADHERENCE.md:17` — `| §4.2 | RFC 5531 record marking (auth bit handling) | Drift | ... receiver uses it as a tag ... effectively repurposed as auth-flag (henyey + stellar-core both do this) |`
- `crates/overlay/SPEC_ADHERENCE.md:194` — *"Both the henyey codec and stellar-core treat the high bit as a per-frame `authenticated` flag"*, citing `codec.rs:128-145: is_authenticated = !matches!(...Hello(_))` — **that code does not exist** anywhere in `codec.rs` at any commit on this branch or on `main`.
- `crates/overlay/README.md:121` — `| codec.rs | Length-prefixed XDR framing with the overlay auth bit. |`

Both claims are false in the same way the old `PARITY_STATUS.md` rows were, and the SPEC_ADHERENCE `Drift` classification is wrong post-fix (there is no drift here — henyey now masks-and-ignores exactly like `getIncomingMsgLength()`). Out of the plan's stated file list and not introduced by this diff, so it does not block the merge.

crates/overlay/src/codec.rs
```diff
@@ -12,8 +12,13 @@
...
-//! - **Bit 31 (MSB)**: Final-fragment flag. Henyey emits and accepts only
-//!   single-fragment overlay messages, so this bit must be set.
+//! - **Bit 31 (MSB)**: Record-marking continuation ("last fragment") flag.
+//!   Henyey always sets it on send, since every overlay message is written as
+//!   a single XDR record fragment. On receive the bit is **masked off and
+//!   never rejected on**, matching stellar-core's
+//!   `TCPPeer::getIncomingMsgLength()` (`TCPPeer.cpp:673-686`), which does
+//!   `length &= 0x7f` and never inspects the bit. It is surfaced as
+//!   [`MessageFrame::is_last_fragment`] for diagnostics only.
 //! - **Bits 0-30**: Actual message body length in bytes.
```
tomerweller, line 22: `doc-drift` (non-blocking, pre-existing — filing as follow-up, not a bounce)

Six lines below this hunk, the same module doc block says:

```
//! # Message Size Limits
//!
//! - Minimum: 12 bytes (at least the authenticated message header)
//! - Maximum: 32 MB (prevents memory exhaustion attacks)
```

but `MAX_MESSAGE_SIZE` is `16 * 1024 * 1024` (16 MB) at `codec.rs:36`, and stellar-core `src/overlay/Peer.h:26` is `MAX_MESSAGE_SIZE = 1024 * 1024 * 16; // 16 MB`. So the **constant is right and the doc is wrong by 2x** — a reader auditing this module against upstream would be told henyey is twice as permissive as stellar-core when it is in fact exact-match.

Not introduced by this diff (pre-existing), but it is in the very doc block this PR is correcting, so worth cleaning up in the same follow-up as the SPEC_ADHERENCE.md/README.md drift.

tomerweller:
## 🔍 Reviewer: Parity

**Verdict:** APPROVE

**Summary:** Both upstream claims verified first-hand against the pinned stellar-core v27.0.0 submodule, including the strongest counter-hypothesis (xdrpp *does* contain fragment-rejecting readers — but stellar-core never links them). Post-fix henyey matches `TCPPeer`'s receive path exactly, including in the multi-fragment hypothetical.

<details>

**Cycle:** 1 (no prior `## 🔍 Reviewer: Parity` comment on this PR) — complete, class-labeled change-list below.

**Lens scope:** observable/interop surface only, per `docs/PARITY.md:33` — *"Overlay P2P wire (`StellarMessage` framing, auth handshake, flood/flow-control observable to peers)"* is a MUST-match surface owned by `crates/overlay`. Internal architecture, metrics, logging and perf are explicitly out of scope and not flagged.

**Verified against:** `stellar-core` submodule at `7696c069d` / `v27.0.0`. Every claim below was re-derived from source, not taken from the plan or the PR body.

### (a) `TCPPeer::getIncomingMsgLength()` masks bit 31 and never branches on it — CONFIRMED

`src/overlay/TCPPeer.cpp:672-703`. The mask is at line 679:

```cpp
size_t length = static_cast<size_t>(header[0]);
length &= 0x7f;            // clear the XDR 'continuation' bit
length <<= 8; length |= header[1];
length <<= 8; length |= header[2];
length <<= 8; length |= header[3];
```

`header[0]` is the MSB byte, so `& 0x7f` clears exactly bit 31 of the 32-bit value — identical to henyey's `raw_len & 0x7FFFFFFF`. The bit's value is discarded at that point and is **not stored, returned, or read again** anywhere in the function or its callers. The only rejections in the remainder of the function are:

```cpp
if (length <= 0 ||
    (!ignoreLimits &&
     ((!isAuthenticated(guard) && (length > MAX_UNAUTH_MESSAGE_SIZE)) ||
      length > MAX_MESSAGE_SIZE)))
```

which map 1:1 onto the guards henyey **retains**: `len == 0`; `len > MAX_UNAUTHENTICATED_MESSAGE_SIZE` pre-auth; `len > MAX_MESSAGE_SIZE` post-auth. Constants also match exactly — `Peer.h:26 MAX_MESSAGE_SIZE = 1024*1024*16` vs henyey `16 * 1024 * 1024`; `TCPPeer.h:19 MAX_UNAUTH_MESSAGE_SIZE = 0x1000` vs henyey `4096`. So the *only* thing henyey did differently on this surface was the rejection this PR deletes.

### (b) No fragment-reassembly path anywhere in stellar-core's reader — CONFIRMED

This is the claim that actually matters (if core reassembled fragments, accepting bit-31-clear without reassembly would mis-parse). Three independent checks:

1. **Text search:** `grep -rni "continuation|fragment" src/overlay/` returns **exactly one hit** across the entire overlay directory — the `// clear the XDR 'continuation' bit` comment at `TCPPeer.cpp:679`. No reassembly buffer, no fragment accumulator, no loop.
2. **Control-flow read:** `readHeaderHandler` (`TCPPeer.cpp:713-741`) → `getIncomingMsgLength()` → if non-zero, `getIncomingBody().resize(expected_length)` and a single `asio::async_read` of **exactly** `expected_length` bytes → `readBodyHandler` → `recvMessage()` → `xdr_argpack_archive` + `g.done()` on that one buffer → `Peer::recvAuthenticatedMessage`. One header, one body, one message. There is structurally nowhere for a second fragment to be appended.
3. **Counter-hypothesis, the important one:** xdrpp *does* ship readers that reject a clear bit — `lib/xdrpp/xdrpp/msgsock.cc:86-92` (`if (!(len & 0x80000000)) { std::cerr << "msgsock: message fragments unimplemented"; ...}`) and `lib/xdrpp/xdrpp/srpc.cc:42-45` (`throw xdr_bad_message_size("read_message: message fragments unimplemented")`). If stellar-core used either, the "core accepts a clear bit" premise would be **false** and this PR would be wrong. It does not: `grep -rn "msgsock|srpc|rpcbind|xdrpp/socket" src/` returns **zero hits**, and the only xdrpp header included anywhere in `src/overlay/` is `<xdrpp/marshal.h>`. Those rejecting readers are xdrpp's own RPC transport, dead code with respect to stellar-core's peer socket. Premise survives its strongest challenge.

### Writer side — the bit is genuinely vestigial

`lib/xdrpp/xdrpp/marshal.cc:19-40`:

```
// last record fragment in a record.  If the high bit is clear, it
// means another fragment follows.  We don't currently implement
// continuation fragments, and instead always set the last-record
// bit to produce a single-fragment record.
...
swap32le(size32(size) | 0x80000000);
```

Both marshalling sites OR in `0x80000000` unconditionally. So stellar-core never *emits* a clear bit and never *interprets* one. Henyey's encoder is unchanged by this PR and also always sets it (`codec.rs:141`, `:268`). Send-side parity is preserved; this is a receive-side relaxation only.

### The multi-fragment hypothetical — henyey and core diverge by zero bytes

The operator's stated worry: if a peer really sent multi-fragment records, does accepting bit-31-clear cause henyey to mis-parse where core would not? Traced both:

- **stellar-core:** reads the masked `length` bytes of fragment 1, `xdr_argpack_archive` + `g.done()` throws `xdr_runtime_error` on the truncated message, `catch` sets `errorMsg = "received corrupt XDR"` → `sendErrorAndDrop(ERR_DATA, ...)` → peer dropped (`TCPPeer.cpp:789-829`).
- **henyey post-fix:** reads the masked `len` bytes, `decode_message` → `AuthenticatedMessage::from_xdr` fails → `OverlayError::Message("failed to decode XDR: ...")` → connection dropped.

Same input, same outcome: connection dropped in both. Core does not reassemble, so it has no advantage here. The change therefore introduces **no** divergence even in the case it was flagged for — it only removes the case where core kept the connection and henyey dropped it. That asymmetry is the entire bug: 6,521 dials where core-compatible peers stayed connected and we did not.

### Doc/PARITY_STATUS.md accuracy

Both edited rows now describe verified behavior rather than the old fiction:
- Line 24 — "always set on send, masked and ignored on receive": matches `encode_message`/`Encoder::encode` and the post-fix decoder exactly.
- Line 444 — the `getIncomingMsgLength()` masking + `marshal.cc` always-set claim: both re-verified above, at the cited locations. `MessageFrame::is_last_fragment` is accurately described as "descriptive metadata only; masked, never rejected on" — confirmed, nothing reads it outside `codec.rs`.

The old rows claimed bit 31 was an "auth bit" backed by an `is_authenticated` field. That was doubly wrong (wrong semantics, and no such field on `MessageFrame`) and was stale before this PR. Correcting it satisfies CLAUDE.md's parity-status-update requirement.

### Considered and dismissed as parity concerns

- **`MIN_MESSAGE_SIZE = 12` has no stellar-core counterpart** (core rejects only `length <= 0`). A 1-11 byte frame is rejected by henyey at the length stage but accepted by core, which then fails XDR unmarshal and drops. Both ends drop the connection, so the observable outcome is identical — not a divergence, and pre-existing / out of scope regardless.
- **Core sends `ERR_DATA` before dropping on corrupt XDR; henyey just errors the connection.** Pre-existing, untouched by this diff (a bit-31-clear frame produced an `OverlayError::Message` before this PR too), and orthogonal to the framing bit. Not a reason to block a fix that strictly reduces divergence.
- **History-archive record marking** (`crates/common/src/xdr_stream.rs`, `fs_utils.rs`, `history/download.rs`) also uses bit 31, and this PR does not touch it. Correct: that is the archive-file surface with a different upstream counterpart (`XDRInputFileStream`), not the peer wire. Not bundling it is right scoping.
- **Widened attack surface:** none. The masked `len` was always what the size guards saw, so bit 31 never gated anything an attacker could exploit; the change removes a rejection rather than adding an acceptance path.

**Parity verdict:** post-fix, henyey's overlay receive framing is byte-for-byte behaviourally equivalent to `TCPPeer`'s on this feature. Net divergence on a MUST-match surface: eliminated.

</details>

tomerweller:
## ✅ Merged

**Commit:** `df2abf83d6031ec6a2aecd017a23f08bc73d6de1`

**Follow-up issues filed for unaddressed inline review comments:** #3803, #3804

**Signals:**
- Reviewer A (Correctness): APPROVE — regression test independently re-verified to fail at pre-fix `38e58266` with the exact reported error and pass at `c665be6c`; `is_last_fragment` confirmed to have zero consumers outside `codec.rs`.
- Reviewer B (Parity): APPROVE — both stellar-core claims re-derived from the pinned `v27.0.0` submodule, including the counter-hypothesis that xdrpp's `msgsock.cc`/`srpc.cc` *do* reject a clear bit (stellar-core links neither; `src/` has zero references).
- CI: green — 47/47 checks (46 SUCCESS, 1 SKIPPED, 0 failures).
- External reviewers: none.
- Bounce-backs: 0 head-scoped / 0 lifetime.
- Self-modifying: no.

Reviewed as it would land: PR head merged with `origin/main` `d5a340dc` (clean, no conflicts) before both lens passes.

Issue #3811, opened 2026-08-09
Catchup replay retry resumes from LCL, permanently losing an aborted batch's ledgerheaders/txhistory/txresults rows


tomerweller:
Discovered while planning #3801. This is the structural reason #3801 **cannot** safely propagate a meta-persist failure, so it is filed separately rather than folded in.

## The bug

`ReplayBatchCtx::replay_one_batch` (`crates/history/src/catchup/replay.rs:281`) does, in order:

1. `download_ledger_data(batch_from, batch_to, ...)`
2. `verify_txsets(&ledger_data)`
3. `replay_via_close_ledger(...)` — closes **every** ledger in the batch, advancing the in-memory `LedgerManager` LCL
4. `persist_ledger_history(&ledger_data, ...)` — writes the **whole batch's** `ledgerheaders` / `txhistory` / `txresults` rows in one transaction

The retry loop `download_verify_and_replay_with_retry_inner` (`replay.rs:517-535`) recomputes the replay start on every attempt from the live in-memory snapshot:

```rust
let snap = ledger_manager.header_snapshot();
let current_lcl = snap.header.ledger_seq;
let replay_first = current_lcl + 1;
```

So if step 4 fails — e.g. a transient `SQLITE_BUSY`, which `persist_ledger_history` maps to the **retriable** `HistoryError::CatchupFailed` — the batch's DB rows are never written, but LCL has *already* advanced past every ledger in the batch. The next attempt resumes at `LCL + 1` and **never revisits them**. If that attempt (or a later one) succeeds, catchup reports success with a permanent multi-table hole of up to `checkpoint_frequency()` (64) ledgers in `ledgerheaders`, `txhistory` and `txresults`.

The same reasoning applies to any error raised *after* `replay_via_close_ledger` has advanced LCL but *before* `persist_ledger_history` commits.

## Why it matters now

#3801 considered making `emit_meta` propagate on retry-budget exhaustion. Under this resume model, propagation from mid-`replay_via_close_ledger` would (a) skip `persist_ledger_history` for the entire batch and (b) resume past the failed ledger anyway — converting a one-row `ledger_close_meta` hole into a ≤64-ledger multi-table hole *without* repairing the original. #3801 therefore keeps log-and-continue, and the propagation option stays blocked on this issue.

## Suggested direction (not decided)

Make the replay/persist pair atomic with respect to the resume anchor. Options to weigh:

[...]

Issue #3799, opened 2026-08-08
monitor-tick: a deterministic catchup wedge fires no trigger — (3b)/(3c)/(3e) all correctly decline, check (2) waits 60m, while apply_ledger_chain=0/0 proves it at 110s


tomerweller:
## monitor-tick: a deterministic catchup wedge is invisible to every automatic trigger — all four gates are individually correct and the node sits `CATCHING UP` for 60 min while `apply_ledger_chain_success_total` proves it did nothing

Self-reflection finding from ticks 2871/2873, on a live wedge that is currently reproducing on demand (#3702, #3797). This is a monitor miscalibration, not a node bug — the node side is already filed.

## What happened

After the 8 d 10 h ENOSPC outage (#3797), the validator was relaunched twice. Both launches wedged identically: restore-from-disk completes, `Catchup range: ReplayOnly` is declared, and then nothing — `lcl` frozen, zero pipeline attempts, every peer shed on straggler timeout.

I walked the skill's trigger table against the live state. **None of them can fire**, and each one is individually right to decline:

| gate | requires | live state | fires? |
|---|---|---|---|
| **(3)** process alive | process dead | alive, PID 2464556 | no |
| **(3a/3d)** repeated-FATAL wipe | dead-process path | alive | no |
| **(3c)** soft-fail wipe | `Recovery escalation blocked` ≥5 min **AND** `fatal_wipe_required` | **0** and **0** occurrences | no |
| **(3b)** wedge | `watchdog_freeze=true` **AND** `/info` empty/timeout | **0** occurrences; `/info` answers in <100 ms | no |
| **(3e)** stuck-but-alive | state matches `valid\|synced\|track` — *"never `Catching up`/`Booting`"* | state is literally **`Catching Up`** | no |
| **(2)** sync deadline | `CRASH_RECOVERY=yes` ⇒ **60 m** deadline | uptime 18 m | reports `CATCHING UP` |

So the tick output is `sync: CATCHING UP (uptime=18m, deadline=60m)` — accurate per the skill, and it will stay that way until **00:28Z**, 60 minutes after a launch that provably did nothing after its first 110 seconds.

(3e)'s exclusion of `Catching up` is deliberate and correct in its own terms — a genuinely-replaying node must not be restarted. The gap is that nothing distinguishes *replaying* from *declared-catchup-and-idle*.

## The signature that is already available and unused

The node exports exactly the discriminator needed, and it was readable **110 seconds** into the wedge:

```
[...]

PR #3818, opened 2026-08-09
Emit wipe signal on fatal startup catchup failure

### LINKED ISSUE #3797: URGENT: mainnet validator OFFLINE 8d10h — ENOSPC kill at 2026-07-31T12:34:56Z, zero alarm signal (recurrence of closed #3478)
### URGENT: mainnet validator has been **OFFLINE for 8 d 10 h** — killed by ENOSPC at `2026-07-31T12:34:56Z`, volume is now at **0 bytes free**. This is the exact failure #3749 was filed to prevent, and the bracket published there at tick 2870 contained the actual death time

The node is **down**, not degraded. No `henyey` process exists; `/info` and the RPC port do not answer; the last line written to `monitor.log` is `2026-07-31T12:34:56Z`. As of `2026-08-08T22:45Z` that is **8 d 10 h 10 m** of continuous outage. The monitor loop was itself interrupted at `2026-07-31T03:29Z` (mid-tick-2871) and did not re-arm, so no tick observed the death — tick 2870 is the last row in `tick-history.jsonl`.

## 1. Cause chain, verbatim from the log

```
12:34:44.794  INFO  State transition from=Validating to=Catching Up
12:34:44.794  WARN  Previous catchup failed — forcing full bucket-apply catchup
12:34:45.552  WARN  work failed work_id=2
                    error=failed to save bucket 776c4f61…c68a to disk:
                          No space left on device (os error 28)
12:34:46.957  WARN  Historywork download failed, falling back to direct catchup
12:34:47.586  WARN  Retrying download-and-replay attempt=1 max_attempts=6
                    error=close_ledger failed at ledger 63733632:
                          bucket error: IO error: No space left on device (os error 28)
12:34:48.406  WARN  … attempt=2 … (same)
12:34:49.416  WARN  … attempt=3 … (same)
12:34:50.748  WARN  Pre-commit hash mismatch (replay mode) ledger_seq=63733632
                    expected_header_hash=4426b58c…  actual_header_hash=6f828db4…
12:34:50.748  ERROR HASH_MISMATCH_DEBUG: live_bucket_list_hash=ebdc4a50…
                    levels 1,2,3 curr=0000…  snap=0000…
12:34:56       (last write; process gone)
```

The sequence is unambiguous: ENOSPC → bucket writes fail → the in-memory bucket list is left with empty levels 1–3 → the replay's pre-commit hash check fails → the process exits. **The hash mismatch is a consequence of the disk-full, not an independent parity bug** — do not triage it as one. Levels 1/2/3 reading `0000…` for *both* curr and snap while levels 5/6/7 hold real hashes is the signature of writes that never landed, not of a divergent merge.

## 2. The #3749 bracket held

Tick 2870 published on #3749 at `03:02Z`:

```
EARLIEST exhaustion: the next co-tenant build of any size seen on 07-30.
LATEST   exhaustion: 1.720 GiB / 3.08 GiB/day = 13.4 h  ~= 2026-07-31T16:30Z
The interval is [now, +13.4 h]
```

Actual death: **`12:34:56Z`** — 9 h 33 m after publication, **3 h 55 m inside the upper bound**. The interval was correct and the node died in it. Recording this because three earlier landing points on that thread were withdrawn; the bracket formulation is the one that worked, and the reason it worked is that it carried the arrival term explicitly instead of freezing it.

Terms at death vs at tick 2870 (`07-31T03:02Z`):

```
                          07-31T03:02Z     08-08T22:45Z (now)      delta
df free                      1.720 GiB          0.000 GiB       -1.720 GiB
mainnet.db                 164.43  GiB        166.00  GiB       +1.57   GiB
mainnet/buckets/            24.102 GiB         24.24  GiB       +0.14   GiB
storestate rows              766,925            868,731         +101,806
```

Note `buckets/` **stopped** reclaiming (+0.14 GiB, against the −1.93 GiB/day I measured at tick 2870) — because the merges that do the reclaiming are the same writes that hit ENOSPC. The negative term is not available when you need it most, which weakens §4 of my tick-2870 comment as a runway input.

## 3. What did **not** happen: VE-12 did not recur

```
zero-byte files in mainnet/buckets/ : 0   (of 65 bucket files)
```

The VE-12 guard (`682c1a1` — `save_to_xdr_file` unlinks the partial file on write error, and `start_merge` re-verifies the loaded file's hash) **held under a real, sustained ENOSPC**. This is the second disk-full event on this volume and the first that produced no corrupt permanent bucket. That fix is validated in production; recording it so the next disk-full incident is not mis-triaged as VE-12.

## 4. The volume — at death, and after the operator reclaim

At `2026-08-08T22:45Z`, before any intervention:

```
/dev/nvme1n1  1.7T used  0 bytes free  100%

mainnet/                    172.30 GiB   (mainnet.db 166.00, buckets/ 24.24, wal 0.09)
74535976/  (monitor session)  4.51 GiB
co-tenant trees                305 dirs  1,670.7 GiB
```

Every one of the **305** co-tenant trees was idle. Minimum idle age across all 305 was **9.3 days**; 239 were idle >30 d, the largest (`9eb89c28`, 103.23 GiB) idle 44 d. **No live process on the host held a cwd inside `/home/tomer/data`** — every `/proc/*/cwd` was scanned. Nothing was using any of it.

```
  103.23 GiB  idle 44.0 d  9eb89c28
   86.76 GiB  idle 37.9 d  b0db5fda
   79.82 GiB  idle 47.1 d  20260622-211046
   64.10 GiB  idle 36.9 d  65c209f0
   61.32 GiB  idle 36.0 d  ed3c207b
   49.75 GiB  idle 37.9 d  b4a0fd10
```

At `22:57Z` the operator reclaimed 96 co-tenant `cargo-target/` build caches — see [#3749 comment](https://github.com/stellar-experimental/henyey/issues/3749#issuecomment-IC_kwDOQxeh6M8AAAABN6YVnw) for the full method and exclusion list. Measured from this side at `23:05Z`:

```
                     22:45Z          23:05Z
df free            0.000 GiB      1,155.9 GiB     (31 % used)
mainnet/           172.30 GiB       172.30 GiB     untouched
74535976/            4.51 GiB         4.51 GiB     untouched
```

`mainnet.db`, `mainnet/buckets/`, and the monitor session — including the `0ac84d42` binary and the 742 MB forensic log this report is built from — all survived intact. **Disk is no longer the blocker.**

## 5. `storestate` is now 868,731 rows and the composition has shifted

```
txset:      449,735
scpstate:   418,992    <- ledgers 63,285,879 .. 63,733,600 (447,721 slots retained)
other:            4
```

Prior measurements on this volume attributed the bloat to `txset:`. It is now **48 % `scpstate:`**, retaining ~26 days of SCP slot state. #3722's slot-state trim is the fix and it is still not in the deployed `0ac84d42`. Both classes need the same operator-authorized `DELETE` + `VACUUM`, and the `VACUUM` needs free space roughly equal to the DB — i.e. **the reclaim in §4 is a prerequisite for the DB reclaim**, not an alternative to it.

## 6. Why the outage produced zero signal for 8 days

The #3749 comment documents the *project-loop* side of the standstill (`project-loop-watchdog.sh` could not clone a scratch tree). The *monitor* side failed independently, through three gates that each behaved exactly as designed:

1. **The monitor loop was interrupted mid-tick-2871 at `03:29Z`** and never re-armed. The node died 9 h later, unobserved.
2. **`check_long_stale_session` returns 1** — no process, `.alive` >6 h stale, env >24 h. The guard exists to stop a monitor auto-restarting a node that was intentionally stopped, and it fired correctly. But its exit path emits **no tick report at all**, so a long-stale session is indistinguishable from silence.
3. **The GitHub token was invalid** (`gh auth status`: *"the token … is invalid"*, HTTP 401 on every call, valid again by `23:05Z`) — so even a tick that did run could not have filed.

Both alarm paths — the tick report and issue filing — were independently disabled, on top of a loop that was not running. Worth a follow-up scoped to that specifically: **a liveness check that does not depend on the monitor loop being healthy**, since here the monitor and the node failed together and each of the three gates above is individually correct.

This is also a **recurrence of closed #3478** (*"OFFLINE: validator aborted on ENOSPC (disk full) — bucket-flush fatal abort"*, 2026-06-18) and of the operator-gated proposal in closed **#3498** (*"In-process retry-with-backoff / pause-and-resume on ENOSPC during bucket ops"*). §1 shows the current binary retries the replay 3× against a full disk and then exits — #3498's pause-and-resume is exactly the behaviour that would have kept the process alive to be rescued 8 days later.

Recovery: the persisted lcl (`ledgerheaders` max = **63,733,600**) is 8.4 d stale and §1's replay already failed on it, so a restart-in-place may re-enter #3702's near-tip stall rather than catch up. The monitor is attempting the restart now (disk is clear, dataset intact per §4) and will report the outcome on this thread either way.

## Reproduce

```bash
df -B1 /home/tomer/data | tail -1
tail -60 ~/data/74535976/logs/monitor.log        # the ENOSPC → hash-mismatch chain
find ~/data/mainnet/buckets -maxdepth 1 -type f -size 0 | wc -l   # 0 — VE-12 guard held
sqlite3 "file:$HOME/data/mainnet/mainnet.db?mode=ro" \
  "select case when statename like 'scpstate:%' then 'scpstate:'
               when statename like 'txset:%' then 'txset:' else 'other' end k,
          count(*) from storestate group by k order by 2 desc;"
# every co-tenant tree's idle age + holder check
for p in /proc/[0-9]*; do readlink $p/cwd; done | grep -c '^/home/tomer/data/'   # 0
```

## Node state

**OFFLINE.** No process. Last log write `2026-07-31T12:34:56Z`; last persisted ledger `63,733,600`; last observed `Validating` ledger (tick 2870) `63,727,372`. Deployed binary was `0ac84d42` (`henyey-v27.0.0-alpha.1`), uptime at death 29 d 14 h. Volume `0` bytes free / 100 %. Deploys remain held on **#3702**.

Filed `urgent`: this is a validator-blocking symptom by definition — the validator is not running.

— `/monitor-tick` 2871

tomerweller:
Closes #3797

## Summary

On a fatal boot-time (startup) catchup failure — a knit-to-LCL mismatch with `LCL > genesis`, or a verification/hash-integrity corruption — the node now emits the `fatal_wipe_required=true` signal (via `trigger_fatal_shutdown`) before propagating the error, so the monitor's existing `(3a)` auto-wipe self-heal fires automatically. Previously the startup catchup path `?`-propagated the same fatal class silently (the online `handle_catchup_result` already emitted the signal), so the monitor never auto-recovered and an operator had to wipe by hand — the ~9 h manual-intervention gap that kept the mainnet validator OFFLINE in #3797.

The change adds a shared classifier `App::startup_catchup_failure_requires_wipe` that mirrors `handle_catchup_result`'s exact #3410 stateless-fresh-genesis knit carve-out (a genesis node has no committed state to protect, so it is not wiped), plus a thin `App::signal_startup_catchup_failure` seam the `run_cmd` startup Err arm calls. Transient/download and ENOSPC-class errors are not `is_fatal_catchup_failure()`, so they are never classified wipe-required — a disk-full during startup catchup is not mis-signalled as corruption (#3478 semantics preserved).

This fixes only the recovery-signal leg of the umbrella incident; the residual root causes (ENOSPC row-loss #3811, monitor-side signal gap #3799) are tracked separately and #3797 should not be silently resolved by this PR alone.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3797#issuecomment-5233422838)

## Test plan

- [x] cargo fmt --check
- [x] cargo test -p henyey-app (1180 + integration tests pass)

## Regression test

- **Test:** `crates/app/src/app/catchup_impl.rs::test_startup_catchup_fatal_emits_wipe_field` (primary) + 4 predicate unit tests
- **Pre-fix:** committed as `9dcd55b` — verified FAILED (seam methods absent; startup path emits no wipe field)
- **Post-fix:** verified PASSES after `5140ad3`

## Deviations from plan

- The primary regression test drives the production seam `signal_startup_catchup_failure` (the exact method `run_cmd`'s Err arm calls) rather than reimplementing the match-arm logic inline, so it locks the real wired behavior. Because that method is new, the failing artifact on `origin/main` is a compile failure (cannot find method) rather than a behavioral assertion failure; the behavioral gap it guards (no `fatal_wipe_required` on the startup path) is real and matches the incident log.
- Did not refactor `handle_catchup_result`'s fatal branch onto the shared predicate (the plan marked this optional / defer-if-risky); the online path's branch has additional interleaved genesis-knit and self-heal handling, so deduplication was left out to avoid disturbing those branches.

### DIFF
--- a/crates/app/src/app/catchup_impl.rs
+++ b/crates/app/src/app/catchup_impl.rs
@@ -7941,4 +7941,202 @@ mod tests {
              herder has externalized target+1 — the run_mode gate is broken"
         );
     }
+
+    // -- #3797: startup (boot-time) catchup must emit the wipe signal ---------
+    //
+    // The online catchup handler (`handle_catchup_result`) already emits
+    // `fatal_wipe_required=true` (via `trigger_fatal_shutdown`) on the fatal
+    // catchup/integrity class, which the monitor greps to drive its (3a)
+    // auto-wipe self-heal. The STARTUP catchup path in `run_cmd` historically
+    // `?`-propagated the same fatal class WITHOUT emitting that signal, so the
+    // monitor never auto-recovered and the operator had to wipe by hand — the
+    // ~9 h outage in #3797. These tests lock the fix: the startup path now
+    // classifies the failure via `startup_catchup_failure_requires_wipe` and
+    // emits the signal via `signal_startup_catchup_failure`.
+
+    /// Helper: build an `App` whose LCL is past genesis (a node with committed
+    /// state worth protecting), matching the #3797 incident node.
+    #[cfg(test)]
+    async fn app_past_genesis() -> App {
+        let dir = tempfile::tempdir().expect("temp dir");
+        let db_path = dir.path().join("rs-stellar-test.db");
+        let config = crate::config::ConfigBuilder::new()
+            .database_path(db_path)
+            .build();
+        let app = App::new(config).await.unwrap();
+        let mut header = app.ledger_manager.current_header();
+        header.ledger_seq = 63_733_600;
+        app.ledger_manager
+            .set_header_for_test(header, henyey_common::Hash256::ZERO);
+        assert!(
+            app.current_ledger_seq() > GENESIS_LEDGER_SEQ,
+            "precondition: node must be past genesis"
+        );
+        // Keep the tempdir alive for the app's lifetime by leaking it — the app
+        // holds an open DB handle into it and the process exits at test end.
+        std::mem::forget(dir);
+        app
+    }
+
+    /// PRIMARY REGRESSION (#3797). Drives the exact production seam that
+    /// `run_cmd`'s startup-catchup Err arm calls — `signal_startup_catchup_failure`
+    /// — under a capturing subscriber, feeding a realistically `.context()`-wrapped
+    /// knit-to-LCL `HistoryError` at `LCL > genesis` (the incident shape). Asserts
+    /// the rendered event carries `fatal_wipe_required=true`.
+    ///
+    /// FAILS on `origin/main`: the startup path there `?`-propagates the error
+    /// with no wipe field (grep over the incident crash log returned 0 matches),
+    /// so the monitor's (3a) auto-wipe never fired. PASSES after the fix.
+    #[tokio::test]
+    async fn test_startup_catchup_fatal_emits_wipe_field() {
+        use std::sync::atomic::Ordering;
+        use std::sync::{Arc, Mutex};
+        use tracing::{
+            field::{Field, Visit},
+            subscriber::with_default,
+            Event, Metadata, Subscriber,
+        };
+
+        const FATAL_WIPE_FIELD: &str = "fatal_wipe_required";
+
+        #[derive(Default)]
+        struct CapturedBool {
+            value: Option<bool>,
+        }
+        impl Visit for CapturedBool {
+            fn record_bool(&mut self, field: &Field, value: bool) {
+                if field.name() == FATAL_WIPE_FIELD {
+                    self.value = Some(value);
+                }
+            }
+            fn record_debug(&mut self, _: &Field, _: &dyn std::fmt::Debug) {}
+        }
+
+        #[derive(Default, Clone)]
+        struct WipeFieldSubscriber {
+            captured: Arc<Mutex<Option<bool>>>,
+        }
+        impl Subscriber for WipeFieldSubscriber {
+            fn enabled(&self, _: &Metadata<'_>) -> bool {
+                true
+            }
+            fn new_span(&self, _: &tracing::span::Attributes<'_>) -> tracing::span::Id {
+                tracing::span::Id::from_u64(1)
+            }
+            fn record(&self, _: &tracing::span::Id, _: &tracing::span::Record<'_>) {}
+            fn record_follows_from(&self, _: &tracing::span::Id, _: &tracing::span::Id) {}
+            fn event(&self, event: &Event<'_>) {
+                let mut cap = CapturedBool::default();
+                event.record(&mut cap);
+                if let Some(v) = cap.value {
+                    *self.captured.lock().unwrap() = Some(v);
+                }
+            }
+            fn enter(&self, _: &tracing::span::Id) {}
+            fn exit(&self, _: &tracing::span::Id) {}
+        }
+
+        let app = app_past_genesis().await;
+
+        // Realistic incident shape: a typed knit-to-LCL HistoryError wrapped in
+        // an anyhow `.context()` layer, as `?` would add on the way out.
+        let err: anyhow::Error = anyhow::Error::from(henyey_history::HistoryError::KnitLclHashMismatch {
+            expected: "4426b58c".into(),
+            actual: "6f828db4".into(),
+        })
+        .context("close_ledger failed at ledger 63733632");
+
+        // Guard the downcast-through-context path used by the classifier.
+        assert!(
+            app.startup_catchup_failure_requires_wipe(&err),
+            "classifier must fire on a .context()-wrapped knit-to-LCL error at LCL > genesis"
+        );
+
+        let sub = WipeFieldSubscriber::default();
+        let captured = sub.captured.clone();
+
+        // Drive the exact production seam run_cmd's Err arm calls.
+        with_default(sub, || {
+            app.signal_startup_catchup_failure(&err);
+        });
+
+        assert_eq!(
+            *captured.lock().unwrap(),
+            Some(true),
+            "startup catchup fatal failure must emit {FATAL_WIPE_FIELD}=true so the \
+             monitor's (3a) auto-wipe self-heal fires (#3797)"
+        );
+        assert!(
+            app.fatal_state_failure.load(Ordering::SeqCst),
+            "fatal_state_failure must be set on a fatal startup catchup failure"
+        );
+    }
+
+    /// #3797 predicate: a knit-to-LCL fatal at `LCL > genesis` requires a wipe.
+    #[tokio::test]
+    async fn test_startup_catchup_knit_fatal_beyond_genesis_requires_wipe() {
+        let app = app_past_genesis().await;
+        let err: anyhow::Error = henyey_history::HistoryError::KnitLclHashMismatch {
+            expected: "aa".into(),
+            actual: "bb".into(),
+        }
+        .into();
+        assert!(
+            app.startup_catchup_failure_requires_wipe(&err),
+            "knit-to-LCL at LCL > genesis is unrecoverable in-process — must wipe"
+        );
+    }
+
+    /// #3797 preserves #3410: a knit-to-LCL fatal at genesis (stateless fresh
+    /// node) must NOT wipe — there is no committed local state to protect.
+    #[tokio::test]
+    async fn test_startup_catchup_knit_fatal_at_genesis_no_wipe() {
+        let dir = tempfile::tempdir().expect("temp dir");
+        let db_path = dir.path().join("rs-stellar-test.db");
+        let config = crate::config::ConfigBuilder::new()
+            .database_path(db_path)
+            .build();
+        let app = App::new(config).await.unwrap();
+        assert_eq!(
+            app.current_ledger_seq(),
+            GENESIS_LEDGER_SEQ,
+            "precondition: fresh app must be at genesis"
+        );
+        let err: anyhow::Error = henyey_history::HistoryError::KnitLclHashMismatch {
+            expected: "aa".into(),
+            actual: "bb".into(),
+        }
+        .into();
+        assert!(
+            !app.startup_catchup_failure_requires_wipe(&err),
+            "knit-to-LCL at genesis must NOT wipe (preserves #3410)"
+        );
+    }
+
+    /// #3797: a transient (download/archive/ENOSPC-class) error must NOT be
+    /// classified wipe-required — no false wipe on disk-full.
+    #[tokio::test]
+    async fn test_startup_catchup_transient_error_no_wipe() {
+        let app = app_past_genesis().await;
+        let err: anyhow::Error =
+            henyey_history::HistoryError::ArchiveUnreachable("timeout".into()).into();
+        assert!(
+            !app.startup_catchup_failure_requires_wipe(&err),
+            "transient errors must not trigger a wipe"
+        );
+    }
+
+    /// #3797: a non-knit verification/integrity fatal at `LCL > genesis`
+    /// (genuine local corruption) requires a wipe.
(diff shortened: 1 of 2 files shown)


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The fix correctly emits the `fatal_wipe_required=true` signal on a fatal startup-catchup failure so the monitor's `(3a)` auto-wipe self-heal fires, closing the ~9 h manual-intervention gap in #3797. The classifier, the `run_cmd` seam, and the 5 unit tests are all sound; a single non-blocking doc-precision note is filed inline.

<details>

Cycle 1 (no prior Correctness comment on this PR) — complete class-labeled change-list.

**Traced and verified correct:**
- `signal_startup_catchup_failure` → `trigger_fatal_shutdown` is the *only* emitter of `fatal_wipe_required=true` (lifecycle.rs:2329), matching the monitor's grep contract. Setting `fatal_state_failure` + a receiver-less `shutdown()` broadcast at startup is harmless; `run_cmd` still returns `Err`, so the supervisor sees exit-nonzero.
- `run_cmd.rs`: the `?` → explicit `match` conversion preserves error propagation (`Err(e) => { signal; return Err(e) }`) and the `Ok(_result)` binding. Correct.
- `startup_catchup_failure_requires_wipe` downcasts through the anyhow source chain, so a `.context()`-wrapped `HistoryError` is still classified — the primary test exercises exactly this.
- `handle_catchup_result` (the online sibling) is reached only from lifecycle.rs:848; the startup path calls `catchup_with_run_mode` directly and never routes through it, so this new interception is the correct (and only) classification point for the boot path — not a double-signal.

**Test-verification gate (bug-fix `kind:`):** PASS. Primary regression `test_startup_catchup_fatal_emits_wipe_field` drives the real production seam under a capturing subscriber with the incident-shaped `.context()`-wrapped `KnitLclHashMismatch` at `LCL > genesis`, asserting `fatal_wipe_required=true`. Pre-fix it fails to compile (seam absent) — acceptable for a new-seam bug-fix; the behavioral gap it guards is real. Four predicate tests cover: knit-fatal beyond genesis (wipe), knit-fatal at genesis (no wipe, preserves #3410), transient `ArchiveUnreachable` (no wipe), and non-knit `VerificationFailed` (wipe). Good coverage of every branch of the classifier.

**Non-blocking (filed inline, doc-precision):** The docstring says the classifier "mirrors ... the exact classification the online catchup handler already applies." It mirrors the fatal + `#3410` genesis-knit carve-outs, but intentionally *omits* the `#3282` seeded-from-local-clone divergence carve-out (`is_local_vs_archive_divergence`), which the online path treats as self-healing without a wipe. This omission is **correct** for the startup context — at boot the in-process self-heal (in-memory `catchup_needs_full_reset` + retry) is unavailable because `run_cmd` exits on `Err`, and the flag is not persisted for the divergence case (`CATCHUP_PERSIST_PENDING` is a separate crash-window sentinel, cleared on successful persist), so a no-wipe restart would re-seed from the same divergent local clone and crash-loop; wiping is the right recovery. The wording should just say it mirrors the *fatal + genesis-knit* classification rather than the "exact" classification, to avoid implying full parity. Class: `doc-drift`. Non-blocking.
</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Auto-wiping on a fatal startup-catchup failure matches the online handler's existing posture and is a strict improvement over the 8-day manual-intervention outage. No infinite-wipe loop, and transient/ENOSPC-class failures are correctly excluded so #3478 semantics are preserved.

<details>

Cycle 1 — complete class-labeled risk assessment. Files touched are `crates/app/` only (not a parity-critical crate per docs/PARITY.md), so this is a risk lens, not a parity lens. No observable-surface (hash / XDR / wire / archive / RPC) change.

**Regression risk — none identified:**
- No wipe-loop: startup fatal → wipe → reboot at genesis → `genesis_recovery_catchup_plan` returns `(Minimal, seed_local_state=false)` → archive-authoritative bucket-apply with no local clone → the knit-to-LCL-against-local class cannot recur. A bad/unreachable archive yields a *transient* error (not `is_fatal_catchup_failure`) → no wipe → retry. A genesis knit is carved out (no wipe). So the loop terminates.
- Aggressiveness vs. the online path is justified: the online path's non-destructive self-heal requires staying in-process; the startup path exits, and the force-full flag is in-memory only, so wiping is the correct startup recovery for a genuinely divergent local state (see Correctness inline note).

**Operational safety:**
- ENOSPC / disk-full during startup catchup (e.g. via `try_resolve_pending_bucket_merges()?`, #3478) surfaces as an IO error, which is not `is_fatal_catchup_failure()` → never mis-signalled as wipe-required. Covered by `test_startup_catchup_transient_error_no_wipe`. #3478 "free space, no wipe" semantics preserved.
- The change is scoped to the boot path Err arm; the running-node event loop is untouched.

**Security / perf:** No new external inputs, no allocation/perf-sensitive paths, no unsafe. N/A.

No blocking risks.
</details>

crates/app/src/app/catchup_impl.rs
```diff
@@ -3004,6 +3004,68 @@ impl App {
...
 
+    /// #3797: classify a **startup (boot-time) catchup** failure as
+    /// wipe-required.
+    ///
+    /// Returns `true` iff `err` downcasts to a fatal catchup/integrity
+    /// [`HistoryError`](henyey_history::HistoryError)
+    /// (`is_fatal_catchup_failure()`) **and** it is not the narrow
+    /// stateless-fresh-genesis knit carve-out (`current_ledger_seq() ==
+    /// GENESIS_LEDGER_SEQ && is_knit_to_lcl_failure()`). This mirrors, for the
+    /// startup path, the exact classification the online catchup handler
```
tomerweller, line 3015: [non-blocking, doc-precision] This says the startup classifier applies "the exact classification the online catchup handler already applies," but it intentionally omits the #3282 seeded-from-local-clone carve-out (`is_local_vs_archive_divergence`): the online `handle_catchup_result` treats that class as self-healing (in-memory `catchup_needs_full_reset` + in-process retry, no wipe), whereas this startup path signals a wipe for it.

That divergence is **correct** — at boot the in-process self-heal is unavailable (`run_cmd` exits on `Err`, the force-full flag is in-memory only and not persisted for the divergence case), so a no-wipe restart would re-seed from the same divergent clone and crash-loop; wiping is the right startup recovery. The only ask is wording: say it mirrors the *fatal + genesis-knit* classification rather than the "exact" classification, so future readers don't assume full parity with the online handler's seeded-clone self-heal.

tomerweller:
## ✅ Merged

**Commit:** 217f230b74f2d5a9af4fbda26216af204a108582

**Follow-up issues filed for unaddressed inline review comments:** #3819

Issue #3789, opened 2026-07-30
monitor: canonicalize monitor-watchdog.sh into the repo and fix over-firing (STALE_SECS/flock/fail-closed)


tomerweller:
Follow-up from #3757 (evaluator too-fresh guard). The nominal root cause of #3757 — the crontab watchdog firing a duplicate headless `/monitor-tick` on essentially every cycle — lives in the **deployed-only** `monitor-watchdog.sh`, which has no canonical copy in this repo (`git ls-files | grep watchdog` returns only `scripts/project-loop-watchdog.sh`, a different watchdog). It therefore could not be fixed in the #3757 PR.

Scope:
1. Canonicalize the deployed `monitor-watchdog.sh` into the repo (e.g. `scripts/monitor-watchdog.sh`) so it is reviewable/testable.
2. Resize `STALE_SECS` off the observed ~44 min cadence, not the nominal 20 min sleep (#3757 body suggests ~4200s/70min).
3. Replace the `.alive`-mtime in-flight guard with an `flock`-based exclusive lock (pattern: `.monitor-tick-filing.lock`); a watchdog tick that cannot take the lock logs `skipped: tick in flight` and exits 0. This closes the interleaving race that #3757 comment 5100384841 showed corrupts archive `metadata.env`.
4. Fail **closed**, not open, on an unparseable history tail, and accept both `ts`/`timestamp` keys (#3757 comment on the 56,590-year staleness / unconditional fire).

See #3757 for full evidence.

tomerweller:
## Over-firing reproduced with a concrete cost: a stale duplicate posted misleading guidance to an urgent issue

New instance, and this time the harm is not just wasted work — the duplicate tick **published stale
advice on an urgent incident issue** (#3905) that would have cost the operator 108 G and a redundant
investigation.

### What happened

At `12:04:57Z` a headless watchdog tick posted a "Monitor status" comment on #3905. It reads as
current but is **two ticks out of date**:

| the duplicate said | actual state at the time |
|---|---|
| "needs the offline `verify-execution --from 64069440 --to 64069520` repro to localize the diverging header field" | **already run at 11:06Z** — 81/81 ledgers matched, 0 mismatches |
| "undiagnosed root cause" | localized to `bucketListHash`, then to **one 36-byte-short XDR record** in live level-0 `curr` |
[...]

PR #3788, opened 2026-07-30
Key VE deploy-target greenness on each sha's most-recent verdict

tomerweller:
Closes #3740

## Summary

`select_latest_green_deploy_target` (`scripts/lib/monitor-decisions.sh`) selected a deploy target off **any historical** `Verify Execution (Mainnet)` success for a sha, *skipping* (rather than disqualifying) a **more-recent `failure`** of that same sha — a fail-open in a correctness-critical deploy gate (#3740). VE is deterministic per-sha, so a newer failure on a previously-green sha is either a genuine regression surfaced by newer mainnet data (e.g. #3730) or a flake; either way "latest success ever" is the wrong signal for "safe to ship now."

The fix keys greenness on each sha's **most-recent definitive verdict**. Because `gh run list` feeds records newest-first, the first definitive (`success`/`failure`) row per sha is authoritative; an older `success` can no longer resurrect a sha whose newest VE run `failure`d. Non-definitive conclusions (`cancelled`/`skipped`/…) remain cancel-per-head noise and let older rows decide the sha (preserves the #3351 `HEAD|cancelled` → older-green behavior). Per-sha bookkeeping uses whitespace-delimited accumulator strings rather than a bash-4 associative array, because the deploy gate invokes this under `emulate -L zsh` (#3581/#3592). A new terminal reason token `green-superseded-by-failure` is emitted when a newer failure superseded an older success and no other usable green remained; it flows through the caller's existing `defer-no-green` branch, so there is **no caller change** and the fail-closed contract is preserved.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3740#issuecomment-5133312132)

## Test plan

- [x] `bash -n` on both changed files; `zsh -n` on the lib (zsh-invoked path)
- [x] `shellcheck -S warning` — no new warnings in the selector region
- [x] `scripts/test-monitor-skill-snippets.sh` passes (0 failures; #3351/#3581/#3592 cases preserved)

## Regression test (kind: bug-fix)

- **Tests:** `scripts/test-monitor-skill-snippets.sh` — Case 7 (`sha with newer failure is disqualified`) and Case 8 (`newer failure + no other green → green-superseded-by-failure`); plus a guard Case 9 (`cancelled newest does not mask older green same sha`, passes on main too).
- **Pre-fix:** committed as `a5417de` — verified Cases 7 & 8 FAILED (selector returned `G` off the older success: `got 'G' (expected 'G_OLD')`, `out='G' err=''`).
[...]

Issue #3758, opened 2026-07-27
eval-alarms.py: maybe_reset_counter_snapshot destroys the counter-streak baseline on a 'missing process identity' skip, defeating the #3279 guard


tomerweller:
## Symptom

A `/monitor-tick` that omits the `PID` / `START_TICKS` env vars **destroys the
counter-streak baseline**, silently resetting every sustained-breach alarm's
streak to 0 — even though `eval_counter_streak` has an explicit guard (#3279)
whose stated contract is that the baseline is *preserved untouched* in exactly
this case.

Hit live on tick 2805: `counter_streak_snapshot` went from

```
version=1
pid=1512116
start_ticks=983181529
counter_value=3413
breach_streak=22      ← 22 ticks of sustained #3748 breach
```

to a **1-byte file** after one identity-less evaluator run. The `breach_streak=22`
evidence on #3748 was erased. (Recovered by hand this tick from the pre-rotation
read; had the tick not happened to log the value first, it would be unrecoverable.)

## Root cause

Two functions disagree, and the destructive one runs second.

**The guard** — `eval_counter_streak`, `scripts/lib/eval-alarms.py:1122-1135`:

```python
    # Missing process identity guard (#3279): an abbreviated tick that skips
    # exporting PID/START_TICKS passes EMPTY strings here. An empty identity is
    # NOT a valid distinct incarnation — ...
    # Establish nothing, decide nothing: skip BEFORE any snapshot/metric I/O so
    # the prior valid baseline is preserved untouched.
    if not pid or not start_ticks:
        return make_result(alarm, "skipped",
                           skip_reason="missing process identity",
                           extra_values=ev_default)
```

The comment's promise — *"the prior valid baseline is preserved untouched"* — is
correct **within this function**. It returns without touching the snapshot.

**The destroyer** — `maybe_reset_counter_snapshot`, `eval-alarms.py:411-461`,
called from the main loop at **line 1644**:

```python
        maybe_reset_counter_snapshot(alarm, kind, result["state"], state_dir)
```

```python
    if state != "skipped":
        return
    ...
    elif kind == "counter-streak":
        snapshot_file = alarm.get("snapshot_file", "counter_streak_snapshot")
        snapshot_path = state_dir / snapshot_file
        snapshot = read_snapshot(snapshot_path)
        if snapshot:
            # Clear the entire snapshot to force a full "collecting_baseline" ...
            write_snapshot(snapshot_path, {})
```

[...]

PR #3821, opened 2026-08-09
Add symmetric too-fresh guard to counter-family alarm evaluators

### LINKED ISSUE #3757: monitor-watchdog fires a duplicate tick every cycle (STALE_SECS=1800 vs ~44m real cadence) — burns alarm cooldowns, 2x streak/archive advance, 113 tick-history schema variants
## Symptom

The crontab watchdog (`/home/tomer/data/monitor-watchdog.sh`, `*/15`) is firing a
**duplicate headless `/monitor-tick` on essentially every interactive loop cycle**,
not as a rare backstop. Today (2026-07-27) it fired 13 times:

```
07:30:01 watchdog: loop stale (2047s since last tick) — launching headless tick
08:15:01 watchdog: loop stale (2259s since last tick) — launching headless tick
09:00:01 watchdog: loop stale (2316s since last tick) — launching headless tick
09:45:02 watchdog: loop stale (2290s since last tick) — launching headless tick
10:30:01 … 11:15:01 … 12:00:01 … 12:45:01 … 13:30:01 … 14:15:01 … 15:00:01
18:00:01 watchdog: loop stale (2521s since last tick) — launching headless tick
```

The watchdog is doing exactly what it is configured to do. The **configuration is
mis-sized against the actual tick cadence**, so the backstop has become a second
concurrent producer.

This is not a node defect — the validator is healthy throughout (`Validating`,
L63675933, quorum `agree=18 disagree=0`, 27 peers, RPC `healthy`). It is a defect in
the monitoring apparatus, and it is **actively suppressing real alarm filings**
(§Harm 1).

## Root cause

Two independent mis-calibrations compound.

### (a) `STALE_SECS` is shorter than the real tick period

`monitor-watchdog.sh`:

```
STALE_SECS=1800        # fire when last completed tick is older than 30 min
ALIVE_FRESH_SECS=600   # skip if a tick STARTED within the last 10 min (in-flight)
REFIRE_COOLDOWN=1800   # never fire more than once per 30 min
TICK_TIMEOUT=1500      # cap a headless tick at 25 min
```

`STALE_SECS=1800` assumes a ~20-minute cadence. But an interactive tick is
*20 min of `ScheduleWakeup` sleep **plus** the tick's own runtime*. Measured
inter-row deltas in `tick-history.jsonl` for 2026-07-27:

| tick `ts` | Δ from previous |
|---|---:|
| 09:51:03Z | +44.2m |
| 10:35:15Z | +44.2m |
| 11:21:56Z | +46.7m |
| 12:04:39Z | +42.7m |
| 12:52:31Z | +47.9m |
| 13:37:29Z | +45.0m |
| 14:22:00Z | +44.5m |
| 15:06:52Z | +44.9m |
| 18:00:00Z | +42.0m |
| 18:08:48Z | **+8.8m** ← the duplicate |

Mean ≈ **44 min** against a 30-min staleness bar. The watchdog therefore trips on
every cycle, bounded only by `REFIRE_COOLDOWN=1800`. The `2047–2521s` values in its
own log are all just "one normal cycle".

### (b) The in-flight guard cannot see a long-running tick

`ALIVE_FRESH_SECS=600` is meant to suppress a fire while a tick is in progress, via
`.alive` mtime. But per `SKILL.md` §Session-alive marker, `.alive` is touched **once,
at tick start**. A tick whose body runs longer than 10 minutes has a stale `.alive`
for the remainder of its run, so the guard silently stops protecting exactly the
ticks that most need protection.

Observed at 18:00:01 — interactive tick 2803 started ~17:22Z and ran until ~18:05Z.
`.alive` was ~38 min old, so the guard passed and a headless tick launched **into a
still-running tick**:

```
tick 2803 (interactive)   17:22Z ──────────────────────────── 18:05Z
headless tick             18:00:01Z ──────────────────── 18:09:30Z (exit=0)
                                     └── ~5 min overlap ──┘
```

### 1. Duplicate ticks consume alarm cooldowns, suppressing filings

The headless tick ran the canonical `eval-alarms.py`, saw `recovery-stalled` firing,
filed/stamped it, and wrote the cooldown at **18:05:40Z** with
`cooldown_seconds=7200`.

Consequence for the very next interactive tick (2804, 18:29Z): `recovery-stalled` is
still firing (`delta=5 streak=22/3`) but is now **suppressed until 20:05:40Z** —
because a duplicate of itself already burned the cooldown. Tick 2803 had explicitly
recorded that the next fire would be comment-eligible; it no longer is.

This is a general failure mode, not a one-off: any alarm with a cooldown can have its
budget spent by the duplicate rather than the tick that reasons about it.

### 2. Streak/ratio snapshots advance at 2× the intended rate

`counter_streak_snapshot` went `breach_streak=20` (tick 2803) → **22** (tick 2804)
across a single interactive cycle, because the headless tick consumed an increment.
Streak thresholds expressed in *ticks* (`streak=22/3`) therefore no longer mean what
the catalog says they mean — they run at double speed and are not comparable across
periods when the watchdog is/isn't firing.

This is the hazard #3209 warned about in the "run the evaluator exactly once per tick"
caution. That caution is enforced *within* a tick; nothing enforces it *across*
concurrent ticks.

### 3. Archive retention window is consumed 2× as fast (interacts with #3724)

Two snapshots per cycle instead of one:

```
2026-07-27T18:06:01.938208947Z   ← headless tick
2026-07-27T17:49:50.796197818Z   ← interactive tick 2803
2026-07-27T17:16:17Z             ← (compact format — see #3724)
```

At `archive_count=501` (the rolling-500 ceiling, already at its off-by-one fixed
point per #3724), doubling the production rate halves the real retention horizon.
This is also the mechanism behind #3724's "364/501 archives from a second producer"
and #3739's replay abort — **the "second producer" is the watchdog**, so those two
issues have a shared upstream cause that neither currently names.

### 4. `tick-history.jsonl` schema has fragmented into 113 variants

`SKILL.md` §Tick history capture specifies exactly 9 keys
(`ts, status, ledger, build, deploys, warnings, actions, self_reflect, watch`).
Census of all 2,945 rows:

| | count | share |
|---|---:|---:|
| rows total | 2,945 | |
| **exactly canonical** | **1,446** | **49.1%** |
| distinct key-signatures | **113** | |
| missing `status` | 550 | 18.7% |
| missing `ledger` | 674 | 22.9% |
| missing `ts` | 0 | 0% |

The largest non-canonical signature (n=444) is missing
`status, ledger, build, deploys, warnings, actions` — i.e. carries a timestamp and
almost nothing else. Since `/daily-summary` and the per-tick 24h digest count by
`status` and sum `deploys`, **every such row is silently dropped from the
aggregate** — the digest under-reports, and does so by a fraction that varies with
how often the watchdog fired.

(`ts` is universally present, so nothing fails to parse. The loss is silent, not
loud.)

## Suggested fix

1. **Size `STALE_SECS` off the observed cadence, not the nominal sleep.** With a
   20-min `ScheduleWakeup` and a 15–25 min tick body, staleness starts at ~50 min.
   `STALE_SECS=4200` (70 min) still catches a genuinely dead loop within ~1.5
   cycles while never tripping on a healthy one. (Note the nominal-vs-actual gap is
   itself worth surfacing: the loop's real period is ~2.2× its configured interval.)
2. **Make the in-flight guard actually track in-flight.** Either touch `.alive`
   periodically through the tick (e.g. at each block boundary) rather than only at
   start, or — better — take an exclusive `flock` for the duration of a tick and
   have the watchdog `flock -n` test it. A lock is race-free where an mtime
   heuristic is not; `.monitor-tick-filing.lock` already establishes the pattern in
   this codebase.
3. **Serialize ticks unconditionally.** Even with (1) and (2) correct, a headless
   and interactive tick must never run `eval-alarms.py` concurrently. The same
   `flock` from (2) gives this for free: a watchdog tick that cannot take the lock
   should log `skipped: tick in flight` and exit 0.
4. **Validate the tick-history row on append.** Assert the key set equals the
   canonical 9 before writing, and fail loudly on mismatch. 113 variants accumulated
   precisely because a non-compliant append is indistinguishable from a compliant one.
   `scripts/test-monitor-skill-snippets.sh` is the natural home for a schema test.
5. **Cross-reference from #3724 / #3739.** Both attribute their symptom to an
   unidentified "second producer" of archive dirs. It is the watchdog. Fixing (3)
   removes the duplicate-producer half of both.

## Regression test suggestion

A test that launches two `monitor-tick` bodies concurrently and asserts (a) exactly
one advances `counter_streak_snapshot`, (b) exactly one archive dir is produced, and
(c) exactly one cooldown stamp is written, would have caught all of §Harm 1–3. A
second, cheaper test asserting `set(json.loads(line).keys()) == CANONICAL` over
`tick-history.jsonl` covers §Harm 4.

## Severity note

Filed **without** `urgent` per the monitor label policy — the validator is
`Validating` and unaffected; nothing here is validator-blocking. But it degrades the
monitor's own integrity in a way that *hides* node problems (a suppressed alarm reads
identically to an absent alarm), so it is worth more than cosmetic priority.

## Cross-references

- **#3209** — "run the evaluator exactly once per tick"; that invariant is violated
  across concurrent ticks, which the existing guard does not cover.
- **#3724** — mixed archive timestamp formats from a "second producer"; the second
  producer is the watchdog.
- **#3739** — replay aborts on non-compliant archives; same upstream.
- **#3748** — the alarm whose cooldown was consumed by the duplicate this cycle.
- Prior watchdog incident (2026-07-22): cron `PATH` lacked `~/.local/bin`, so the
  watchdog silently no-opped for ~29 h; fixed with an absolute `CLAUDE_BIN`. That fix
  is present and working — this issue is the opposite failure mode (over-firing).

Surfaced by `/monitor-tick` 2804, 2026-07-27T18:35Z.

tomerweller:
Closes #3757

## Summary

A duplicate / back-to-back monitor-tick (the watchdog firing into a still-running interactive tick, per #3757) samples an inter-scrape interval far below the real ~44m cadence. The cross-tick counter-family evaluators in `scripts/lib/eval-alarms.py` had only an UPPER bound (`gap_stale`); with no lower bound, such a tick would reset a breach streak to 0 (delta=0) or advance it (delta>=1), burning alarm cooldowns and double-advancing streak/ratio/archive snapshots.

This adds the symmetric lower-bound mirror: `compute_too_fresh(prev_scrape_age, min_eval_window)` derived in `main()` from `MIN_EVAL_WINDOW_SECONDS` (default 600), threaded into `eval_counter`, `eval_counter_dynamic`, `eval_histogram_p99`, `eval_counter_ratio`, and `eval_counter_streak`. Each returns a NON-destructive skip (`SKIP_INTERVAL_TOO_SHORT`, no snapshot write), placed after the identity/counter-reset/gap_stale branches so a genuine restart is still handled first. `maybe_reset_counter_snapshot` gains a `skip_reason` parameter and early-returns on the sentinel so the centralized reset preserves the streak rather than wiping it. Age -1/absent stays fail-safe (not too-fresh); default `too_fresh=False` keeps every existing path byte-identical.

Scope note: this is the evaluator-level *floor* that neutralizes the duplicate-tick harm. The nominal watchdog resizing (`STALE_SECS`/`flock`), the acceleration case, and tick-history schema validation are deferred follow-ups #3789/#3790/#3791 per the plan.

## Plan reference

[Converged Plan comment](https://github.com/stellar-experimental/henyey/issues/3757#issuecomment-5134241612)

## Test plan

- [x] `python3 scripts/lib/test_eval_alarms_min_eval_window.py` — 6 passed
- [x] existing eval-alarms suites: `test_eval_alarms_counter_reset` (24), `_converged_loop` (9), `_histogram_rebucket` (10), `_inbound_auth_severity` (7) — all pass
- [x] `bash scripts/test-monitor-skill-snippets.sh` — exit 0
- [x] catalog validate-only + end-to-end too-fresh smoke — exit 0

## Regression test (bug-fix)

- **Test:** `scripts/lib/test_eval_alarms_min_eval_window.py`
- **Pre-fix:** committed as `c8ff807` — verified FAILED (AttributeError: `compute_too_fresh` / `SKIP_INTERVAL_TOO_SHORT` / the `too_fresh` param do not exist on main)
- **Post-fix:** verified PASSES after `d7d3863` (6 tests: streak preservation on delta=0, no-false-advance on delta>=1, ratio streak/baseline preservation, centralized-reset preservation, derivation boundaries, and default-off unregressed paths)

## Deviations from plan

- Extracted the derivation into a small module-level `compute_too_fresh()` helper (called from `main()`) instead of an inline expression, so the `[0, min)` boundary is directly unit-testable (`test_unknown_age_not_too_fresh`). Behavior is identical to the planned inline form.
- Guarded all five evaluators (including the non-load-bearing `eval_counter`/`eval_histogram_p99`) per the plan's primary approach; each returns a well-formed `extra_values` on the skip so `render_aggregate` cannot KeyError.


tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** Clean, well-scoped symmetric too-fresh lower bound for the counter-family evaluators. The regression test provably fails at pre-fix (new symbols absent) and passes post-fix; all existing eval-alarms suites and monitor-skill snippet tests stay green; default `too_fresh=False` keeps every existing path byte-identical.

<details>

Cycle 1 (no prior `## 🔍 Reviewer: Correctness` comment) — complete class-labeled change-list below. No blocking concerns found.

**Correctness — verified:**
- `compute_too_fresh(prev_scrape_age, min_eval_window)` = `0 <= prev_scrape_age < min_eval_window`. Half-open window `[0, min)` is correct: age 0 (same instant) is too-fresh, exactly the window is not, and age -1/absent ⇒ unknown ⇒ NOT too-fresh (fail-safe, mirrors gap_stale). Directly unit-tested at the boundaries.
- Sequencing verified by reading the full `eval_counter_streak` / `eval_counter_ratio` bodies: every `write_snapshot` preceding the `too_fresh` guard is inside a branch that returns (identity-change re-baseline, `collecting_baseline`, counter-reset re-baseline, `gap_stale` re-baseline). When control reaches the `too_fresh` branch, none of those returned, so the skip returns *before* any streak/baseline write and the snapshot is preserved untouched. A genuine restart/gap is still handled first (re-baseline wins over too-fresh), which is the intended precedence.
- `maybe_reset_counter_snapshot` early-returns on the `SKIP_INTERVAL_TOO_SHORT` sentinel only when `state == "skipped"`, so the centralized reset preserves the snapshot for a too-fresh skip and is a no-op for the non-snapshot kinds (`counter`, `histogram-p99`). No other skip reason is affected.
- Each evaluator's too-fresh skip returns a well-formed `extra_values` (`ev_default`), so `render_aggregate` cannot KeyError.
- `MIN_EVAL_WINDOW_SECONDS` parsed with a `ValueError` → 600 fallback; default 600.

**Test-verification gate (kind: bug-fix):** PASS. Regression test `scripts/lib/test_eval_alarms_min_eval_window.py` was run against pristine `origin/main`'s evaluator and fails (`AttributeError: compute_too_fresh` — the new API is absent), and passes (6/6) at PR head merged with current `origin/main`. The behavioral assertions (breach_streak preserved on delta=0, no false advance on delta>=1, ratio streak/baseline preservation, centralized-reset preservation, default-off unregressed) exercise the exact #3757 harm.

**Regression check:** `test_eval_alarms_counter_reset` (24), `_converged_loop` (9), `_histogram_rebucket` (10), `_inbound_auth_severity` (7) and `test-monitor-skill-snippets.sh` (439 checks) all pass at PR head merged with main.

Non-blocking nit (not gating; left as an observation, not an inline): the pre-fix failure mode is an import-level `AttributeError` rather than an assertion failure, because the behavioral API doesn't exist on main. That's acceptable for this fix since the assertions are only meaningful with the new parameter, but a note in the test docstring that "pre-fix failure is import-level" would aid future readers.

</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Non-parity change (touches only `scripts/lib/` monitoring tooling — no observable/interop surface). Purely additive, default-off, env-gated with a safe fallback, and fail-safe on unknown age. No regression, security, or operational risk identified.

<details>

Cycle 1 (no prior `## 🔍 Reviewer: Risk` comment) — complete class-labeled change-list below. No blocking concerns found.

**Parity:** N/A. Changed paths are `scripts/lib/eval-alarms.py` and a new test — internal monitoring/alarm tooling. Per `docs/PARITY.md` this is not on the observable/interop surface (no ledger/bucket hashes, tx result/meta XDR, SCP/overlay wire bytes, history archive format, HTTP/RPC/CLI contract, or crypto output). Nothing to gate.

**Regression-risk:** low. The new `too_fresh` parameter defaults to `False` on all five evaluators, so every existing call path that predates this PR is byte-identical unless `too_fresh` is explicitly threaded through from `main()`. The `main()` derivation is env-gated (`MIN_EVAL_WINDOW_SECONDS`, default 600) and only triggers when the prev baseline is genuinely younger than the window — the exact duplicate-tick condition from #3757.

**Operational-risk:** low. A mis-set `MIN_EVAL_WINDOW_SECONDS` only widens/narrows the suppression window; a `ValueError` falls back to 600. Unknown age (-1/absent) is never treated as too-fresh, so the guard can never suppress a legitimately-spaced tick. The change can only *suppress* a spurious cross-tick delta on a sub-window duplicate — it cannot cause a false fire.

**Scope:** the PR body correctly scopes this as the evaluator-level floor; the watchdog `STALE_SECS`/`flock` resize, acceleration case, and tick-history schema validation are deferred to follow-ups #3789/#3790/#3791 per the converged plan. That matches the issue's suggested-fix decomposition.

</details>

tomerweller (commented):
Addressed the bounce (Cycle 1): rebased the branch onto current `origin/main` (now atop #3788) to clear the flaky CANCELLED `test (testnet, ...)` Quickstart job. No code change — the diff is byte-identical, only the base moved. New head `3d1b32e0a`.

Local re-verification on the rebased tree:
- `test_eval_alarms_min_eval_window.py` — 6 passed
- existing eval-alarms suites (counter_reset 24, converged_loop 9, histogram_rebucket 10, inbound_auth_severity 7) — all pass
- `test-monitor-skill-snippets.sh` — exit 0 (439 checks)

Both agent reviewers already APPROVED; re-triggering CI for a clean testnet run. Ready for re-review.

tomerweller:
## 🔍 Reviewer: Correctness

**Verdict:** APPROVE

**Summary:** The symmetric too-fresh guard is correctly implemented, well sequenced in every counter-family evaluator, and the new regression test provably fails on `main` and passes on the PR head. Full eval-alarms suite is green.

<details>

Reviewed against PR head merged with current `origin/main` (branch already up to date with main; clean merge).

**Verified correct:**
- `compute_too_fresh(prev_scrape_age, min_eval_window)` uses a half-open window `[0, min)`: age `0` is too-fresh, exactly the window is not, and unknown age `-1`/absent ⇒ NOT too-fresh. This mirrors the `gap_stale` fail-safe (never suppress on unknown age) and defaults (600s lower bound vs 3600s upper bound) do not overlap.
- Sequencing in `eval_counter_ratio` and `eval_counter_streak`: the `too_fresh` branch is placed AFTER the identity / counter-reset / `gap_stale` re-baseline branches and BEFORE any snapshot write, so a PID/start-ticks change or counter reset is still handled first and the valid baseline (streak, counter_value, num/den) is preserved untouched. Confirmed by reading lines around the insertions.
- `maybe_reset_counter_snapshot` early-returns on the `SKIP_INTERVAL_TOO_SHORT` sentinel, placed first after the `state != "skipped"` guard, so the centralized reset does not wipe the streak/ratio/dynamic snapshot for a too-fresh skip. This is the load-bearing second half of the guard.
- `eval_counter` / `eval_counter_dynamic` / `eval_histogram_p99` return a plain SKIP on too-fresh (their prev.prom baseline is recomputed each tick, so no persisted state to protect). Correct.
- Default `too_fresh=False` on all six evaluators ⇒ normal path unregressed (confirmed by the default-off tests).

**Test-verification gate (bug-fix):** PASS. `test_eval_alarms_min_eval_window.py` is provably RED on `main` (`AttributeError: module 'eval_alarms' has no attribute 'compute_too_fresh'`) and GREEN on the PR head (6/6). It covers the two headline harms from #3757 — counter-streak zero-delta preservation, no-false-advance, and counter-ratio streak/baseline preservation — plus `compute_too_fresh` derivation and the `maybe_reset` early-return. Full suite: 5 files / 56 tests all green in-worktree.

**Non-blocking note (optional, not a bounce reason):** the `too_fresh` skip branches in `eval_counter`, `eval_counter_dynamic`, and `eval_histogram_p99` are not directly exercised by a unit test (only counter-streak and counter-ratio are). These three are simple plain-SKIP paths with no persisted baseline, so the risk is low — but a one-line assertion per branch would close the coverage gap. Optional follow-up.
</details>

tomerweller:
## 🔍 Reviewer: Risk

**Verdict:** APPROVE

**Summary:** Non-parity monitoring tooling only; no observable/interop surface touched. The suppression window is tight, fail-safe on unknown age, and env-configurable. No security, perf, or data-format concerns.

<details>

Lens: **Risk** (auto-detected — changed files are `scripts/lib/eval-alarms.py` and a test; no parity-critical crate under `crates/{scp,herder,ledger,tx,overlay}` is touched, so no `docs/PARITY.md` observable surface is in scope). Not self-modifying (eval-alarms.py is not in the pipeline-skills/scripts self-mod set).

**Regression / behavioral risk — bounded:**
- The only behavioral change is that counter-family alarms now SKIP (non-destructively) when the previous baseline is younger than `MIN_EVAL_WINDOW_SECONDS` (600s). Could this suppress a genuine alarm? Only when two scrapes land <10 min apart — which, per the issue's measured ~44 min real cadence, happens only on a duplicate/back-to-back tick (the exact pathology). Normal ticks are unaffected.
- Fail-safe: unknown age (`-1`/absent) is NOT treated as too-fresh, so a missing scrape-age never silences an alarm.
- `MIN_EVAL_WINDOW_SECONDS` is env-overridable with a `ValueError` fallback to 600 — no crash on bad input.

**Security / perf / data-format:** none. No wire bytes, XDR, hashes, or persisted-format changes; snapshot schema (`SCHEMA_VERSION = 1`) is unchanged; the guard only adds a new `skip_reason` string.

**Scope observation (non-blocking):** This PR implements a defensive guard at the evaluator layer, addressing Harm §1–3 of #3757 (cooldown burn, 2× streak/archive advance). The issue's root causes — `STALE_SECS` mis-sizing, the in-flight `.alive` guard, and unconditional tick serialization via `flock`, plus tick-history schema validation (§Harm 4) — are NOT in this PR and remain open. That is a reasonable defense-in-depth slice, not a defect; flagging so the remaining root-cause work isn't assumed done by closing #3757.
</details>

tomerweller:

**Reason:** CI failed (unrelated flaky test, will rebase/retry)

**Reviewer A (Correctness):** APPROVE — guard correctly implemented and sequenced; regression test provably red on `main`, green on head; full suite green.
**Reviewer B (Risk):** APPROVE — non-parity tooling; suppression window tight and fail-safe; no observable-surface/security/perf/format impact.
**External reviewers:** none (only the agent's own COMMENTED review; no unresolved inline threads).
**CI:** red — but **not diff-attributable**.

Both reviewer lenses APPROVE. The sole CI failure is an **unrelated timing-boundary flake**, not caused by this diff:

```
not ok 40 - long-stale: .alive at boundary (21600s) → not stale
# 1 test(s) failed  (439 total)
```

This test lives in `scripts/test-monitor-skill-snippets.sh` (test 20h) and exercises the watchdog `.alive` staleness check in `scripts/lib/monitor-decisions.sh` — **neither file is touched by this PR** (the diff is confined to `scripts/lib/eval-alarms.py` + a new test). The test mocks `.alive` mtime to exactly `now - 21600` and asserts `alive_age -le 21600`; any wall-clock drift between mock creation and the check makes `alive_age = 21600 + δ`, flipping the boundary to "stale". It is inherently racy at the exact boundary and fails non-deterministically regardless of this change.

Triple-green cannot be declared with a red check, so routing back to `ready-for-doing`. `/do` Mode B: no code changes are required for the reviewers — just rebase on `origin/main` and re-push to re-trigger CI (the flake should clear on re-run). If the same unrelated `.alive` boundary test keeps failing, it is a pre-existing flaky test worth a separate hardening issue (make the boundary test deterministic, e.g. freeze the clock or use `-lt`), not a change to this PR.

tomerweller:
## Do: CI Flake Cleared (no code change)

The bounce was CI-red only, from an **unrelated flaky test** — `not ok 40 - long-stale: .alive at boundary (21600s) → not stale` in `scripts/test-monitor-skill-snippets.sh`. That test mocks `.alive` mtime to `now − 21600` and then re-reads `date +%s`; when a one-second boundary is crossed between the mock and the check, `alive_age` becomes `21601` and the `-le 21600` guard (in `scripts/lib/monitor-decisions.sh`) reports the session stale. It is a pre-existing timing flake in files this PR does not touch (this PR only changes `scripts/lib/eval-alarms.py` + adds `test_eval_alarms_min_eval_window.py`).

Re-ran the failed jobs — Monitor-Tick Skill Snippets is now green. No code change pushed (fixing the unrelated flake would expand PR scope). Both reviewers already APPROVED. Ready for re-review / merge once the remaining long-running Quickstart testnet job finishes (currently green-tracking).

tomerweller:
Rebased onto `origin/main` (post-#3838) and merging. Four convergent conflicts in `scripts/lib/eval-alarms.py` against #3758/#3834 and the just-landed #3838; resolved as follows.

**The elegant part:** rather than keeping two parallel early-returns, I folded the sentinel into main's existing set —

```python
PRESERVE_BASELINE_SKIP_REASONS = {"missing process identity", SKIP_INTERVAL_TOO_SHORT}
```

— so main's single `if skip_reason in PRESERVE_BASELINE_SKIP_REASONS: return` subsumes this PR's special case. `SKIP_INTERVAL_TOO_SHORT` is defined above the set (Python evaluation order), and the set's comment now records why the too-fresh reason belongs there: the evaluator already returned *without* a snapshot write, so the membership is what stops the centralized reset from wiping streak/ratio/dynamic state.

**Carry-forward from #3838, applied:** `eval_histogram_bucket_rate` is a `prev.prom`-differencing family, so it needed the lower bound as well as the upper one. It arrived (via #3838) with `gap_stale` only. I added `too_fresh: bool = False` to its signature, threaded it at the dispatch site, and added the guard body mirroring `eval_histogram_p99` exactly. Without this, the newest alarm family would have had asymmetric window guarding from birth.

**Verified on the resolved tree, real exit codes:** `ast.parse` OK; `pytest scripts/lib/ scripts/ci/test_check_alarm_versions.py` → **138 passed**; `test-monitor-skill-snippets.sh` → rc=0, **465/465**, no plan mismatch; `.claude`/`.agents` catalogs still byte-identical.

⚠️ **Substantive correction to this PR's own docstring — please don't propagate it.** The comment claims that for the `prev.prom` families "the next normal-cadence tick spans the real interval". That is **false**: `SKILL.md:1183` step 2 runs `mv current.prom prev.prom` **unconditionally every tick**, and this PR doesn't touch it. So the below-window sub-interval is not deferred — it is **permanently dropped**. Consequence: an increment that lands entirely inside a too-fresh window becomes **unalertable**, which is exactly the fire-then-die class at `crates/app/src/app/persist.rs:709`, where `record_db_busy_drop` fires immediately before `shutdown.trigger`. The effect is bounded, and skipping beats evaluating a meaningless sub-second delta, so this still merges — but either the comment should be corrected or step 2 gated on the same guard.

**Does not fix #3757's root cause** (the out-of-repo `STALE_SECS` configuration), so **#3757 should stay open** rather than being auto-closed by this.

**Symmetry itself is correct:** half-open `[0, min)` window with a `-1` fail-safe, and the snapshot families are genuinely non-destructive.

**Self-modifying:** `scripts/lib/eval-alarms.py` and its new test file — both self-modifying. No Rust, so #3830's clippy gate is vacuous. Not a parity concern.

**Pre-existing gap worth its own issue:** `scripts/lib/test_eval_alarms_*.py` are executed by **no workflow** (only `test-monitor-skill-snippets.sh` and `scripts/ci/test_check_alarm_versions.py` are wired), so this PR's new tests — and #3838's and #3847's — are unenforced in CI. That's how #3838's `.agents` mirror divergence would have shipped silently.

Sole non-green check was the chronic `test (testnet, …)` shard (#3768), which touches nothing here.

